PR #80, opened 2024-03-18
Fix: Initial and Future A value string check

OnlyJousting:
Changes:
- Fixed comparison of initial and future A values in pool/advanced tab to determine "up or "down"


michwill (approved):

bout3fiddy (commented):
but why doesn't `future_A! > initial_A!` work? it should work in principle, yes?

amytsang (approved):

PR #127, opened 2024-04-02
Pool list - Only render visible rows

### LINKED ISSUE #108: Improve pool list load time
On the pool list page: https://curve.fi/#/ethereum/pools

when you try to sort the list of pools by clicking on each tab, it takes quite a bit of delay before the new arranged list is rendered. This needs to be instantaneous.

So: the task is to optimise the rendering of this list.

Location of code for the page folder: https://github.com/curvefi/curve-frontend/tree/main/apps/main/src/components/PagePoolList

the array is rendered on this line: https://github.com/curvefi/curve-frontend/blob/38a4bf5993792abc090754b6cce2305341a45a1b/apps/main/src/components/PagePoolList/index.tsx#L286

DanielSchiavini:
Fixes #108

### DIFF
--- a/apps/main/src/store/createPoolListSlice.ts
+++ b/apps/main/src/store/createPoolListSlice.ts
@@ -1,7 +1,6 @@
 import type { GetState, SetState } from 'zustand'
 import type { State } from '@/store/useStore'
-import type { FormStatus, FormValues, SearchParams, SortKey } from '@/components/PagePoolList/types'
-import type { FilterKey, Order } from '@/components/PagePoolList/types'
+import type { FilterKey, FormStatus, FormValues, Order, SearchParams, SortKey } from '@/components/PagePoolList/types'
 
 import Fuse from 'fuse.js'
 import chunk from 'lodash/chunk'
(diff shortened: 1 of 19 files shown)


apps/main/src/store/createPoolListSlice.ts
```diff
@@ -195,14 +194,14 @@ const createPoolListSlice = (set: SetState<State>, get: GetState<State>) => ({
       userPoolList: UserPoolListMapper | undefined
     ) => {
       const activeKey = getPoolListActiveKey(rChainId, searchParams)
-      const storedFormValues = get()[sliceKey].formValues
-      const storedResults = get()[sliceKey].result[activeKey]
+      const state = get()[sliceKey]
```
DanielSchiavini, line 197: Is there a good reason to call `get()[sliceKey]` repeatedly?

amytsang (approved):

apps/main/src/store/createPoolListSlice.ts
amytsang: Depends on where it is being called, if it is used after an await, then the state can be state, or else I don't think it is necessary. I also did not measure, so not sure if it is causing a memory issue.

PR #191, opened 2024-06-04
chore: upgrade lending api 2.0.9

amytsang:
Add Leverage symbol
Screenshot 2024-06-05 at 5 10 51 AM
Screenshot 2024-06-05 at 5 11 06 AM
Screenshot 2024-06-05 at 5 11 16 AM

### DIFF
--- a/apps/lend/src/components/PageMarketList/Page.tsx
+++ b/apps/lend/src/components/PageMarketList/Page.tsx
@@ -41,10 +41,16 @@ const Page: NextPage = () => {
 
   const { signerAddress } = api ?? {}
 
-  const FILTER_MAPPER: FilterMapper = {
-    all: { id: 'all', displayName: t`All` },
-    user: { id: 'user', displayName: t`My markets` },
-  }
+  const FILTER_MAPPER: FilterMapper = signerAddress
+    ? {
+        all: { id: 'all', displayName: t`All` },
+        leverage: { id: 'leverage', displayName: t`Leverage` },
+        user: { id: 'user', displayName: t`My markets` },
+      }
+    : {
+        all: { id: 'all', displayName: t`All` },
+        leverage: { id: 'leverage', displayName: t`Leverage` },
+      }
 
   const FILTER_TYPE_MAPPER: FilterMapper = {
     borrow: { id: 'borrow', displayName: t`Borrow` },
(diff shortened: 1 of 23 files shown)


michwill (dismissed):

apps/lend/src/components/LeverageIcon.tsx
ignromanov, line 1: Perhaps we should add all style components to the UI library at once. Specifically, create an Icons folder and put such components there.

apps/lend/src/components/PageMarketList/Page.tsx
```diff
@@ -34,16 +35,30 @@ const Page: NextPage = () => {
...
   const [loaded, setLoaded] = useState(false)
   const [parsedSearchParams, setParsedSearchParams] = useState<SearchParams | null>(null)
 
   const { signerAddress } = api ?? {}
 
-  const FILTER_MAPPER: FilterMapper = {
-    all: { id: 'all', displayName: t`All` },
-    user: { id: 'user', displayName: t`My markets` },
-  }
+  const FILTER_MAPPER: FilterMapper = signerAddress
```
ignromanov, line 45: It might be worth creating 2 constants outside of this component, and using only them in this condition
```
const SIGNER_FILTER_MAPPER: FilterMapper = {...} as const
const DEFAULT_FILTER_MAPPER: FilterMapper = {...} as const
```

```suggestion
  const FILTER_MAPPER = signerAddress ? SIGNER_FILTER_MAPPER : DEFAULT_FILTER_MAPPER
```

apps/lend/src/store/createAppSlice.ts
```diff
@@ -107,26 +109,18 @@ const createAppSlice = (set: SetState<State>, get: GetState<State>): AppSlice =>
     state.updateGlobalStoreByKey('api', api)
     state.updateGlobalStoreByKey('isLoadingCurve', false)
 
-    const { owmDatasMapper } = await state.markets.fetchMarkets(api)
+    await state.markets.fetchMarkets(api)
```
ignromanov, line 112: It may be worth querying this data in parallel, without waiting for the first query to be executed
```suggestion
    await Promise.all([
        state.markets.fetchMarkets(api)
```

ignromanov (approved):

apps/lend/src/store/createAppSlice.ts
amytsang: No, I need to call the market list first, which I use to get a list of tokens to call usdRates on.

ignromanov (approved):

Macket (approved):

PR #233, opened 2024-07-04
task: display campaign rewards for lending markets

OnlyJousting:
Changes:
- Display campaigns from https://github.com/curvefi/curve-external-reward on Lend

### DIFF
--- a/packages/ui/src/CampaignRewards/CampaignRewardsComp.tsx
+++ b/packages/ui/src/CampaignRewards/CampaignRewardsComp.tsx
@@ -1,10 +1,11 @@
-import type { CampaignRewardsCompProps } from 'ui/src/CampaignRewards/types'
+import type { CampaignRewardsCompProps, RewardsPool } from 'ui/src/CampaignRewards/types'
 
 import styled from 'styled-components'
 import Image from 'next/image'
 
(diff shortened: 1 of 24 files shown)


packages/ui/src/CampaignRewards/CampaignRewardsComp.tsx
```diff
@@ -25,18 +27,42 @@ const RewardsCompSmall: React.FC<CampaignRewardsCompProps> = ({ rewardsPool, hig
...
-  dashboardLink,
-}: {
-  platform: string
-  description: string
-  dashboardLink: string
-}) => {
+const TooltipMessage = ({ rewardsPool }: { rewardsPool: RewardsPool }) => {
+  const { campaignName, platform, description, dashboardLink, campaignStart, campaignEnd } = rewardsPool
+
+  const options: Intl.DateTimeFormatOptions = {
```
ignromanov, line 33: It would be nice to take this definition out of the component so that it is only defined once, rather than being created on every render. It won't affect performance now, but if in the future this variable will be in dependencies, it will lead to unnecessary re-rendering.

ignromanov (approved):

ignromanov (approved):

amytsang (approved):

PR #245, opened 2024-07-09
fix: show error if submit form is unsuccessful

amytsang:


amytsang:
move changes to refactor/deposit-withdraw-swap-forms

PR #322, opened 2024-08-16
fix: add custom number formatting to all chart series

OnlyJousting:
Changes:
- Added custom number formatting for all liquidation range series in the ohlc chart component to fix edge case price scale bug.


bout3fiddy (approved):
Maybe if this comes back to us we can consider using szabo or gwei or whatever decimal units that makes it legible

mimaklas (approved):

PR #331, opened 2024-08-28
perf: improve cache performance

### LINKED ISSUE #330: Unnecessary serialization causes performance issues
I am looking at the performance of the curve-frontend app, and I noticed that some unnecessary time is being spent in a function setItem. I tracked it down to [this line](https://github.com/curvefi/curve-frontend/blob/1e5d0d74508a4cec7cd230873f2fed53ddf1a902/apps/main/src/store/useStore.ts#L89).

`JSON.stringify` is quite slow, mainly if the payload can end up being 5MB!
It's unclear why we need this limit here, if necessary we could use [something like this](https://stackoverflow.com/a/11900218)

Also:
- We should document why we have the 5MB limit
- Why is the result of JSON.stringify multiplied by 2?
- `maxSize` calculation only needs to be done once

DanielSchiavini:
Fixes #330 by debouncing the localStorage cache.
Updated zustand to v4 so we can customize the `persist` storage.


apps/main/src/store/createPoolsSlice.ts
```diff
@@ -348,32 +347,21 @@ const createPoolsSlice = (set: SetState<State>, get: GetState<State>): PoolsSlic
...
-      for (const idx in results) {
-        const rewardsApy = results[idx]
-        rewardsApyMapper[rewardsApy.poolId] = rewardsApy
-      }
-      get()[sliceKey].setStateByActiveKey('rewardsApyMapper', chainId.toString(), rewardsApyMapper)
-      return rewardsApyMapper
-    },
     fetchPoolsRewardsApy: async (chainId, poolDatas) => {
       log('fetchPoolsRewardsApy', chainId, poolDatas.length)
+      let rewardsApyMapper: RewardsApyMapper = { ...get()[sliceKey].rewardsApyMapper[chainId] }
```
DanielSchiavini, line 352: I merged the 2 functions so we don't need to reclone this variable every iteration

apps/main/src/store/useStore.ts
```diff
@@ -74,40 +73,31 @@ const store = (set: SetState<State>, get: GetState<State>): State => ({
...
 const useStore =
   process.env.NODE_ENV === 'development'
     ? create<State>(devtools(persist(store, cache)))
     : create<State>(persist(store, cache))
-// const useStore = isDevelopment ? create(devtools(store)) : create(store)
-
-// check if there are too much in storage
-function canAddToStorage<T extends Object>(storageObj: T) {
-  const maxSize = 5 * 1024 * 1024 // 5MB limit
-  const itemSize = (JSON.stringify(storageObj)?.length ?? 0) * 2
```
DanielSchiavini, line 108: storage.setItem was also calling JSON.stringify, so this was repeated 2x every loop

amytsang (approved):

OnlyJousting (approved):

OnlyJousting (approved):

PR #355, opened 2024-09-23
docs: translate to simplified Chinese

yuzuswapteam005:
translate to simplified Chinese


DanielSchiavini:
Thanks, however we are currently leaving the translation to browser plugins

PR #356, opened 2024-09-23
docs: translate to traditional Chinese

yuzuswapteam005:
translate to traditional Chinese


DanielSchiavini:
Thanks, however we are currently leaving the translation to browser plugins

PR #372, opened 2024-10-01
[refactor] move shared code

DanielSchiavini:
In preparation to implement tanstack for llamalend, we can move some shared files to curve-lib and ui, so they can be reused.


apps/loan/src/lib/apiCrvusd.ts
```diff
@@ -479,7 +479,7 @@ const loanCreate = {
         : null,
     ])
 
-    const maxRecvs = fulfilledValue(maxRecvsResults) ?? null
+    const maxRecvs: any = fulfilledValue(maxRecvsResults) ?? null
```
ignromanov, line 482: Try to avoid using `any`, better use `as type` or, if you really don't know the type, use `as unknown`.

packages/curve-lib/src/shared/index.ts
ignromanov, line 1: Based on FSD, there should be no layer exports. All exports should be from `slices`, and if we are talking specifically about a `shared` layer, then from `segments`.

packages/curve-lib/src/shared/utils.ts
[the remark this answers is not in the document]
ignromanov: And good to split that file by modules

packages/curve-lib/src/shared/validation/lib.ts
ignromanov, line 1: Add aliases for paths in TS config to use full paths

packages/ui/src/FormErrorsDisplay/FormErrorsDisplay.tsx
```diff
@@ -1,9 +1,9 @@
-import AlertFormError from '@/components/AlertFormError'
+import AlertFormError from '../../../../apps/main/src/components/AlertFormError'
```
ignromanov, line 1: It shouldn't be like this

apps/main/src/entities/chain/model/chain-validation.ts
```diff
@@ -1,5 +1,5 @@
 import type { ChainQueryParams } from '@/entities/chain/types'
-import { createValidationSuite } from '@/shared/validation'
+import { createValidationSuite } from '@/shared/curve-lib'
```
ignromanov, line 2:
```suggestion
import { createValidationSuite } from '@/shared/curve-lib/validation'
```
Once you fix the exports in curve-lib this and similar imports should look like this to comply with FSD.

apps/main/src/features/add-gauge-reward-token/ui/AddRewardToken.tsx
```diff
@@ -10,11 +10,12 @@ import TxInfoBar from '@/ui/TxInfoBar'
 
 import { useSignerAddress } from '@/entities/signer'
 import { DistributorInput, EstimatedGasInfo, FormActions, TokenSelector } from '@/features/add-gauge-reward-token/ui'
-import { formDefaultOptions } from '@/shared/model/form'
-import { FormErrorsDisplay } from 'shared/ui/forms'
+import { formDefaultOptions } from '@/shared/curve-lib'
+import { FormErrorsDisplay } from '@/ui/FormErrorsDisplay'
```
ignromanov, line 14:
```suggestion
import { FormErrorsDisplay } from '@shared/ui-kit/FormErrorsDisplay'
```
Try doing these imports via FSD.

ignromanov (changes requested):

apps/loan/src/lib/apiCrvusd.ts
DanielSchiavini: This is old code I was just fixing the build error. The real type is actually pretty convoluted. I'll fix it though.

packages/curve-lib/src/shared/utils.ts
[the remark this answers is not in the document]
DanielSchiavini: I was just moving this, but given it's only logging I can rename it

packages/ui/src/FormErrorsDisplay/FormErrorsDisplay.tsx
DanielSchiavini: yeah for sure, this is unused now, I'll remove

apps/main/src/features/add-gauge-reward-token/ui/AddRewardToken.tsx
DanielSchiavini: This seems a bit unrelated to this PR. We could just replace `@/ui` with `@/shared/ui`

ignromanov (approved):

OnlyJousting (approved):

PR #376, opened 2024-10-03
refactor: replace usdRate slice with react-query

DanielSchiavini:
- Start-up react-query for llamalend
- Remove usdRate slice, use react-query instead
- Remove some unused code


packages/ui/src/QueryProvider/index.tsx
```diff
@@ -9,7 +9,7 @@ type QueryProviderWrapperProps = {
 }
 
 export function QueryProvider({ children, persister, queryClient }: QueryProviderWrapperProps) {
-  if (persister) {
+  if (persister && !window?.localStorage?.getItem('react-query-no-persist')) {
```
ignromanov, line 12: What's the reason? 🤔

DanielSchiavini: it's annoying to develop when all the hooks get cached, so when I change them they don't get called

ignromanov: Try to implement devtools for that
https://tanstack.com/query/latest/docs/framework/react/devtools

apps/lend/src/entities/token/lib/last-results.ts
```diff
@@ -0,0 +1,7 @@
+import { queryClient } from '@/shared/api/query-client'
+import { tokenKeys } from '@/entities/token'
+
+export const getLastTokenUsdRate = (chainId: ChainId, tokenAddress: string) => queryClient.getQueryData<number>(tokenKeys.usdRate({
```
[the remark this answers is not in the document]
DanielSchiavini: The reason I wanted to keep this function isolated is that it should be temporary. It's a bit of an anti-pattern to use the last value instead of using the query.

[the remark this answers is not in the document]
ignromanov: No, I think it's fine, Tanstack even advises it. You have no other way to get the value inside an API function other than using that function or an asynchronous one to get the value precisely.

apps/lend/src/entities/token/lib/validation.ts
```diff
@@ -0,0 +1,15 @@
...
+import { assertValidity, checkValidity } from '@/shared/lib/validation'
+import { tokenValidationSuite } from '@/entities/token/model/validation'
+
+export const assertTokenValidity = <T extends CombinedTokenParams>(data: T, fields?: Extract<keyof T, string>[]) =>
+  assertValidity(tokenValidationSuite, data, fields)
+
+export const checkTokenValidity = <T extends CombinedTokenParams>(data: T, fields?: Extract<keyof T, string>[]) =>
+  checkValidity(tokenValidationSuite, data, fields)
+
+export const extractTokenParams = ([, chainId, , givenAddress]: TokenQueryKeyType<'usdRate'>) =>
```
ignromanov, line 11: Implement a factory generic function for this. It should be useful for all kind of keys. And I would call it `assertQueryKey`

apps/lend/src/store/createMarketListSlice.ts
```diff
@@ -86,15 +88,18 @@ const createMarketListSlice = (set: SetState<State>, get: GetState<State>): Mark
     filterSmallMarkets: (api, owmDatas) => {
       const { chainId } = api
       const capAndAvailableMapper = get().markets.statsCapAndAvailableMapper[chainId] ?? {}
-      const usdRatesMapper = get().usdRates.tokens
       const { smallMarketAmount, marketListShowOnlyInSmallMarkets } = networks[chainId]
       return owmDatas.filter(({ owm }) => {
         const { cap } = capAndAvailableMapper[owm.id] ?? {}
-        const token = owm.borrowed_token
-        const usdRate = usdRatesMapper[token.address]
+        const usdRate = queryClient.getQueryData<number>(
```
ignromanov, line 94: Wrap it with a factory function to don't use keys outside the entity
```suggestion
        const usdRate = getTokenQueryData('usdRate', { chainId, tokenAddress })
```

packages/curve-lib/src/shared/lib/queries/types.ts
```diff
@@ -10,10 +10,16 @@ export type QueryResultsArray<T extends QueryOptionsArray> = {
...
+  'data' | 'isLoading' | 'isPending' | 'isError' | 'isFetching'
+>
+
+export type CombinedQueriesResult<T extends QueryOptionsArray> = PartialQueryResult<CombinedDataType<T>>
+
+export type CombinedQueryMappingResult<T extends QueryOptionsArray, K extends string[]> = PartialQueryResult<
+  Record<K[number], CombinedDataType<T>[number]>
+>
+
+export const FETCHING = { isError: false, isLoading: true, isPending: true, isFetching: true, data: undefined } as const
```
ignromanov, line 24: it should be in `config.ts`

apps/lend/src/entities/token/lib/validation.ts
DanielSchiavini: how can I extract the keys in a generic way? We only have functions for params=>key but not the other way around.

apps/lend/src/store/createMarketListSlice.ts
DanielSchiavini: this one is also not going to work, we need the tokenKeys to be able to create the key

apps/lend/src/entities/token/lib/query-data.ts
```diff
@@ -0,0 +1,6 @@
+import { tokenKeys } from '../model'
+import { queryClient } from '@/shared/api/query-client'
+import { TokenKey } from '@/entities/token'
+
+export const getTokenQueryData = <Result, Key extends TokenKey = TokenKey>(key: TokenKey, keyData: Parameters<typeof tokenKeys[Key]>[0]) =>
```
ignromanov, line 5: Take the generic factory function out to the shared modules to only specify the keys object here and have a special function for tokens.
```ts
export const getTokenQueryData = getQueryDataFactory(tokensKeys)
```

ignromanov: Another architectural problem to think about is how to get the type of the return value depending on the key. To avoid using the `Result` type. This can be done on pure TS types

DanielSchiavini: yeah the result type is only defined by the query function, and then connected via query options. So it's not possible to get the return type just from the keys

DanielSchiavini: [That](https://github.com/curvefi/curve-frontend/pull/376/commits/4b69aae9eff2531738c5206dd0b8608064777b69) kind of makes things more complicated :disappointed:

ignromanov: Is it? It's a once-written factory function that was written and forgotten. Or what are you talking about?

ignromanov (approved):

ignromanov (approved):

PR #378, opened 2024-10-04
refactor: move one way markets to react-query

DanielSchiavini:
- remove one way markets from the store slice, retrieve them via react-query
- get rid of the OWMData structure, which wasn't necessary
- rename references from 'owm' to 'market' for readability
- update references
- ~Depends on https://github.com/curvefi/curve-frontend/pull/376~


apps/lend/src/components/ChartOhlcWrapper/index.tsx
```diff
@@ -15,11 +15,12 @@ import Box from '@/ui/Box'
 import PoolActivity from '@/components/ChartOhlcWrapper/PoolActivity'
 import TextCaption from '@/ui/TextCaption'
 import AlertBox from '@/ui/AlertBox'
+import { useOneWayMarket } from '@/entities/chain'
 
 const ChartOhlcWrapper: React.FC<ChartOhlcWrapperProps> = ({ rChainId, userActiveKey, rOwmId }) => {
+  const market = useOneWayMarket(rChainId, rOwmId)?.data
```
ignromanov, line 21: Why can a hook be undefined?  🤔

apps/lend/src/components/DetailsUser/components/DetailsUserSupply.tsx
```diff
@@ -50,7 +50,7 @@ const DetailsUserSupply = (pageProps: PageContentProps) => {
       ) : foundVaultShares ? (
         <ContentWrapper paddingTop>
           <Wrapper>
-            <CellUserMain {...pageProps} type="supply" />
+            <CellUserMain {...pageProps} market={market!} type="supply" />
```
ignromanov, line 53: 🙈

apps/lend/src/components/PageMarketList/index.tsx
```diff
@@ -13,19 +13,20 @@ import MarketListNoResult from '@/components/PageMarketList/components/MarketLis
 import MarketListItemContent from '@/components/PageMarketList/components/MarketListItemContent'
 import TableSettings from '@/components/PageMarketList/components/TableSettings/TableSettings'
 import usePageVisibleInterval from '@/ui/hooks/usePageVisibleInterval'
+import { useOneWayMarketMapping } from '@/entities/chain'
 
 const MarketList = (pageProps: PageMarketList) => {
   const { rChainId, isLoaded, searchParams, api, updatePath } = pageProps
 
   const activeKey = _getActiveKey(rChainId, searchParams)
+  const {data: marketMapping} = useOneWayMarketMapping(rChainId)
```
ignromanov, line 22:
```suggestion
  const { data: marketMapping } = useOneWayMarketMapping(rChainId)
```
What's wrong with prettier?

apps/lend/src/entities/chain/api/markets-api.ts
ignromanov, line 1: I would name it `markets-query`

apps/lend/src/entities/chain/api/markets-api.ts
```diff
@@ -0,0 +1,17 @@
...
+export const queryOneWayMarketNames: QueryFunction<
+  string[],
+  ChainQueryKeyType<'markets'>
+> = async ({ queryKey }) => {
+  logQuery(queryKey)
+  const [, chainId, , ] = queryKey
+  assertChainValidity({ chainId })
+
+  const {api} = useStore.getState()
+  return api!.oneWayfactory.getMarketList()
```
ignromanov, line 16: You cannot return undefined from queries. Throw an error if `api` doesn't exists

apps/lend/src/entities/chain/lib/chain-info.ts
```diff
@@ -2,7 +2,12 @@ import useStore from '@/store/useStore'
 import { useMemo } from 'react'
 import { FETCHING, PartialQueryResult, READY } from '@/shared/lib/queries'
 
-export const useChainId = (): PartialQueryResult<ChainId> => {
+export const useApi = (): PartialQueryResult<Api> => {
```
ignromanov, line 5: Put jsdoc `@deprecated` tag or something like this. To don't use it and don'd forget to delete in a future.

apps/lend/src/entities/chain/lib/data-hooks.ts
```diff
@@ -0,0 +1,26 @@
...
+    chainValid && marketNames && api ? Object.fromEntries(
+      marketNames
+        .filter(marketName => !networks[chainId].hideMarketsInUI[marketName])
+        .map(name => [name, api.getOneWayMarket(name)])
+    ) : undefined, [api, chainId, chainValid, marketNames]
+  )
+  return { data, ...rest }
+}
+
+export const useOneWayMarket = (chainId: ChainId, marketName: string) => {
```
ignromanov, line 23: Good to add type to the marketName

apps/lend/src/entities/chain/model/query-keys.ts
```diff
@@ -1,8 +1,11 @@
-import type { ChainQueryParams } from '@/entities/chain/types'
+import type { ChainQueryParams } from '../types'
 import type { ExtractQueryKeyType } from '@/shared/types/api'
 
 export const chainKeys = {
   root: ({ chainId }: ChainQueryParams) => ['chain', chainId] as const,
+  markets: ({ chainId }: ChainQueryParams) => ['chain', chainId, 'markets'] as const,
```
ignromanov, line 6:
```suggestion
  markets: ({ chainId }: ChainQueryParams) => [...chainKeys.root({ chainId }), 'markets'] as const,
```

apps/lend/src/components/ChartOhlcWrapper/index.tsx
DanielSchiavini: yeah this looks incorrect

apps/lend/src/entities/chain/api/markets-api.ts
DanielSchiavini: this will not return undefined, it would throw an error if api is undefined. However, the enabled condition checks for the API being in the store

apps/lend/src/components/DetailsUser/components/DetailsUserSupply.tsx
DanielSchiavini: so the pageProps are defined globally, but in this case we know the market is set, so it's safe to use `!`
however, it would be best to refactor and remove market from pageProps

apps/lend/src/components/PageMarketList/index.tsx
DanielSchiavini: seems it's not part of the pipeline

DanielSchiavini: If I run `npm run format` I get 244 files changed

apps/lend/src/entities/chain/lib/data-hooks.ts
DanielSchiavini: what do you mean? it shouldn't be `: string`?

apps/lend/src/entities/chain/lib/chain-info.ts
DanielSchiavini: :thinking: why?

ignromanov: I think it's a temporary hook and it shouldn't work that way. So I would label it somehow so I don't use it too often.

apps/lend/src/entities/chain/lib/data-hooks.ts
ignromanov: I assumed that `getMarketList` returns a special type, not just a string. This does not seem to be the case.

ignromanov (approved):

PR #415, opened 2024-10-24
chore: upgrade next to fix warnings

DanielSchiavini:
This warning is being printed every time in development mode.
Seems it’s been fixed in 14.2.4: https://github.com/vercel/next.js/issues/65161

The upgrade caused issues with "@lingui/swc-plugin"

image


apps/lend/package.json
```diff
@@ -41,7 +41,7 @@
     "lodash": "4.17.21",
     "make-plural": "^7.1.0",
     "memoizee": "^0.4.15",
-    "next": "^13.5.6",
+    "next": "*",
```
ignromanov, line 44: You should also put it to `peerDependencies` section

PR #427, opened 2024-11-06
refactor: theme setup

DanielSchiavini:
- Based on ~https://github.com/curvefi/curve-frontend/pull/421~ https://github.com/curvefi/curve-frontend/pull/429
- Do not read the JSON file during runtime. Instead, generate a TypeScript file.
- Get rid of FSD structure for themes
- Get rid of import forwarding to MUI, any customization should be done via theme


DanielSchiavini:
Storybook is failing because we need to enable github pages. That shouldn't block the merge

OnlyJousting (approved):

PR #490, opened 2024-11-30
Update messages.po

yusufky63:
Fixed typos in 'Withdraw' translations


DanielSchiavini (changes requested):
.po files are generated we do not write to them manually

PR #493, opened 2024-12-03
styling: fix button on hover border styling

OnlyJousting:
Changes:
- Fixed button border styling bug on hover


0xAlunara (approved):
Tested & worked fine. Alternatively you could've just changed `border: 1px solid var(--button--border-color);` to `border: 1px solid transparent;` instead, given `--button--border-color` isn't really used anywhere else and its sole purpose is now to fix some jumping issues. But I think these components will be replaced by MUI in the future anyway.

0xtutti (approved):

PR #486, opened 2024-11-29
feat: implement design system

DanielSchiavini:
- Import Figma design system into our codebase
- See [b222d3](https://github.com/curvefi/curve-frontend/commit/b222d320001515783b75fb944eda5138d1b57f96#diff-25b96aa16f8b9f68c692d0db061713707e9cfc67b296e0ea30908928fb3ffcd6) for the source code. I guess we don't need that in git.
- The design system is assumed to be maintained in code from now on

PR #500, opened 2024-12-04
feat: add new footer

0xAlunara:
- Removes the old footer.
- Adds a new footer in `curve-ui-kit/src/widgets/Footer`.
- Use the new footer in every app and always show it, regardless of device.
- Uses the new design system made by Daniel.
- Responsive on mobile, tablet and desktop by using MUI Grid.
- Added a simple Cypress test to get familiar with it. It's very basic and only tests whether it's shown on desktop, tablet and mobile and contains more than one link.
- Uses existing MUI tokens, imported custom (SVG) icons for X, Llama Airforce, CrvHub and Dune Analytics.
- Depends on https://github.com/curvefi/curve-frontend/pull/486

@0xtutti I didn't felt like bothering to add the flag masks to the Telegram icons for Chinese and Russian channels.

https://github.com/user-attachments/assets/5ae5c1e4-ffab-4432-a227-2fd0404619b2


apps/main/src/layout/default/index.tsx
```diff
@@ -59,10 +56,10 @@ const BaseLayout = ({ children }: { children: React.ReactNode }) => {
         maintenanceMessage={maintenanceMessage}
         handleNetworkChange={handleNetworkChange}
       />
-      <Container className={isMdUp ? 'hasFooter' : ''} globalAlertHeight={layoutHeight?.globalAlert}>
+      <Container globalAlertHeight={layoutHeight?.globalAlert}>
```
DanielSchiavini, line 59: so much duplication :sob: 
these index files are almost the same (nothing related to this PR)

packages/curve-ui-kit/src/shared/ui/CrvHubIcon.tsx
```diff
@@ -0,0 +1,14 @@
+import { createSvgIcon } from '@mui/material/utils'
```
DanielSchiavini, line 1: I think it would be nice to create a separate folder for icons

packages/curve-ui-kit/src/widgets/Footer/Description.tsx
```diff
@@ -0,0 +1,57 @@
...
+  margin: 8,
+  alt: 'Curve',
+})
+
+export const Description = () => (
+  <Box
+    sx={{
+      display: 'flex',
+      flexDirection: 'column',
+      maxWidth: '330px',
```
DanielSchiavini, line 21: Hardcoded width isn't ideal.. Can't we have it size automatically?
The parent already has a fixed size

@0xtutti it looks hardcoded at 328px in the design

packages/curve-ui-kit/src/widgets/Footer/Description.tsx
```diff
@@ -0,0 +1,57 @@
...
+      mr: '1.5rem',
+    }}
+  >
+    <Link
```
DanielSchiavini, line 25: This looks like a duplicate of `<HeaderLogo appName="main" />` maybe we should [move this component](https://github.com/curvefi/curve-frontend/blob/ebd51245b139c68a1f4ad6703d27d998e749637d/packages/curve-common/src/widgets/Header/HeaderLogo.tsx#L18) to shared/ui too?

packages/curve-ui-kit/src/widgets/Footer/Description.tsx
```diff
@@ -0,0 +1,57 @@
...
+        Curve
+      </Typography>
+    </Link>
+
+    <Typography
+      variant="bodySRegular"
+      color="textSecondary"
+      sx={{
+        lineHeight: '1rem',
+      }}
```
DanielSchiavini, line 51: The lineHeight in Figma is sm which is the default for [bodySRegular](https://github.com/curvefi/curve-frontend/blob/3718d3940ca5890016a884fca9d7b95fc3a379b2/packages/curve-ui-kit/src/themes/typography/create-typography.ts#L63)
```suggestion
```

packages/curve-ui-kit/src/widgets/Footer/Footer.tsx
```diff
@@ -0,0 +1,56 @@
...
+      position: 'relative',
+      paddingInline: SizesAndSpaces.Spacing.lg,
+      paddingBlock: SizesAndSpaces.Spacing.xl,
+      backgroundColor: t.design.Layer[3].Fill,
+    })}
+  >
+    <Grid
+      size={{
+        mobile: 12,
+        tablet: 12,
```
DanielSchiavini, line 34: No need to repeat if it's the same.. This prop `mobile: 12` can be read as `>mobile` ([mobile breakpoint is 0](https://github.com/curvefi/curve-frontend/blob/88281be8d15553b2718323305d3c3445a80f6656/packages/curve-ui-kit/src/themes/basic-theme/basic.ts#L9))

```suggestion
```

packages/curve-ui-kit/src/widgets/Footer/Link.tsx
```diff
@@ -0,0 +1,56 @@
...
+        fontFamily: Fonts['Mona Sans'],
+        fontSize: SizesAndSpaces.FontSize.md,
+        // In dark mode the bolder font is very hard to read.
+        fontWeight: (t) =>
+          t.palette.mode === 'dark' ? SizesAndSpaces.FontWeight.Normal : SizesAndSpaces.FontWeight.Medium,
+      }}
+    >
+      {label}
+    </Typography>
+  </LinkMui>
```
DanielSchiavini, line 55: I think we can just use the mui button with ghost color and the startIcon prop

```suggestion
  <Button
    component={MuiLink}
    href={href}
    target="_blank"
    rel="noreferrer"
    size="small"
    color="ghost"
    startIcon={icon}
  >
      {label}
  </Button>
```

packages/curve-ui-kit/src/widgets/Footer/Link.tsx
```diff
@@ -0,0 +1,56 @@
...
+    <Typography
+      variant="ghost"
+      sx={{
+        display: 'flex',
+        alignItems: 'center',
+        height: '1.5rem',
+        fontFamily: Fonts['Mona Sans'],
+        fontSize: SizesAndSpaces.FontSize.md,
+        // In dark mode the bolder font is very hard to read.
+        fontWeight: (t) =>
```
DanielSchiavini, line 49: We should never customize fontSize, fontFamily and fontWeight, all customization should be done in [the typography variants](https://github.com/curvefi/curve-frontend/blob/3718d3940ca5890016a884fca9d7b95fc3a379b2/packages/curve-ui-kit/src/themes/typography/create-typography.ts#L53)

packages/curve-ui-kit/src/widgets/Footer/Section.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const Section = ({ title, links }: SectionProps) => (
+  <Grid container spacing={1}>
+    <Grid size={12}>
+      <Typography
+        variant="headingXsBold"
+        sx={{
+          display: 'flex',
+          alignItems: 'center',
+          height: '2.5rem',
```
DanielSchiavini, line 19:
```suggestion
          height: SizesAndSpaces.ButtonSize.sm,
```

packages/curve-ui-kit/src/widgets/Footer/Sections.tsx
```diff
@@ -0,0 +1,118 @@
...
+import BugReportOutlinedIcon from '@mui/icons-material/BugReportOutlined'
+
+import { XIcon } from 'curve-ui-kit/src/shared/ui/XIcon'
+import { DiscordIcon } from 'curve-ui-kit/src/shared/ui/DiscordIcon'
+import { LAFIcon } from 'curve-ui-kit/src/shared/ui/LAFIcon'
+import { DuneIcon } from 'curve-ui-kit/src/shared/ui/DuneIcon'
+import { CrvHubIcon } from 'curve-ui-kit/src/shared/ui/CrvHubIcon'
+import { SectionProps } from './Section'
+
+export const sections: SectionProps[] = [
```
DanielSchiavini, line 17: missing translations (which means we cannot create these in the module level

packages/curve-ui-kit/src/widgets/Footer/Description.tsx
```diff
@@ -0,0 +1,57 @@
...
+    >
+      Curve DAO is building the software that powers the future world economy: decentralised, trustless, inclusive and
+      autonomous.
```
DanielSchiavini, line 54: We should add translation tags

```suggestion
      {t`Curve DAO is building the software that powers the future world economy: decentralised, trustless, inclusive and autonomous.`}
```

DanielSchiavini (commented):
It looks great! Thanks

packages/curve-ui-kit/src/widgets/Footer/Description.tsx
0xAlunara: I took the new header logo component as inspiration indeed, but had to make slight modifications as it's slightly different according to the Figma design. Up to @0xtutti if we actually want a carbon copy of said logo.

packages/curve-ui-kit/src/shared/ui/CrvHubIcon.tsx
0xAlunara: I put them in `curve-ui-kit\src\shared\icons` for now, together with `DodoIcon.tsx`.

packages/curve-ui-kit/src/widgets/Footer/Description.tsx
0xAlunara: It's been removed

packages/curve-ui-kit/src/widgets/Footer/Footer.tsx
0xAlunara: It's been removed

packages/curve-ui-kit/src/widgets/Footer/Link.tsx
0xAlunara: I gave it a try but it opens a new can of worms with respect to default button style and behavior like hovers that make it look nothing like the Figma reference. Undoing that styling wouldn't make it any easier, so I've resorted back to using a MUI Link. I have however been able to remove some styling.

https://github.com/user-attachments/assets/dfed90fc-6f9c-4796-bfb5-7fb50493895d

packages/curve-ui-kit/src/widgets/Footer/Link.tsx
0xAlunara: Setting the variant to `bodyMRegular` made these unnecessary, so they've been removed.

packages/curve-ui-kit/src/widgets/Footer/Section.tsx
0xAlunara: I'm using `SizesAndSpaces.ButtonSize.sm` for the height now in `Section.tsx`. In similar vain I'm using `height: SizesAndSpaces.ButtonSize.xs` instead of 1.5rem for `Link.tsx`.

packages/curve-ui-kit/src/widgets/Footer/Description.tsx
0xAlunara: I've added `import { t } from '@lingui/macro'` and changed it into what you've suggested.

packages/curve-ui-kit/src/widgets/Footer/Footer.tsx
```diff
@@ -0,0 +1,55 @@
...
+    {sections.map((section, i) => (
+      <Grid
+        key={i}
+        size={{
+          mobile: 12,
+          tablet: 4,
+          desktop: 3,
+        }}
+      >
+        <Section key={section.title} {...section} />
```
DanielSchiavini, line 49: We only need the key in the item generated from the loop. Better to avoid the index if we can, so react can still work well if the order changes (not the case here I guess)

```suggestion
        key={section.title}
        size={{
          mobile: 12,
          tablet: 4,
          desktop: 3,
        }}
      >
        <Section {...section} />
```

packages/curve-ui-kit/src/widgets/Footer/Link.tsx
```diff
@@ -0,0 +1,47 @@
...
+  >
+    <Box
+      sx={{
+        '& svg': {
+          width: '100%',
+          height: '100%',
+        },
+      }}
+    >
+      {icon ? icon : null}
```
DanielSchiavini, line 33: I don't think there's a point in using `?` and `: null` unless `icon` can be `0`
```suggestion
      {icon}
```

packages/curve-ui-kit/src/widgets/Footer/Link.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { SizesAndSpaces } from 'curve-ui-kit/src/themes/design/1_sizes_spaces'
+
+export type LinkProps = {
+  label: string
+  href: string
+  icon?: ReactNode
+}
+
+export const Link = ({ label, href, icon }: LinkProps) => (
+  <LinkMui
```
DanielSchiavini, line 14: I see your response in [this comment](https://github.com/curvefi/curve-frontend/pull/500#discussion_r1872260759) however, I think it should follow the design system (yeah that video seems correct according to that)

packages/curve-ui-kit/src/widgets/Footer/Sections.tsx
```diff
@@ -0,0 +1,118 @@
...
+        href: 'https://t.me/crvrussianchat',
+        icon: <TelegramIcon />,
+      },
+    ],
+  },
+  {
+    title: 'Documentation',
+    links: [
+      {
+        label: 'News',
```
DanielSchiavini, line 62: how about the translations here?

packages/curve-ui-kit/src/widgets/Footer/Section.tsx
```diff
@@ -0,0 +1,34 @@
...
+          alignItems: 'center',
+          height: SizesAndSpaces.ButtonSize.sm,
+        }}
+      >
+        {title}
+      </Typography>
+    </Grid>
+
+    {links.map((link, i) => (
+      <Grid key={i} size={{ mobile: 6, tablet: 12, desktop: 12 }}>
```
DanielSchiavini, line 29:
```suggestion
      <Grid key={i} size={{ mobile: 6, desktop: 12 }}>
```

packages/curve-ui-kit/src/widgets/Footer/Link.tsx
0xAlunara: As discussed in chat, will need tutti opinion. Can be done in diff PR.

packages/curve-ui-kit/src/widgets/Footer/Section.tsx
0xAlunara: Heh, actually had to remove `desktop: 12` so that it's still 12 in tablet as well.

packages/curve-ui-kit/src/widgets/Footer/Sections.tsx
0xAlunara: Added

packages/curve-ui-kit/src/widgets/Footer/Footer.tsx
0xAlunara: Should be fixed in https://github.com/curvefi/curve-frontend/pull/500/commits/17db9684935a0e9b4f8a356860f9eb1fdff0dc50

packages/curve-ui-kit/src/widgets/Footer/Sections.tsx
```diff
@@ -0,0 +1,120 @@
...
+import BugReportOutlinedIcon from '@mui/icons-material/BugReportOutlined'
+
+import { XIcon } from 'curve-ui-kit/src/shared/icons/XIcon'
+import { DiscordIcon } from 'curve-ui-kit/src/shared/icons/DiscordIcon'
+import { LAFIcon } from 'curve-ui-kit/src/shared/icons/LAFIcon'
+import { DuneIcon } from 'curve-ui-kit/src/shared/icons/DuneIcon'
+import { CrvHubIcon } from 'curve-ui-kit/src/shared/icons/CrvHubIcon'
+import { SectionProps } from './Section'
+
+export const sections: SectionProps[] = [
```
DanielSchiavini, line 19: Calling `t` on the module level can give [issues when the user changes their language](https://lingui.dev/ref/macro#using-macros). However, just easily finding `t` references is good enough for now.

DanielSchiavini (dismissed):

packages/curve-ui-kit/src/widgets/Footer/Sections.tsx
0xAlunara: I figured that'd be the case and thought it was good enough as well for the `t` reference finding. But then if we do decide to keeps translations I'm sure we'd forget about this, so I tried to fix it in https://github.com/curvefi/curve-frontend/pull/500/commits/8ba7f676a402493526e93a25e7d3f73be4a7d009. Wasn't able to test it, but when locale changes the footer should re-render and `getSection()` should be execugted again I think.

DanielSchiavini (dismissed):

packages/curve-ui-kit/src/themes/button/mui-button.ts
```diff
@@ -65,7 +66,20 @@ export const defineMuiButton = ({ Button, Text }: DesignSystem): Components['Mui
...
+        variants: [
+          ...Object.entries(colors).map(([color, style]) => ({ props: { color }, style })),
+          {
+            props: {
+              variant: 'link',
+            },
+            style: {
+              textTransform: 'none',
+              padding: 0,
+              height: `${Sizing[400]} !important`,
```
DanielSchiavini, line 78: We should document the `!important` hack
```suggestion
              height: `${Sizing[400]} !important`, // unfortunately the sizes below are more specific
```

packages/curve-ui-kit/src/themes/design/1_sizes_spaces.ts
```diff
@@ -306,4 +306,5 @@ export const SizesAndSpaces = {
     minHeight: '300px',
   },
   OutlineWidth: '0.125rem', // 2px
+  FooterWidth: '96rem', // 1536px
```
DanielSchiavini, line 309: we should make the header consistent too. I used the [default mui `<Container>`](https://github.com/curvefi/curve-frontend/blob/5545c375e9bcf0f5d46c39d7645728a90b59d698/packages/curve-common/src/widgets/Header/DesktopHeader.tsx#L37) which [follows `var(--width)`](https://github.com/curvefi/curve-frontend/blob/da6bf3f7d877cb6357e0c2a5c6819f76c46247cb/packages/curve-ui-kit/src/themes/components.ts#L40), currently mapped to [1440px](https://github.com/curvefi/curve-frontend/blob/9caa81d15b97dc4486d42e340c0a9674967713bb/packages/ui/src/styles/layout.css#L24)

packages/curve-ui-kit/src/widgets/Footer/Description.tsx
```diff
@@ -0,0 +1,47 @@
...
+import Typography from '@mui/material/Typography'
+
+import { RCLogoSM } from 'ui/src/images'
+
+export const Description = () => (
+  <Box
+    sx={{
+      display: 'flex',
+      flexDirection: 'column',
+      mr: '1.5rem',
```
DanielSchiavini, line 13: We can use the theme spacing.  `theme.spacing(6) == 1.5rem`
```suggestion
      marginRight: 6,
```

packages/curve-ui-kit/src/widgets/Footer/Description.tsx
```diff
@@ -0,0 +1,47 @@
...
+      href="/"
+      sx={{
+        display: 'flex',
+        alignItems: 'center',
+        textDecoration: 'none',
+      }}
+    >
+      <RCLogoSM
+        style={{
+          width: '48px',
```
DanielSchiavini, line 26: we should use ButtonSize.lg or IconSize.xxl

packages/curve-ui-kit/src/themes/design/1_sizes_spaces.ts
DanielSchiavini: as discussed with @0xtutti we should update `--width` to 96rem

DanielSchiavini (approved):

bout3fiddy (approved):

PR #501, opened 2024-12-05
chore: add dao to cypress

DanielSchiavini:
- adds the new dao application to the cypress tests


OnlyJousting (approved):

tests/cypress/e2e/all/header.cy.ts
```diff
@@ -92,4 +91,9 @@ describe('Header', () => {
       })
     })
   })
+
+  function waitIsLoaded() {
```
0xAlunara, line 95: Works for now but it's a bit weird and random to either select a proposal title or connect button to determine if the page is loaded. Perhaps there's a better way somehow?

0xAlunara (approved):

PR #503, opened 2024-12-05
task: added refresh proposal data call after execute proposal action

OnlyJousting:
Changes:
- Added api call to refresh proposal data after successful execute proposal action

### DIFF
--- a/apps/dao/src/store/createProposalsSlice.ts
+++ b/apps/dao/src/store/createProposalsSlice.ts
@@ -477,19 +477,50 @@ const createProposalsSlice = (set: SetState<State>, get: GetState<State>): Propo
 
           await helpers.waitForTransaction(voteResponseHash, provider)
 
-          get()[sliceKey].setStateByKey('executeTxMapper', {
-            ...get()[sliceKey].executeTxMapper,
-            [voteIdKey]: {
-              status: 'SUCCESS',
-              hash: voteResponseHash,
-              txLink: networks[1].scanTxPath(voteResponseHash),
-              error: null,
-            },
-          })
-
           dismissDeploying()
           const successNotificationMessage = t`Proposal executed successfully!`
           notifyNotification(successNotificationMessage, 'success', 15000)
+
+          // update proposal executed status
+          const response = await fetch(
+            `https://prices.curve.fi/v1/dao/proposals/details/${voteType.toLowerCase()}/${voteId}?tx_hash=${voteResponseHash}`,
+          )
+          const data: PricesProposalResponse = await response.json()
+
+          if ('detail' in data) {
+            return
+          }
+
+          const formattedData = {
+            ...data,
+            votes_for: +data.votes_for / 1e18,
+            votes_against: +data.votes_against / 1e18,
+            support_required: +data.support_required / 1e18,
+            min_accept_quorum: +data.min_accept_quorum / 1e18,
+            total_supply: +data.total_supply / 1e18,
+            creator_voting_power: +data.creator_voting_power / 1e18,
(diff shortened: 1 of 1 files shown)


apps/dao/src/store/createProposalsSlice.ts
```diff
@@ -477,19 +477,50 @@ const createProposalsSlice = (set: SetState<State>, get: GetState<State>): Propo
...
+
+          // update proposal executed status
+          const response = await fetch(
+            `https://prices.curve.fi/v1/dao/proposals/details/${voteType.toLowerCase()}/${voteId}?tx_hash=${voteResponseHash}`,
+          )
+          const data: PricesProposalResponse = await response.json()
+
+          if ('detail' in data) {
+            return
+          }
```
DanielSchiavini, line 492: What does this mean? Maybe a comment would be nice

OnlyJousting: sure, when the is no data the api returns an object with only a "detail" field

apps/dao/src/store/createProposalsSlice.ts
```diff
@@ -477,19 +477,50 @@ const createProposalsSlice = (set: SetState<State>, get: GetState<State>): Propo
...
-              error: null,
-            },
-          })
-
           dismissDeploying()
           const successNotificationMessage = t`Proposal executed successfully!`
           notifyNotification(successNotificationMessage, 'success', 15000)
+
+          // update proposal executed status
+          const response = await fetch(
```
0xAlunara, line 488: Can't you just call `getProposal()` again to fetch fresh data? It feels as if this is code duplication at this point. Alternatively, just add an optional `tx_hash` parameter to `getProposal()`.

0xAlunara (changes requested):

apps/dao/src/store/createProposalsSlice.ts
OnlyJousting: I could, I don't want to initiate a loading state so this runs in the background without triggering any spinners but that can also be optionalised ofc

OnlyJousting: Yea, this is a good suggestion. I wanted to do a silent refetch of data but can do it within the getProposal method.

0xAlunara (dismissed):

apps/dao/src/store/createProposalsSlice.ts
```diff
@@ -158,17 +158,18 @@ const createProposalsSlice = (set: SetState<State>, get: GetState<State>): Propo
...
-    getProposal: async (voteId: number, voteType: ProposalType, silentFetch = false) => {
+    getProposal: async (voteId: number, voteType: ProposalType, silentFetch = false, txHash?: string) => {
       if (!silentFetch) {
         get()[sliceKey].setStateByKey('proposalLoadingState', 'LOADING')
       }
 
       try {
         const proposal = await fetch(
-          `https://prices.curve.fi/v1/dao/proposals/details/${voteType.toLowerCase()}/${voteId}`,
+          `https://prices.curve.fi/v1/dao/proposals/details/${voteType.toLowerCase()}/${voteId}${!!txHash ? `?tx_hash=${txHash}` : ''}`,
```
DanielSchiavini, line 168:
```suggestion
          `https://prices.curve.fi/v1/dao/proposals/details/${voteType.toLowerCase()}/${voteId}${txHash ? `?tx_hash=${txHash}` : ''}`,
```

DanielSchiavini (approved):

0xAlunara (approved):

PR #508, opened 2024-12-06
Styling/remove chad mode button font size rule

OnlyJousting:
Changes:
- Removed chad mode specific button component font-size rule
- Removed chad mode specific pool page tab name for Withdraw/Claim

### DIFF
--- a/packages/ui/src/styles/theme-chad.css
+++ b/packages/ui/src/styles/theme-chad.css
@@ -54,7 +54,6 @@ body.theme-chad {
   /* Buttons */
   --button--color: var(--text-600);
   --button--background-color: var(--primary-700);
-  --button--font-size: small;
   --button--shadow-color: var(--text-600);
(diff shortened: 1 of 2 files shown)


packages/ui/src/styles/theme-chad.css
```diff
@@ -54,7 +54,6 @@ body.theme-chad {
   /* Buttons */
   --button--color: var(--text-600);
   --button--background-color: var(--primary-700);
-  --button--font-size: small;
```
DanielSchiavini, line 57: Won't this leave some undefined references?

DanielSchiavini (approved):

packages/ui/src/styles/theme-chad.css
OnlyJousting: it will use the rule from the default theme

michwill (approved):

PR #515, opened 2024-12-11
feat: add new metric widget

0xAlunara:
* New Metric widget as per Figma, to be used in for example the user profile. Started working on this as part of that feature, but spinned off into a new PR for earlier discussion & merging.
* I've decided not to implement the progress bar and icon yet as there's no need for them yet. I'd rather implement them when a specific use case arises.
* The `abbr` and `suffix` functions are number helper functions I've decided to use instead of `formatNumber`, because of the way we make them stand out in the UI with slightly different styling. I'd rather not dwell too much on the specifics; decimal number harmonization can become a tedious discussion.
* I couldn't decide between one prop for object vs multiple properties, like one `notionalConfig` prop vs multiple like `notionalAbbreviate` and `notionalDecimals`. I've chosen for a flat structure to make using the component easier as many are optional. The unit property however is an object property, as all of them are required when you want a custom unit.
* You can find many examples in the Storybook. There's no cypress tests as the new widget will not yet be used anywhere on the website.

image

### DIFF
--- a/packages/curve-ui-kit/src/shared/ui/Metric.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/Metric.tsx
@@ -13,6 +13,7 @@ import { TypographyVariantKey, TYPOGRAPHY_VARIANTS } from 'curve-ui-kit/src/them
 import { abbreviateNumber, scaleSuffix } from 'curve-ui-kit/src/utils'
 
 const { Spacing } = SizesAndSpaces
+const COPY_ALERT_DURATION = 6000
 
 // Correspond to flexbox align items values.
 export const ALIGNMENTS = ['start', 'center', 'end'] as const
@@ -223,7 +224,7 @@ export const Metric = ({
         </Typography>
       )}
 
-      <Snackbar open={openCopyAlert} autoHideDuration={6000} onClose={() => setOpenCopyAlert(false)}>
+      <Snackbar open={openCopyAlert} autoHideDuration={COPY_ALERT_DURATION} onClose={() => setOpenCopyAlert(false)}>
         <Alert variant="filled" severity="success">
           Copied metric value: {value}
         </Alert>
--- a/packages/curve-ui-kit/src/shared/ui/Metric.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/Metric.tsx
@@ -1,5 +1,6 @@
 import Box from '@mui/material/Box'
 import Skeleton from '@mui/material/Skeleton'
+import Tooltip from '@mui/material/Tooltip'
 import Typography from '@mui/material/Typography'
 
 import { SizesAndSpaces } from 'curve-ui-kit/src/themes/design/1_sizes_spaces'
@@ -143,28 +144,32 @@ export const Metric = ({
           }}
         />
       ) : (
-        <Box display="flex" alignItems="baseline" gap={Spacing.xxs}>
-          {unit?.position === 'prefix' && (
-            <Typography variant={MetricSize[size]} color="textSecondary">
-              {unit.symbol}
-            </Typography>
-          )}
+        <Box display="flex" gap={Spacing.xs} alignItems="baseline">
+          <Tooltip arrow title={value.toLocaleString()}>
+            <Box display="flex" gap={Spacing.xxs} alignItems="baseline">
+              {unit?.position === 'prefix' && (
+                <Typography variant={MetricSize[size]} color="textSecondary">
+                  {unit.symbol}
+                </Typography>
+              )}
 
-          <Typography variant={MetricSize[size]} color="textPrimary">
-            {formatter(abbreviate ? abbreviateNumber(value) : value)}
-          </Typography>
+              <Typography variant={MetricSize[size]} color="textPrimary">
+                {formatter(abbreviate ? abbreviateNumber(value) : value)}
+              </Typography>
 
-          {abbreviate && (
-            <Typography variant={MetricSize[size]} color="textPrimary" textTransform="capitalize">
-              {scaleSuffix(value)}
-            </Typography>
-          )}
+              {abbreviate && (
+                <Typography variant={MetricSize[size]} color="textPrimary" textTransform="capitalize">
+                  {scaleSuffix(value)}
+                </Typography>
+              )}
 
-          {unit?.position === 'suffix' && (
-            <Typography variant={MetricSize[size]} color="textSecondary">
-              {unit.symbol}
-            </Typography>
-          )}
+              {unit?.position === 'suffix' && (
+                <Typography variant={MetricSize[size]} color="textSecondary">
+                  {unit.symbol}
+                </Typography>
+              )}
+            </Box>
+          </Tooltip>
 
           {(change || change === 0) && (
             <Typography
--- a/packages/curve-ui-kit/src/shared/ui/Metric.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/Metric.tsx
@@ -151,7 +151,7 @@ export const Metric = ({
         {tooltip && (
           <Tooltip arrow placement="top" title={tooltip}>
             <span>
-              &nbsp;
+              {' '}
               <InfoOutlinedIcon sx={{ fontSize: '1.25em' }} />
             </span>
           </Tooltip>
--- a/packages/curve-ui-kit/src/shared/ui/Metric.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/Metric.tsx
@@ -149,7 +149,7 @@ export const Metric = ({
       <Typography variant="bodyXsRegular" color="textTertiary">
         {label}
         {tooltip && (
-          <Tooltip title={tooltip}>
+          <Tooltip arrow placement="top" title={tooltip}>
             <span>
               &nbsp;
               <InfoOutlinedIcon sx={{ fontSize: '1.25em' }} />
@@ -170,6 +170,7 @@ export const Metric = ({
         <Box display="flex" gap={Spacing.xs} alignItems="baseline">
           <Tooltip
             arrow
+            placement="bottom"
             title={value.toLocaleString()}
             onClick={copyValue}
             sx={{
--- a/packages/curve-ui-kit/src/shared/ui/Metric.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/Metric.tsx
@@ -6,6 +6,7 @@ import Skeleton from '@mui/material/Skeleton'
 import Snackbar from '@mui/material/Snackbar'
 import Tooltip from '@mui/material/Tooltip'
 import Typography from '@mui/material/Typography'
+import InfoOutlinedIcon from '@mui/icons-material/InfoOutlined'
 
 import { SizesAndSpaces } from 'curve-ui-kit/src/themes/design/1_sizes_spaces'
 import { TypographyVariantKey, TYPOGRAPHY_VARIANTS } from 'curve-ui-kit/src/themes/typography'
@@ -79,6 +80,8 @@ const formatChange = (value: number) => {
 type Props = {
   /** Label that goes above the value */
   label: string
+  /** Optional tooltip content shown next to the label */
+  tooltip?: string
 
   /** The actual metric value to display */
   value: number
@@ -108,6 +111,7 @@ type Props = {
 
 export const Metric = ({
   label,
+  tooltip,
 
   value,
   formatter = (value: number) => formatValue(value, decimals),
@@ -144,6 +148,14 @@ export const Metric = ({
     <Box display="flex" flexDirection="column" alignItems={alignment} gap={Spacing.xs}>
       <Typography variant="bodyXsRegular" color="textTertiary">
         {label}
+        {tooltip && (
+          <Tooltip title={tooltip}>
+            <span>
+              &nbsp;
+              <InfoOutlinedIcon sx={{ fontSize: '1.25em' }} />
+            </span>
+          </Tooltip>
+        )}
       </Typography>
 
       {loading ? (
--- a/packages/curve-ui-kit/src/shared/ui/Metric.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/Metric.tsx
@@ -48,20 +48,18 @@ const UNIT_MAP: Record<(typeof UNITS)[number], UnitOptions> = {
 } as const
 
 // Default value formatter.
-const formatValue = (value: number, decimals?: number): string => {
-  return value.toLocaleString(undefined, {
+const formatValue = (value: number, decimals?: number): string =>
+  value.toLocaleString(undefined, {
     minimumFractionDigits: decimals,
     maximumFractionDigits: decimals,
   })
-}
 
 // Default notional value formatter.
-const formatNotionalValue = (value: number, decimals?: number): string => {
-  return value.toLocaleString(undefined, {
+const formatNotionalValue = (value: number, decimals?: number): string =>
+  value.toLocaleString(undefined, {
     minimumFractionDigits: decimals,
     maximumFractionDigits: decimals,
   })
-}
 
 const formatChange = (value: number) => {
   // Looks aesthetically more pleasing without decimals.
--- a/packages/curve-ui-kit/src/shared/ui/Metric.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/Metric.tsx
@@ -1,5 +1,9 @@
+import { useState } from 'react'
+
+import Alert from '@mui/material/Alert'
 import Box from '@mui/material/Box'
 import Skeleton from '@mui/material/Skeleton'
+import Snackbar from '@mui/material/Snackbar'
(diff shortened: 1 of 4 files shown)


packages/curve-ui-kit/src/features/user-profile/UserProfile.tsx
```diff
@@ -0,0 +1,27 @@
...
+  open: boolean
+  onClose: () => void
+  walletAddress: NonNullable<ConnectWalletIndicatorProps['walletAddress']>
+}
+
+export const UserProfile = ({ open, onClose, walletAddress }: Props) => (
+  <Drawer open={open} anchor="right" onClose={onClose}>
+    <Box
+      sx={{
+        minWidth: '500px', // Hardcoded for now as per Figma design
```
DanielSchiavini, line 18: We should use SizesAndSpaces.ModalWidth.lg

packages/curve-ui-kit/src/features/user-profile/home/Header.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { ConnectWalletIndicatorProps } from 'curve-common/src/features/connect-wallet'
+
+type Props = {
+  walletAddress: NonNullable<ConnectWalletIndicatorProps['walletAddress']>
+  onClose: () => void
+}
+
+const LlamaImageSrc = (LlamaImg as unknown as { src: string }).src
+const Llama = styled('img')({
+  height: '40px',
```
DanielSchiavini, line 20: Can't we use some theme variable?

packages/curve-ui-kit/src/features/user-profile/home/Header.tsx
```diff
@@ -0,0 +1,47 @@
...
+})
+
+/**
+ * Gap deviates from Figma as personally 'xs' is too narrow.
+ *
+ * I know there's <AddressLabel> but not sure how to use it with a different typography,
```
DanielSchiavini, line 26: hmm weird one, indeed :thinking:
at least we should extract the substringing to a shared function imo
that AddressLabel is too simple and specific anyway it could be just inlined

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -0,0 +1,195 @@
+import Box from '@mui/material/Box'
+import Typography from '@mui/material/Typography'
+import { SizesAndSpaces } from 'themes/design/1_sizes_spaces'
+import { Fonts } from 'themes/typography/fonts'
```
DanielSchiavini, line 4: Fonts are already defined in the Typography, they should change depending on the theme and the variant

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -0,0 +1,195 @@
...
+type Unit = (typeof UNITS)[number] | UnitOptions
+
+const UNIT_MAP: Record<(typeof UNITS)[number], UnitOptions> = {
+  dollar,
+  percentage,
+} as const
+
+// I don't know if this is the right place or approach for size mapping.
+// Perhaps it's suited somewhere better in the design folder.
+const SizeMapping: Record<Size, string> = {
```
DanielSchiavini, line 41: we could just receive `variant: TypographyVariantKey` instead of a separate mapping

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -0,0 +1,195 @@
...
+  small: 'highlightM',
+  medium: 'highlightL',
+  large: 'highlightXl',
+  extraLarge: 'highlightXxl',
+}
+
+/**
+ * Returns the appropriate unit suffix for a given number value.
```
DanielSchiavini, line 49: I'm amazed if we don't have this somewhere already. Otherwise we should reuse it in many more places

DanielSchiavini (commented):
looks nice!

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -0,0 +1,193 @@
...
+      ) : (
+        <Box display="flex" alignItems="baseline" gap={Spacing.xxs}>
+          {unit?.position === 'prefix' && (
+            <Typography variant={MetricSize[size]} color="textSecondary">
+              {unit.symbol}
+            </Typography>
+          )}
+
+          <Typography variant={MetricSize[size]} color="textPrimary">
+            {formatter(valueAbbr ? abbr(value) : value)}
```
DanielSchiavini, line 156: How about we render them as separate dom elements, for nicer updates/animations :grin: 
```suggestion
            {formatter(valueAbbr ? abbr(value) : value).split('').map((letter, index) => (<span key={index}>{letter}</span>))}
```

packages/curve-ui-kit/src/utils/number.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @example
+ * unit(1500000000000) // Returns 't'
+ * unit(2000000000) // Returns 'b'
+ * unit(3000000) // Returns 'm'
+ * unit(4000) // Returns 'k'
+ * unit(500) // Returns ''
+ * unit(-1000000) // Returns 'm'
+ * unit(0) // Returns ''
+ */
+export function suffix(value: number): string {
```
DanielSchiavini, line 16: This name is too vague. How about `scaleSuffix`

packages/curve-ui-kit/src/utils/number.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+/**
+ * Abbreviates a number such that it can go along with a suffix like k, m, b or t.
+ *
+ * @example
+ * round(1234.5678) // Returns "1.23", goes with suffix "k"
+ * round(1000000) // Returns "1.0", goes with suffix "m"
+ */
+export function abbr(value: number): number {
```
DanielSchiavini, line 30: How about `abbreviateNumber`?

packages/curve-ui-kit/src/utils/number.ts
```diff
@@ -0,0 +1,36 @@
...
+  const exp = Math.floor(Math.log10(Math.abs(value)) / 3) * 3
```
DanielSchiavini, line 31: `Math.floor(Math.log10(Math.abs(value)) / 3) * 3` is repeated multiple times, it should be a separate function and maybe it would be best for performance if we only called it once

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -0,0 +1,193 @@
...
+  alignment?: Alignment
+  loading?: boolean
+}
+
+export const Metric = ({
+  label,
+
+  value,
+  formatter = (value: number) => formatValue(value, decimals),
+  abbreviate,
```
DanielSchiavini, line 111: maybe this should just be a formatter?

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -0,0 +1,193 @@
...
+  percentage,
+} as const
+
+// Default value formatter.
+const formatValue = (value: number, decimals?: number): string => {
+  return value.toLocaleString(undefined, {
+    minimumFractionDigits: decimals,
+    maximumFractionDigits: decimals,
+  })
+}
```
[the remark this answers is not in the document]
0xAlunara: Prettier doesn't support this. We can configure eslint with `"as-needed" enforces no braces where they can be omitted (default)` as per https://eslint.org/docs/latest/rules/arrow-body-style, but because of the fact that we use the same eslint config for different projects it means we have to convert quite a lot of arrow notation usages. I can do this if you want to, but I'd create a new PR for it as it's outside of the scope here.

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -0,0 +1,193 @@
...
+  notionalAbbreviate,
+  notionalUnit,
+  notionalDecimals,
+
+  size = 'medium',
+  alignment = 'start',
+  loading = false,
+}: Props) => {
+  unit = typeof unit === 'string' ? UNIT_MAP[unit as keyof typeof UNIT_MAP] : unit
+  const valueAbbr = abbreviate ?? unit?.abbreviate ?? false
```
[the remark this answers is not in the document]
0xAlunara: Figured out I don't need a new `const` after all and can just re-use the existing ones (since they're primitives)

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -0,0 +1,193 @@
...
+  label: string
+
+  /** The actual metric value to display */
+  value: number
+  /** Optional formatter for metric value */
+  formatter?: (value: number) => string
+  /** If the value should be abbreviated to 1.23k or 3.45m */
+  abbreviate?: boolean
+  /** The number of decimals the value should contain */
+  decimals?: number
```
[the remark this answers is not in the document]
0xAlunara: I think precision is favored for singular numbers like user deposits, but for a collection of numbers consistency is probably better? That's just my personal take on it though.

packages/curve-ui-kit/src/shared/ui/Metric.tsx
0xAlunara: You mean directly the contents of `formatValue` in this case? When I made this it wasn't a one-liner and didn't want to put a whole function definition as a default. Figured this would be a bit more clear? Not sure if you still want me to inline the function here.

packages/curve-ui-kit/src/utils/number.ts
0xAlunara: I've extracted it to a smaller function, but can't really avoid calling it multiple times as otherwise the code becomes unreadable; would rather keep having the original value as its only input.

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  return (
+    <Box display="flex" flexDirection="column" alignItems={alignment} gap={Spacing.xs}>
+      <Typography variant="bodyXsRegular" color="textTertiary">
+        {label}
+        {tooltip && (
+          <Tooltip title={tooltip}>
+            <span>
+              &nbsp;
```
0xAlunara, line 154: I only wanted to add a space when the tooltip was shown, and wanted exactly the size of "one spacebar" which I couldn't replace with margin (1ch is wider than a space). Had to wrap it in a span because Tooltip doesn't accept text as children, and `<></>` would break stuff.

0xAlunara:
The snackbar works, but because of Storybook using frames it only looks correct when you open a specific story and not the overview page

image

0xAlunara:
There's tooltips, but I'm using the default styling for now. If you want to style it I'd rather want to create a separate PR for that at a later stage

image
image

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -0,0 +1,233 @@
...
+}: Props) => {
+  unit = typeof unit === 'string' ? UNIT_MAP[unit] : unit
+  abbreviate ??= unit?.abbreviate ?? false
+
+  notionalUnit = typeof notionalUnit === 'string' ? UNIT_MAP[notionalUnit] : notionalUnit
+  notionalAbbreviate ??= notionalUnit?.abbreviate ?? false
+
+  const [openCopyAlert, setOpenCopyAlert] = useState(false)
+
+  const copyValue = () => {
```
0xAlunara, line 142: Was debating on using `useCallback`, but Claude told me it's unnecessary given it has no dependencies and isn't passed on to child components.

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -0,0 +1,193 @@
...
+          {unit?.position === 'suffix' && (
+            <Typography variant={MetricSize[size]} color="textSecondary">
+              {unit.symbol}
+            </Typography>
+          )}
+
+          {(change || change === 0) && (
+            <Typography
+              variant="highlightM"
+              color={change > 0 ? 'success' : change < 0 ? 'error' : 'textHighlight'}
```
DanielSchiavini, line 208: nb. there is a problem with the feedback colors: https://www.notion.so/Fix-feedback-colors-15b599aae06480b4a40ac99c6c163726?pvs=4

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -0,0 +1,193 @@
...
+              sx={{ marginInlineStart: Spacing.xs }} // Not part of Figma, but my own addition
```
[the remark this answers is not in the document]
DanielSchiavini: Maybe the right place is in a figma comment. We can leave it there and merge this one

packages/curve-ui-kit/src/shared/ui/Metric.tsx
DanielSchiavini: I think this comment is related to what you mentioned regarding the prop flattening.
Basically `formatter`, `abbreviate`, `unit` and `decimals` are 4 variables defining how to convert a number to string. Same for the notional value.
While flexibility is nice, it creates many different ways to use this component which can lead to inconsistency.
IMO we should have only one way to display dollars, amounts, percentages, etc
This is not a blocker to merge, though! We can always improve/change later.

packages/curve-ui-kit/src/shared/ui/Metric.tsx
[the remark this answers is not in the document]
DanielSchiavini: yeah interesting take. I don't know how this will be used so we can leave it for now. It might still be worth to support precision in this component (when we need it?)

packages/curve-ui-kit/src/shared/ui/Metric.tsx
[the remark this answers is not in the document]
DanielSchiavini: yeah 100% out of scope. I like the single-line functions but maybe not worth the effort/errors

packages/curve-ui-kit/src/shared/ui/Metric.tsx
DanielSchiavini: This might be a bit clearer
```suggestion
              {' '}
```

DanielSchiavini (dismissed):

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -0,0 +1,233 @@
...
+      {notional !== undefined && (
+        <Typography variant="highlightXsNotional" color="textTertiary">
+          {notionalUnit?.position === 'prefix' && notionalUnit.symbol}
+          {notionalFormatter(notionalAbbreviate ? abbreviateNumber(notional) : notional)}
+          {notionalAbbreviate && scaleSuffix(notional)}
+          {notionalUnit?.position === 'suffix' && notionalUnit.symbol}
+        </Typography>
+      )}
+
+      <Snackbar open={openCopyAlert} autoHideDuration={6000} onClose={() => setOpenCopyAlert(false)}>
```
DanielSchiavini, line 226: the `autoHideDuration` shouldn't be hardcoded. Let's either use a constant or configure this in `components.ts` like `muiSnackbar: {defaultProps: {autoHideDuration: 6000}}`

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -0,0 +1,233 @@
...
+        <Alert variant="filled" severity="success">
+          Copied metric value: {value}
```
DanielSchiavini, line 228: I just noticed the font family seems broken here, because we don't use the global `CssBaseline` (as that could conflict with the existing code). We might need to wrap this in `Typography` or configure the `Alert`'s `styleOverrides`

DanielSchiavini: btw, do we need to add the actual value to the snackbar?

0xAlunara: Debatable, I figured it's nice as an extra confirmation to see what you've just copied to the clipboard. Might be information overload, I don't have a strong opinion on this matter. @0xtutti ?

packages/curve-ui-kit/src/shared/ui/Metric.tsx
0xAlunara: I've decided to create a const variable on top called `COPY_ALERT_DURATION`. I don't like the idea of setting it default to 6000ms for every snackbar, as we might also use it in the future for more important web3 transaction related alerts which you don't want to hide away as quick.

Was thinking about putting it in `0_primitives.ts` but wasn't sure if that's the right place, so just put it here as it's only related to this specific component for now.

DanielSchiavini (approved):

0xtutti (approved):

PR #517, opened 2024-12-11
fix: warning icon visibility

0xAlunara:
@0xtutti I've been struggling with making it visible with the yellow primary color, so I've settled with the red from error. This way it's way more in all themes compared to the warning's yellow. I've tried to find a different icon or add a black outline/stroke but wasn't able to find one that looks good.

@DanielSchiavini I've added a hardcoded gap of 0.125 to make it look nicer; I can't be arsed making it responsive with the mui theme given that this will prob get redesigned in the future anyway. In fact, the whole TooltipAlert probably gets replaced some time in the future. Consider this PR a quick fix until we get there.

https://mui.com/material-ui/material-icons/?query=warning

image

### DIFF
--- a/packages/ui/src/Tooltip/TooltipAlert.tsx
+++ b/packages/ui/src/Tooltip/TooltipAlert.tsx
@@ -1,31 +1,24 @@
+import WarningOutlinedIcon from '@mui/icons-material/WarningOutlined'
+
 import type { AlertType } from 'ui/src/AlertBox/types'
 import type { TooltipProps } from 'ui/src/Tooltip/types'
 
 import * as React from 'react'
-import Icon from 'ui/src/Icon'
 import IconTooltip from 'ui/src/Tooltip/TooltipIcon'
 
 const TooltipAlert = ({
   alertType,
+  isDeprecated,
   ...props
 }: React.PropsWithChildren<
   TooltipProps & {
     alertType: AlertType
+    isDeprecated?: boolean
   }
 >) => {
-  return (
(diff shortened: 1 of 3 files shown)


packages/ui/src/Tooltip/TooltipAlert.tsx
```diff
@@ -1,31 +1,24 @@
...
 import type { AlertType } from 'ui/src/AlertBox/types'
 import type { TooltipProps } from 'ui/src/Tooltip/types'
 
 import * as React from 'react'
-import Icon from 'ui/src/Icon'
 import IconTooltip from 'ui/src/Tooltip/TooltipIcon'
 
 const TooltipAlert = ({
   alertType,
+  isDeprecated,
```
DanielSchiavini, line 11: is this set anywhere?

DanielSchiavini: oh I see it was set already [here](https://github.com/curvefi/curve-frontend/blob/9209fb035661b3205539c4c7ac293794ee01ae7c/apps/loan/src/hooks/useCollateralAlert.tsx#L13). That comparison with '' also looks fishy

0xAlunara: Yeah the comparison is for mapping the default of nothing to the existing default of 'info'. The other mapping is from 'danger' to 'error' as there's only three colors we want to use: `info | warning | error`.

DanielSchiavini (approved):

bout3fiddy (approved):

PR #521, opened 2024-12-13
chore: more precision for borrowed amount

DanielSchiavini:


apps/lend/src/hooks/useVaultShares.ts
```diff
@@ -24,7 +24,7 @@ function useVaultShares(rChainId: ChainId, rOwmId: string, vaultShares: string |
       const borrowedAmtUsd = +pricePerShare * +vaultShares * +usdRate
 
       return {
-        borrowedAmount: `${formatNumber(borrowedAmt, { showDecimalIfSmallNumberOnly: true })} ${symbol}`,
+        borrowedAmount: `${formatNumber(borrowedAmt, { minimumFractionDigits: 2, maximumFractionDigits: 2 })} ${symbol}`,
```
DanielSchiavini, line 27:
```suggestion
        borrowedAmount: `${formatNumber(borrowedAmt, { minimumSignificantDigits: 6, maximumSignificantDigits: 6 })} ${symbol}`,
```

michwill (approved):

apps/lend/src/hooks/useVaultShares.ts
0xAlunara: With it being non-perfect, is it 'good enough' or are we in need for a better solution? If it's 'good enough' can you quickly add a comment explaining this edge case where it's behaving slightly different than expected?

0xAlunara (changes requested):

apps/lend/src/hooks/useVaultShares.ts
DanielSchiavini: I was talking about the suggested change, not the committed one. I think the committed change is good enough for now.
We can improve it further later

0xAlunara (approved):

PR #522, opened 2024-12-13
fix: app links reset network

DanielSchiavini:
- Add the current network to the cross-app urls
- Add tests


apps/lend/src/layout/Header.tsx
```diff
@@ -32,7 +32,7 @@ const Header = ({ chainId, sections }: HeaderProps) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [footerHeight])
 
-  const { rLocalePathname } = getParamsFromUrl()
+  const { rLocalePathname, rNetwork } = getParamsFromUrl()
```
0xAlunara, line 35: You're using `getNetworkFromUrl()` in `dao` and `loan` apps, but `getParamsFromUrl()` in `lend` and `main` apps. I was expecting it to be the same for all 4. Probably still works and might be something to fix once we harmonize all apps into a single layout.

EDIT: ah nvm I see `getParamsFromUrl()` calls `getNetworkFromUrl()` and some more.

apps/main/src/layout/default/Header.tsx
```diff
@@ -57,7 +58,7 @@ export const Header = ({ sections }: HeaderProps) => {
         () =>
           _parseRouteAndIsActive(
             [
-              ...(hasRouter && (!network || network?.showRouterSwap)
+              ...(hasRouter && (!network || network.showRouterSwap)
```
0xAlunara, line 61: Still puzzles me how the linter doesn't catch these needless null checks.

apps/main/src/layout/default/Header.tsx
```diff
@@ -69,7 +70,7 @@ export const Header = ({ sections }: HeaderProps) => {
             ],
             rLocalePathname,
             routerPathname,
-            routerNetwork,
+            rNetwork,
           ),
         [hasRouter, network, networks, rChainId, rLocalePathname, routerCached, routerNetwork, routerPathname],
```
0xAlunara, line 75: It's a bit confusing how there's now both `rNetwork` from the url and `routerNetwork` also from the url. Can this be harmonized / simplified?

0xAlunara (commented):
Really can't wait to merge the apps so there's hopefully less router state hacks required.

It works when navigating between `dex` and `lend` app, but once you go to `crvusd` or `dao` it basically loses whatever the network was and defaults back to `ethereum`, probably because these only live on mainnet and there's no network switcher there.

If that is acceptable that's fine, although I'm not sure that was the intent given you've also changed Header.tsx in both the `dao` and `loan` app.

apps/main/src/layout/default/Header.tsx
DanielSchiavini: This actually was not working with routerNetwork

It is something to dig into

apps/main/src/layout/default/Header.tsx
DanielSchiavini: Well this was not bad just redundant

0xAlunara (approved):

PR #527, opened 2024-12-15
fix: open footer integrations link as new tab incorrectly opens swap page

0xAlunara:
This fix makes the footer use router links when the href doesn't start with HTTP like in the header/nav. This does not fix the issue of opening 'integrations' in a new tab opening the swap page, but that's a different issue.


packages/curve-ui-kit/src/widgets/Footer/Sections.tsx
```diff
@@ -75,7 +75,7 @@ export const getSections = (): SectionProps[] => [
       },
       {
         label: t`Integrations`,
-        href: '#/integrations',
+        href: window.location.origin + '#/integrations',
```
DanielSchiavini, line 78: this might crash on server-side renders (no window)
this will also ignore the react router and force a full page refresh

DanielSchiavini (changes requested):

packages/curve-ui-kit/src/widgets/Footer/Sections.tsx
```diff
@@ -75,7 +75,8 @@ export const getSections = (): SectionProps[] => [
       },
       {
         label: t`Integrations`,
-        href: '#/integrations',
+        // Would've loved to have used ROUTE.PAGE_INTEGRATIONS, but they differ per app.
+        href: '/integrations',
         icon: <IntegrationsIcon />,
         target: '_self',
```
DanielSchiavini, line 81: Unrelated, but seems pretty useless
```suggestion
```

packages/curve-ui-kit/src/widgets/Footer/Link.tsx
```diff
@@ -9,7 +11,16 @@ export type LinkProps = {
 }
 
 export const Link = ({ label, href, icon, target = '_blank' }: LinkProps) => (
-  <Button component="a" href={href} target={target} rel="noreferrer" color="ghost" variant="link" startIcon={icon}>
+  <Button
+    component={href.startsWith('http') ? LinkMui : RouterLink}
+    to={href}
+    href={href}
```
DanielSchiavini, line 17: doesn't this generate errors in the console, as the invalid `to` prop will be passed to the `a` element?
might be worth to implement a separate component with the 'http' logic

packages/curve-ui-kit/src/widgets/Footer/Sections.tsx
```diff
@@ -75,7 +75,8 @@ export const getSections = (): SectionProps[] => [
       },
       {
         label: t`Integrations`,
-        href: '#/integrations',
+        // Would've loved to have used ROUTE.PAGE_INTEGRATIONS, but they differ per app.
```
DanielSchiavini, line 78: Can we somehow make them the same per app?
Currently, this is losing the selected network and switching to eth (at least for DEX).
It might be necessary to pass the `integrationUrl` or `networkName` to the `getSections` function

[Screencast from 18-12-24 09:21:29.webm](https://github.com/user-attachments/assets/8fd65578-956a-4a4f-aa0f-ead1658eb81d)

DanielSchiavini (changes requested):

packages/curve-ui-kit/src/widgets/Footer/Sections.tsx
0xAlunara: Remnant from the past; removed

packages/curve-ui-kit/src/widgets/Footer/Link.tsx
0xAlunara: Didn't give an error but I'll replace it with spreading property objects depending on the conditional

packages/curve-ui-kit/src/widgets/Footer/Sections.tsx
0xAlunara: I couldn't really get the `selectedApp` from the header as it's set there and couldn't pass it to the footer easily, so for now I've settled with drilling down `networkName`. If present in the link together with a non http link, it'll use that `networkName` in the generated url. It's almost like `externalAppUrl` from `PageTabs.tsx`, but without the app. Getting tired of this issue lol.

0xAlunara:
I also fixed the integrations link in the sidebars, except for the DAO app, because that app doesn't have an intgerations link at all. Need to create a task for that perhaps.

0xtutti (approved):

PR #529, opened 2024-12-16
feat: responsive typography

DanielSchiavini:


packages/curve-ui-kit/src/themes/design/1_sizes_spaces.ts
```diff
@@ -255,12 +255,12 @@ const MappedLineHeight = {
 
 const MappedTypography = {
   FontSize: {
-    xs: Sizing[125],
-    sm: Sizing[150],
-    md: Sizing[200],
-    lg: Sizing[250],
-    xl: Sizing[350],
-    xxl: Sizing[500],
+    xs: { mobile: Sizing[125], tablet: Sizing[125], desktop: Sizing[125] },
```
0xAlunara, line 258: Newline spacing differs from the rest but really not a blocking issue

packages/curve-ui-kit/src/themes/design/1_sizes_spaces.ts
```diff
@@ -255,12 +255,12 @@ const MappedLineHeight = {
 
 const MappedTypography = {
```
0xAlunara, line 256: Verified with Style Guide ✔️

packages/curve-ui-kit/src/themes/design/1_sizes_spaces.ts
```diff
@@ -272,12 +272,36 @@ const MappedTypography = {
     Extra_Bold: 800,
   },
   LineHeight: {
-    xs: '0.75rem', // 12px
-    sm: '0.875rem', // 14px
-    md: '1rem', // 16px
-    lg: '1.5rem', // 24px
-    xl: '2rem', // 32px
-    xxl: '2.5rem', // 40px
+    xs: {
```
0xAlunara, line 275: Verified with Style Guide ✔️

tests/cypress/e2e/all/header.cy.ts
```diff
@@ -102,7 +102,7 @@ describe('Header', () => {
     cy.get(`[data-testid='chain-icon-${eth}']`).should('be.visible')
     cy.get(`[data-testid='btn-change-chain']`).click()
     cy.get(`[data-testid='menu-item-chain-${arbitrum}']`).click()
-    cy.get(`[data-testid^='menu-item-chain-']`).should('not.be.visible')
+    cy.get(`[data-testid^='menu-item-chain-']`).should('not.exist')
```
0xAlunara, line 105: Heh I think you've also fixed this in your network switching PR. Can stay, all fine.

packages/curve-ui-kit/src/themes/typography/mui-typography.ts
```diff
@@ -1,22 +1,45 @@
...
+  tableCellL: 'span',
+  tableCellMRegular: 'span',
+  tableCellMBold: 'span',
+  tableCellSRegular: 'span',
+  tableCellSBold: 'span',
+
+  highlightXsNotional: 'em',
+  highlightXs: 'em',
+  highlightS: 'em',
+  highlightM: 'strong',
```
0xAlunara, line 34: Are you sure about using `em` and `strong` here? There's two issues:

1. `xs` and `xsnotional` are now also italic, because by default browsers apply italic styling to `<em>` tags. 
2. If you *do* decide to keep `strong`, you no longer need `fontWeight: 'Bold'` in `TYPOGRAPHY_VARIANTS` as browser apply a `font-weight: bold` by default to `<strong>` elements.

image

packages/curve-ui-kit/src/themes/typography/mui-typography.ts
```diff
@@ -1,22 +1,45 @@
...
+  bodySRegular: 'p',
+  bodySBold: 'p',
+  bodyXsRegular: 'p',
+  bodyXsBold: 'p',
+
+  buttonXs: 'span',
+  buttonS: 'span',
+  buttonM: 'span',
+
+  tableHeaderM: 'dfn',
```
0xAlunara, line 23: Similarly to the issue with highlight font styling, browser apply italic styling by default to `<dfn>` elements, which is not what should happen to according to the Figma design.

packages/curve-ui-kit/src/widgets/Footer/Description.tsx
```diff
@@ -39,7 +39,7 @@ export const Description = () => (
         color="textPrimary"
         sx={{
           fontSize: Sizing[600], // This is a one-off specifically for the logo text
-          fontWeight: 'bold',
+          fontWeight: 'bold !important',
```
0xAlunara, line 42: This doesn't appear to be necessary, `bold` works fine without the `!important`.

packages/curve-ui-kit/src/widgets/Footer/Description.tsx
```diff
@@ -39,7 +39,7 @@ export const Description = () => (
         color="textPrimary"
         sx={{
           fontSize: Sizing[600], // This is a one-off specifically for the logo text
```
0xAlunara, line 41: This does need adjustment now though. Because you can't easily apply `!important` to this and that's a dirty global hack anyway, you can use the self-referencing `&` operator to increase specificity:

```ts
'&': {
  fontSize: Sizing[600], // This is a one-off specifically for the logo text
},
```

packages/curve-common/src/features/switch-advanced-mode/AdvancedModeSwitcher.tsx
```diff
@@ -24,7 +24,7 @@ export const AdvancedModeSwitcher: FunctionComponent<AdvancedModeSwitcherProps>
       <Typography
         variant="headingXsBold"
         display="inline-block"
-        sx={{ marginLeft: 2, marginRight: 4, lineHeight: '37px' }}
+        sx={{ marginLeft: 2, marginRight: 4, lineHeight: '37px !important' }}
```
0xAlunara, line 27: `'&': { lineHeight: '37px' }`

packages/curve-common/src/widgets/Header/HeaderLogo.tsx
```diff
@@ -27,7 +27,11 @@ export const HeaderLogo = ({ appName, isLite, sx }: HeaderLogoProps) => (
     {appName ? (
       <Box display="inline-flex" flexDirection="column">
         {/* Note: Anti-pattern to change the line height, however we want the subtitle to fit nicely */}
-        <Typography variant="headingSBold" sx={{ lineHeight: '1.2rem', textTransform: 'none' }} color="textPrimary">
+        <Typography
+          variant="headingSBold"
+          sx={{ lineHeight: '1.2rem !important', textTransform: 'none' }}
```
0xAlunara, line 32: `'&': { lineHeight: '1.2rem' }`

0xAlunara (changes requested):
In general `!important` should be avoided if possible as it's a global hammer that screws everything over; Mui seems to recommend the self-referencing `&` operator instead to increase specificity. So instead of `fontSize: '1rem !important'` we should use `'&': { fontSize: '1rem' }`. You can increase specificity by increase the amount of `&`, like `'&&&&': { fontSize: '1rem' }`

Because typography is now reactive it might be that some `sx={{ fontSize: 'foo'}}` have broken and need the `&` specificity workaround, but it's hard to determine which.

packages/curve-ui-kit/src/widgets/Footer/Description.tsx
DanielSchiavini: yeah I changed the wrong field it was supposed to be on the fontSize :fearful:

packages/curve-ui-kit/src/themes/typography/mui-typography.ts
DanielSchiavini: yeah I was thinking about semantic HTML but this is a PITA. I'll use `<p>`

0xAlunara (approved):

0xtutti (approved):

PR #535, opened 2024-12-17
chore: enable tanstack devtools when running locally

0xAlunara:
I think we can enable TanStack Devtools again; not sure why it was disabled in the first place in fd3f719333b6d405defefaccbe6b0bf3852eb01e as I find it very helpful, and in production builds it gets completely excluded by default (it only gets included if `process.env.NODE_ENV === 'development'`)

Because Cypress tests run with development mode enabled, I've added an extra check to not use the devtools in case Cypress is running. This is because the devtool button's positioning may break some tests. It's not really a big deal as in production the devtools get omitted anyway.

### DIFF
--- a/packages/ui/src/QueryProvider/index.tsx
+++ b/packages/ui/src/QueryProvider/index.tsx
@@ -1,6 +1,7 @@
 import { type QueryClient, QueryClientProvider } from '@tanstack/react-query'
 import { type Persister, PersistQueryClientProvider } from '@tanstack/react-query-persist-client'
-import React, { ReactNode } from 'react'
+import { ReactQueryDevtools } from '@tanstack/react-query-devtools'
+import { ReactNode } from 'react'
 
 type QueryProviderWrapperProps = {
   children: ReactNode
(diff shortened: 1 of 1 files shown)


DanielSchiavini:
I found this plugin slow and not very useful
It's fine to include if you wanna use it

I think it's nice to run cypress in dev mode because the errors get obfuscated in production builds

packages/ui/src/QueryProvider/index.tsx
```diff
@@ -9,13 +10,23 @@ type QueryProviderWrapperProps = {
 }
 
 export function QueryProvider({ children, persister, queryClient }: QueryProviderWrapperProps) {
-  // children = (<>{children}<ReactQueryDevtools buttonPosition="bottom-left" /></>)
+  // Cypress runs in dev mode which means the devtools button is visible, which may break tests.
+  const isCypress = typeof window !== 'undefined' && (window as any).Cypress
```
DanielSchiavini, line 14: there's no reason to keep this in the function, it can be defined globally

0xAlunara: old vue habits die hard

DanielSchiavini (approved):

0xtutti (approved):

PR #537, opened 2024-12-18
fix: show banner when failing to init curveJS

DanielSchiavini:
- show banner when failing to init curveJS
  - move the banner to under the header, as the header is now sticky so the banner would be hidden behind it
  - properly propagate curveJS errors so they are caught and converted to network errors

image
image
image


0xAlunara (approved):

0xtutti (dismissed):

packages/curve-common/src/widgets/Header/types.ts
```diff
@@ -35,6 +36,8 @@ export type BaseHeaderProps<TChainId = number> = {
   isLite?: boolean
   ChainProps: ChainSwitcherProps<TChainId>
   WalletProps: ConnectWalletIndicatorProps
+  BannerProps: GlobalBannerProps
+  bannerHeight: number | undefined
```
0xAlunara, line 40: Couldn't this be part of BannerProps? I know it's for cypress testing, but it's still banner related.

packages/curve-common/src/widgets/Header/DesktopHeader.tsx
```diff
@@ -31,6 +34,8 @@ export const DesktopHeader = <TChainId extends number>({
   return (
     <>
       <AppBar color="transparent" ref={mainNavRef}>
+        <GlobalBanner {...BannerProps} />
```
0xAlunara, line 37: What's your opinion on using slots for this instead of tightly coupling `Banner.tsx` to `DesktopHeader.tsx`? Instead of passing `BannerProps` and importing, coupling and using a different component we could just have a `ReactNode` as use it directly. The user would then be able to pass whatever he wants as a banner component and whatever props are necessary.

Reference: https://www.dhiwise.com/post/mastering-react-slot-patterns-for-flexible-ui-components

0xAlunara (approved):
Approved and non-blocking, but I do have some questions.

packages/curve-common/src/widgets/Header/types.ts
DanielSchiavini: this is to adjust the mobile drawer and the desktop header so they go under the banner. It's not passed to the banner, but it could be part of the object

packages/curve-common/src/widgets/Header/DesktopHeader.tsx
DanielSchiavini: yeah it's also a good idea, but cannot be combined with the previous comment. This prop isn't adding much

DanielSchiavini: How about we convert the ChainProps, WalletProps and BannerProps to slots?

DanielSchiavini: Oh too bad, that doesn't work in this case :thinking: 
https://github.com/curvefi/curve-frontend/blob/fcef114d408b568e3980fe78d56370f1073f47e6/packages/curve-common/src/widgets/Header/MobileHeader.tsx#L112

michwill (approved):

PR #544, opened 2024-12-19
refactor: arrow-style-linting

0xAlunara:


DanielSchiavini (dismissed):
Nice! Do we need to add eslint to the [format scripts](https://github.com/curvefi/curve-frontend/blob/main/package.json#L18-L19)? So it also runs in the [ci job](https://github.com/curvefi/curve-frontend/blob/main/.github/workflows/ci.yaml#L16).

0xAlunara:
> Nice! Do we need to add eslint to the [format scripts](https://github.com/curvefi/curve-frontend/blob/main/package.json#L18-L19)? So it also runs in the [ci job](https://github.com/curvefi/curve-frontend/blob/main/.github/workflows/ci.yaml#L16).

I don't think that's needed as formatting with prettier is a different tool and modifies code instead of just analyzing it.

0xtutti (approved):

Macket (approved):

PR #545, opened 2024-12-19
styling: copy metric alert outline variant

0xAlunara:
Did not yet decide to change snackbar position, given dani is probably working no calculating the correct top offset in another PR.


DanielSchiavini (approved):

DanielSchiavini:
> Did not yet decide to change snackbar position, given dani is probably working no calculating the correct top offset in another PR.

From #552: https://github.com/curvefi/curve-frontend/blob/de90615ac4eca32d862acd3785d740c918bebe34/packages/curve-common/src/widgets/Header/Header.tsx#L17

0xAlunara:
Closed in favor of the upcoming PR that'll tackle various Snackbar alert variants

PR #548, opened 2024-12-21
feat: add new risk disclaimers page

0xAlunara:
This one took waaayyy longer than I hoped. Very simple concept, not much logic per se, but hard to get right with the huge amount of text, headers, titles, paragraphs, lists and links in between paragraphs. At first I made components with properties like `<Section title="foo" paragraphs={["bar", "baz"]}>` but then rewrote it to be more like
```tsx
<Section>
  <Title>foo</Title>
  <Paragraph>bar</Paragraph>
  <Paragraph>baz</Paragraph>
</Section>
```
which is probably a bit more modula and easier to work with.

* Adds new /disclaimer endpoint with new disclaimer tabs
* Added link to footer which updates open tab depending on app and current URL
* Completely removes the old disclaimer code and page

image

I've taken the liberty here and there to increase the gap / spacing a bit because imo a lot of text was crumpled a bit too much together.

I've made a beginning with the header, but decided to omit it from this PR. The 'issue' is that the page specific header lives outside of the Next page component. We first need to build a proper way to control that component from somewhere deep in the page. Perhaps with a Teleport / Portal.

image


apps/lend/src/pages/index.tsx
```diff
@@ -35,7 +35,7 @@ const App: NextPage = () => {
       {SubRoutes}
       <Route path=":locale">{SubRoutes}</Route>
       <Route path="/markets/*" element={<Navigate to={`/ethereum${ROUTE.PAGE_MARKETS}`} replace />} />
-      <Route path="/risk-disclaimer" element={<Navigate to={`/ethereum${ROUTE.PAGE_RISK_DISCLAIMER}`} replace />} />
```
DanielSchiavini, line 38: do we need to set up a redirect for old links living elsewhere?

apps/main/src/layout/default/index.tsx
```diff
@@ -72,6 +72,7 @@ const getSections = (locale: Locale, network: string) => [
       { route: 'https://news.curve.fi/', label: t`News` },
       { route: 'https://resources.curve.fi/lending/understanding-lending/', label: t`User Resources` },
       { route: 'https://docs.curve.fi', label: t`Developer Resources` },
+      { route: `${network ? `/${network}` : ''}${ROUTE.PAGE_DISCLAIMER}`, label: t`Risk Disclaimers` },
```
DanielSchiavini, line 75: unrelated to this PR, but this is so ugly :cry:

packages/curve-ui-kit/src/widgets/Disclaimer/Disclaimer.tsx
```diff
@@ -0,0 +1,73 @@
...
+    const params = new URLSearchParams(window.location.hash.split('?')[1])
+    const tabId = params.get('tab')
+    return tabs.find((tab) => tab.id === tabId)?.value ?? DEFAULT_TAB_INDEX
+  }
+
+  const handleTabChange = (newTabIndex: number) => {
+    setTabIndex(newTabIndex)
+    const tabId = tabs.find((tab) => tab.value === newTabIndex)?.id
+    const currentHash = window.location.hash.split('?')[0]
+    window.location.hash = `${currentHash}?tab=${tabId}`
```
DanielSchiavini, line 47: Might be safer to use `navigate` in case we depart from the hash router (which would be great).
For reading the URL `useLocation`

packages/curve-ui-kit/src/widgets/Disclaimer/Disclaimer.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+const DEFAULT_TAB_INDEX = 0
+const tabs = [
+  { id: 'dex', label: t`Dex` },
+  { id: 'lend', label: t`LlamaLend` },
+  { id: 'crvusd', label: t`crvUSD` },
+  { id: 'scrvusd', label: t`Savings crvUSD` },
+].map((tab, value) => ({
+  ...tab,
+  value,
```
DanielSchiavini, line 27: this is confusing, why do we need a value here? `index` would be clearer
we could avoid the indexes completely and just use the ID
```
const DEFAULT_TAB = 'dex'
const tabs = {
  dex: t`Dex`,
  ......
}

File: packages/curve-ui-kit/src/widgets/Disclaimer/HeaderPage.tsx [hunk 5]
```diff
@@ -0,0 +1,55 @@
...
+
+export const HeaderPage = () => {
+  const router = useRouter()
+
+  const [canGoBack, setCanGoBack] = useState(false)
+  // Check if window.history has previous entries from our app
+  // TODO: this doesn't work properly yet.
+  useEffect(() => {
+    setCanGoBack(window.history.length > 1 && document.referrer.includes(window.location.host))
+  }, [])
```
DanielSchiavini, line 23: useMemo is simpler here
```suggestion
  const canGoBack = useMemo(() => window.history.length > 1 && document.referrer.includes(window.location.host), [])
```

packages/curve-ui-kit/src/widgets/Disclaimer/Section.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import { SizesAndSpaces } from 'curve-ui-kit/src/themes/design/1_sizes_spaces'
+
+const { Spacing } = SizesAndSpaces
+
+export const Header = ({ children }: { children?: ReactNode }) => (
+  <Stack
+    justifyContent="end"
+    sx={{
+      minHeight: '56px', // TODO: place as a constant somewhere.
```
DanielSchiavini, line 14: Can't we use `lineHeight` or `padding` here?

packages/curve-ui-kit/src/widgets/Disclaimer/Section.tsx
```diff
@@ -0,0 +1,55 @@
...
+export const Paragraph = ({ children }: { children?: ReactNode }) => (
+  <Typography component="div" variant="bodyMRegular" color="textSecondary">
+    {children}
+  </Typography>
+)
+
+export const Section = ({ children }: { children?: ReactNode }) => {
+  const childArray = Children.toArray(children)
+  const title = childArray.find((child) => typeof child === 'object' && 'type' in child && child.type === Title)
+  const paragraphs = childArray.filter((child) => typeof child === 'object' && 'type' in child && child.type !== Title)
```
DanielSchiavini, line 38: this children magic isn't ideal. We could use a `title` slot for the title element then just use `children` for the rest

0xAlunara: I couldn't decide. I don't think it's ideal, but deviating from the rest of the style where I use children using a property again looks weird (I had it as a prop before). I really miss named slots like Vue in React. At least the tabs look aesthetically pleasing and easier to modify when it's all slots instead of properties; looks a bit like shadcn 😅 . Even Claude suggested I'd use this for composability instead of keeping the prop.

packages/curve-ui-kit/src/widgets/Disclaimer/Disclaimer.tsx
0xAlunara: I had it as `index` before but `TabSwitcher` really wants a `value` prob. So I could use the ids, but I'd still have to rename the property from `id` to `value` as otherwise the compiler will complain

apps/main/src/layout/default/index.tsx
0xAlunara: Do you think using a query parameter for network selection would be better, as it's more a thing about the app state than it is about an actual route? Not that it would completely solve the ugliness, but it might be a tad bit better and prevent some routing problems we have currently.

apps/lend/src/pages/index.tsx
0xAlunara: Oh I just replaced all the old links (I could find) with the new one

apps/dao/src/components/PageDisclaimer/Page.tsx
```diff
@@ -0,0 +1,39 @@
+import type { NextPage } from 'next'
+import { useEffect } from 'react'
+import { t } from '@lingui/macro'
+
```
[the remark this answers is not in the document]
0xAlunara: I'm being too orderly with respect to import grouping. Just personal preference.

packages/curve-ui-kit/src/widgets/Disclaimer/Disclaimer.tsx
DanielSchiavini: How sbout `Object.entries(tabs).map(([value, label] => ....)`

apps/dao/src/components/PageDisclaimer/Page.tsx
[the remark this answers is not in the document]
DanielSchiavini: Ohmy you do imports manually?
You need a real IDE 😂

apps/main/src/layout/default/index.tsx
DanielSchiavini: Well if it's optional it should indeed be in the query or in the end of the route 🤔

packages/curve-ui-kit/src/widgets/Disclaimer/Section.tsx
DanielSchiavini: Ok you though hard about this 👌

packages/curve-ui-kit/src/widgets/Disclaimer/Section.tsx
0xAlunara: Sadly not, as line height ads space before and after (top and bottom); in this case we only want space on top. However, I did replace this with `paddingBlockStart: Spacing.lg` which makes it responsive.

packages/curve-ui-kit/src/widgets/Disclaimer/HeaderPage.tsx
```diff
@@ -15,12 +15,9 @@ const { Spacing, Sizing, MaxWidth } = SizesAndSpaces
...
   const router = useRouter()
 
-  const [canGoBack, setCanGoBack] = useState(false)
-  // Check if window.history has previous entries from our app
-  // TODO: this doesn't work properly yet.
-  useEffect(() => {
-    setCanGoBack(window.history.length > 1 && document.referrer.includes(window.location.host))
-  }, [])
+  // Check if window.history has previous entries from our app.
+  // TODO: this doesn't work properly yet, but also isn't used yet.
```
DanielSchiavini, line 19: why isn't this used? Can we delete it?

packages/curve-ui-kit/src/widgets/Disclaimer/Disclaimer.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+type Props = {
+  className?: string
+}
+
+export const Disclaimer = ({ className }: Props) => {
+  const navigate = useNavigate()
+  const location = useLocation()
+
+  const getTabFromUrl = () => {
```
DanielSchiavini, line 37: This should be  moved to the global scope to avoid the eslint warning (`useEffect has a missing dependency: 'getTabFromUrl'`)

DanielSchiavini (commented):
Can we add a simple cypress test?

packages/curve-ui-kit/src/widgets/Disclaimer/Disclaimer.tsx
0xAlunara: Should we configure lint warnings as errors? Missed this one as it's hidden away as a small step in `yarn lint` on GitHub Actions.

packages/curve-ui-kit/src/widgets/Disclaimer/HeaderPage.tsx
0xAlunara: It's this 3rd bar, the 'page header'. This bar needs a different approach as it has to become part of the layout Header and can't be part of the page itself. We'll need a different task and PR for that. I didn't exactly want to remove it yet as the styling and layout might get re-used, but since it's not being used right now and will need to be remade anyway I've opted to just delete the entire file.

packages/curve-ui-kit/src/widgets/Disclaimer/Disclaimer.tsx
0xAlunara: Moved `getTabFromUrl` outside of the function component. This amde me realize `handleTabChange` is dependent on both `navigate` and `location` and is passed as a callback to `TabSwitcher` so I've wrapped it in a `useCallback` as well.

packages/curve-ui-kit/src/widgets/Disclaimer/Disclaimer.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+type Props = {
+  className?: string
+}
+
+export const Disclaimer = ({ className }: Props) => {
+  const navigate = useNavigate()
+  const location = useLocation()
+
+  const [tab, setTab] = useState(getTabFromUrl(location.search))
```
DanielSchiavini, line 44: it's usually simpler to avoid a state variable and simply take the url as the source-of-truth

tests/cypress/support/ui.ts
```diff
@@ -14,3 +14,5 @@ export const oneTabletViewport = () =>
 
 export const oneMobileOrTabletViewport = () =>
   [randomInt(MIN_WIDTH, DESKTOP_BREAKPOINT), randomInt(MIN_HEIGHT, MAX_HEIGHT)] as const
+
+export const allViewports = () => [oneDesktopViewport(), oneMobileViewport(), oneTabletViewport()] as const
```
DanielSchiavini, line 18: Testing everything with all viewports is generally wasteful, i.e. unnecessarily increasing test times. IMO unless the test has viewport dependencies, testing with one random viewport is good enough to catch edge cases

packages/curve-ui-kit/src/widgets/Disclaimer/Disclaimer.tsx
DanielSchiavini: This is what I did for the table sorting:
https://github.com/curvefi/curve-frontend/blob/c9530c0b13a2e3a63b784f893d5ddffe64f0354d/packages/curve-ui-kit/src/hooks/useSortFromQueryString.ts#L5-L14

tests/cypress/support/ui.ts
DanielSchiavini: This is what I'm using to test the new llamalend tables: https://github.com/curvefi/curve-frontend/blob/c9530c0b13a2e3a63b784f893d5ddffe64f0354d/tests/cypress/support/ui.ts#L19

DanielSchiavini (approved):

packages/curve-ui-kit/src/widgets/Disclaimer/Disclaimer.tsx
0xAlunara: Cheers, good catch, the state is a leftover from when tabs weren't derived from the URL yet.

* I've removed the state variable and introduced `useTabFromQueryString`.
* I've given up on insisting on a tab property called `id`, as `TabsSwitcher` really wants a `value` property that acts as the tab id. Creating a copy of the tabs for its option parameter with a duplicate property felt nasty.
* However, I had to keep `options={TABS.map((x) => x)}` as I can't pass a `const` array to the `options` parameter, and I need the `as const` for correct type inference for the hook's return values. Instead of type safe unions of ids you'd merely get `string`, losing strict id typing.

packages/curve-ui-kit/src/hooks/useTabFromQueryString.ts
```diff
@@ -0,0 +1,49 @@
+import { useCallback } from 'react'
+import { useNavigate, useLocation } from 'react-router-dom'
+
+type Tab = { value: string; label: string }
```
DanielSchiavini, line 4: We could add the component here as well to avoid [repeating all the options](https://github.com/curvefi/curve-frontend/blob/795e0f5163eede1ba0590587498a1333ed0784d2/packages/curve-ui-kit/src/widgets/Disclaimer/Disclaimer.tsx#L47-L50)
```suggestion
type Tab = { value: string; label: string, render: FunctionComponent }
```

DanielSchiavini (approved):

packages/curve-ui-kit/src/hooks/useTabFromQueryString.ts
0xAlunara: That's what I had on the first attempt, but then I realized the TabSwitcher itself doesn't render the component, and having to find the active tab with `.find()` was more cumbersome than just lines of easy code 😅

packages/curve-ui-kit/src/widgets/Disclaimer/Section.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <Stack
+      gap={Spacing.sm}
+      sx={{
+        '& + &': {
```
DanielSchiavini, line 44: This deserves an explaining comment

packages/curve-ui-kit/src/widgets/Disclaimer/Section.tsx
```diff
@@ -0,0 +1,55 @@
...
+export const Paragraph = ({ children }: { children?: ReactNode }) => (
+  <Typography component="div" variant="bodyMRegular" color="textSecondary">
+    {children}
+  </Typography>
+)
+
+export const Section = ({ children }: { children?: ReactNode }) => {
+  const childArray = Children.toArray(children)
+  const title = childArray.find((child) => typeof child === 'object' && 'type' in child && child.type === Title)
+  const paragraphs = childArray.filter((child) => typeof child === 'object' && 'type' in child && child.type !== Title)
```
DanielSchiavini, line 38: What about string children for example?

packages/curve-ui-kit/src/widgets/Disclaimer/Tabs/CrvUsd.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+      <Button
+        component={Link}
+        variant="link"
+        href="https://docs.curve.fi/references/audits/#crvusd-and-lending"
+        target="_blank"
+        endIcon={<ArrowOutwardIcon />}
+        sx={{
+          padding: Spacing.xs,
+          '&': { fontWeight: '500' },
```
DanielSchiavini, line 54: Do we really need to depart from the variant @0xtutti ?

packages/curve-ui-kit/src/themes/tabs/mui-tabs.ts
```diff
@@ -47,6 +47,7 @@ export const defineMuiTabs = ({
   styleOverrides: {
     root: {
       minHeight: 0,
+      ['&.MuiTabs-root [role="tablist"]']: { flexWrap: 'wrap' },
       [`&.${contained} .MuiTab-root`]: tabVariant(Contained),
```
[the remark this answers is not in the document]
0xAlunara: It's been removed in a later commit, was experimenting with overflows but it broke the tab overflows in kther places.

packages/curve-ui-kit/src/widgets/Disclaimer/Section.tsx
0xAlunara: Added a few lines of comments to explain what it does and why it's there

packages/curve-ui-kit/src/widgets/Disclaimer/Section.tsx
0xAlunara: Good question, I've made it such those won't be supported. Only titles and paragraphs should be rendered in sections.

DanielSchiavini (approved):

0xtutti (approved):

OnlyJousting (approved):

PR #552, opened 2024-12-23
feat: tanstack table

DanielSchiavini:
- [x] implement tanstack table with mui components
- [x] make headers sticky in desktop and mobile
- [x] add sorting
- [x] add line graphs 
- [x] add utilization graph
- [x] make table horizontal scrollable on mobile (minwidth)
- [x] cypress tests


packages/curve-ui-kit/src/shared/ui/tanstack-table.overrides.d.ts
```diff
@@ -0,0 +1,7 @@
+import '@tanstack/react-table'
+
+declare module '@tanstack/table-core' {
```
0xAlunara, line 3: Is this still needed or were you testing something? I find it odd you'd need to manually interfere with tanstack react typings.

packages/curve-ui-kit/src/shared/icons/ArrowIcon.tsx
```diff
@@ -0,0 +1,11 @@
+import { createSvgIcon } from '@mui/material/utils'
+
+export const ArrowDownIcon = createSvgIcon(
+  <svg viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <path
+      d="M10 2.66675C10.3682 2.66675 10.6667 2.96522 10.6667 3.33341L10.6667 15.0573L14.8619 10.862C15.1223 10.6017 15.5444 10.6017 15.8047 10.862C16.0651 11.1224 16.0651 11.5445 15.8047 11.8048L10.4714 17.1382C10.3464 17.2632 10.1768 17.3334 10 17.3334C9.82319 17.3334 9.65362 17.2632 9.5286 17.1382L4.19526 11.8048C3.93491 11.5445 3.93491 11.1224 4.19526 10.862C4.45561 10.6017 4.87772 10.6017 5.13807 10.862L9.33333 15.0573L9.33333 3.33341C9.33333 2.96522 9.63181 2.66675 10 2.66675Z"
+      fill="currentColor"
+    />
+  </svg>,
+  'Arrow',
```
0xAlunara, line 10: `ArrowDown`

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
```diff
@@ -0,0 +1,155 @@
...
+const { Sizing, Spacing, MinWidth } = SizesAndSpaces
+
+const getAlignment = <T extends any>({ columnDef }: Column<T>) => (columnDef.meta?.type == 'numeric' ? 'right' : 'left')
+
+const DataCell = <T extends unknown>({ cell }: { cell: Cell<T, unknown> }) => {
+  const column = cell.column
+  return (
+    <Typography
+      variant="tableCellMBold"
+      color="text.primary"
```
0xAlunara, line 34: Not related to this PR, but should we add explicit typing here? Sometimes we use `text.primary`, sometimes `textPrimary`; either works. But there's no auto-complete so it's not consistent across the board and you have to look it up somewhere each time.

EDIT: likewise for the variant

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
```diff
@@ -0,0 +1,155 @@
...
+  Row,
+  SortingState,
+  useReactTable,
+} from '@tanstack/react-table'
+import { useSortFromQueryString } from '../../hooks/useSortFromQueryString'
+import { ArrowDownIcon } from '../icons/ArrowIcon'
+
+const { Sizing, Spacing, MinWidth } = SizesAndSpaces
+
+const getAlignment = <T extends any>({ columnDef }: Column<T>) => (columnDef.meta?.type == 'numeric' ? 'right' : 'left')
```
0xAlunara, line 27: `const getAlignment = ({ columnDef }: Column<any>) => (columnDef.meta?.type == 'numeric' ? 'right' : 'left')`

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+const DataCell = <T extends unknown>({ cell }: { cell: Cell<T, unknown> }) => {
```
0xAlunara, line 29: `const DataCell = ({ cell }: { cell: Cell<any, unknown> }) => {`

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
```diff
@@ -0,0 +1,155 @@
...
+        marginBlock: 0,
+        height: Sizing['3xl'],
+        borderBottom: '1px solid',
+        borderColor: (t) => t.design.Layer[1].Outline,
+      }}
+      ref={ref}
+      // title={JSON.stringify(row.original)}
+      data-testid={`data-table-row-${row.id}`}
+    >
+      {entry?.isIntersecting && row.getVisibleCells().map((cell) => <DataCell key={cell.id} cell={cell} />)}
```
0xAlunara, line 64: This doesn't seem to work? When I decrease the width of my browser such that the cells are out of renderview, the dom elements are still there.

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const sort = column.getIsSorted()
+  return (
+    <TableCell
+      sx={{
+        textAlign: getAlignment(column),
+        alignContent: 'end',
+        padding: Spacing.sm,
+        paddingBlockStart: 0,
+        ...getExtraColumnPadding(column),
+        cursor: 'pointer',
```
0xAlunara, line 89: I think this only makes sense if it's sortable?

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
```diff
@@ -0,0 +1,155 @@
...
+    state: { sorting },
+    onSortingChange,
+    maxMultiSortColCount: 3, // allow 3 columns to be sorted at once
+  })
+  return (
+    <Table sx={{ minWidth: MinWidth.table }} data-testid="data-table">
+      <TableHead
+        sx={(t) => ({
+          zIndex: t.zIndex.appBar - 1,
+          position: 'sticky',
```
0xAlunara, line 134: I wonder how well this works if you have two tables on one page (possible stacked vertically)

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
```diff
@@ -0,0 +1,155 @@
...
+  return (
+    <TableRow
+      sx={{
+        marginBlock: 0,
+        height: Sizing['3xl'],
+        borderBottom: '1px solid',
+        borderColor: (t) => t.design.Layer[1].Outline,
+      }}
+      ref={ref}
+      // title={JSON.stringify(row.original)}
```
0xAlunara, line 61: Can this be removed?

packages/curve-ui-kit/src/hooks/useSortFromQueryString.ts
```diff
@@ -0,0 +1,29 @@
...
+  const sort = new URLSearchParams(search)
+    .getAll('sort')
+    .map((id) => (id.startsWith('-') ? { id: id.slice(1), desc: true } : { id, desc: false }))
+  return sort.length ? sort : defaultSort.map(({ id, desc }) => ({ id: id.replace(/\./g, '_'), desc }))
+}
+
+export function updateSort(search: string, state: SortingState): string {
+  const params = new URLSearchParams(search)
+  params.delete('sort')
+  state.forEach(({ id, desc }) => params.append('sort', `${desc ? '-' : ''}${id}`))
```
0xAlunara, line 26: This works for one column, but for now multi-column sorting will not work.

packages/curve-common/src/widgets/Header/MobileTopBar.tsx
```diff
@@ -21,9 +19,8 @@ export const MobileTopBar = <TChainId extends number>({
   isSidebarOpen,
   toggleSidebar,
   isLite,
-  sx,
 }: MobileTopBarProps<TChainId>) => (
-  <Box display="flex" flexDirection="row" flexGrow={1} paddingX={2} sx={sx}>
+  <Box display="flex" flexDirection="row" flexGrow={1} paddingX={2}>
```
0xAlunara, line 23: Works, but could use `<Stack>`

packages/curve-common/src/widgets/Header/Header.tsx
```diff
@@ -3,8 +3,21 @@ import { DesktopHeader } from './DesktopHeader'
...
+
+/**
+ * Helper function to calculate the header height based on the banner height and the current screen size
+ */
+export const useHeaderHeight = (bannerHeight: number | undefined) => {
+  const isDesktop = useMediaQuery(isDesktopQuery, { noSsr: true })
+  const theme = useTheme()
+  return isDesktop
+    ? `calc(96px + ${bannerHeight ?? 0}px)` // note: hardcoded height is tested in cypress
+    : `calc(2 * ${theme.spacing(3)} + ${DEFAULT_BAR_SIZE} + ${bannerHeight}px)`
```
0xAlunara, line 22: The times two feels a bit hardcoded and I'm unsure what it's for.

apps/loan/src/components/PageLlamaMarkets/LendingMarketsTable.tsx
```diff
@@ -0,0 +1,77 @@
...
+  data: LendingVault[]
+  headerHeight: string
+}) => (
+  <Stack
+    sx={{
+      marginBlockStart: Spacing.xl,
+      marginBlockEnd: Spacing.xxl,
+      maxWidth: MaxWidth.table,
+      minWidth: MinWidth.table,
+      width: '100%',
```
0xAlunara, line 60: Can be removed I think

apps/loan/src/components/PageLlamaMarkets/cells/LineGraphCell.tsx
```diff
@@ -0,0 +1,58 @@
...
+    contractAddress: vault.controllerAddress,
+  })
+  const { design } = useTheme()
+  const value = vault.rates[`${type}ApyPcent`]
+  if (value == null) {
+    return '-'
+  }
+
+  return (
+    <Box
```
0xAlunara, line 32: `<Stack>`?

apps/loan/src/components/PageLlamaMarkets/cells/LineGraphCell.tsx
```diff
@@ -0,0 +1,58 @@
+import { LendingSnapshot, useLendingSnapshots } from '@/entities/lending'
+import { LendingVault } from '@/entities/vaults'
+import { Line, LineChart } from 'recharts'
+import { useTheme } from '@mui/material/styles'
+import { DesignSystem } from '@ui-kit/themes/design'
+import Box from '@mui/material/Box'
+import Skeleton from '@mui/material/Skeleton'
+
+const graphSize = { width: 172, height: 48 }
```
0xAlunara, line 9: Would be nice if this could be made responsive and / or determined by the table that uses this cell instead of having the size hardcoded in the cell itself.

apps/loan/src/components/PageLlamaMarkets/cells/UtilizationCell.tsx
```diff
@@ -1,4 +1,18 @@
...
+export const UtilizationCell = ({ getValue }: CellContext<LendingVault, LendingVault['utilizationPercent']>) => {
+  const value = getValue()
+  if (value == null) {
+    return '-'
+  }
+  return (
+    <>
+      {value.toFixed(2) + '%'}
+      <Box sx={(t) => ({ border: `1px solid ${t.design.Color.Primary[600]}`, height: 16 })}>
+        <Box sx={(t) => ({ backgroundColor: `${t.design.Color.Primary[500]}`, width: `${value}%`, height: '100%' })} />
```
0xAlunara, line 14: Double `<Box>` confuses me, is it really necessary?

apps/loan/src/components/PageLlamaMarkets/cells/PoolTitleCell.tsx
```diff
@@ -1,20 +1,27 @@
...
     <Box display="flex">
       <TokenIcons
-        // todo: is imageBaseUrl dependent on the chain? It should come from the API then.
-        imageBaseUrl="https://cdn.jsdelivr.net/gh/curvefi/curve-assets/images/assets/"
+        imageBaseUrl={`https://cdn.jsdelivr.net/gh/curvefi/${urlPath}`}
         tokens={coins.map((c) => c.symbol)}
         tokenAddresses={coins.map((c) => c.address)}
       />
-      <Typography variant="tableCellL">{coins.map((coin) => coin.symbol).join(' - ')}</Typography>
+      <Typography variant="tableCellL" sx={{ paddingInlineStart: Spacing.sm }}>
```
0xAlunara, line 22: Should we consider the extra padding to be part of the tableCellL variant for Typography, so that it becomes part of the default styling?

apps/loan/src/entities/lending.ts
```diff
@@ -29,23 +31,34 @@ type LendingSnapshotFromApi = {
...
+export type LendingSnapshot = LendingSnapshotFromApi
+
 type LendingSnapshotsFromApi = {
   chain: string
   market_id: number
-  data: LendingSnapshotFromApi[]
+  data: LendingSnapshot[]
 }
 
+const getSupportedChains = memoize(async () => {
```
0xAlunara, line 42: Not important now but ought to use tanstack for this some day as this fetch query is rather slow and can take a few hundred milliseconds. I know it's used in the query below but would rather have it be its own standalone query.

apps/loan/src/entities/lending.ts
```diff
@@ -29,23 +31,34 @@ type LendingSnapshotFromApi = {
...
 export const { useQuery: useLendingSnapshots } = queryFactory({
   queryKey: (params: ContractParams) => [...rootKeys.contract(params), 'lendingSnapshots'] as const,
-  queryFn: async ({ blockchainId, contractAddress }: ContractQuery): Promise<LendingSnapshotsFromApi> => {
-    // call with units=none it returns 4h points
-    const url = `https://prices.curve.fi/v1/lending/markets/${blockchainId}/${contractAddress}/snapshots?units=none`
+  queryFn: async ({ blockchainId, contractAddress }: ContractQuery): Promise<LendingSnapshot[]> => {
+    const chains = await getSupportedChains()
+    if (!chains.includes(blockchainId)) return [] // backend gives 404 for optimism
+
+    // const start = Math.floor(Date.now() - TIME_FRAMES.WEEK / 1000)
```
0xAlunara, line 54: Can remove this if it's not used

0xAlunara (changes requested):

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
DanielSchiavini: It is lol

packages/curve-common/src/widgets/Header/Header.tsx
DanielSchiavini: Aha good that you say. It's for up and down margins

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
DanielSchiavini: Actually unknown or never should be used not any

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
DanielSchiavini: The types for typography seem to be a bit screwed up. Variants work fine for e.g. in buttons giving errors because of the d.ts overrides

packages/curve-ui-kit/src/shared/ui/tanstack-table.overrides.d.ts
DanielSchiavini: Yeah this came from their docs actually. It's the way to customize the metadata type

apps/loan/src/components/PageLlamaMarkets/cells/LineGraphCell.tsx
DanielSchiavini: Yeah good point. We should discuss with tutti

apps/loan/src/components/PageLlamaMarkets/cells/UtilizationCell.tsx
DanielSchiavini: One is for the border, the one inside is the filled part, say 50% width for 50% utilization

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
0xAlunara: I tried both `never` and `unknown` but it opens up another can of worms sadly

apps/loan/src/components/PageLlamaMarkets/cells/UtilizationCell.tsx
0xAlunara: I present you, modern CSS:
```tsx
<Box
  sx={(t) => ({
    height: 16,
    border: `1px solid ${t.design.Color.Primary[600]}`,
    background: `linear-gradient(to right, ${t.design.Color.Primary[500]} ${value}%, transparent ${value / 2}%)`,
  })}
/>
```

0xAlunara: I dislike how the percentage is also not part of the progress bar itself. I'll ask @0xtutti about it in Slack, but you can try:
```tsx
    <Box
      sx={(t) => ({
        height: 24,
        border: `1px solid ${t.design.Color.Primary[600]}`,
        background: `linear-gradient(to right, ${t.design.Color.Primary[500]} ${value}%, transparent ${value / 2}%)`,
        display: 'flex',
        alignItems: 'center',
        justifyContent: 'center',
        color: 'white',
        textShadow: '0 0 4px rgba(0,0,0,0.8)',
      })}
    >
      {value.toFixed(2) + '%'}
    </Box>
```
image
image
image

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
0xAlunara: I've created a PR to fix typography variant typing in https://github.com/curvefi/curve-frontend/pull/557

DanielSchiavini: thanks!!

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
DanielSchiavini: I think the correct is `<T extends unknown>({ columnDef }: Column<T>)` but I kind of agree `({ columnDef }: Column<any>)` is simpler

DanielSchiavini: I'd still like to argue for the more complex signature because any will just disable any type checks. Unknown forces us to check when using the values

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
DanielSchiavini: As I said above this would disable typescript checks

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
DanielSchiavini: I think it should work fine, but if we ever need it we can customize

DanielSchiavini: [Screencast from 30-12-24 11:32:44.webm](https://github.com/user-attachments/assets/5a173e3a-2ec4-4f1b-b0b5-b489b81a99ba)

packages/curve-ui-kit/src/hooks/useSortFromQueryString.ts
DanielSchiavini: Multi-column sorting works fine, you need to hold the shift when clicking on the 2nd column though (default tanstack behavior). For example I get `/#/ethereum/beta-markets?sort=-rates_borrowApyPcent&sort=rates_lendApyPcent`

apps/loan/src/components/PageLlamaMarkets/cells/UtilizationCell.tsx
DanielSchiavini: not sure the linear-gradient solution is simpler, but I like it :nerd_face:

DanielSchiavini: for the alternative design lets wait tutti

apps/loan/src/components/PageLlamaMarkets/cells/PoolTitleCell.tsx
DanielSchiavini: I don't think so, this is just to separate the text from the icons. I'll change to use flex gap instead

DanielSchiavini: Also will move the typography to the column meta

apps/loan/src/entities/lending.ts
DanielSchiavini: I'll add a todo. I can fix it in another PR

packages/curve-ui-kit/src/hooks/useSortFromQueryString.ts
0xAlunara: Huh, I wonder who else knows about this...

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const sort = column.getIsSorted()
+  return (
+    <TableCell
+      sx={{
+        textAlign: getAlignment(column),
+        alignContent: 'end',
+        padding: Spacing.sm,
+        paddingBlockStart: 0,
+        ...getExtraColumnPadding(column),
+        ...(column.columnDef.enableSorting && {
```
0xAlunara, line 90: This doesn't work, because if `enableSorting` isn't set the default behavior is indeed to be `true`, but the actual value of the property is `undefined` thus this is false. Use `column.getCanSort()` instead.

packages/curve-ui-kit/src/hooks/useSortFromQueryString.ts
DanielSchiavini: yeah it's not obvious, we'd need to provide better user support. But I also dont like the idea to enable multisort by default

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
```diff
@@ -0,0 +1,158 @@
...
+          cursor: 'pointer',
+        }),
+      }}
+      colSpan={header.colSpan}
+      width={header.getSize()}
+      onClick={column.getToggleSortingHandler()}
+      data-testid={`data-table-header-${column.id}`}
+    >
+      <Typography variant="tableHeaderS" color={`text.${sort ? 'primary' : 'secondary'}`}>
```
0xAlunara, line 99: Not part of this PR perhaps, but
```tsx
      <Typography
        variant="tableHeaderS"
        color={`text.${sort ? 'primary' : 'secondary'}`}
        sx={{
          ...(column.getCanSort() && {
            '&:hover': {
              color: 'text.primary',
            },
          }),
        }}
      >
```
would add some nice highlighting when you hover clickable table header cells. This isn't the right place, probably better suited in the default style configuration of header cells. It's a "nice to have".

0xAlunara (changes requested):

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
DanielSchiavini: yeah it should have been enableSorting != false but getCanSort is better indeed

0xAlunara (approved):

OnlyJousting (approved):

0xAlunara (approved):

PR #553, opened 2024-12-23
refactor: merge curve-common and curve-lib into curve-ui-kit

0xAlunara:
Draft PR to see if something fails on Vercel (which seems to be the case)

* I've put most of `curve-lib/shared` into `curve-ui-kit/lib`, some of the root utils have been put into `curve-ui-kit/utils`
* Moved web3-onboard, tanstack, ethers, safe-sdk and babel package dependencies to curve-ui-kit's `package.json`.


0xAlunara:
Closed in favor of the smaller https://github.com/curvefi/curve-frontend/pull/556

PR #557, opened 2024-12-27
fix: broken typography variant typing

0xAlunara:
I've managed to fix the typing for typography variants. However, I've not been able to fix the typing for typography colors, I think it needs a different approach given that the colors can be used for much more than just colors.

### DIFF
--- a/packages/curve-ui-kit/src/themes/typography/override-typography.d.ts
+++ b/packages/curve-ui-kit/src/themes/typography/override-typography.d.ts
@@ -1,22 +1,14 @@
-declare module 'global' {
-  type TypographyConfig = import('./create-typography')
-  type TypographyVariantKey = TypographyConfig['TypographyVariantKey']
+import type { CSSProperties } from 'react'
+import type { TypographyVariantKey, DisabledTypographyVariantKey } from './create-typography'
 
-  type NewTypographyVariants<T> = { [key in TypographyVariantKey]: T }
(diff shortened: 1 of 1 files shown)


packages/curve-ui-kit/src/themes/typography/override-typography.d.ts
```diff
@@ -1,22 +1,14 @@
...
-    interface TypographyVariants extends NewTypographyVariants<React.CSSProperties> {}
-    interface TypographyVariantsOptions extends Partial<NewTypographyVariants<React.CSSProperties>> {}
-  }
-
-  module '@mui/material/Typography' {
-    interface TypographyPropsVariantOverrides extends NewTypographyVariants<true>, DisabledTypographyVariants {}
-  }
+declare module '@mui/material/styles' {
+  interface TypographyVariants extends NewTypographyVariants<React.CSSProperties> {}
+  interface TypographyVariantsOptions extends Partial<NewTypographyVariants<React.CSSProperties>> {}
```
DanielSchiavini, line 9:
```suggestion
  interface TypographyVariants extends NewTypographyVariants<CSSProperties> {}
  interface TypographyVariantsOptions extends Partial<NewTypographyVariants<CSSProperties>> {}
```

DanielSchiavini (approved):
Nice! Can we test that an invalid typography breaks the github action?

0xAlunara:
Github Actions will invoke the build script, which will fail due to a typecheck error:

image
image
image

DanielSchiavini (approved):

PR #561, opened 2024-12-30
fix: z-index conflict

DanielSchiavini:
- Fixes z-index issue in the mobile header
- Uses the correct z-index for the app bar based on the theme


DanielSchiavini:
this was already changed in https://github.com/curvefi/curve-frontend/pull/552 I'll move this there

PR #570, opened 2024-12-31
refactor: move i18n lib to curve-ui-kit

0xAlunara:
In order to be able to change the locale (language) in the user profile feature that will reside in the curve-ui-kit package, some of the i18n code that's copy-pasted among the 4 apps has to be moved and consolidated into that same curve-ui-kit package.

This PR's concern is about migrating this code, not improving it; all original code except `dynamicActivate` remain intact. Because of the way `dynamicActivate` used dynamic import, which doesn't really work in a package library, I had to replace its invocations with
```tsx
;(async () => {
  let data = await import(`@/locales/${parsedLocale}/messages`)
  dynamicActivate(parsedLocale, data)
})()
```
Is it ugly? Yes, and it can be improved, but I'd rather have it work and improve later when necessary.


packages/curve-ui-kit/src/lib/i18n.ts
```diff
@@ -55,14 +54,12 @@ export function parseLocale(locale?: string): { parsedLocale: Locale['value']; p
...
-export async function dynamicActivate(locale: string) {
-  const { parsedLocale } = parseLocale(locale)
-  let data = await import(`../locales/${parsedLocale}/messages`)
+export async function dynamicActivate(locale: string, data: any) {
   i18n.load(locale, data.messages)
   i18n.activate(locale)
 }
 
-export function updateAppLocale(locale: string, updateGlobalStoreByKey: <T>(key: DefaultStateKeys, value: T) => void) {
+export function updateAppLocale(locale: string, updateGlobalStoreByKey: <T>(key: 'locale', value: T) => void) {
```
DanielSchiavini, line 62: Unused parameter
```suggestion
export function updateAppLocale(locale: string, updateGlobalStoreByKey: <T>(value: T) => void) {
```

apps/main/src/pages/_app.tsx
```diff
@@ -90,7 +90,10 @@ function CurveApp({ Component }: AppProps) {
     const { rLocale } = getLocaleFromUrl()
     const parsedLocale = rLocale?.value ?? 'en'
     initTranslation(i18n, parsedLocale)
-    dynamicActivate(parsedLocale)
+    ;(async () => {
```
DanielSchiavini, line 93: Is it possible to remove repetitions of this? Just one per app should be enough. Your explanation in the PR description might come in handy, too

packages/curve-ui-kit/src/lib/i18n.ts
0xAlunara: I can't remove it as it's used below: `updateGlobalStoreByKey('locale', locale)`. It's an injected function whose type signature is `updateGlobalStoreByKey: <T>(key: DefaultStateKeys, value: T) => void`. I guess you can remove it but then you need to not forget and copypaste the `updateGlobalStoreByKey` invocation every time `updateAppLocale` is called.

apps/main/src/pages/_app.tsx
0xAlunara: There is only one invocation per app on load, but sadly we have 4 apps so there's some repetition there.

Then each app also has its own `usePageOnMount` where there's a watch / useEffect to see if the locale has changed. You can argue that should be enough and the call in `_app.tsx` can be removed, but I'm afraid somebody somewhere doesn't use that hook and that specific page would fail to load the locale.

However, I've added documentation explaining the situation to the `dynamicActivate` function.

DanielSchiavini: There're multiple invovations per app if I'm not mistaken?

DanielSchiavini (approved):

apps/main/src/pages/_app.tsx
0xAlunara: Per app there's an invocation in _app.tsx inside a useEffect with [] as it's dependency, so I'm assuming that happens once on initial mount, unless I'm missing something.

DanielSchiavini: For example 
https://github.com/curvefi/curve-frontend/blob/b067240955dc2ae2134848760d80150e02a67de8/apps/loan/src/hooks/usePageOnMount.ts#L259

0xAlunara: That's for locale changes and I'm not sure if every page even uses that hook. If they don't and you'd remove the invocation in _app.tsx you'd have no locale initialization at all. And since we have 4 diff apps that are slightly different I don't want to move that code from the hook to the four _app.tsx's. This needs a refactor but imo is outside the scope of the PR.

OnlyJousting (approved):

PR #576, opened 2025-01-03
Fix/modal dialog small screen sizes

OnlyJousting:
Changes:
- Adjust modal size on smaller screens to utilise 100% width and height

### DIFF
--- a/packages/curve-ui-kit/src/shared/ui/ModalDialog.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/ModalDialog.tsx
@@ -22,11 +22,16 @@ export const ModalDialog = ({ children, open, onClose, title, titleAction, foote
   <Dialog open={open} onClose={onClose}>
     <Card
       sx={{
-        ...SizesAndSpaces.ModalHeight,
-        width: SizesAndSpaces.ModalWidth.md,
+        ...SizesAndSpaces.ModalHeight.sm,
+        width: SizesAndSpaces.ModalWidth.sm,
         maxWidth: '100vw',
         display: 'flex',
         flexDirection: 'column',
+
+        [`@media (min-width: ${SizesAndSpaces.ModalWidth.md})`]: {
+          ...SizesAndSpaces.ModalHeight.md,
+          width: SizesAndSpaces.ModalWidth.md,
+        },
       }}
     >
       <CardHeader
--- a/packages/curve-ui-kit/src/themes/components.ts
+++ b/packages/curve-ui-kit/src/themes/components.ts
@@ -43,7 +43,7 @@ export const createComponents = (design: DesignSystem): ThemeOptions['components
   },
   MuiDialog: {
     styleOverrides: {
-      paper: { [basicMuiTheme.breakpoints.down('tablet')]: { margin: SizesAndSpaces.Spacing.md.mobile } },
+      paper: { maxHeight: '100vh', [basicMuiTheme.breakpoints.down('tablet')]: { margin: 0 } },
     },
   },
   MuiIconButton: defineMuiIconButton(design),
--- a/packages/curve-ui-kit/src/themes/components.ts
+++ b/packages/curve-ui-kit/src/themes/components.ts
@@ -43,7 +43,7 @@ export const createComponents = (design: DesignSystem): ThemeOptions['components
   },
   MuiDialog: {
     styleOverrides: {
-      paper: { maxHeight: '100vh', [basicMuiTheme.breakpoints.down('tablet')]: { margin: 0 } },
+      paper: { maxHeight: '100dvh', [basicMuiTheme.breakpoints.down('tablet')]: { margin: 0 } },
     },
   },
   MuiIconButton: defineMuiIconButton(design),
--- a/packages/curve-ui-kit/src/themes/design/1_sizes_spaces.ts
+++ b/packages/curve-ui-kit/src/themes/design/1_sizes_spaces.ts
@@ -330,11 +330,24 @@ const MappedTypography = {
 } as const
 
 const MappedModalWidth = {
-  xs: '19rem', // 304px
-  sm: '24rem', // 384px
+  xs: '100vw', // 100%
+  sm: '100vw', // 100%
   md: '28rem', // 448px
   lg: '32rem', // 512px
   xl: '36rem', // 576px
+}
+
+const MappedModalHeight = {
+  sm: {
+    maxHeight: '100vh',
+    height: '100vh',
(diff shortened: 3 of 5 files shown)


packages/curve-ui-kit/src/themes/components.ts
```diff
@@ -43,7 +43,7 @@ export const createComponents = (design: DesignSystem): ThemeOptions['components
   },
   MuiDialog: {
     styleOverrides: {
-      paper: { [basicMuiTheme.breakpoints.down('tablet')]: { margin: SizesAndSpaces.Spacing.md.mobile } },
+      paper: { maxHeight: '100vh', [basicMuiTheme.breakpoints.down('tablet')]: { margin: 0 } },
```
DanielSchiavini, line 46: Isn't 0 the default? :thinking:
```suggestion
      paper: { maxHeight: '100vh' },
```

packages/curve-ui-kit/src/themes/design/1_sizes_spaces.ts
```diff
@@ -330,13 +330,26 @@ const MappedTypography = {
 } as const
 
 const MappedModalWidth = {
-  xs: '19rem', // 304px
-  sm: '24rem', // 384px
+  xs: '100vw', // 100%
```
DanielSchiavini, line 333: This is originating from figma. It referes to the modal size not the screen size. It's also supposed to be responsive :thinking: 
image

packages/curve-ui-kit/src/shared/ui/ModalDialog.tsx
```diff
@@ -22,11 +22,16 @@ export const ModalDialog = ({ children, open, onClose, title, titleAction, foote
...
       sx={{
-        ...SizesAndSpaces.ModalHeight,
-        width: SizesAndSpaces.ModalWidth.md,
+        ...SizesAndSpaces.ModalHeight.sm,
+        width: SizesAndSpaces.ModalWidth.sm,
         maxWidth: '100vw',
         display: 'flex',
         flexDirection: 'column',
+
+        [`@media (min-width: ${SizesAndSpaces.ModalWidth.md})`]: {
```
DanielSchiavini, line 31: I find it quite confusing to use sm/md for the screen size, cause we have been consistently using the t-shirt sizes for elements, not screens. However, I don't know what a better name would be. Maybe `ModalWidth.fullScreenBreakpoint`?

packages/curve-ui-kit/src/themes/design/1_sizes_spaces.ts
DanielSchiavini: Disregard that, I think xs-mobile is incorrect in figma. So all screens have the same modal sizes

packages/curve-ui-kit/src/themes/components.ts
0xAlunara, line 46: It's better to use `100dvh` as this takes the mobile top bar into account, for which is max height is targeted for.

0xAlunara (changes requested):

packages/curve-ui-kit/src/shared/ui/ModalDialog.tsx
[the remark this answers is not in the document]
DanielSchiavini: yeah this is not enough, he wants a different breakpoint here

packages/curve-ui-kit/src/themes/components.ts
DanielSchiavini: I see other 42 usages of 'vh' in the codebase :cry:

OnlyJousting: No, there is another default, it was quite a lot of margin at that hence why I think 0.5rem was set initially

packages/curve-ui-kit/src/themes/design/1_sizes_spaces.ts
OnlyJousting: I see, these are not responsive values though. It doesn't make sense to micro manage like this when setting 100% offers a better experience for the user.

packages/curve-ui-kit/src/shared/ui/ModalDialog.tsx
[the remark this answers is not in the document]
OnlyJousting: This is much too crude for the modal. The modal shouldn't be 100% width at the tablet breakpoint. Using three breakpoints offers too little control over the UI when judging on a case by case makes for a much nicer user experience.

OnlyJousting: Renaming it makes sense yea. That value is the last fixed width of the modal so it's a good breakpoint.

packages/curve-ui-kit/src/themes/components.ts
OnlyJousting: wow I didn't know about dvh, seems great

packages/curve-ui-kit/src/shared/ui/ModalDialog.tsx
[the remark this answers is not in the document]
0xAlunara: In that case `[theme.breakpoints.up('md')]:` works as well

packages/curve-ui-kit/src/themes/components.ts
DanielSchiavini: aha that's why I've overridden it

packages/curve-ui-kit/src/shared/ui/ModalDialog.tsx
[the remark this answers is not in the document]
OnlyJousting: This would lead to the modal being 600px (full screen) wide rather then 448px for a bit. There will be more cases where sticking to some arbitrary value for a breakpoint will come at the expense of a superior UI. I don't see the benefit to sticking to these default values.

DanielSchiavini: the thing is that the modal (in the future) needs to receive 'sm', 'md' as prop so it can be displayed properly. This implementation works fine for this one size but it won't work for different modal sizes

packages/curve-ui-kit/src/themes/design/1_sizes_spaces.ts
DanielSchiavini: These values are kind of a desired width, but ofc we need to scale down to 100dvh when on small screens.
The problem here is that we are mixing screen size with component size

0xAlunara (approved):

PR #579, opened 2025-01-04
feat: add toggle button styling

0xAlunara:
For the user profile, theme selection will need to happen through a group of toggle buttons, so I had to style those according to the Figma design first. You can find examples in the storybook.

* I've added a custom property to the toggle button group component called `compact` that removes the gap between toggle buttons in that group. Sometimes, in case of multiple non-exclusive buttons, the gap is kinda ugly.
* I've added custom 'square' sizes. This isn't part of the Figma design, but is aesthetically pleasing when the toggle buttons contain mere icons. Goes well with the compact view as seen in the "Multiple" story in the storybook.
* Hover state doesn't work in the Storybook's 'Docs' overview page. However, if you click a specific story like 'Text' it will work. Not sure why that happens.
image


packages/curve-ui-kit/src/themes/stories/ToggleButtons.stories.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+import CheckIcon from '@mui/icons-material/Check'
+
+const meta: Meta<typeof ToggleButtonGroup> = {
+  title: 'UI Kit/Primitives/ToggleButton',
+  component: ToggleButtonGroup,
+  argTypes: {
+    size: {
+      control: 'select',
+      options: ['extraSmall', 'small', 'medium', 'extraSmallSquare', 'smallSquare', 'mediumSquare'],
```
DanielSchiavini, line 25: Do we really need this many sizes?
Afaik all buttons are square so I don't understand the difference

0xAlunara:
The togglebuttons are rectangular with only a min width, most noticable
with the text variant in the storybook and Figma. The block padding is also
slightly larger than the inline padding.

The square sizes are really square with equal block and inline padding,
fixed (responsive) width and hidden overflow. This variant isn't on the
btn-toggle Figma page, but it's used in the user profile one.

packages/curve-ui-kit/src/themes/stories/ToggleButtons.stories.tsx
0xAlunara: The togglebuttons are rectangular with only a min width, most noticable with the text variant in the storybook and Figma. The block padding is also slightly larger than the inline padding.

The square sizes are really square with equal block and inline padding, fixed (responsive) width and hidden overflow. This variant isn't on the btn-toggle Figma page, but it's used in the user profile one.

DanielSchiavini (approved):

packages/curve-ui-kit/.storybook/preview.tsx
```diff
@@ -9,7 +9,7 @@ import Box from '@mui/material/Box'
 const themes = {
   light: lightTheme(),
   lightInverted: darkTheme({ inverted: true }),
-  dark: lightTheme(),
+  dark: darkTheme(),
```
DanielSchiavini, line 12: what :exploding_head:

DanielSchiavini (dismissed):

0xtutti (dismissed):

packages/curve-ui-kit/.storybook/preview.tsx
```diff
@@ -30,7 +30,6 @@ export const decorators: Decorator[] = [
           font-family: ${monaSans.style.fontFamily};
           font-style: normal;
           font-display: swap;
```
DanielSchiavini, line 32: I'm not sure if we need these 2 lines, either

DanielSchiavini (approved):

DanielSchiavini (approved):

OnlyJousting (approved):

PR #575, opened 2025-01-02
feat: table filters

DanielSchiavini:

PR #583, opened 2025-01-06
feat: column visibility settings

DanielSchiavini:
- Add option to control the visibility of columns
- Add tests
- Based off https://github.com/curvefi/curve-frontend/pull/575


packages/curve-ui-kit/src/themes/components.ts
```diff
@@ -47,6 +48,12 @@ export const createComponents = (design: DesignSystem): ThemeOptions['components
       paper: { maxHeight: '100dvh', [basicMuiTheme.breakpoints.down('tablet')]: { margin: 0 } },
     },
   },
+  MuiFormControlLabel: {
+    styleOverrides: {
+      root: { margin: '0' }, // by default there is a negative margin 🤦
```
0xAlunara, line 53: lol what's up with that, I encountered the same thing with toggle buttons, specifically the middle button in a group had a `margin-left: -1px` 🤷

packages/curve-ui-kit/src/shared/ui/TableColumnVisibilityPopover.tsx
```diff
@@ -0,0 +1,129 @@
...
+  label: string
+}
+export type ColumnVisibilityGroup = {
+  columns: ColumnVisibility[]
+  label: string
+}
+const { Spacing } = SizesAndSpaces
+
+// when we define columns in nested objects, tanstack replaces dots with underscores internally
+export const cleanColumnId = (field: string) => field.replaceAll('.', '_')
```
0xAlunara, line 22: Such a weird little detail you need to know about TanStack Table, hidden in the docs at https://tanstack.com/table/latest/docs/guide/column-defs#unique-column-ids

0xAlunara: As per ChatGPT:

TanStack Table replaces . with _ in column IDs to ensure the IDs are valid JavaScript property keys when they are used internally.

In JavaScript, . is interpreted as a property accessor, so having . in an ID could cause issues when accessing properties programmatically. Replacing it with _ avoids potential conflicts or errors when TanStack Table constructs internal object paths or uses these IDs in operations.

This transformation ensures that column IDs are always safe and consistent for internal use while still being derived from user-provided values.

packages/curve-ui-kit/src/themes/components.ts
DanielSchiavini: Yeah I think they did this for backwards compatibility with some old versions :disappointed:

0xAlunara:
Not sure what's the cause of this, but when there's no wallet connected the layout is broken
image

packages/ui/src/ConnectWalletPrompt/index.tsx
```diff
@@ -23,11 +24,7 @@ const ConnectWalletPrompt: React.FC<ConnectWalletPromptProps> = ({
   loadingText,
   theme = 'light',
 }) => {
-  const BackgroundSvg =
-    theme === 'light'
-      ? 'https://cdn.jsdelivr.net/gh/curvefi/curve-assets/branding/curve_illustration-light.svg'
-      : 'https://cdn.jsdelivr.net/gh/curvefi/curve-assets/branding/curve_illustration-dark.svg'
-
+  const BackgroundSvg = getBackgroundUrl(theme)
```
DanielSchiavini, line 27: Looks like it's broken 😢

DanielSchiavini:
> Not sure what's the cause of this, but when there's no wallet connected the layout is broken

Fixed now, it was a typo

OnlyJousting (approved):

0xtutti (dismissed):

OnlyJousting (approved):

PR #589, opened 2025-01-07
feat: user profile

0xAlunara:
This PR is not even close to being finished, but because I'm touching global settings and stores that are all over place in the apps, I want to keep it small and not keep the PR open too long.

* The user profile button shows *only in beta mode*. That means it won't show up in production
* The user profile button replaces the advanced toggle and theme switcher button
* The user profile is available even when no wallet is connected, because there's global settings there that are important even without a wallet. It'll show the 0x0000 address if there's no wallet connected.
* I've decided to create new theme switching and advanced mdoe toggle components instead of re-using the `switch-theme` and `switch-advanced-mode` toggles. They differ enough that I couldn't re-use them in the user profile. I'm not sure if we want to keep those components as generic 'features', rather than just part of the component that hosts them.
* At the moment to keep things small and simple only the advanced toggle and theme switcher are implemented. No metrics, no disconnect button.
* Both these two settings reside in a new store specific to the feature, `useUserProfileStore()`.
* In addition, the store also contains the configured locale and slippage settings (although not configurable from the UI yet)
* The slippage setting has a hardcoded `global` value. This used to be called `router` but was used globally anyway in all four different apps, hence the rename. Any other 'property' is usually a slippage setting for a certain pool & chain id combination.
* All four properties (advanced mode, theme, locale, slippage) have been removed from the `app` or `global` Zustand slices. Hence the reason this PR is so big: the entire app now needs to use the new store.
* The store is persisted in local storage. This means that until we merge all 4 apps, the different apps don't share a global persisted state. However, this is already the case in production.
* There's no tests yet given the feature isn't released yet and is bound to change as it gets developed over time.
* Current header tests fail of course, can fix that before merging.

image
image


0xtutti (approved):

OnlyJousting (approved):

packages/curve-ui-kit/src/utils/index.ts
```diff
@@ -9,3 +9,5 @@ export * from './searchText'
 export const isBeta =
   typeof window !== 'undefined' &&
   (window.localStorage.getItem('beta') !== null || !window.location.hostname.includes('curve.fi'))
+
+export const isCypress = typeof window !== 'undefined' && (window as any).Cypress
```
DanielSchiavini, line 13:
```suggestion
export const isCypress = typeof window !== 'undefined' && Boolean((window as {Cypress?: boolean}).Cypress)
```

apps/dao/src/layout/Header.tsx
```diff
@@ -27,20 +27,20 @@ export const Header = ({ sections, BannerProps }: HeaderProps) => {
   const connectState = useStore((state) => state.connectState)
   const isMdUp = useStore((state) => state.layout.isMdUp)
   const bannerHeight = useStore((state) => state.layoutHeight.globalAlert)
-  const locale = useStore((state) => state.locale)
   const routerProps = useStore((state) => state.routerProps)
-  const themeType = useStore((state) => state.themeType)
-  const setAppCache = useStore((state) => state.setAppCache)
   const updateConnectState = useStore((state) => state.updateConnectState)
 
+  const theme = useUserProfileStore((state) => state.theme)
```
DanielSchiavini, line 33: isn't it possible to  move this to the `NewHeader` component now?

apps/lend/src/components/ChartOhlcWrapper/index.tsx
```diff
@@ -405,7 +407,7 @@ const ChartOhlcWrapper: React.FC<ChartOhlcWrapperProps> = ({ rChainId, userActiv
       </LpEventsWrapperExpanded>
     </ExpandedWrapper>
   ) : (
-    <Wrapper className={!isAdvanceMode ? 'normal-mode' : ''} chartExpanded={chartExpanded}>
+    <Wrapper className={!isAdvancedMode ? 'normal-mode' : ''} chartExpanded={chartExpanded}>
```
DanielSchiavini, line 410: Double negatives are hard to read
```suggestion
    <Wrapper className={isAdvancedMode ? '' : 'normal-mode'} chartExpanded={chartExpanded}>
```

packages/curve-ui-kit/src/utils/address.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @example
+ * addressShort("0x1234567890abcdef1234567890abcdef12345678") // "0x1234...5678"
+ * addressShort("0x1234567890abcdef1234567890abcdef12345678", 8) // "0x12345...5678"
+ */
+export function addressShort(address?: string, digits = 6): string {
+  if (!address) {
+    return '0x000...000'
+  }
+
+  const pre = digits / 2 + 2
```
DanielSchiavini, line 18:
```suggestion
  const pre = digits / 2 + 2 // +2 for the "0x"
```

packages/curve-ui-kit/src/features/connect-wallet/ui/ConnectedWalletLabel.tsx
```diff
@@ -17,6 +18,8 @@ export const ConnectedWalletLabel: FunctionComponent<ConnectedWalletLabelProps>
   ...props
 }) => (
   <Button size="small" color="ghost" onClick={onDisconnectWallet} {...props}>
-    <AddressLabel value={walletAddress} />
+    <Box maxWidth={200} title={walletAddress}>
```
DanielSchiavini, line 21: This hardcoded width isn't great here. I think we should have it in a separate `AddressLabel` component so it can be reused in other places

packages/curve-ui-kit/src/features/user-profile/UserProfile.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+type Props = {
+  open: boolean
+  onClose: () => void
+}
+
+export const UserProfile = ({ open, onClose }: Props) => (
+  <Drawer open={open} anchor="right" onClose={onClose}>
+    <Box
+      sx={{
```
DanielSchiavini, line 15: The Drawer already renders a `MuiPaper` (but by default that uses Layer[2])
```suggestion
  <Drawer open={open} anchor="right" onClose={onClose} PaperProps={{ sx: {
```

packages/curve-ui-kit/src/features/user-profile/UserProfileButton.tsx
```diff
@@ -0,0 +1,19 @@
+import { useState } from 'react'
+import IconButton from '@mui/material/IconButton'
+import PersonIcon from '@mui/icons-material/Person'
+
+import { UserProfile } from './UserProfile'
+
+export const UserProfileButton = () => {
+  const [open, setOpen] = useState(false)
```
DanielSchiavini, line 8:
```suggestion
import { useSwitch } from '@ui-kit/hooks/useSwitch'

export const UserProfileButton = () => {
  const [isOpen, open, close] = useSwitch()
```

packages/curve-ui-kit/src/features/user-profile/UserProfileButton.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+  return (
+    <>
+      <IconButton size="small" onClick={() => setOpen(true)}>
+        <PersonIcon />
+      </IconButton>
+
+      <UserProfile open={open} onClose={() => setOpen(false)} />
```
DanielSchiavini, line 16: It's nice to lazy render the user profile as most times it will be invisible
```suggestion
      {open != null && <UserProfile open={open} onClose={close} />}
```

packages/curve-ui-kit/src/features/user-profile/home/Header.tsx
```diff
@@ -0,0 +1,52 @@
...
+  onClose: () => void
+}
+
+const LlamaImageSrc = (LlamaImg as unknown as { src: string }).src
+
+/** Gap deviates from Figma as personally 'xs' is too narrow. */
+export const Header = ({ walletAddress, onClose }: Props) => (
+  <Stack direction="row" alignItems="center" gap={SizesAndSpaces.Spacing.sm}>
+    <Box
+      component="img"
```
DanielSchiavini, line 24: missing alt

packages/curve-ui-kit/src/features/user-profile/settings/Setting.tsx
```diff
@@ -0,0 +1,19 @@
+import { ReactNode } from 'react'
+import Stack from '@mui/material/Stack'
+import Typography from '@mui/material/Typography'
+
+import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
+
+const { Spacing } = SizesAndSpaces
+
+export const Label = ({ children }: { children?: ReactNode }) => (
```
DanielSchiavini, line 9:
```suggestion
export const SettingLabel = ({ children }: { children?: ReactNode }) => (
```

packages/curve-ui-kit/src/features/user-profile/home/Header.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+type Props = {
+  walletAddress?: Address
+  onClose: () => void
+}
+
+const LlamaImageSrc = (LlamaImg as unknown as { src: string }).src
+
+/** Gap deviates from Figma as personally 'xs' is too narrow. */
+export const Header = ({ walletAddress, onClose }: Props) => (
```
DanielSchiavini, line 21:
```suggestion
export const UserProfileHeader = ({ walletAddress, onClose }: Props) => (
```

packages/curve-common/src/widgets/Header/SideBarFooter.tsx
```diff
@@ -30,6 +31,7 @@ export const SideBarFooter = ({ themes: [theme, setTheme], advancedMode, WalletP
     >
       <Box display="flex" paddingX={4} marginTop={4}>
         <ConnectWalletIndicator {...WalletProps} sx={{ flexGrow: 1 }} />
+        <UserProfileButton />
```
DanielSchiavini, line 34: The theme switcher is not rendered in two places. The alignment here doesn't seem correct either
image
```suggestion
        {isBeta && <UserProfileButton />}
```

packages/curve-ui-kit/src/features/user-profile/settings/advanced-mode/AdvancedMode.tsx
```diff
@@ -0,0 +1,17 @@
+import { t } from '@lingui/macro'
+import Switch from '@mui/material/Switch'
+
+import { Label, Setting } from '../Setting'
+import useUserProfileStore from '../../store'
+
+export const AdvancedMode = () => {
```
DanielSchiavini, line 7: looks like a duplicate of `AdvancedModeSwitcher`

packages/curve-ui-kit/src/features/user-profile/settings/advanced-mode/AdvancedMode.tsx
```diff
@@ -0,0 +1,17 @@
...
+  const isAdvancedMode = useUserProfileStore((state) => state.isAdvancedMode)
+  const setAdvancedMode = useUserProfileStore((state) => state.setAdvancedMode)
+
+  return (
+    <Setting>
+      <Label>{t`Advanced Mode`}</Label>
+      <Switch checked={isAdvancedMode} onChange={() => setAdvancedMode(!isAdvancedMode)} color="primary" size="small" />
```
DanielSchiavini, line 14: I discovered that the correct way to do this is using the `FormControlLabel`  ([like here](https://github.com/curvefi/curve-frontend/blob/e400655e2ac3c472171b0a7ae4e1f98964302736/packages/curve-ui-kit/src/features/switch-advanced-mode/AdvancedModeSwitcher.tsx#L12))

```suggestion
      <FormControlLabel control={<Switch checked={isAdvancedMode} onChange={() => setAdvancedMode(!isAdvancedMode)} color="primary" size="small" />} label={t`Advanced Mode`}>
```

packages/curve-ui-kit/src/features/user-profile/settings/theme-selector/ThemeSelector.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const ThemeSelector = () => {
+  const theme = useUserProfileStore((state) => state.theme)
+  const setTheme = useUserProfileStore((state) => state.setTheme)
+
+  return (
+    <Setting>
+      <Label>{t`Theme`}</Label>
+
+      <ToggleButtonGroup
```
DanielSchiavini, line 18: Duplicate of `ThemeSwitcherButtons`

apps/main/src/store/createGlobalSlice.ts
```diff
@@ -38,21 +37,17 @@ type GlobalState = {
...
   pageWidthPx: number | null
   pageWidth: PageWidthClassName | null
-  maxSlippage: { [key: string]: string }
   routerProps: RouterProps | null
   showScrollButton: boolean
-  themeType: Theme
 }
 
 // prettier-ignore
 export interface GlobalSlice extends GlobalState {
```
DanielSchiavini, line 47: Unrelated, but this is pretty concerning: https://github.com/curvefi/curve-frontend/blob/9f0db9e97443d20d7eab9f01afb99679d61db1ce/apps/main/src/store/useStore.ts#L1

packages/curve-ui-kit/src/features/user-profile/store.ts
```diff
@@ -0,0 +1,106 @@
...
+  setAdvancedMode: (isAdvancedMode) => set((state) => ({ ...state, isAdvancedMode })),
+})
+
+const cache: PersistOptions<Store> = {
+  name: 'user-profile',
+  merge: (persistedState, currentState) => merge(currentState, persistedState),
+  version: 1,
+}
+
+const useUserProfileStore =
```
DanielSchiavini, line 103: Ah cool so apparently they use global data rather than context. IMO using "profile" here isn't the right name but I'm not super strong on it
```suggestion
const useUserPreferences =
```

packages/curve-ui-kit/src/features/connect-wallet/ui/ConnectedWalletLabel.tsx
0xAlunara: I've precisely removed the `AddressLabel` component because it was too simple too be warranted a component. Sometimes it's a header, sometimes just a span, sometimes it should be in a `<Box>` etc. Just using the `addressShort` should suffice.

But the box with maxWidth is useless regardless given the usage of `addressShort` so I removed it.

packages/curve-ui-kit/src/features/user-profile/UserProfileButton.tsx
0xAlunara: `<Drawer open={open}` already lazily renders the user profile. Applying your suggestion also breaks the draw transition anyway.

packages/curve-ui-kit/src/features/user-profile/settings/theme-selector/ThemeSelector.tsx
0xAlunara: Same reasoning as for the 'duplicate' advanced theme component

apps/main/src/store/createGlobalSlice.ts
0xAlunara: Huh, I removed it and prettier only applied to one line... Was introduced near the initial commit at the beginning of 2024; probably a leftover debugging thing?

packages/curve-ui-kit/src/features/user-profile/store.ts
0xAlunara: Not super strong on it either, but for the moment it's 'coupled' with the user profile setting so I'd rather keep until it's no longer suitable in the future.

packages/curve-common/src/widgets/Header/SideBarFooter.tsx
0xAlunara: Ho lol, a remainder from earlier experimentations. I've removed this button from the sidebar altogether, there's no need for it as there's already settings to change theme and advanced mode.

packages/curve-ui-kit/src/features/user-profile/settings/advanced-mode/AdvancedMode.tsx
0xAlunara: image

Not sure it works neatly with the grid I've just set up. If you insist on using it together with a proper form I'd rather do that in a diff PR as I'm sure other things will need to be changed as well.

packages/curve-ui-kit/src/features/connect-wallet/ui/ConnectedWalletLabel.tsx
```diff
@@ -16,7 +16,7 @@ export const ConnectedWalletLabel: FunctionComponent<ConnectedWalletLabelProps>
   onDisconnectWallet,
   ...props
 }) => (
-  <Button size="small" color="ghost" onClick={onDisconnectWallet} {...props}>
-    <AddressLabel value={walletAddress} />
+  <Button size="small" color="ghost" onClick={onDisconnectWallet} title={walletAddress} {...props}>
```
DanielSchiavini, line 19: Title is nice, but a [tooltip](https://mui.com/material-ui/react-tooltip/) is probably nicer

packages/curve-ui-kit/src/features/user-profile/UserProfileButton.tsx
```diff
@@ -0,0 +1,19 @@
...
+export const UserProfileButton = () => {
+  const [isOpen, open, close] = useSwitch()
+
+  return (
+    <>
+      <IconButton size="small" onClick={open}>
+        <PersonIcon />
+      </IconButton>
+
+      <UserProfile open={isOpen ?? false} onClose={close} />
```
DanielSchiavini, line 16: This is fine, but if you want false to be the default, it could be passed as arg to `useSwitch`

packages/curve-ui-kit/src/features/user-profile/home/Home.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { UserProfileHeader } from './Header'
+import { Settings } from '../settings'
+
+const { Spacing } = SizesAndSpaces
+
+type Props = {
+  onClose: () => void
+}
+
+export const Home = ({ onClose }: Props) => {
```
DanielSchiavini, line 15: This component's name isn't super clear

packages/curve-ui-kit/src/features/user-profile/settings/advanced-mode/AdvancedMode.tsx
DanielSchiavini: I don't like to have 4 components that essentially do the same thing

DanielSchiavini (approved):

packages/curve-ui-kit/src/features/user-profile/home/Header.tsx
```diff
@@ -0,0 +1,53 @@
...
+    <Box
+      component="img"
+      src={LlamaImageSrc}
+      sx={{
+        height: SizesAndSpaces.IconSize.xl,
+      }}
+      alt="Llama Icon"
+    />
+
+    <Typography variant="headingMLight" flexGrow={1}>
```
DanielSchiavini, line 32: title/tooltip would be helpful here too

packages/curve-ui-kit/src/features/user-profile/home/Home.tsx
0xAlunara: It derives from the multiple pages that are planned once the user profile is fully implemented. But it's subject to change.

packages/curve-ui-kit/src/features/connect-wallet/ui/ConnectedWalletLabel.tsx
0xAlunara: I'm not even sure we need a tooltip or title at all. This is the wallet connect button that shows an abbreviated address once connected; I've never seen it before where a tooltip shows the entire address on hover. Imo it should be removed and if the user wants to copy their address they should be able to in the upcoming user profile menu where it's displayed on top as a big header.

packages/curve-ui-kit/src/features/user-profile/UserProfileButton.tsx
0xAlunara: Should we update `useSwitch` such that it's strictly a boolean and the type of `const isOpen: boolean | undefined` becomes `const isOpen: boolean`?

packages/curve-ui-kit/src/features/connect-wallet/ui/ConnectedWalletLabel.tsx
DanielSchiavini: The tooltip can be useful (imo), what's the drawback?

packages/curve-ui-kit/src/features/user-profile/UserProfileButton.tsx
DanielSchiavini: I want it to be undefined in the places where I used it, meaning it hasn't been initialized yet

packages/curve-ui-kit/src/features/user-profile/home/Header.tsx
0xAlunara: Note to self: click to copy with `cursor: pointer` would be even better.

packages/curve-ui-kit/src/features/connect-wallet/ui/ConnectedWalletLabel.tsx
0xAlunara: It's multifaceted, the information displayed wouldn't lead to any meaningful action to the user or give information they already don't know. On top of that the address is relatively large, and so would the tooltip be; a bit distracting.

OnlyJousting (approved):

PR #595, opened 2025-01-13
styling: sync tab styling with Figma design

0xAlunara:
* I tried to match the Figma design as much as possible.
* I've added some transitions.
* Decided to use a ::after pseudo element for the borders, see docs in `mui-tabs.ts` for the reasons why.
* Default tabswitcher size changed to small, text variants when not configured match tabswitcher size (small, medium or large).
* Storybook tab page has only one story, play around with variant & sizes.
* Also affects disclaimer page tabs and navigation pages tab in the header.

Man, this one was finicky as HELL, I spent a lot of times getting it right. Looked very easy but the details took 90% of the time. Haven't look at scrolling yet.


0xtutti (dismissed):
Inverted theme is displaying funny, but it's not an issue as there's currently no setup where tabs would need to be inverted.

packages/curve-ui-kit/src/themes/tabs/mui-tabs.ts
```diff
@@ -13,51 +18,142 @@ export const TABS_HEIGHT_CLASSES = { small, medium, large }
...
+  },
 })
 
+const tabPadding = (
+  blockStart: keyof typeof Spacing,
+  blockEnd: keyof typeof Spacing,
+  inlineStart: keyof typeof Spacing,
+  inlineEnd: keyof typeof Spacing,
+) =>
+  Object.entries(basicMuiTheme.breakpoints.keys)
```
DanielSchiavini, line 95: can't we use `handleBreakpoints`?

packages/curve-ui-kit/src/themes/design/2_theme.ts
```diff
@@ -5,13 +5,17 @@ const { plain, inverted } = SurfacesAndText
 
 const Transition = `color ${TransitionFunction}, background-color ${TransitionFunction}, filter ${TransitionFunction}`
 
+const InsetOverline = '0 0 auto' as const // Top border only
```
DanielSchiavini, line 8: I don't understand these :thinking: 
`0 0 auto` means `{top: 0, right: 0, bottom: 'auto', left: 0}`, how is that top border only?

0xAlunara: 'auto' is simply the default CSS value in order to be able to use the 4 value syntax of top-right-bottom-left.

`0 0 auto` is equivalent to `top: 0; right: 0` which means on top from left to right; so top border.
`auto 0 0` is equivalent to `bottom: 0; right: 0` which means on bottom from left to right; so bottom border.

DanielSchiavini: ah thx, tbh I never used inset :fearful:

packages/curve-ui-kit/src/themes/tabs/mui-tabs.ts
0xAlunara: right, beautiful!

packages/curve-ui-kit/src/themes/tabs/mui-tabs.ts
```diff
@@ -13,51 +18,134 @@ export const TABS_HEIGHT_CLASSES = { small, medium, large }
...
+    paddingBlockStart: Spacing[blockStart],
+    paddingBlockEnd: Spacing[blockEnd],
+    paddingInlineStart: Spacing[inlineStart],
+    paddingInlineEnd: Spacing[inlineEnd],
+  })
+
+// Overlined and Underlined have common paddings and sizes.
+const tabSizesNonContained = {
+  [`&.${medium} .MuiTab-root`]: tabPadding('md', 'xs', 'md', 'md'),
+  [`&.${large} .MuiTab-root`]: tabPadding('md', 'xs', 'md', 'md'),
```
DanielSchiavini, line 105: These calls are not readable without checking the method definition.
Also, the order of the args doesn't follow the css standard (top, right, bottom, left).

```suggestion
  [`&.${medium} .MuiTab-root`]: handleBreakpoints({ padding: 'md', paddingBlockEnd: 'xs' }),
  [`&.${large} .MuiTab-root`]: handleBreakpoints({ padding: 'md', paddingBlockEnd: 'xs' }),
```

Or...
```suggestion
  [`&.${medium} .MuiTab-root`]: tabPadding({ blockStart: 'md', blockEnd: 'xs', inlineStart: 'md', inlineEnd: 'md' }),
  [`&.${large} .MuiTab-root`]: tabPadding({ blockStart: 'md', blockEnd: 'xs', inlineStart: 'md', inlineEnd: 'md'),
```

But yeah they are quite simple so not a big deal.

DanielSchiavini (approved):

OnlyJousting (approved):

PR #600, opened 2025-01-14
refactor: improve translation in packages

0xAlunara:
Moving some package dependencies around to curve-ui-kit and having a shared lingui config file.

Depends on #598


apps/dao/lingui.config.ts
```diff
@@ -0,0 +1,6 @@
+import { config, defaultCatalog } from '../../lingui.config'
+
+export default {
```
DanielSchiavini, line 3: all apps export the same thing, can't we move defaultCatalog to the `config` object?

apps/lend/package.json
```diff
@@ -27,25 +27,24 @@
   },
   "dependencies": {
     "@curvefi/lending-api": "^2.4.0",
-    "@lingui/detect-locale": "^4.6.0",
-    "@lingui/react": "^4.6.0",
+    "@lingui/cli": "^4.6.0",
```
DanielSchiavini, line 30: If it worked with dev dependencies, why move it here?

packages/ui/src/utils/utilsFormat.ts
```diff
@@ -1,12 +1,20 @@
...
 
+export function detectLocale() {
+  if (typeof window !== 'undefined') {
+    const DEFAULT_FALLBACK = () => 'en'
+    return detect(fromUrl('lang'), fromNavigator(), DEFAULT_FALLBACK) ?? 'en-US'
+  } else {
+    return 'en-US'
+  }
+}
+
```
DanielSchiavini, line 17:
```suggestion
export const detectLocale = () =>
  (typeof window !== 'undefined' && detect(fromUrl('lang'), fromNavigator(), () => 'en')) ?? 'en-US'
```

DanielSchiavini (approved):
It looks like a bad idea to move `devDependencies` to `dependencies` but I don't see a difference in the bundle size

packages/ui/src/utils/utilsFormat.ts
0xAlunara: Decided to inline it directly as `const localeDetected =  (typeof window !== 'undefined' && detect(fromUrl('lang'), fromNavigator(), () => 'en')) || 'en-US'` given the function is literally used only once.

apps/dao/lingui.config.ts
0xAlunara: Was debating, couldn't chose, figured maybe you want to extend later. Either way `defaultCatalog` is now moved inside the config directly.

apps/lend/package.json
0xAlunara: Ah cheers for noticing. I did some shuffling around and moved those dev dependencies out, only to realize they couldn't. So I moved them back, but did not notice it wasn't dev dependencies. They're moved back now.

apps/lend/lingui.config.ts
```diff
@@ -1,6 +1,5 @@
-import { config, defaultCatalog } from '../../lingui.config'
+import { config } from '../../lingui.config'
 
 export default {
   ...config,
-  catalogs: [defaultCatalog],
 }
```
DanielSchiavini, line 5:
```suggestion
export default config
```

DanielSchiavini (approved):

apps/lend/lingui.config.ts
0xAlunara: 😅

DanielSchiavini (approved):

pilotvietnam (approved):

OnlyJousting (approved):

PR #609, opened 2025-01-17
styling: update metric size and spacing according to figma

0xAlunara:


OnlyJousting:
I think it could be nice if the loading skeleton used the same height as the loaded component to reduce stuttering.

0xAlunara:
@OnlyJousting Not sure what you mean as it's already happening: ```<Skeleton variant="text" width="100%" sx={(t) => ({ ...t.typography[fontVariant] })} />```. The Skeleton takes over the typography styling such that the height won't jump around when `loading` is set from `true` to `false`. I see no jumping around either when toggling it in the storybook.

OnlyJousting (approved):

PR #616, opened 2025-01-22
feat: add peg keepers to crvUSD menu

DanielSchiavini:
- add peg keepers to crvUSD menu
- make sure the design does not break when there are many options

### DIFF
--- a/packages/curve-ui-kit/src/widgets/Header/HeaderStats.tsx
+++ b/packages/curve-ui-kit/src/widgets/Header/HeaderStats.tsx
@@ -9,7 +9,8 @@ export type HeaderStatsProps = {
 export const HeaderStats: FunctionComponent<HeaderStatsProps> = ({ appStats }) =>
   appStats?.map(({ label, value }) => (
     <Box key={label} display="inline-flex" alignItems="baseline">
-      <Typography variant="bodyMRegular" color="grey.600">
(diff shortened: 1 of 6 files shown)


0xAlunara (approved):

0xtutti (approved):

packages/curve-ui-kit/src/widgets/Header/HeaderStats.tsx
```diff
@@ -9,7 +9,8 @@ export type HeaderStatsProps = {
 export const HeaderStats: FunctionComponent<HeaderStatsProps> = ({ appStats }) =>
   appStats?.map(({ label, value }) => (
     <Box key={label} display="inline-flex" alignItems="baseline">
-      <Typography variant="bodyMRegular" color="grey.600">
+      {/* add ellipsis*/}
```
DanielSchiavini, line 12:
```suggestion
```

OnlyJousting:
SCR-20250123-nhgc
there is some clipping in the nav bar at some size but nothing critical

OnlyJousting (approved):

DanielSchiavini:
@OnlyJousting yeah I changed it so it clips the crvUSD price. Note however that the Llama(beta) item isn't visible normally. And the "risk disclaimer" is going to be removed

PR #618, opened 2025-01-22
feat: add api status page link to footer

0xAlunara:
image

### DIFF
--- a/packages/curve-ui-kit/src/widgets/Footer/Sections.tsx
+++ b/packages/curve-ui-kit/src/widgets/Footer/Sections.tsx
@@ -5,6 +5,7 @@ import GitHubIcon from '@mui/icons-material/GitHub'
 import BeenhereOutlinedIcon from '@mui/icons-material/BeenhereOutlined'
 import BugReportOutlinedIcon from '@mui/icons-material/BugReportOutlined'
 import HelpOutlineIcon from '@mui/icons-material/HelpOutline'
+import MonitorHeartOutlinedIcon from '@mui/icons-material/MonitorHeartOutlined'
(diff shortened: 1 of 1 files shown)


packages/curve-ui-kit/src/widgets/Footer/Sections.tsx
```diff
@@ -143,6 +144,11 @@ export const getSections = (appName: AppName): Omit<SectionProps, 'networkName'>
         href: 'https://dune.com/mrblock_buidl/Curve.fi',
         icon: <DuneIcon />,
       },
+      {
+        label: t`API Status`,
+        href: 'https://statuspage.freshping.io/59335-CurveAPI',
```
DanielSchiavini, line 149: how about the prices api?

DanielSchiavini (approved):

OnlyJousting (approved):

packages/curve-ui-kit/src/widgets/Footer/Sections.tsx
0xAlunara: Should ask @mimaklas if we can get a fixed URL that points to a status page that can be updated to include new endpoints.

0xAlunara:
Do not merge yet, tutti is having second thoughts.

packages/curve-ui-kit/src/widgets/Footer/Sections.tsx
mimaklas: hmm i am not aware of a status page for curve-prices api

0xtutti (approved):

PR #620, opened 2025-01-23
fix: console logging

DanielSchiavini:
- fixed missing `%c` in the format string
- escape %c in the data
- removed logs that were too long
- remove unused code
- export `LogStatus` for custom logging

### DIFF
--- a/apps/main/src/dex/components/PageRouterSwap/utils.ts
+++ b/apps/main/src/dex/components/PageRouterSwap/utils.ts
@@ -7,7 +7,7 @@ import orderBy from 'lodash/orderBy'
 import uniq from 'lodash/uniq'
 
 import { NETWORK_TOKEN } from '@main/constants'
-import { log } from '@ui-kit/lib'
+import { log, LogStatus } from '@ui-kit/lib'
 import { weiToEther } from '@ui-kit/utils'
 import { Pool, UserBalancesMapper, PoolData, UsdRatesMapper } from '@main/types/main.types'
 
@@ -65,7 +65,7 @@ export function parseRouterRoutes(
         try {
           pool = getPool(route.poolId)
         } catch (error) {
-          log('routerBestRouteAndOutput missing poolName', route)
+          log('routerBestRouteAndOutput missing poolName', LogStatus.ERROR, route.poolId)
         }
       }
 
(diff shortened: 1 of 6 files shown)


OnlyJousting (approved):

packages/curve-ui-kit/src/features/switch-chain/ui/ChainSwitcher.tsx
```diff
@@ -51,7 +51,7 @@ export const ChainSwitcher = <TChainId extends number>({
   return (
     <>
       <IconButton size="small" disabled={disabled} onClick={onClick} data-testid="btn-change-chain">
-        <ChainSwitcherIcon chain={selectedNetwork} />
+        {selectedNetwork && <ChainSwitcherIcon chain={selectedNetwork} />}
```
0xAlunara, line 54: uuuh ok

apps/main/src/dex/components/PageRouterSwap/utils.ts
```diff
@@ -65,7 +65,7 @@ export function parseRouterRoutes(
         try {
           pool = getPool(route.poolId)
         } catch (error) {
-          log('routerBestRouteAndOutput missing poolName', route)
+          log('routerBestRouteAndOutput missing poolName', LogStatus.ERROR, route.poolId)
```
0xAlunara, line 68: Not important but I'd be in favor of removing specific function variants like `logSuccess` for the more generic variant with logstatus parameter like `log("foo", LogStatus.SUCCESS)

Or instead of this variant use `logError`. Either way pick a side 😄

0xAlunara (approved):

packages/curve-ui-kit/src/features/switch-chain/ui/ChainSwitcher.tsx
DanielSchiavini: This would crash sometimes during load

apps/main/src/dex/components/PageRouterSwap/utils.ts
DanielSchiavini: Makes sense. The `logSuccess` and `logError` functions expect a query key though (we could change that)

DanielSchiavini: I merged now, but we can improve separately

PR #629, opened 2025-01-24
feat: hide USD values > 100T

DanielSchiavini:
- sometimes our APIs will return incorrect dollar values
- ideally we want to be warned about such errors; however, we don't yet have a mechanism for it
- for now, we'll just hide the values from the UI


packages/ui/src/utils/utilsConstants.ts
```diff
@@ -4,6 +4,8 @@ export const CURVE_ASSETS_URL = `${CURVE_CDN_URL}/curve-assets`
 export const CURVE_LOGO_URL = `${CURVE_ASSETS_URL}/branding/logo.png`
 export const NOT_FOUND_IMAGE_URL = `${CURVE_ASSETS_URL}/branding/four-oh-llama.jpg`
 
+export const MAX_USD_VALUE = 100_000_000_000_000 // $ 100T 🤑
```
DanielSchiavini, line 7:
```suggestion
// Sometimes API returns overflowed USD values. Don't show them!
export const MAX_USD_VALUE = 100_000_000_000_000 // $ 100T 🤑
```

michwill (approved):

bout3fiddy (approved):

PR #630, opened 2025-01-24
feat: add prices api

0xAlunara:
Heavily refactored on the Curve Monitor repo to make it as suitable as possible for curve.fi repo.

Package isn't ready for public publishing to the npm registry yet, internal usage is more important in the short term.
It doesn't have to be perfect from the start, we can add and modify params and typing as we use it over time.

I have split the actual refactoring that makes use of this package up into multiple smaller PRs which makes testing and reviewing easier. In those same PRs I will also modify the package code such that suits our needs (think of stuff like function parameters and such).

As such, this PR only *adds* code without using it yet.


DanielSchiavini (changes requested):
Sorry, I don't think we should add all this code to this repository.
If we want to use this code, it should either be a separate repository/npm package or we should only include the code that we use at that point

README.md
```diff
@@ -50,6 +50,7 @@ This repository is organized as follows:
 - `/apps/lend`: Lend [React](https://react.dev/) application.
 - `/tests`: DApp tests
 - `/packages/curve-ui-kit`: Shared UI kit created using Material UI, mapped as `@ui-kit`
+- `/packages/prices-api`: Package for consuming the Prices API, mapped as `@curvefi/prices-api`
```
DanielSchiavini, line 53:
```suggestion
- `/packages/prices-api`: Package for consuming the Prices API, mapped as `@curvefi/prices-api`. Soon to be to separated its own NPM package.
```

packages/prices-api/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "license": "MIT",
+  "private": true,
+  "scripts": {
+    "lint": "eslint \"**/*.ts*\"",
+    "typecheck": "tsc --noEmit"
+  },
+  "peerDependencies": {
+    "typescript": "*"
+  },
+  "devDependencies": {
```
DanielSchiavini, line 16: no dependencies :fire:

packages/prices-api/src/index.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+/**
+ * Gets the API host URL, using the provided host or falling back to default
+ * @param options - Optional configuration options
+ * @returns Promise that resolves to the host URL
+ * @example
+ * const host = await getHost() // "https://prices.curve.fi"
+ */
+export const getHost = (options?: Options): Required<Options>['host'] =>
```
DanielSchiavini, line 21: Why would `getHost` need to be async :thinking:

packages/prices-api/src/timestamp.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Converts a timestamp string or number to UTC unix timestamp in seconds
+ * @param timestamp - Timestamp as string (ISO format) or number (unix seconds)
+ * @returns UTC timestamp in seconds
+ * @example
+ * toUTC(1234567890) // Returns 1234567890
+ * toUTC("2024-01-01T00:00:00Z") // Returns 1704067200
+ * toUTC("2024-01-01T00:00:00.123Z") // Returns 1704067200 (ignores ms)
+ */
+export function toUTC(timestamp: string | number): number {
```
DanielSchiavini, line 10: why do we use timestamps instead of the `Date` object?

packages/prices-api/src/solver/parsers.ts
```diff
@@ -0,0 +1,27 @@
...
+  }, {}),
+  solutions: x.solutions.map((sol) => ({
+    solver: sol.solver,
+    solverAddress: sol.solverAddress,
+    score: BigInt(sol.score),
+    ranking: sol.ranking,
+    clearingPrices: Object.entries(sol.clearingPrices).reduce<Record<Address, bigint>>((acc, [key, value]) => {
+      acc[key as Address] = BigInt(value)
+      return acc
+    }, {}),
```
DanielSchiavini, line 20: We could have a nice helper fn to make these calls clear

```suggestion
    clearingPrices: mapRecord(sol.clearingPrices, (key, value) => [key as Address, BigInt(value)])
```

packages/prices-api/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+  "compilerOptions": {
+    "target": "ES2020",
+    "lib": ["dom", "esnext"],
```
DanielSchiavini, line 4: do we need `dom` here?

image

apps/main/tsconfig.json
```diff
@@ -11,7 +11,8 @@
       "@ui": ["../../../packages/ui/src/index.ts"],
       "@ui/*": ["../../../packages/ui/src/*"],
       "@ui-kit/*": ["../../../packages/curve-ui-kit/src/*"],
-      "@external-rewards": ["../../../packages/external-rewards/src/index.ts"]
+      "@external-rewards": ["../../../packages/external-rewards/src/index.ts"],
+      "@curvefi/prices-api/*": ["../../../packages/prices-api/src/*"]
```
DanielSchiavini, line 15:
```suggestion
      "@curvefi/prices-api": ["../../../packages/prices-api/src/index.ts"],
      "@curvefi/prices-api/*": ["../../../packages/prices-api/src/*"]
```

DanielSchiavini: image

packages/prices-api/src/timestamp.ts
DanielSchiavini: Oh please do not use second timestamps that's confusing as JS usually expects milliseconds

packages/prices-api/src/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  'xdai',
+  'matic',
+  'harmony',
+  'moonbeam',
+  'base',
+  'polygon',
+  'fraxtal',
+] as const
+
+export type Chain = (typeof chains)[number]
```
DanielSchiavini, line 44: another chain type :scream: 
It's not super specific to this package, but we have so many ways to describe chains.
The sad part is that in the other PRs the type is not really used except for casting (which doesn't check anything)

Curve UI has a different one:
https://github.com/curvefi/curve-frontend/blob/5fccbeea6173cfaae837e5f84610de046005f3f2/packages/curve-ui-kit/src/utils/network.ts#L1-L0

Dao uses `1`:
https://github.com/curvefi/curve-frontend/blob/036aeaaac28efa59e269064a9c87c28bc0e4d58b/apps/main/src/dao/types/dao.types.ts#L18

crvUSD too
https://github.com/curvefi/curve-frontend/blob/036aeaaac28efa59e269064a9c87c28bc0e4d58b/apps/main/src/loan/types/loan.types.ts#L16

Dex uses just any number lol
https://github.com/curvefi/curve-frontend/blob/036aeaaac28efa59e269064a9c87c28bc0e4d58b/apps/main/src/dex/types/main.types.ts#L17

Lending uses an alias of this
https://github.com/curvefi/curve-lending-js/blob/2958c2efaaa383a27f71ab53ca1e4fc293eee374/src/interfaces.ts#L8

packages/prices-api/src/index.ts
0xAlunara: Good call; a leftover from before the refactor from class to functions-only. Property and function are now made non-async. I keep the `const host = getHost(options)` for readability in the api functions. When there's more query parameters being used inlining becomes a bit unreadable.

packages/prices-api/tsconfig.json
0xAlunara: I live in the past where fetch used to be part of dom, but it's a global now, so it's removed 👍

packages/prices-api/src/solver/parsers.ts
0xAlunara: Gets quite hairy wrt typing and generics and I don't want to introduce a lodash dependency. However I've changed the function from using `reduce` to `fromEntries(entries)`, it's a little bit less verbose this way.

packages/prices-api/src/index.ts
0xAlunara: Preferably I'd keep the (exported) chain type of this package internal, but sadly TYpescript doesn't have anything to mark things as `internal` without maybe resorting to hacks 😢

packages/prices-api/src/timestamp.ts
0xAlunara: This function now returns a `Date` object, and all API model timestamps have been updated accordingly.

OnlyJousting (approved):

packages/prices-api/src/solver/parsers.ts
DanielSchiavini: fromEntries is nice indeed. FYI I created this for another project:
```typescript
export function mapRecord<K extends string, V, R>(obj: Record<K, V>, mapper: (key: K, value: V) => R): Record<K, R> {
  const entries = Object.entries(obj).map(([k, v]) => [k, mapper(k as K, v as V)]);
  return Object.fromEntries(entries) as Record<K, R>;
}
```

packages/prices-api/src/fetch.ts
```diff
@@ -0,0 +1,48 @@
...
+      const json = (await resp.json()) as T
+
+      return json
+    }
+  } catch (err) {
+    if (err instanceof Error) {
+      throw err
+    }
+
+    throw new Error(`Unknown fetch error`)
```
DanielSchiavini, line 46: Why hide the underlying exception?
```suggestion
    throw err
```

DanielSchiavini (approved):

packages/prices-api/src/solver/parsers.ts
0xAlunara: Will copy this for the next PR like the try/catch that can be removed.

PR #632, opened 2025-01-25
refactor: use prices api for llamma ohlc

0xAlunara:
- depends on #630

### DIFF
--- a/apps/main/src/global-extensions.ts
+++ b/apps/main/src/global-extensions.ts
@@ -1,0 +1,13 @@
+import type { UTCTimestamp } from 'lightweight-charts'
+
+declare global {
+  interface Date {
+    getUTCTimestamp(): UTCTimestamp
+  }
+}
+
+Date.prototype.getUTCTimestamp = function () {
+  return (this.getTime() / 1000) as UTCTimestamp
+}
+
+export {}
--- a/apps/main/src/lend/store/createOhlcChartSlice.ts
+++ b/apps/main/src/lend/store/createOhlcChartSlice.ts
@@ -422,15 +422,19 @@ const createOhlcChart = (set: SetState<State>, get: GetState<State>) => ({
         let ohlcDataArray: LpPriceOhlcDataFormatted[] = []
 
         for (const item of ohlc) {
-          baselinePriceArray.push({
-            time: item.time.getUTCTimestamp(),
-            base_price: item.basePrice,
-          })
+          if (item.basePrice) {
+            baselinePriceArray.push({
+              time: item.time.getUTCTimestamp(),
+              base_price: item.basePrice,
+            })
+          }
 
-          oraclePriceArray.push({
-            time: item.time.getUTCTimestamp(),
-            value: item.oraclePrice,
-          })
+          if (item.oraclePrice) {
+            oraclePriceArray.push({
+              time: item.time.getUTCTimestamp(),
+              value: item.oraclePrice,
+            })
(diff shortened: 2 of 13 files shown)


apps/main/src/lend/store/createOhlcChartSlice.ts
```diff
@@ -308,35 +309,40 @@ const createOhlcChart = (set: SetState<State>, get: GetState<State>) => ({
...
-            base_price: item.base_price,
+            time: item.time as UTCTimestamp,
+            base_price: item.priceBase,
           })
 
           oraclePriceArray.push({
-            time: convertToLocaleTimestamp(item.time) as UTCTimestamp,
-            value: item.oracle_price,
+            time: item.time as UTCTimestamp,
+            value: item.priceOracle,
```
DanielSchiavini, line 341: priceOracle sounds like the address of the oracle.
oraclePrice seems like the correct name here

apps/main/src/lend/store/createOhlcChartSlice.ts
```diff
@@ -308,35 +309,40 @@ const createOhlcChart = (set: SetState<State>, get: GetState<State>) => ({
...
-        const llammaOhlcFetch = await fetch(
-          `https://prices.curve.fi/v1/lending/llamma_ohlc/${network}/${poolAddress}?agg_number=${interval}&agg_units=${timeUnit}&start=${start}&end=${end}`,
-        )
-        const llammaOhlcResponse: LlammaOhlcApiResponse = await llammaOhlcFetch.json()
+        const ohlc = await getOHLC({
+          endpoint: 'lending',
+          chain: network as Chain,
+          llamma: poolAddress as Address,
+          interval,
+          units: timeUnit as Parameters<typeof getOHLC>[0]['units'],
```
DanielSchiavini, line 317: Should we introduce a `TimeUnit` type?

OnlyJousting (dismissed):

apps/main/src/lend/store/createOhlcChartSlice.ts
0xAlunara: The API is rather inconsistent with its time options, hence I 'hardcoded' them instead of defining a type.

0xAlunara:
Also had to add support for /v1/lending/oracle_ohlc as it was used by the llame lend market ohlc

apps/main/src/global-extensions.ts
```diff
@@ -0,0 +1,13 @@
+import type { UTCTimestamp } from 'lightweight-charts'
+
+declare global {
+  interface Date {
+    getUTCTimestamp(): UTCTimestamp
```
DanielSchiavini, line 5: This does nothing related to UTC.
For consistency with `getTime` I would suggest
```suggestion
    getTimeInSeconds(): UTCTimestamp
```

DanielSchiavini (approved):

apps/main/src/global-extensions.ts
0xAlunara: It also converts from Javascript's milliseconds to 'normal' UTC timestamp in seconds, as expected by charting libraries. Given the prices API now uses Date's that's a lot of instances of `(x.getTime() / 1000) as UTCTimestamp` without the prototype. The casting is necessary btw, as it's a nominal type and lightweight charts doesn't accept a raw `number`.

OnlyJousting (approved):

PR #636, opened 2025-01-25
refactor: use prices api for dao daily locks

0xAlunara:
- depends on #635

PR #635, opened 2025-01-25
refactor: use prices api for dao fees

0xAlunara:

PR #637, opened 2025-01-25
refactor: use prices api for dao lockers

0xAlunara:
- depends on #636

### DIFF
--- a/apps/main/src/dao/components/PageAnalytics/CrvStats/index.tsx
+++ b/apps/main/src/dao/components/PageAnalytics/CrvStats/index.tsx
@@ -25,9 +25,7 @@ const CrvStats: React.FC = () => {
     }
   }, [veCrvData.totalCrv, veCrvData.fetchStatus, getVeCrvData, provider])
 
-  const veCrvApr = aprLoading
-    ? 0
-    : calculateApr(veCrvFees.fees[1].feesUsd, Number(veCrvData.totalVeCrv) / 10 ** 18, crv)
+  const veCrvApr = aprLoading ? 0 : calculateApr(veCrvFees.fees[1].feesUsd, veCrvData.totalVeCrv.fromWei(), crv)
 
   const loading = Boolean(provider && veCrvLoading)
   return (
@@ -40,7 +38,7 @@ const CrvStats: React.FC = () => {
             title={t`Total CRV`}
             data={
               <MetricsColumnData>
-                {noProvider ? '-' : formatNumber(Number(veCrvData.totalCrv) / 10 ** 18, { notation: 'compact' })}
+                {noProvider ? '-' : formatNumber(veCrvData.totalCrv.fromWei(), { notation: 'compact' })}
               </MetricsColumnData>
             }
           />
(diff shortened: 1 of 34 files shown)


apps/main/src/dao/components/PageAnalytics/CrvStats/index.tsx
```diff
@@ -39,7 +41,7 @@ const CrvStats: React.FC = () => {
             title={t`Total CRV`}
             data={
               <MetricsColumnData>
-                {noProvider ? '-' : formatNumber(veCrvData.totalCrv, { notation: 'compact' })}
+                {noProvider ? '-' : formatNumber(Number(veCrvData.totalCrv) / 10 ** 18, { notation: 'compact' })}
```
DanielSchiavini, line 44: Why do we have `10**18` all over the code instead of the parsing code?

packages/prices-api/src/dao/api.ts
```diff
@@ -24,9 +25,21 @@ export async function getUserLocks(user: string, options?: Options) {
...
   const host = await getHost(options)
-  const resp = await fetch<Responses.GetLockersResponse>(`${host}/v1/dao/lockers/${top}`)
+  const fs = (page: number) =>
+    fetch<Responses.GetLockersResponse>(`${host}/v1/dao/lockers?pagination=1000&page=${page}`).then((resp) =>
+      resp.locks.map(Parsers.parseLockers),
+    )
+
+  const lockers = await paginate(fs, 1, 100)
+
+  return lockers
```
DanielSchiavini, line 37:
```suggestion
  return paginate(fs, 1, 100)
```

apps/main/src/dao/components/PageAnalytics/TopHoldersChart/TopHoldersBarChartComponent.tsx
```diff
@@ -39,14 +40,20 @@ const TopHoldersBarChart: React.FC<TopHoldersBarChartProps> = ({ data, filter })
     return user.length > 15 ? (shortenTokenAddress(user)?.toString() ?? user) : user
   }
 
+  const dataFormatted = data.map((x) => ({
+    ...x,
+    weight: Number(x.weight) / 10 ** 18,
```
DanielSchiavini, line 45: 10**18 doesn't belong in components IMO

apps/main/src/dao/components/PageAnalytics/CrvStats/index.tsx
0xAlunara: See Slack discussion wrt formatting values from the store. As suggested, I could create a util function (`fromGwei`), but I'd prefer to implement that in another PR, together with some other global fixes that I need to apply as a result of the other refactor PRs and their review feedback.

DanielSchiavini (commented):
Looks nice, please have a look at this [date comment](https://github.com/curvefi/curve-frontend/pull/635#pullrequestreview-2594927535) and let's create the `fromWei` function as discussed

DanielSchiavini (dismissed):

OnlyJousting (approved):

DanielSchiavini (dismissed):

DanielSchiavini (approved):

OnlyJousting (approved):

PR #639, opened 2025-01-27
refactor: use wallet hook instead of store

DanielSchiavini:
- replace wallet store by wallet hook
- remove notify function from hook
- for reviewers: almost everything are just import differences. Important changes are all in [hooks.ts](https://github.com/curvefi/curve-frontend/blob/refactor/wallet-hook/packages/curve-ui-kit/src/features/connect-wallet/lib/hooks.ts)
- includes https://github.com/curvefi/curve-frontend/pull/651


0xAlunara (dismissed):
Probably could've just used `import { notify } from '@ui-kit/features/connect-wallet'` with out `as notifyNotification`, but not too important.

DanielSchiavini:
> Probably could've just used `import { notify } from '@ui-kit/features/connect-wallet'` 

Possible but it would result in many more changes

DanielSchiavini:
> Possible but it would result in many more changes

WIP in https://github.com/curvefi/curve-frontend/pull/651

OnlyJousting (approved):

0xAlunara (approved):

PR #644, opened 2025-01-28
feat: remove llamalend preheader

DanielSchiavini:
- removes the large explanation from the top of the table


0xtutti (approved):

0xAlunara (approved):
Figured you'd just not use the component or set display to `none`, but deleting works as well I guess. If you ever need to restore it you'll have to do a bit of searching.

PR #645, opened 2025-01-28
Added support for 'multiplier' field visibility in token campaigns

OnlyJousting:
Enable external rewards campaign that use the type 'token' to display the multiplier field if the field is not empty or null.


mimaklas (dismissed):

packages/ui/src/CampaignRewards/CampaignRewardsComp.tsx
```diff
@@ -10,7 +10,7 @@ import Icon from 'ui/src/Icon'
 const RewardsCompSmall: React.FC<CampaignRewardsCompProps> = ({ rewardsPool, highContrast, mobile, banner }) => {
   const { platform, multiplier, platformImageSrc } = rewardsPool
 
-  const isPoints = rewardsPool.tags.includes('points')
+  const hasMultiplier = !!multiplier && multiplier !== ''
```
DanielSchiavini, line 13: `''` is falsey (i.e. `!!'' === false`)

```suggestion
  const hasMultiplier = !!multiplier
```

mimaklas (approved):

DanielSchiavini (approved):

PR #648, opened 2025-01-28
feat: separate graph columns

DanielSchiavini:
- separate the chart columns from the rate columns
- get rid of unnecessary custom feature hiding logic


apps/main/src/loan/components/PageLlamaMarkets/cells/LineGraphCell.tsx
```diff
@@ -34,65 +28,40 @@ const calculateDomain =
...
+          <Line
+            type="monotone"
+            dataKey={snapshotKey}
+            stroke={getColor(design, snapshots, snapshotKey)}
+            strokeWidth={1}
+            dot={<></>}
+          />
+        </LineChart>
+      ) : isLoading ? (
+        <Skeleton {...graphSize} />
```
0xAlunara, line 59: There's no need for the hardcoded `graphSize` when you use the linechart as its child; the skeleton will infer the size. You can see an example in `Metric.tsx`:
```tsx
      {loading ? (
        <Skeleton variant="text">
          <MetricValue {...metricValueProps} />
        </Skeleton>
      ) : (
        <MetricValue {...metricValueProps} />
      )}
```
You'll have to extract the LineChart above into a component that you can invoke twice. Or you can create a Skeleton helper component that basically just wraps its `children` property inside a `<Skeleton>` if the provided property `isLoading` is set to true. I didn't do that for the metric component, but I could've.

0xAlunara (changes requested):
You can now 'sort' the table by the line chart, which doesn't make sense to me. Not sure what the sorting algorithm even does in this scenario.

apps/main/src/loan/components/PageLlamaMarkets/cells/LineGraphCell.tsx
DanielSchiavini: graphSize was hardcoded for the chart lib, not necessarily for the skeleton

DanielSchiavini:
> You can now 'sort' the table by the line chart, which doesn't make sense to me. Not sure what the sorting algorithm even does in this scenario.

It sorts by the average. I can disable sorting though, that's easy

0xAlunara (approved):

0xAlunara (approved):

PR #658, opened 2025-01-31
Make wallet connect prompt spicy

bout3fiddy:


0xAlunara:
No clear goal, tutti is against I think and we're trying to get rid of this prompt anyway.

PR #662, opened 2025-02-03
feat: improve Tether search

DanielSchiavini:
- by default, replace all '₮' with 'T' in the data
- however, allow users to search tether-only by using the `₮` symbol

### DIFF
--- a/apps/main/package.json
+++ b/apps/main/package.json
@@ -39,7 +39,7 @@
     "ethers": "^6.13.4",
     "external-rewards": "*",
     "focus-visible": "5.2.0",
-    "fuse.js": "^6.6.2",
+    "fuse.js": "^7.1.0",
     "immer": "^9.0.12",
     "lightweight-charts": "4.1.2",
     "lodash": "4.17.21",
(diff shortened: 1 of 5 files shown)


0xAlunara (commented):
I'm a bit confused:

1. Searching for aUSD₮ works fine
image

2. Searching for aUSDT works fine too, although I'm getting more results than with aUSD₮
image

3. Searching for T gives lots of results
image

4. Searching for ₮ gives zero results.
image

My understanding would be that ₮ simple would be replaced with a T, but the search results aren't what I'd expect them to be.

DanielSchiavini:
@0xAlunara I am not replacing ₮ in the search term. I only replace ₮ in the pools when there is no ₮ in the search terms.

The search currently requires 2 minimum character matches - so d₮ shows results, but ₮ does not. "T" will only give results in pools where the letter is there at least 2 times :rofl:  which doesn't make much sense but it's not new

0xAlunara (approved):

OnlyJousting (approved):

PR #668, opened 2025-02-05
chore: update to next 14

DanielSchiavini:
- update to next [v14](https://nextjs.org/docs/pages/building-your-application/upgrading/version-14)
- minor updates other dependencies
- removed lingui
  - it is not used 
  - its wasm plugin keeps on [crashing](https://github.com/lingui/swc-plugin/issues/25) on build
  - created fake `t` and `Trans`, mapped `"@lingui/macro"` imports to `"curve-ui-kit/src/lib/i18n.ts"`
  - we can replace these imports in a later PR
- removed unused dependencies
- refactor react-aria SelectProps type issues
  - removed some unused selects
  - this is the part that needs some testing


packages/ui/src/Select/select.overrides.d.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Override the types of the `Select` and `ListBox` components to add the `children` prop.
+ * This is for backwards compatibility with the existing components that are currently passing the children prop.
+ *
+ * --- Some background
+ * They changed they API like e.g.: `AriaSelectOptions<T> = Omit<AriaSelectProps<T>, 'children'>`
+ * Their source code, however, always passes all given props back to the caller (at this moment).
+ * The components seem to work fine, but the types are broken.
+ *
+ * TODO Refactor components to render their own children, or replace with MUI
```
DanielSchiavini, line 10: Select is a good candidate for MUI replacement

apps/main/src/pages/lend.tsx
```diff
@@ -2,31 +2,24 @@ import type { NextPage } from 'next'
...
 import GlobalStyle from '@/lend/globalStyle'
 import Page from '@/lend/layout'
-import { dynamicActivate, initTranslation } from '@ui-kit/lib/i18n'
-import { messages as messagesEn } from '@/locales/en/messages.js'
 import networks from '@/lend/networks'
 import { getPageWidthClassName } from '@/lend/store/createLayoutSlice'
 import useStore from '@/lend/store/useStore'
 import { QueryProvider } from '@ui/QueryProvider'
 import { isMobile, removeExtraSpaces } from '@/lend/utils/helpers'
-import { getLocaleFromUrl } from '@/lend/utils/utilsRouter'
```
0xAlunara, line 23: Not important for this URL, but lots of locale code remaining that we can remove now in a new PR.

apps/main/src/pages/dex.tsx
```diff
@@ -97,11 +91,6 @@ const App: NextPage = () => {
     // init locale
     const { rLocale } = getLocaleFromUrl()
     const parsedLocale = rLocale?.value ?? 'en'
-    initTranslation(i18n, parsedLocale)
```
0xAlunara, line 100:
```tsx
const { rLocale } = getLocaleFromUrl()
const parsedLocale = rLocale?.value ?? 'en'
```
can be removed too

apps/main/src/pages/dao.tsx
```diff
@@ -77,14 +71,7 @@ const App: NextPage = () => {
...
     const { rLocale } = getLocaleFromUrl()
-    const parsedLocale = rLocale?.value ?? 'en'
-    initTranslation(i18n, parsedLocale)
-    ;(async () => {
-      let data = await import(`@/locales/${parsedLocale}/messages`)
-      dynamicActivate(parsedLocale, data)
-    })()
-    setLocale(parsedLocale)
-    updateAppLocale(parsedLocale)
+    setLocale(rLocale?.value ?? 'en')
```
0xAlunara, line 74:
```tsx
const { rLocale } = getLocaleFromUrl()
setLocale(rLocale?.value ?? 'en')
```
can be removed too

packages/curve-ui-kit/src/lib/i18n.ts
```diff
@@ -54,40 +37,21 @@ export function parseLocale(locale?: string): { parsedLocale: Locale['value']; p
...
+/**
+ * Placeholder for a translation function, so that we can introduce a real one later.
+ * @param key Key to translate. This is an array if there are variables to interpolate.
+ * @param template Variables to interpolate. They are appended to the keys of the same index.
+ * @returns The original string, or the string with the variables interpolated.
+ */
+export const t = (key: string | readonly string[], ...template: unknown[]) =>
+  Array.isArray(key) ? key.map((k, i) => (i < template.length ? `${k}${template[i]}` : k)).join('') : (key as string)
 
-  if (numbroLang) {
```
0xAlunara, line 87: If we're no longer configuring numbro I wonder if we can just get rid of the package alltogether maybe? Or at least get rid of the top import `import 'numbro/dist/languages.min.js'`. Maybe get rid of Dayjs too...

packages/ui/src/SelectNetwork/SelectNetwork.tsx
```diff
@@ -34,7 +34,7 @@ export const SelectNetwork: React.FC<SelectNetworkProps> = ({
 }) => (
   // TODO: add darkTheme icon or make sure icon work for both dark and light theme
   <Select {...props} className={className} items={items} aria-label="Select network" label="">
-    {({ chainId, src, label }) => (
+    {({ chainId, src, label }: SelectNetworkItem) => (
```
0xAlunara, line 37: Is this component actually still used now we have a new network selector in curve-ui-kit?

0xAlunara (changes requested):
There seems to be a difference wrt some font sizes:

New:
image

Old:
image

Similarly on the pool pages:
image
image

<hr >
On the pool page the disabled style of the buttons is blue, not grey?

New:
image

Old:
image

<hr>

Tooltip [i] seems to be aligned incorrect? (on pool page)

New:
image

Old:
image

<hr>
Select component on crvusd page hover effect looks wonky

New:
image

Old:
image

<hr>

It seems the chips have no hover effect at all anymore
image

<hr>

The proposal page colored status circle has disappeared

New
image

Old
image

apps/main/src/pages/lend.tsx
DanielSchiavini: sounds like a good plan

apps/main/src/pages/dex.tsx
DanielSchiavini: let's do that separate cause then I wanna change the store too

packages/curve-ui-kit/src/lib/i18n.ts
DanielSchiavini: I'll get rid of the languages import. It's still used in 2 files, we can fix that separately
https://github.com/search?q=repo%3Acurvefi%2Fcurve-frontend%20numbro&type=code

DanielSchiavini:
@0xAlunara thanks for the findings, they were caused by a breaking change in styled components
Fixed in [c0f496c](https://github.com/curvefi/curve-frontend/commits/c0f496c404d96a6f8c6946ba8ff0e9c820b94f37)

0xAlunara (approved):
There's no need to separate it—it's too much effort when it's already here. I quickly reviewed the entire website, and everything appears resolved; I couldn't find any discrepancies. If anything else shows up as broken, we can fix it in another PR.

0xtutti (approved):

OnlyJousting (approved):

PR #669, opened 2025-02-05
refactor: prices api minor feedback

0xAlunara:
* Turns out lightweight charts prefer to have the actual data adjusted to the local time zone as per https://tradingview.github.io/lightweight-charts/docs/time-zones#how-to-add-time-zone-support-to-your-chart.
* Adds `bigint.prototype.fromWei()` to easily convert bigints to a presentational number
* Renames 'toUTC' to 'toDate' such that the function's purposes is clearer
* Lodash doesn't have a suitable alternative for the `combinations` function we've implemented ourselves, so that has been kept untouched.
* Depends on https://github.com/curvefi/curve-frontend/pull/637

### DIFF
--- a/apps/main/src/dao/components/PageAnalytics/CrvStats/index.tsx
+++ b/apps/main/src/dao/components/PageAnalytics/CrvStats/index.tsx
@@ -25,9 +25,7 @@ const CrvStats: React.FC = () => {
     }
   }, [veCrvData.totalCrv, veCrvData.fetchStatus, getVeCrvData, provider])
 
-  const veCrvApr = aprLoading
-    ? 0
-    : calculateApr(veCrvFees.fees[1].feesUsd, Number(veCrvData.totalVeCrv) / 10 ** 18, crv)
+  const veCrvApr = aprLoading ? 0 : calculateApr(veCrvFees.fees[1].feesUsd, veCrvData.totalVeCrv.fromWei(), crv)
 
   const loading = Boolean(provider && veCrvLoading)
   return (
@@ -40,7 +38,7 @@ const CrvStats: React.FC = () => {
             title={t`Total CRV`}
             data={
               <MetricsColumnData>
-                {noProvider ? '-' : formatNumber(Number(veCrvData.totalCrv) / 10 ** 18, { notation: 'compact' })}
+                {noProvider ? '-' : formatNumber(veCrvData.totalCrv.fromWei(), { notation: 'compact' })}
               </MetricsColumnData>
             }
           />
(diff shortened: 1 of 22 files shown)


OnlyJousting (approved):

apps/main/src/dao/components/PageAnalytics/CrvStats/index.tsx
```diff
@@ -25,9 +25,7 @@ const CrvStats: React.FC = () => {
     }
   }, [veCrvData.totalCrv, veCrvData.fetchStatus, getVeCrvData, provider])
 
-  const veCrvApr = aprLoading
-    ? 0
-    : calculateApr(veCrvFees.fees[1].feesUsd, Number(veCrvData.totalVeCrv) / 10 ** 18, crv)
+  const veCrvApr = aprLoading ? 0 : calculateApr(veCrvFees.fees[1].feesUsd, veCrvData.totalVeCrv.fromWei(), crv)
```
DanielSchiavini, line 28: Ideally we should move all `fromWei` operation to the API parsing IMO. That would have 2 advantages:
- separating the business logic from the UI
- only repeating for every response, not for every render

You mentioned that you wanted to keep `bignumber` for eventual extra calculations, and maybe `fromWei` could even keep it a bignumber (although by our tests, Number is precise enough).

apps/main/src/global-extensions.ts
```diff
@@ -3,11 +3,24 @@ import type { UTCTimestamp } from 'lightweight-charts'
 declare global {
   interface Date {
     getUTCTimestamp(): UTCTimestamp
+    getLocalTimestamp(): UTCTimestamp
```
DanielSchiavini, line 6: there is a risk someone else implemented the same extension :warning:

packages/prices-api/src/timestamp.ts
```diff
@@ -2,15 +2,16 @@ const TZ_OFFSET = /[+-]\d{2}:?\d{2}$/
...
- * toUTC("2024-01-01T00:00:00.000Z+01:00")
- * toUTC("2024-01-01T00:00:00") // (assumes UTC)
+ * toDate(1234567890) // Unix timestamp number
+ * toDate("1234567890") // Unix timestamp string
+ * toDate("2024-01-01T00:00:00.000Z") // ISO with UTC
+ * toDate("2024-01-01T00:00:00.000+01:00") // ISO with timezone
+ * toDate("2024-01-01T00:00:00") // ISO without timezone (assumes UTC)
  */
-export function toUTC(timestamp: string | number): Date {
+export function toDate(timestamp: string | number): Date {
```
DanielSchiavini, line 14: `parseDate` would be clearer imo, but this is good enough

DanielSchiavini (approved):

apps/main/src/dao/components/PageAnalytics/CrvStats/index.tsx
0xAlunara: I was thinking of maybe making toGwei lazy and apply memoization, but the calculation is so small and fast I wonder if it makes any difference at all 🤔

PR #671, opened 2025-02-05
feat: add discussion link to dao submenu

0xAlunara:
Not the prettiest PR, but now that the apps are merged we should probably revise routing and menu generation anyway in a different PR (and it's also a bit buggy as well)

Either way, this PR adds a link to the governance forum to the DAO app. It used to be a separate top level link, but it makes more sense to group it under the DAO app.
image


OnlyJousting (approved):

DanielSchiavini (approved):
> we should probably revise routing 

:100:

PR #678, opened 2025-02-10
chore: update fantom explorer

DanielSchiavini:
Note that the new explorer does not seem to have a token page, so that results in 404.


0xtutti (approved):

DanielSchiavini:
Mima wants to wait until a bit closer to feb 20

PR #682, opened 2025-02-12
feat: add user llamalend columns

DanielSchiavini:
- Add columns with user data to the llamalend table
  - Add "my markets" filter
  - Show collateral erosion warning
- Allow hiding a list of columns at once
- Change minimum filter to work as a range filter
- Create InvertOnHover component
  - Rewrite the multiselect dropdown because hover broke the mui select
  - Use hover also on chain switcher


packages/curve-ui-kit/src/shared/ui/DataTable.tsx
```diff
@@ -30,17 +33,20 @@ const getAlignment = <T extends TableItem>({ columnDef }: Column<T>) =>
...
-        variant={column.columnDef.meta?.variant ?? 'tableCellMBold'}
+        variant={variant ?? 'tableCellMBold'}
         color="text.primary"
         component="td"
         sx={{
           textAlign: getAlignment(column),
           paddingInline: Spacing.sm,
           paddingBlock: Spacing.xs, // `md` removed, content should be vertically centered
           ...getExtraColumnPadding(column),
+          transition: `color ${TransitionFunction}, border-right ${TransitionFunction}`,
```
0xAlunara, line 48: Is this transition still necessary now that it's part of `create-typography.ts` on line 64?

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
```diff
@@ -134,8 +130,9 @@ const HeaderCell = <T extends TableItem>({ header }: { header: Header<T, unknown
             '&:hover': {
               color: `text.highlight`,
             },
-            transition: `color ${TransitionFunction}`,
+            transition: `color ${TransitionFunction}, border-right ${TransitionFunction}`,
```
0xAlunara, line 133: Same here, still required?

apps/main/src/loan/entities/chains.ts
```diff
@@ -1,18 +1,22 @@
...
 
-export const { getQueryOptions: getSupportedLendingChainOptions } = queryFactory({
+const { getQueryOptions: getSupportedLendingChainOptions } = queryFactory({
   queryKey: () => ['prices-api', 'supported-lending-chains'] as const,
   queryFn: getChains,
   staleTime: '1d',
   validationSuite: EmptyValidationSuite,
 })
+
+export const getSupportedLendingChains = () => queryClient.fetchQuery(getSupportedLendingChainOptions({}))
```
0xAlunara, line 21: imo `fetchQuery` should be returned from `queryFactory` similarly to how it returns `getQueryData`, but can do that in a diff PR

apps/main/src/loan/components/PageLlamaMarkets/cells/index.ts
```diff
@@ -2,4 +2,5 @@ export * from './CompactUsdCell'
 export * from './MarketTitleCell'
 export * from './LineGraphCell'
 export * from './RateCell'
-export * from './UtilizationCell'
+export * from './PercentageCell'
+export * from '@/loan/components/PageLlamaMarkets/cells/ErrorCell'
```
0xAlunara, line 6: Relative import?

apps/main/src/loan/components/PageLlamaMarkets/Page.tsx
```diff
@@ -26,16 +30,22 @@ const { Spacing, MaxWidth, ModalHeight } = SizesAndSpaces
...
-  const { data, isLoading, isError } = useLlamaMarkets() // todo: show errors and loading state
+  const { signerAddress } = useWallet()
+  const { data, isError, isLoading } = useLlamaMarkets(signerAddress)
   const bannerHeight = useStore((state) => state.layout.height.globalAlert)
   const headerHeight = useHeaderHeight(bannerHeight)
   usePageOnMount() // required for connecting wallet
   return (
     <Box sx={{ marginBlockEnd: Spacing.xxl }}>
-      {isLoading ? (
+      {!data && (!isError || isLoading) ? (
```
0xAlunara, line 40: why is simply using isLoading not sufficient? i'd expect isLoading to be false if there's no data but an error regardless

0xAlunara (commented):
The background change (in my case from black to white) on hover is instant, while the rest of the color transitions aren't. Broken or deliberate?

image

<hr>

image

1. `$x%` being the average rate is new to me... Not sure if this is a mistake or deliberate, either way confusing to me.
2. The 'Current Rate' tooltip gets in the way getting the 'Average Rate' tooltip from the number above. There's no point in making the contents of these tooltips selectable, so you can try the following (copied from `TokenIcon` component)
```tsx
  <Tooltip
    title={tooltip}
    placement="top"
    slotProps={{
      popper: {
        sx: {
          userSelect: 'none',
          pointerEvents: 'none',
        },
      },
    }}
  >
```
Same holds for a few other tooltips on the table, like 'copy market address'.

<hr>

image

1. It's really hard to at first glance see which chips are enabled and which are not. I know the enabled chips have a X icon but it requires me to look, think and process for another 200ms.
2. I can't disable both 'mint markets' and 'lend markets' when I enable one or the other. I explicitly have to click 'Reset Filters'. A bit odd.

<hr>

image

Asked tutti in slack but don't want it to get lost: should we really add `-` everywhere? Looks probably calmer if we opt for keeping cells empty if there's no data.

<hr>

image

1. We should add a debounce to these slider values, because if you currently slide it will instantly re-render the table for every miniscule step change. This kills the performance.
2. Should we add a shadow to the dropdown? The way it overlaps with the datatable columns headers makes it hard to distinguish, given they live on the same layer.

packages/prices-api/src/crvusd/api.ts
```diff
@@ -49,19 +49,15 @@ export async function getKeepers(chain: Chain, options?: Options) {
 
 export async function getUserMarkets(userAddr: string, chain: Chain, options?: Options) {
   const host = getHost(options)
-  const resp = await fetch<Responses.GetUserMarketsResponse>(
-    `${host}/v1/crvusd/users/${chain}/${userAddr}?page=1&per_page=100`,
-  )
-
+  const resp = await fetch<Responses.GetUserMarketsResponse>(`${host}/v1/crvusd/users/${chain}/${userAddr}`)
```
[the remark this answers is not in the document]
DanielSchiavini: great catch! stats returns a single record but I'll make this 100

packages/curve-ui-kit/src/utils/dom.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Check if the target has a parent with the given class name.
+ * Optionally, you can specify a tag name to stop the search at.
+ * @param target The target element to start the search from.
+ * @param className The class name to search for.
+ * @param untilTag The tag name to stop the search at.
+ */
+export function hasParentWithClass(target: EventTarget, className: string, { untilTag }: { untilTag?: string } = {}) {
+  let element = target as HTMLElement
+  while (element?.tagName != untilTag?.toUpperCase()) {
```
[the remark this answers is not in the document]
DanielSchiavini: I tested it, it fails on `Uncaught TypeError: Cannot read properties of null (reading 'classList')` - will fix

packages/curve-ui-kit/src/shared/ui/DataTable.tsx
DanielSchiavini: indeed, can be removed

apps/main/src/loan/entities/chains.ts
DanielSchiavini: yeah I agree, let's keep separate this is large

apps/main/src/loan/components/PageLlamaMarkets/Page.tsx
DanielSchiavini: on initial render isLoading is still false. I'll add a comment

DanielSchiavini:
@0xAlunara I fixed your comments. It's hard to reply to the items in the main post but here we go

> I know columns aren't technically 'filters', but yet I expected this button to also reset the column visibility 🤔

@0xtutti wanted it like this and I agree - visibility should be in local storage, filters should ultimately be in the url

> The background change on hover is instant

Fixed

> 1. `$x%` being the average rate is new to me...

It was a typo, fixed

> 2. The 'Current Rate' tooltip gets in the way getting the 'Average Rate' tooltip from the number above. There's no point in making the contents of these tooltips selectable, so you can try the following (copied from `TokenIcon` component)

I added this to a separate task for implementing the tooltip from the design system

> 1. It's really hard to at first glance see which chips are enabled and which are not

The border was broken, now it should be bordered blue in the dark theme

> 2. I can't disable both 'mint markets' and 'lend markets' when I enable one or the other. I explicitly have to click 'Reset Filters'. A bit odd.

Fixed

> Asked tutti in slack but don't want it to get lost: should we really add `-` everywhere? Looks probably calmer if we opt for keeping cells empty if there's no data.

Made them all empty, it looks nicer

> 1. We should add a debounce to these slider values, because if you currently slide it will instantly re-render the table for every miniscule step change. This kills the performance.

Fixed

> 2. Should we add a shadow to the dropdown? The way it overlaps with the datatable columns headers makes it hard to distinguish, given they live on the same layer.

Fixed

OnlyJousting (approved):

0xtutti (approved):

OnlyJousting (approved):

OnlyJousting (approved):

packages/curve-ui-kit/src/hooks/useDebounce.ts
```diff
@@ -70,4 +70,25 @@ function useDebounce<T>(initialValue: T, debounceMs: number, callback: (value: T
   return [value, setDebouncedValue] as const
 }
 
-export default useDebounce
+const SearchDebounceMs = 166 // 10 frames at 60fps
+
+/**
+ * A hook that debounces a search value and calls a callback when the debounce period has elapsed.
+ */
+export function useSearchDebounce<T>(defaultValue: T, callback: (value: T) => void, debounceMs = SearchDebounceMs) {
```
0xAlunara, line 78: Thinking about this, maybe we should rename it to 'useUniqueDebounce' or something? It's not that related to searching anymore (well, until the return of `if (typeof value === 'string')`.

0xAlunara (approved):

OnlyJousting (approved):

PR #686, opened 2025-02-14
chore: upgrade eslint and react

DanielSchiavini:
- upgrade eslint according to [the guide](https://eslint.org/docs/latest/use/configure/migration-guide)
- upgrade react to 19 and fix errors
  - I was mistaken, and left one reference to react 18 leftover in https://github.com/curvefi/curve-frontend/pull/677
- update main app deployment to node 22
- sync project dependencies
  - got rid of chai-bigint


yarn.lock
```diff
@@ -19517,12 +19495,10 @@ __metadata:
   languageName: node
   linkType: hard
 
-"react@npm:^18.3.1":
```
DanielSchiavini, line 19520: oopsie

eslint.config.mjs
```diff
@@ -0,0 +1,25 @@
...
+const config = [
+  ...compat.extends('custom'),
+  {
+    settings: {
+      next: {
+        rootDir: ['apps/*/', 'packages/ui/*/'],
+      },
+    },
+  },
+]
```
0xAlunara, line 23: I think you can get rid of the `@eslint/eslintrc` dependency if you dont use the compat layer and just use the new v9 presets bundled with `@eslint/js`. In addition we might also be able to just inline the config from `eslint-config-custom` and get rid of that internal package and `package.json` altogether.

If you want you can take a peek at how I've implemented it for Llama Airforce and CurveMonitor over at https://github.com/Llama-Airforce/Llama-Airforce/blob/master/eslint.config.mjs. Might come in handy if you need to reconfigure typescript linting in the native v9 style.

```suggestion
import eslint from "@eslint/js";
import tseslint from "typescript-eslint";

const config = [
  eslint.configs.recommended,
  ...tseslint.configs.strictTypeChecked,
  {
    rules: {
      // stuff from eslint-config-custom
    }
  },
]
```

Showstoppers might be
```
extends: ['next', 'turbo', 'prettier'],
plugins: ['no-only-tests', 'unused-imports', 'import'],
```

OnlyJousting (approved):

0xAlunara (approved):
Approved, but worth investigating if we can get rid of `eslint-config-custom` and `@eslint/eslintrc`.

eslint.config.mjs
DanielSchiavini: We can do that separately

PR #687, opened 2025-02-14
feat: table row hovering

DanielSchiavini:
- invert the theme when the row is hovered
- add a row hover background
- cancel the click when clicking on the copy or favorite buttons
- show lend page with address in URL


0xAlunara (changes requested):
1) Clicking the crvUSD - ETH mint market doesn't bring you to the market page, but the market list page. Odd, because other markets are working fine. Probably a bug related to WETH/ETH?

2) A bit sadge there's no hover background transition (because of performance issues I believe?). The favorite icon showing up does have a transition which feels odd as the timing's off.

Rest looks fine, just need to fix 1)

0xAlunara (approved):

OnlyJousting (approved):

PR #693, opened 2025-02-19
chore: publish prices-api package to npm

0xAlunara:
Exporting API response types as well in the barrels as I need them whenever I create a proxy server for CurveMonitor's back-end.
Added the new oracle endpoint
Publishing of the NPM package happens manually by me or Macket
Tested with Curve Monitor by publishing test versions first on my public npm account, and later as 0.0.1 and 0.0.2 on the curve.fi npm account.

Current draft version available on https://www.npmjs.com/package/@curvefi/prices-api


DanielSchiavini (approved):
I expected a separate repository for this, but this is easy.

For frontend work I guess we don't need the npm version? Otherwise we need to automate the releasing process (it's [simple](https://github.com/curvefi/curve-js/blob/8fd7cddef841478f25450c688caecc670dd985e1/.github/workflows/publish.yml#L6))

OnlyJousting (dismissed):

.github/workflows/prices-api.yaml
```diff
@@ -0,0 +1,27 @@
+name: Prices API
+description: Builds and publishes the Prices API package to NPM
+
+on:
+  push:
+    branches:
+      - main
```
DanielSchiavini, line 7: We dont need a new version every push. This would probably fail because of duplicated versions
[on: tags](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs) is better in this case I think
```suggestion
    tags:
      - prices-api-v**
```

0xAlunara: opted for the more specific `prices-api/[0-9]+.[0-9]+.[0-9]+` tag

DanielSchiavini (approved):

OnlyJousting (approved):

PR #694, opened 2025-02-20
fix: link to correct network scanners for gauges on /dao

OnlyJousting:
Extended networks.ts in dao to access scanAddressPath for more networks to link to the correct network scanner when a user follows the "View on explorer" links on gauge pages.


apps/main/src/dao/types/dao.types.ts
```diff
@@ -1,4 +1,4 @@
-import type { INetworkName } from '@curvefi/api/lib/interfaces'
+import type { IChainId, INetworkName } from '@curvefi/api/lib/interfaces'
```
0xAlunara, line 1: Ser can you please fix your prettier 😭 
The PR reviews get really unwieldy if we keep up like this lol

0xAlunara (changes requested):

apps/main/src/dao/networks.ts
```diff
@@ -3,18 +3,21 @@ import { ChainOption } from '@ui-kit/features/switch-chain'
...
-  api: curvejsApi,
-  isActiveNetwork: true,
-  showInSelectNetwork: true,
-}
+const DEFAULT_NETWORK_CONFIG = { api: curvejsApi, isActiveNetwork: true, showInSelectNetwork: true }
 
-const networksConfig = {
-  1: {},
-}
+const networksConfig = Object.values(Chain).reduce(
```
DanielSchiavini, line 12: This variable doesn't make any sense. Why does it exist? We could iterate over `Chain` directly

apps/main/src/dao/types/dao.types.ts
```diff
@@ -14,7 +14,7 @@ export type PageWidthClassName =
   | 'page-small-x'
   | 'page-small-xx'
 export type CurveApi = typeof curveApi & { chainId: 1 }
-export type ChainId = 1
+export type ChainId = IChainId | number
```
DanielSchiavini, line 17: if it can be any number it might as well just be number?

apps/main/src/dao/utils/index.ts
```diff
@@ -59,3 +57,10 @@ export function sleep(ms?: number) {
 }
 
 export const httpFetcher = (uri: string) => fetch(uri).then((res) => res.json())
+
+export function getChainIdFromGaugeData(gaugeData: GaugeFormattedData | undefined) {
+  if (!gaugeData) return 1
```
DanielSchiavini, line 62: if no data then eth seems like bad logic

apps/main/src/dao/types/dao.types.ts
DanielSchiavini:
```suggestion
export type ChainId = number
```

apps/main/src/dao/utils/index.ts
OnlyJousting: I'm concerned that lite-networks will have unknown network names and feel more comfortable defaulting to ethereum as the root gauge will be on ethereum

DanielSchiavini (approved):

apps/main/src/dao/utils/index.ts
OnlyJousting: it's for when neither a pool or market object doesn't exist in gaugeData, this might happen if the side chain has not yet adopted by the API

DanielSchiavini (dismissed):

0xAlunara (approved):

PR #695, opened 2025-02-20
fix: show errors when dashboard fails loading

DanielSchiavini:
- improve error handling so the dashboard shows errors when the RPC fails

### DIFF
--- a/apps/main/src/dex/components/PageDashboard/components/SummaryTotal.tsx
+++ b/apps/main/src/dex/components/PageDashboard/components/SummaryTotal.tsx
@@ -13,8 +13,9 @@ import {
 import { Chip } from '@ui/Typography'
 import Spinner from '@ui/Spinner'
 
-const SummaryTotal: React.FC = () => {
+const SummaryTotal = () => {
   const { dashboardDataMapper, isValidAddress } = useDashboardContext()
+  const { isLoading } = useDashboardContext()
 
   const total = useMemo(() => {
     if (typeof dashboardDataMapper === 'undefined') return null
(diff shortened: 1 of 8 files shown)


OnlyJousting (approved):

apps/main/src/dex/components/PageDashboard/components/SummaryTotal.tsx
```diff
@@ -13,8 +13,9 @@ import {
 import { Chip } from '@ui/Typography'
 import Spinner from '@ui/Spinner'
 
-const SummaryTotal: React.FC = () => {
+const SummaryTotal = () => {
   const { dashboardDataMapper, isValidAddress } = useDashboardContext()
+  const { isLoading } = useDashboardContext()
```
0xAlunara, line 18: You're already using it above, could've been put on a single line?

apps/main/src/dex/lib/curvejs.ts
```diff
@@ -1310,12 +1311,21 @@ const poolWithdraw = {
   },
 }
 
+/**
+ * TODO: Work around for getPoolList being called twice from UserSlice and DashboardSlice
```
0xAlunara, line 1315: Prob temporary until we refactor to use tanstack instead of stores

0xAlunara (approved):

apps/main/src/dex/lib/curvejs.ts
DanielSchiavini: indeed #696

apps/main/src/dex/components/PageDashboard/components/SummaryTotal.tsx
DanielSchiavini: oh sorry I'll do that

DanielSchiavini: https://github.com/curvefi/curve-frontend/pull/696/commits/9ef9505f055c9a3d4268c11e9dc5dda56c9b4c5d

PR #696, opened 2025-02-20
feat: user pools via prices API to fix dashboard

DanielSchiavini:
- curve-js calls every single pool on chain, that results in too many RPC calls
- instead we can use the new prices API endpoint to find the pools that a user is part of
- lib is still used for unsupported chains
- add `fetchQuery` to factory to remove some duplication

### DIFF
--- a/packages/prices-api/src/pools/responses.ts
+++ b/packages/prices-api/src/pools/responses.ts
@@ -34,6 +34,15 @@ export type GetPoolsResponse = {
   data: Pool[]
 }
 
+export type GetUserPoolsResponse = {
+  chain: string
+  positions: {
+    pool_name: string
+    pool_address: Address
+    lp_token_balance: string
+  }[]
+}
+
 export type GetPoolResponse = Pool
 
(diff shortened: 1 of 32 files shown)


packages/prices-api/src/pools/responses.ts
```diff
@@ -34,6 +34,15 @@ export type GetPoolsResponse = {
   data: Pool[]
 }
 
+export type GetUserPoolsResponse = {
+  chain: string
```
0xAlunara, line 38:
```suggestion
  chain: Chain
```

DanielSchiavini: https://github.com/curvefi/curve-frontend/pull/696/commits/9ef9505f055c9a3d4268c11e9dc5dda56c9b4c5d

fedorovdg (approved):

DanielSchiavini:
This will be fixed in curvejs by calling the RPC in chunks. The API doesn't return all user pools.

PR #701, opened 2025-02-21
chore: test flakyness

DanielSchiavini:
- get rid of all the retries
- run in parallel in different browsers
- fix all test failures
- use production build

PR #707, opened 2025-02-25
chore: migrate to next app router

DanielSchiavini:
- migrate all apps to the new app router
  - create a client component for the app
  - simply render the client component for now
  - note that next now shows every console.error in the UI in dev mode
- fix all styles with styled-components
- fix cypress tests by ignoring the hydration errors
  - unfortunately the error is still displayed but doesn't block the tests
- ~depends on https://github.com/curvefi/curve-frontend/pull/701~


OnlyJousting (approved):

OnlyJousting (commented):
🙌

apps/main/next-env.d.ts
```diff
@@ -1,5 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
+/// <reference types="next/navigation-types/compat/navigation" />
 
 // NOTE: This file should not be edited
-// see https://nextjs.org/docs/pages/api-reference/config/typescript for more information.
+// see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
```
0xAlunara, line 6: Not sure if this is an accident or what the final value should be at this point

packages/curve-ui-kit/src/shared/ui/TokenIcons.tsx
```diff
@@ -20,10 +20,10 @@ export function TokenIcons({ tokens, ...props }: Props) {
       display="inline-grid"
       sx={{
         gridTemplateColumns: `repeat(${iconsPerRow}, auto)`,
-        [`& > *:nth-child(${iconsPerRow}n-1)`]: {
+        [`& > *:nth-of-type(${iconsPerRow}n-1)`]: {
```
0xAlunara, line 23: `*:nth-of-type` is basically the same as `*:nth-child` right? Wondering why you changed this

0xAlunara (approved):

PR #711, opened 2025-02-26
feat: new token selection modal

0xAlunara:
Various:
* Fixes incorrect border colors for clickable chips
* Implements correct colors for dividers
* Removes outlines on focused lists (was a black border around the entire list)
* Fixes card headers being vertically aligned to the bottom

Main:
* Adds new token select feature
* Adds stories for the new token select feature
* Removes old token select code and replaces it with the new one

Highlights:
* User search is debounced for performance
* Tried Tanstack Virtual, but had to abandon due to glitchy behaviour and even worse performance than native. The good thing is, the token icons are loaded lazily even in native React, which means they won't be loaded until the users scrolls it into view
* Adds support for token favorites. A good example use case is when creating a pool and selected a token, two tokens can be considered favorite.
* Token list is sorted first by user dollar balance, putting anything with a dollar value on top. If one has no usd value, it switches to ordering by token balance. Lastly, if that also doesn't work, it switches to ordering by token symbol
* I've added the 'hide small pools' settings to the user profile, because it makes sense to me that it's a global setting across different pages.
* The token selector can be disabled
* Specific tokens in the list can be disabled (perhaps due to blacklisting because of a hack)
* You can pass a custom settings component to the selector in case you want to add extra settings between the list and the favorite chips yourself.

I've refrained from refactoring old code as much as possible, which means that the initial fetching of the token list is still slow as hell. I don't want to inflate the PR with even more code changes than it already has.


OnlyJousting (dismissed):

packages/curve-ui-kit/src/features/select-token/ui/modal/TokenList.tsx
```diff
@@ -0,0 +1,163 @@
...
+import { t } from '@ui-kit/lib/i18n'
+import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
+
+const { Spacing } = SizesAndSpaces
+
+import type { TokenOption as Option } from '../../types'
+import { FavoriteTokens } from './FavoriteTokens'
+import { TokenOption } from './TokenOption'
+
+// Prevent all the token options from re-rendering if only the balance of a single one has changed.
```
DanielSchiavini, line 19: This shouldn't be necessary, it should happen automatically :thinking:

packages/curve-ui-kit/src/features/select-token/ui/modal/TokenOption.tsx
```diff
@@ -0,0 +1,61 @@
...
+      onClick={disabled ? undefined : onToken}
+      sx={{
+        minHeight: IconSize.xxl,
+        ...(disabled && {
+          opacity: 0.5,
+          cursor: 'not-allowed',
+        }),
+      }}
+    >
+      {/* I'm abusing the symbol here to deliberately show a tooltip for the address instead */}
```
DanielSchiavini, line 41: why? :thinking:

apps/main/src/dex/components/PageCreatePool/TokensInPool/SelectTokenButton.tsx
```diff
@@ -0,0 +1,289 @@
...
+  chainId: ChainId
+  disabledKeys?: string[]
+  haveSigner: boolean
+  blockchainId: string
+  selectedAddress: string
+  tokens: CreateToken[]
+  onSelectionChange: (selectedAddress: React.Key) => void
+}
+
+const ComboBoxTokenPicker = ({
```
DanielSchiavini, line 32: Was this file moved? What was the name? Diff isn't showing.

apps/main/src/dex/components/PagePool/Swap/index.tsx
```diff
@@ -88,15 +86,32 @@ const Swap = ({
...
-  )
+  const { selectList, swapTokensMapper } = useMemo(() => {
+    const { selectList, swapTokensMapper } = getSwapTokens(tokensMapper, poolDataCacheOrApi)
+
+    return {
+      selectList: selectList
+        .filter((token) => !!token)
+        .map((token) => ({
+          chain: network?.networkId ?? '',
+          address: token?.address as `0x${string}`,
```
DanielSchiavini, line 104:
```suggestion
          address: token?.address as Address,
```

apps/main/src/dex/components/PagePool/Swap/index.tsx
```diff
@@ -88,15 +86,32 @@ const Swap = ({
...
+          symbol: token?.symbol,
+          label: '',
```
DanielSchiavini, line 106: Why is this empty? And why do we use an empty string instead of undefined?
```suggestion
```

apps/main/src/dex/components/PagePool/Swap/index.tsx
```diff
@@ -88,15 +86,32 @@ const Swap = ({
...
-    [poolDataCacheOrApi, tokensMapper],
-  )
+  const { selectList, swapTokensMapper } = useMemo(() => {
+    const { selectList, swapTokensMapper } = getSwapTokens(tokensMapper, poolDataCacheOrApi)
+
+    return {
+      selectList: selectList
+        .filter((token) => !!token)
+        .map((token) => ({
+          chain: network?.networkId ?? '',
```
DanielSchiavini, line 103: if chain is optional it should be optional not an empty string

apps/main/src/dex/components/PageRouterSwap/index.tsx
```diff
@@ -98,30 +97,54 @@ const QuickSwap = ({
...
   const tokensMapperNonSmallTvlStr = useMemo(
     () => getTokensMapperStr(tokensMapperNonSmallTvl),
     [tokensMapperNonSmallTvl],
   )
 
-  const selectFromTokensList = useMemo(
-    () => getTokensObjList(selectFromList ?? selectToList, tokensMapper),
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [selectFromList, selectToListStr, tokensMapperStr],
+  function getTokensObjList(tokensList: string[] | undefined, tokensMapper: TokensMapper | undefined) {
```
DanielSchiavini, line 112: this should be in the global scope or be included in the `useMemo` dependencies, not sure why we don't get a warn

apps/main/src/dex/components/PageRouterSwap/index.tsx
```diff
@@ -98,30 +97,54 @@ const QuickSwap = ({
...
+    if (!tokensList || tokensList.length === 0 || !tokensMapper || Object.keys(tokensMapper).length === 0) return []
```
DanielSchiavini, line 113:
```suggestion
    // import { isEmpty } from 'lodash'
    if (isEmpty(tokensList) || isEmpty(tokensMapper)) return []
```

apps/main/src/dex/components/PageRouterSwap/index.tsx
```diff
@@ -98,30 +97,54 @@ const QuickSwap = ({
...
+  const tokensFrom = useMemo(
+    () =>
+      getTokensObjList(selectFromList ?? selectToList, tokensMapper)
+        .filter((token) => !!token)
+        .map((token) => ({
+          chain: network?.networkId ?? '',
+          address: token?.address as `0x${string}`,
+          symbol: token?.symbol,
+          label: '',
+          volume: token?.volume ?? 0,
```
DanielSchiavini, line 126: Zero and empty strings should not be the fallback

```suggestion
          chain: network?.networkId,
          address: token?.address as Address,
          symbol: token?.symbol,
          volume: token?.volume,
```

apps/main/src/dex/components/PageRouterSwap/index.tsx
```diff
@@ -98,30 +97,54 @@ const QuickSwap = ({
...
+  function getTokensObjList(tokensList: string[] | undefined, tokensMapper: TokensMapper | undefined) {
+    if (!tokensList || tokensList.length === 0 || !tokensMapper || Object.keys(tokensMapper).length === 0) return []
+    return tokensList.map((address) => tokensMapper[address])
+  }
+
+  const tokensFrom = useMemo(
+    () =>
+      getTokensObjList(selectFromList ?? selectToList, tokensMapper)
+        .filter((token) => !!token)
+        .map((token) => ({
```
DanielSchiavini, line 121: Why isn't `filter` and `map` part of `getTokensObjList` as it's repeated

apps/main/src/dex/features/add-gauge-reward-token/ui/TokenSelector.tsx
```diff
@@ -24,23 +25,34 @@ export const TokenSelector = ({
...
+        (token): token is Token =>
+          token !== undefined &&
+          token.decimals === 18 &&
+          !aliasesCrv &&
+          ![...gaugeRewardTokens, zeroAddress, NETWORK_TOKEN, aliasesCrv].some((rewardToken) =>
+            isAddressEqual(rewardToken as Address, token.address as Address),
+          ),
+      )
+      .map((token) => ({
+        chain: network?.networkId ?? '',
```
DanielSchiavini, line 47: this is repeated code

apps/main/src/dex/features/deposit-gauge-reward/ui/AmountTokenInput.tsx
```diff
@@ -54,30 +53,40 @@ export const AmountTokenInput = ({ chainId, poolId }: { chainId: ChainId; poolId
...
-        activeRewardTokens.some((rewardToken) => isAddressEqual(rewardToken as Address, token.address as Address)),
-    )
+    const filteredTokens = Object.values(tokensMapper)
+      .filter(
+        (token): token is Token =>
+          token !== undefined &&
+          activeRewardTokens.some((rewardToken) => isAddressEqual(rewardToken as Address, token.address as Address)),
+      )
+      .map((token) => ({
+        chain: networkId ?? '',
```
DanielSchiavini, line 70: repeated code

apps/main/src/dex/features/deposit-gauge-reward/ui/AmountTokenInput.tsx
```diff
@@ -54,30 +53,40 @@ export const AmountTokenInput = ({ chainId, poolId }: { chainId: ChainId; poolId
...
 
-    const filteredTokens = Object.values(tokensMapper).filter(
-      (token): token is Token =>
-        token !== undefined &&
-        activeRewardTokens.some((rewardToken) => isAddressEqual(rewardToken as Address, token.address as Address)),
-    )
+    const filteredTokens = Object.values(tokensMapper)
+      .filter(
+        (token): token is Token =>
+          token !== undefined &&
```
DanielSchiavini, line 66: What if it's null?
```suggestion
          token &&
```

apps/main/src/dex/features/deposit-gauge-reward/ui/HelperFields.tsx
```diff
@@ -1,18 +1,25 @@
...
-  } = useTokensUSDRates([rewardTokenId])
+  const usdRatesMapper = useStore((state) => state.usdRates.usdRatesMapper)
+  const tokens = [rewardTokenId]
+  const tokensKey = JSON.stringify(tokens)
+
+  const [tokenUsdRate] = useMemo(
+    () => tokens.map((token) => (token ? usdRatesMapper[token] : undefined)),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [tokensKey, usdRatesMapper],
+  )
```
DanielSchiavini, line 22: Why do we need this complicated JSON stringify and ignoring eslint rules?

```suggestion
  const tokenUsdRate = rewardTokenId && usdRatesMapper[rewardTokenId]
```

apps/main/src/dex/store/createQuickSwapSlice.ts
```diff
@@ -410,11 +409,9 @@ const createQuickSwapSlice = (set: SetState<State>, get: GetState<State>): Quick
 
       const { chainId, signerAddress } = curve
 
-      const selectToList = orderBy(
-        Object.entries(tokensMapper).map(([_, v]) => v!),
-        ({ volume }) => (typeof volume !== 'undefined' ? +volume : 0),
-        ['desc'],
-      ).map(({ address }) => address)
+      const selectToList = Object.entries(tokensMapper)
+        .map(([_, v]) => v!)
```
DanielSchiavini, line 413: This looks incorrect - if `tokensMapper` has no `null`/`undefined` why is it typed so?

packages/curve-ui-kit/src/features/select-token/ui/TokenSelectButton.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+const { Spacing, ButtonSize, MinWidth } = SizesAndSpaces
+
+import type { TokenOption } from '../types'
+
+const ButtonContent = ({ token, disabled }: { token: TokenOption; disabled: boolean }) => (
+  <Stack direction="row" gap={Spacing.xxs} alignItems="center">
+    <TokenIcon
+      blockchainId={token.chain}
+      symbol="" // No need for tooltip of symbol as it's displayed right next to it
```
DanielSchiavini, line 19:
```suggestion
      // No need for tooltip of symbol as it's displayed right next to it
```

packages/curve-ui-kit/src/features/user-profile/settings/hide-small-pools/HideSmallPools.tsx
```diff
@@ -0,0 +1,22 @@
+import { t } from '@ui-kit/lib/i18n'
+import Switch from '@mui/material/Switch'
+
+import { SettingLabel, Setting } from '../Setting'
+import useUserProfileStore from '../../store'
+
+export const HideSmallPools = () => {
```
DanielSchiavini, line 7:
```suggestion
export const HideSmallPoolsSetting = () => {
```

packages/curve-ui-kit/src/themes/components.ts
```diff
@@ -97,6 +104,15 @@ export const createComponents = (design: DesignSystem, typography: TypographyOpt
       },
     },
   },
+  MuiList: {
+    styleOverrides: {
+      root: {
+        '&:focus': {
+          outline: 'none',
```
DanielSchiavini, line 111: Are you sure we want to disable this? It might be useful for accessibility. How about `focus-visible`?
For buttons and icon buttons I use `'&:focus-visible': { borderColor: Focus_Outline }` as discussed with UX.

0xAlunara: The menu list has an auto-focus so you can use keyboard navigation immediately to navigate the list. However, that means it immediately renders with a damn awful black border. I could change `&:focus` to `&:focus-visible`, but showing any kind of outline is not an option, so neither would be using `Focus_Outline`

image

packages/curve-ui-kit/src/features/select-token/ui/TokenSelectButton.tsx
0xAlunara: Had to make symbol an optional property that defaults to the empty string for this

packages/curve-ui-kit/src/features/select-token/ui/modal/TokenList.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+          // Compare USD balances first (including existence check)
+          const aBalanceUsd = (tokenPrices[a.address] ?? 0) * aBalance
+          const bBalanceUsd = (tokenPrices[b.address] ?? 0) * bBalance
+          if (aBalanceUsd !== bBalanceUsd) return bBalanceUsd - aBalanceUsd
+
+          // If USD balances are equal, compare raw balances
+          if (aBalance !== bBalance) return bBalance - aBalance
+
+          // Sort by volume in descending order
```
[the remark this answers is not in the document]
0xAlunara: It matches the original ordering though, so at least it's not different from what people are used to 😅

packages/curve-ui-kit/src/features/select-token/ui/modal/TokenList.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+      {customOptions && (
+        <>
+          {customOptions}
+          <Divider />
+        </>
+      )}
+
+      {error ? (
+        <Alert variant="filled" severity="error">
```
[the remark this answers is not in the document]
0xAlunara: I've made the error message grow in height as needed. The height is actually fixed to `IconSize.sm` in `defineMuiAlertTitle`. I didn't really want to introduce an entire variant for this one off, so I've opted for `sx={{ '&': { height: 'auto' } }}` for now.

The error will also go away when you close the modal or change the search value.

Not sure what's the best way to tack on a friendly error message. I could change any error to "invalid token" but maybe it's not related to the token but the RPC. Or maybe there's something else, and then you'd want to know the full error. Not sure we can just look at the error message and search for "Multicall3", replace it with "Invalid token" and call it a day

packages/curve-ui-kit/src/features/select-token/ui/modal/TokenList.tsx
0xAlunara: I tried with and without and didn't notice / measure any difference, so I've removed it.

0xAlunara: I really need to give virtualization a second try though, when you first open the modal with all tokens the performance is dogshit imo. Not the case when I limit the list to just 5 items. It's also not the re-renders causing the performance hog, checked it with a `console.log(symbol)` inside `TokenOption.tsx`.

apps/main/src/dex/components/PageCreatePool/TokensInPool/SelectTokenButton.tsx
0xAlunara: Lol seems to have been a copy-paste error from before my time. I'll fix it though

apps/main/src/dex/components/PagePool/Swap/index.tsx
0xAlunara: I prefer explicitness and want to avoid null & undefined. Will make this one optional though

apps/main/src/dex/features/deposit-gauge-reward/ui/AmountTokenInput.tsx
0xAlunara: That doesn't work, but `!!token` does

apps/main/src/dex/components/PagePool/Swap/index.tsx
0xAlunara: TokenIcon chain used be required, but I've now made it optional and default to `''` inside the component itself

apps/main/src/dex/store/createQuickSwapSlice.ts
0xAlunara: Probably because `export type TokensMapper = { [tokenAddress: string]: Token | undefined }`

If I remove the exclamation mark I actually get a type error that the `address` in `.map(({ address }) => address)` might be `undefined` as the type is `Token | undefined`. It's a bit silly because then I don't even know if you can have a key whose value is undefined?

apps/main/src/dex/features/deposit-gauge-reward/ui/HelperFields.tsx
0xAlunara: Refactoring this away with a tanstack query in a future PR

apps/main/src/dex/components/PageRouterSwap/index.tsx
0xAlunara: Had to add an exclamation mark to `return tokensList!.map()` as isEmpty works with `undefined`, but doesn't alter the type of `tokensList` from `string[] | undefined` to `string[]` even though we know for sure it's not undefined.

packages/curve-ui-kit/src/features/select-token/ui/modal/TokenOption.tsx
0xAlunara: I've experimented with showing the address on hover and keyboard focus at the place where the label would be, but because we're not using labels it causes the token symbol to move up. See video attached, the symbol typography has to move up to make room for the address to show up. When no longer hovering or focusing the item the address would go away and the symbol would move down again.

This looks very jumpy and distracting so I've opted to keep the token on hover for now

https://github.com/user-attachments/assets/5233f53e-e299-4904-b3e5-6764be2b39cd

packages/curve-ui-kit/src/features/select-token/ui/modal/TokenList.tsx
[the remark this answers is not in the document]
DanielSchiavini: I think we should only have a generic error as title, the details can be in the alert body

packages/curve-ui-kit/src/features/select-token/ui/modal/TokenList.tsx
DanielSchiavini: The comment is still left. Did you check the performance profiler?

packages/curve-ui-kit/src/features/select-token/ui/modal/TokenList.tsx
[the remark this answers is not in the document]
0xAlunara: You're right, I was so focused on the title I forgot you can use an actual body message.

I've added the copy button (smallest sizen, ghost) in an unconventional way, because the button would take up an entire column, leaving a lot of unused whitespace that the message could've used. So for now I've manually experimented with some absolute positioning. This is a one off, and if we want to make this the default behavior I'd rather implement it properly in a different PR.

I've also added word wrapping to the message. I have a hunch not all users actually use the copy function and send a screenshot instead. This would make the error more readable as well; scrollbars would be annoying in this case.

image

packages/curve-ui-kit/src/features/select-token/ui/modal/TokenList.tsx
0xAlunara: I've checked the profiler, and could not find any meaningful slowdown in the token list itself. The performance is not bad at all in production, so I chalk it up to a slow laptop on top of local dev mode. It seems the headers and footers with their icons are bigger performance hogs as well. On top of that, after discussion with tutti we're no longer showing the entire token list at once, but will have a "show more" button, as per the new Figma design. Will resolve this for now.

packages/curve-ui-kit/src/features/select-token/ui/modal/TokenOption.tsx
0xAlunara: Renamed the property from 'symbol' to 'tooltip' to more clearly indicate its intent.
As discussed with tutti, we've decided to implement two sections in the modal: "My tokens" and "All tokens"

In "My tokens", we will show balances as we currently do. In "All tokens" we will show the token address (where currently nothing is shown).

apps/main/src/dex/features/add-gauge-reward-token/ui/TokenSelector.tsx
0xAlunara: My bad, there were multiple instances so I lost track.

For the mapping of 'old' token objects to the TokenOption type required by the token selector feature I've written a small utility function that, given a network, converts it into such:
```ts
export const toTokenOption =
  (chain?: string) =>
  (token: Token): TokenOption => ({
    chain,
    address: token.address as Address,
    symbol: token.symbol,
    volume: token.volume,
  })
```
Example usage:
```ts
tokens.map(toTokenOption(network?.networkId))
```

The `chain`, `label` and `volume` properties in the token option type itself have also been made optional:
```ts
export type TokenOption = {
  address: Address
  symbol: string
  chain?: string
  label?: string
  volume?: number
}
```

apps/main/src/dex/store/createQuickSwapSlice.ts
0xAlunara: I feel like we're working around a typescript quirk here, but nonetheless I've added a `.filter((token) => !!token)`

packages/curve-ui-kit/src/features/select-token/ui/modal/TokenList.tsx
```diff
@@ -0,0 +1,163 @@
...
+      }
+    }, 200)
+    return () => clearTimeout(timer)
+  }, [debouncedSearch, onSearch, search])
+
+  const showFavorites = favorites.length > 0 && !debouncedSearch
+
+  // List of all token options
+  const options = useMemo(() => {
+    const tokensFiltered = [...tokens].filter((token) =>
```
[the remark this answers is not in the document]
0xAlunara: Figured I'd keep it more simple, but yeah, maybe common behavior across the board is more desired.

apps/main/src/dex/components/PageCreatePool/TokensInPool/SelectTokenButton.tsx
```diff
@@ -0,0 +1,289 @@
...
+  const isMobile = useStore((state) => state.isMobile)
+  const nativeToken = useStore((state) => state.networks.nativeToken[chainId])
+
+  const userAddedTokens = useStore((state) => state.createPool.userAddedTokens)
+  const updateUserAddedTokens = useStore((state) => state.createPool.updateUserAddedTokens)
+
+  const { basePools, basePoolsLoading } = useStore((state) => state.pools)
+  const { swapType } = useStore((state) => state.createPool)
+
+  const [error, setError] = useState('')
```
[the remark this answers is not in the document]
0xAlunara: Copied it from somewhere else in the code, but sure, made it optional

0xAlunara:
Major changes since last version:

* Tokens are now split up in two sections called "My tokens" and "All tokens" if applicable.
* 'My tokens' shows 5 tokens, 'All tokens' shows 10 tokens. If there's more you can click the 'Show more' buttons.
* Tokens without a balance now show the shortened token address for verification (the idea is, if you don't own it you prob want to verify it's the correct one. And if you do own it you prob don't care about the address anymore as you already trust it)
* Chips with icons have their padding and margin fixed
* Lots of speed optimizations. The biggest one is that `createQuickSwapSlice` no longer has two different 'from' and 'to' lists, and the sorting is gone as it's now done internally by the new select token modal.
* Hover and focus-visible now use inverted themes. Not ideal but I got really tired so I'm calling it a day. Improvements regarding hover and focus states for MUI's MenuItem are for another PR.
* Errors are rendered more nicely.
* A few focus / keyboard navigation related fixed.
* All review feedback has been processed.

0xtutti (approved):

DanielSchiavini (approved):

apps/main/src/dex/components/PageCreatePool/TokensInPool/SelectTokenButton.tsx
```diff
@@ -0,0 +1,289 @@
...
+  chainId: ChainId
+  disabledKeys?: string[]
+  haveSigner: boolean
+  blockchainId: string
+  selectedAddress: string
+  tokens: CreateToken[]
+  onSelectionChange: (selectedAddress: React.Key) => void
+}
+
+const SelectTokenButton = ({
```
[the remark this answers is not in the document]
0xAlunara: It's indeed the old file, but moved and trimmed down. A lot of code has been deleted now we have `TokenSelectorModal`, but not everything for the token selection page when creating a pool. So what you see is what remained and could not be replaced by the new `TokenSelectorModal` in its entirety.

apps/main/src/dex/features/deposit-gauge-reward/ui/AmountTokenInput.tsx
```diff
@@ -54,30 +54,34 @@ export const AmountTokenInput = ({ chainId, poolId }: { chainId: ChainId; poolId
...
 
-    const filteredTokens = Object.values(tokensMapper).filter(
-      (token): token is Token =>
-        token !== undefined &&
-        activeRewardTokens.some((rewardToken) => isAddressEqual(rewardToken as Address, token.address as Address)),
-    )
+    const filteredTokens = Object.values(tokensMapper)
+      .filter(
+        (token): token is Token =>
+          !!token &&
```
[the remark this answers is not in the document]
0xAlunara: Doesn't work:
```
Type 'boolean | undefined' is not assignable to type 'boolean'.
  Type 'undefined' is not assignable to type 'boolean'.
```

apps/main/src/dex/features/deposit-gauge-reward/ui/HelperFields.tsx
0xAlunara: It's not new, I just inlined whatever `useTokensUSDRates` used to be as this was the only place it was called. As said, will be a tanstack query in the future.

packages/curve-ui-kit/src/features/select-token/ui/modal/SearchInput.tsx
```diff
@@ -0,0 +1,38 @@
...
+type Props = {
+  onSearch: (search: string) => void
+  debounceMs?: number
+}
+
+export const SearchInput = ({ onSearch, debounceMs = 200 }: Props) => {
+  const [search, setSearch] = useState('')
+  const lastSearch = useRef(search)
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
0xAlunara: Been going back and forth of this as I came across multiple debounce implemenations. So naturally, I created another version :D
Will create a small PR

packages/curve-ui-kit/src/features/select-token/ui/modal/TokenList.tsx
```diff
@@ -0,0 +1,236 @@
...
+        {hasMore && !showAll && (
+          <Button
+            fullWidth
+            variant="link"
+            color="ghost"
+            size="medium"
+            endIcon={<ExpandMoreIcon />}
+            onClick={onShowAll}
+            // Override variant button height to match menu list item height
+            sx={{ '&': { height: `${ButtonSize.md} !important` } }}
```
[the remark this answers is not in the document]
0xAlunara: Ah right, forgot to remove it after '&' didn't do the trick.

apps/main/src/dex/components/PageCreatePool/TokensInPool/SelectTokenButton.tsx
```diff
@@ -0,0 +1,289 @@
...
+        try {
+          const token = await curve.getCoinsData([filterValueLowerCase])
+          const isBasePool = basePools[chainId].some(
+            (basepool) => basepool.token.toLowerCase() === filterValueLowerCase,
+          )
+
+          updateUserAddedTokens(filterValueLowerCase, token[0].symbol, false, isBasePool)
+        } catch (error) {
+          console.log(error)
+          setError(error)
```
DanielSchiavini, line 115: @0xAlunara error is expected to be string!

0xAlunara: I think there should be a lint rule now that warns you when you're not typechecking `error` in a `catch` clause 🤔

DanielSchiavini: yeah that's how I found it in #2611 :nerd_face:

PR #714, opened 2025-02-27
chore: add missing yarn build step for prices api

0xAlunara:
It seems "prepublishOnly": "yarn run build" in the package.json doesn't really work when executed through github actions, so the published package ends up missing its `dist` folder. I can't really test it properly until Github Actions runs this in prod.


DanielSchiavini (approved):

DanielSchiavini:
Should we remove the `prepublishOnly` script?

0xAlunara:
It's still nice to have if you publish locally I suppose, doesn't hurt

OnlyJousting (approved):

PR #717, opened 2025-02-28
chore: remove unused and useless code

DanielSchiavini:
- remove unused code
- simplify useless code
- remove useless comments
- remove global react imports
- remove empty lines in import statements


packages/curve-ui-kit/src/lib/queries/factory.ts
```diff
@@ -1,6 +1,5 @@
 import type { UseQueryOptions } from '@tanstack/react-query'
 import { QueryKey, useQuery } from '@tanstack/react-query'
-import { queryClient } from '@ui-kit/lib/api/query-client'
```
0xAlunara, line 3: Note to self: prob will need it back for a `createFetchQuery` that has to be used in slices which can't use `useQueryClient()`

packages/prices-api/src/crvusd/models.ts
```diff
@@ -43,16 +43,6 @@ export type Snapshot = {
   discountLoan: number
 }
 
-export type PoolPrice = {
```
0xAlunara, line 46: Don't delete these as they have to be exported for package users (like Curve Monitor). Prob best to exclude this package from the automated tools?

packages/ui/src/Stepper/StepAction.tsx
```diff
@@ -10,7 +10,7 @@ import styled, { css } from 'styled-components'
 import { isInProgress, taskStepStatusStyles } from './helpers'
 import Button from 'ui/src/Button'
 import Spinner from 'ui/src/Spinner/Spinner'
-import StepModal from 'ui/src/Stepper/StepModal'
+import { StepModal } from 'ui/src/Stepper'
```
0xAlunara, line 13: Huh, how did your tool not see `import * as React from 'react'` which can be removed as well?

0xAlunara (changes requested):

OnlyJousting (approved):

packages/ui/src/Stepper/StepAction.tsx
DanielSchiavini: This opened a whole can of worms, sorry for the large diff :skull:

packages/prices-api/src/crvusd/models.ts
0xAlunara: I use them in CurveMonitor as prop types for example

OnlyJousting (approved):

packages/prices-api/src/crvusd/models.ts
DanielSchiavini: Are they related to any API endpoints?

PR #735, opened 2025-03-10
feat: minor token selector changes

0xAlunara:
* set max height of token selector to 752px
* change 'all tokens' to 'tokens by 24h volume'
* optimize sorting a bit


DanielSchiavini (approved):

0xAlunara:
Changed property 'limit' to 'preview'. It's no longer just a number, but can also have a filter function. Basically you're getting a 'preview' of all tokens in a list with a 'show more' buttons for whatever falls outside of the scope of the preview.

'My tokens' now no longer have a fixed number of tokens in as a 'limit' for its preview, but the list will consist of all tokens that have a USD value >1% of the user's total USD balance. This way a user will see most of their important tokens, but we have a 'relative' dust filter. One rare exception is that some tokens might not have a USD price; in that case we still put the token in the preview list.

DanielSchiavini (approved):

DanielSchiavini (approved):

0xtutti (approved):

PR #743, opened 2025-03-13
Feat/action-details

0xAlunara:
* extends action info with new features, see storybook
* added custom accordion component, see storybook and docs on why
* replace transaction details in scrvusd page with new accordion and action infos
* add WithSkeleton helper component

scrvusd transactions details now look like
image

and action info component should be able to in a future PR create details such as
image


OnlyJousting (approved):

packages/curve-ui-kit/src/shared/ui/Accordion.tsx
```diff
@@ -0,0 +1,170 @@
...
+}
+
+/**
+ * A customized accordion component that provides a collapsible content section.
+ *
+ * This component differs from the default MUI Accordion in that it's designed as a
+ * single-item constrained collapse component. It's called an Accordion in our Figma
+ * design system for consistency. We created this custom implementation because the
+ * MUI Accordion lacks the customization options needed for our design requirements,
+ * particularly for features like the ghost variant and responsive sizing with
```
DanielSchiavini, line 45:
```suggestion
 * particularly for features like the ghost color and responsive sizing with
```

packages/curve-ui-kit/src/shared/ui/Accordion.tsx
```diff
@@ -0,0 +1,170 @@
...
+            sx={{
+              width: IconSize.md,
+              height: IconSize.md,
+              transform: isOpen ? 'rotate(180deg)' : 'rotate(0deg)',
+              // Create a transition that mimics the collapse transition
+              transition: (t) =>
+                t.transitions.create('transform', {
+                  duration: t.transitions.duration.standard,
+                  easing: t.transitions.easing.easeInOut,
+                }),
```
DanielSchiavini, line 134: Isn't this the same as 
```suggestion
              transition: `transform ${TransitionFunction}`
```
?

packages/curve-ui-kit/src/shared/ui/ActionInfo.tsx
```diff
@@ -1,75 +1,159 @@
...
+              popper: {
+                sx: {
+                  userSelect: 'none',
+                  pointerEvents: 'none',
+                },
+              },
+            }}
+          >
+            <Typography variant={valueSize[size]} color={valueColor ?? 'textPrimary'}>
+              {!loading ? value : typeof loading === 'string' ? loading : MOCK_SKELETON}
```
DanielSchiavini, line 133: Double negatives are hard to read
```suggestion
              {typeof loading === 'string' ? loading : loading ? MOCK_SKELETON : value}
```

DanielSchiavini (approved):

packages/curve-ui-kit/src/shared/ui/Accordion.tsx
0xAlunara: No, that is our own made up transition function. I specifically wanted the transition time and easing that is used by the collapse component. maybe we should use that as well instead of using our own custom TransitionFunction? 🤔

DanielSchiavini: That function (`ease-out 256ms`) comes from @0xtutti's design system.
Actually we should customize the MUI theme to use that, and change the `TransitionFunction` usages to `theme.transitions.create([...])`

PR #746, opened 2025-03-14
refactor: replace isMobile by useMediaQuery

0xAlunara:
Trying to work on the lending market pages I come across quite some duplicate components, and to consolidate these we first need to consolidate some stuff outside app specific stores.

This PR tries to get the 'isMobile' state out of the stores and into a hook so that generic components don't need to import app specific stores just to get this value. Perhaps in the future we won't need this hook at all and rely purely on CSS and/or MUI. But for now that's not the case.

Moving the isMobile state variable into a hook however means that every component subscribes to `window.addEventListener('resize')`. So if 10 components use the hook we'd call `detectIsMobile()` 10 times. I'm unsure if this calcaulation is worthy of a singleton pattern; not used to this in React as the issue doesn't exist at all in previous frameworks I've worked with. 

Can remove the whole singleton shebang if we're ok with multiple listeners all calling `detectIsMobile()`. It's been a learning experience for me as well.

**Note**
I'm literally just moving from state to hook. I'm not touching the original isMobile code detection *at all*, that's not the point of the PR. The point is just to remove duplication.

**Note #2**
After this PR I intend to the same but for `isPageVisible`, `isMdUp` and some other layout specific things that shouldn't be in the app specific stores. This is a big bottleneck causing component duplication.


packages/curve-ui-kit/src/hooks/useIsMobile.ts
```diff
@@ -0,0 +1,96 @@
...
+      setLocalIsMobile(isMobileValue)
+    }
+
+    const callback = (value: boolean) => setLocalIsMobile(value)
+    subscribers.add(callback)
+
+    return () => {
+      subscribers.delete(callback)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
DanielSchiavini, line 92: Why not add `isMobile` to deps?

packages/curve-ui-kit/src/hooks/useIsMobile.ts
```diff
@@ -0,0 +1,96 @@
...
+function setIsMobile(value: boolean): void {
+  if (value !== isMobileValue) {
+    isMobileValue = value
+    subscribers.forEach((callback) => callback(value))
+  }
+}
+
+// Initialize global listener
+if (typeof window !== 'undefined' && !isListenerAttached) {
+  window.addEventListener('resize', () => {
```
DanielSchiavini, line 46: This does not make sense if we are checking `maxTouchPoints` - that won't change regardless of the screen size.
If we want to measure the screen size we should use mui's `useMediaQuery`

0xAlunara: Completely got rid of the custom `useIsMobile` hook and replaced all with const `useMediaQuery((t) => t.breakpoints.down('tablet'))`

DanielSchiavini (approved):
Code looks fine, but let's be careful as the new useMediaQuery has different behavior

OnlyJousting (approved):

PR #747, opened 2025-03-14
feat: preload dex cache server-side

DanielSchiavini:
- Move pool retrieval function to helper file
- Call it from a layout in the `[network]` folder
- Inject preloaded pool list, tvl and volume in the client
- Unfortunately, CurveJS will reload everything once the page is loaded


apps/main/src/app/dex/[network]/InjectServeSideData.tsx
```diff
@@ -0,0 +1,23 @@
...
+  children: ReactNode
+}
+
+export const InjectServeSideData = ({ chainId, pools, tvl, volume, children }: ServerSideStoreProps) => {
+  const setServerPreloadData = useStore((state) => state.storeCache.setServerPreloadData)
+  useEffect(() => {
+    setServerPreloadData(chainId, { pools, tvl, volume })
+    console.log(
+      `Injected ${Object.values(pools).length} pools, ${Object.keys(tvl).length} TVL, and ${Object.keys(volume).length} volume for chain ${chainId}`,
+    )
```
Comment (DanielSchiavini) [inline apps/main/src/app/dex/[network]/InjectServeSideData.tsx:20]:
```suggestion
```

apps/main/src/dex/lib/pools.ts
```diff
@@ -0,0 +1,145 @@
...
+        'tokens',
+        'tokensCountBy',
+        'tokensAll',
+        'tokensLowercase',
+        'tokenAddresses',
+        'tokenAddressesAll',
+        'pool.id',
+        'pool.name',
+        'pool.address',
+        'pool.gauge.address',
```
DanielSchiavini, line 107: This function is pretty much just moved elsewhere. Note though that I changed this line from `'pool.gauge'` - the whole object contains a class that cannot be serialized

OnlyJousting (approved):

PR #748, opened 2025-03-15
fix: slippage management for stableswap routes in router swap

OnlyJousting:
Changes:
- Restore stableswap specific slippage settings for router swap routes that only use stableswap pools

### DIFF
--- a/apps/main/src/dex/utils/utilsSwap.ts
+++ b/apps/main/src/dex/utils/utilsSwap.ts
@@ -111,6 +111,7 @@ export function _parseRoutesAndOutput(
       : getIsLowExchangeRate(haveCryptoRoutes, toAmount, fromAmount),
     isHighImpact: priceImpact > +parsedMaxSlippage,
     isHighSlippage: haveCryptoRoutes ? false : Number(exchangeRates[0]) > 0.98,
+    isStableswapRoute: !haveCryptoRoutes,
     maxSlippage: parsedMaxSlippage,
     priceImpact: priceImpact,
     routes: parsedRouterRoutes.routes,
--- a/packages/curve-ui-kit/src/features/user-profile/store.ts
+++ b/packages/curve-ui-kit/src/features/user-profile/store.ts
@@ -7,8 +7,8 @@ import type { ThemeKey } from '@ui-kit/themes/basic-theme'
 
 type State = {
   theme: ThemeKey
-  /** Key is either 'global', 'stable' or a chainIdPoolId from getChainPoolIdActiveKey. */
-  maxSlippage: { global: string; stable: string } & Partial<Record<string, string>>
+  /** Key is either 'crypto', 'stable' or a chainIdPoolId from getChainPoolIdActiveKey. */
+  maxSlippage: { crypto: string; stable: string } & Partial<Record<string, string>>
   isAdvancedMode: boolean
   hideSmallPools: boolean
 }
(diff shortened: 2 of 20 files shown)


packages/curve-ui-kit/src/features/user-profile/store.ts
```diff
@@ -8,7 +8,7 @@ import type { ThemeKey } from '@ui-kit/themes/basic-theme'
 type State = {
   theme: ThemeKey
   /** Key is either 'global' or a chainIdPoolId from getChainPoolIdActiveKey. */
-  maxSlippage: { global: string } & Partial<Record<string, string>>
+  maxSlippage: { global: string; stable: string } & Partial<Record<string, string>>
```
0xAlunara, line 11: Don't forget to update the docs

packages/curve-ui-kit/src/features/user-profile/store.ts
```diff
@@ -44,7 +47,7 @@ const INITIAL_THEME =
 
 const INITIAL_STATE: State = {
   theme: INITIAL_THEME,
-  maxSlippage: { global: '0.1' },
+  maxSlippage: { global: '0.1', stable: '0.03' },
```
0xAlunara, line 50: should we rename the original 'global' to 'crypto' for clarity now that there's two 'global' ones?

apps/main/src/dex/utils/utilsSwap.ts
```diff
@@ -111,6 +111,7 @@ export function _parseRoutesAndOutput(
       : getIsLowExchangeRate(haveCryptoRoutes, toAmount, fromAmount),
     isHighImpact: priceImpact > +parsedMaxSlippage,
     isHighSlippage: haveCryptoRoutes ? false : Number(exchangeRates[0]) > 0.98,
+    isStableswapRoute: !haveCryptoRoutes,
```
0xAlunara, line 114: could've just gone with the same `haveCryptoRoutes`, but no biggy

0xAlunara (changes requested):
I think if you look for
```tsx
  const maxSlippage = useMemo(() => {
    if (globalMaxSlippage) return globalMaxSlippage
    if (!pool) return ''

    return pool.isCrypto ? '0.1' : '0.03'
  }, [globalMaxSlippage, pool])
```
on the pool page, you'll notice that it uses the global crypto slippage, even if the pool page you're on is for example 3pool.

So in short, we've only fixed it for quickswap but not the pool page I think. I'll ignore lending for now as there's slippage in lending pages as well, but prob outside of the scope of this PR

packages/curve-ui-kit/src/features/user-profile/store.ts
OnlyJousting: I think global is ok, stable is only for pure stableswap routes, global is for routes that has a cryptopool but can also include stableswap pools in the route

OnlyJousting: I change my mind, crypto makes more sense lol

0xAlunara (approved):

OnlyJousting:
> Collaborator

on the pool page. globalMaxSlippage is called using a pool key. if it's not already set it will be checked with isCrypto below and the default slippage value will be set. changes to the slippage setting for this pool, using the pool key, will be then available on next visit and globalMaxSlippage will be true.

OnlyJousting:
> I think if you look for
> 
> ```tsx
>   const maxSlippage = useMemo(() => {
>     if (globalMaxSlippage) return globalMaxSlippage
>     if (!pool) return ''
> 
>     return pool.isCrypto ? '0.1' : '0.03'
>   }, [globalMaxSlippage, pool])
> ```
> 
> on the pool page, you'll notice that it uses the global crypto slippage, even if the pool page you're on is for example 3pool.
> 
> So in short, we've only fixed it for quickswap but not the pool page I think. I'll ignore lending for now as there's slippage in lending pages as well, but prob outside of the scope of this PR

on the pool page. globalMaxSlippage is called using a pool key. if it's not already set it will be checked with isCrypto below and the default slippage value will be set. changes to the slippage setting for this pool, using the pool key, will be then available on next visit and globalMaxSlippage will be true.

mimaklas (approved):

PR #750, opened 2025-03-15
mich requests for token selector

0xAlunara:
1. In quickswap, the 'to' token list no longer uses user balances and token prices. Just a list sorted by volume now.

2. In the 'from' token list, the 'tokens by 24h volume' will now include tokens hidden away as 'dust' if there's a 'my tokens' section. The list is still sorted by volume and ignores user balances. If the user clicks on 'show dust', the very same dust tokens will no longer show up in the '24h volume' section to avoid duplicate entries.

I've been internally debating about whether or not we should show token balances in the bottom section if they're hidden behind the 'show dust' button. I've opted to show balances, even though you can argue it breaks consistency with the other tokens in the list. This way the user can be sure that the token is really the token they're expecting. It would only be confusing if a user has a balance of $0.01 but it wouldn't show; is it the same token or not?

### DIFF
--- a/apps/main/src/dex/components/PageRouterSwap/index.tsx
+++ b/apps/main/src/dex/components/PageRouterSwap/index.tsx
@@ -445,9 +445,7 @@ const QuickSwap = ({
             <TokenSelector
               selectedToken={toToken}
               tokens={tokens}
-              balances={userBalancesMapper}
               disabled={isDisable || !toToken}
-              tokenPrices={usdRatesMapper}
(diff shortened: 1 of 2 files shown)


OnlyJousting (approved):

DanielSchiavini (approved):
This component is getting too complicated :skull_and_crossbones:

0xAlunara:
@DanielSchiavini we haven't even gotten to the lists feature yet 👀

PR #757, opened 2025-03-18
fix: token selector bugs

0xAlunara:
Fixes three smol bugs that appear in edge cases. Brought back compact mode I removed earlier, turns it was a bad idea to make bake it in by default. Whenever you searched and you had only 1 result the entire modal would shrink in height. Oops!

### DIFF
--- a/apps/main/src/dex/components/PageCreatePool/TokensInPool/SelectTokenButton.tsx
+++ b/apps/main/src/dex/components/PageCreatePool/TokensInPool/SelectTokenButton.tsx
@@ -168,6 +168,7 @@ const SelectTokenButton = ({
           disabledTokens={disabledKeys ?? []}
           disableSorting={true}
           disableMyTokens={false}
+          compact={false}
           customOptions={
             swapType === STABLESWAP && (
               <Checkbox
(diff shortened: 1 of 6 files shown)


OnlyJousting (approved):

apps/main/src/dex/components/PagePool/Swap/index.tsx
```diff
@@ -381,6 +381,7 @@ const Swap = ({
                 disabled={isDisabled || selectList.length === 0}
                 showSearch={false}
                 showManageList={false}
+                compact={true}
```
DanielSchiavini, line 384: compact is optional bool, the true could be ommitted
```suggestion
                compact
```

apps/main/src/dex/components/PageCreatePool/TokensInPool/SelectTokenButton.tsx
```diff
@@ -168,6 +168,7 @@ const SelectTokenButton = ({
           disabledTokens={disabledKeys ?? []}
           disableSorting={true}
           disableMyTokens={false}
+          compact={false}
```
DanielSchiavini, line 171: compact is optional bool, the prop could be ommitted
```suggestion
```

DanielSchiavini (approved):

apps/main/src/dex/components/PageCreatePool/TokensInPool/SelectTokenButton.tsx
0xAlunara: It's only optional on the TokenSelector root component, not the modal component. Since this is an old style button only the modal component is needed here, hence the big amount of props. This was a deliberate choice as I'd rather not having a chain of intermediate optionals resorting to different defaults.

PR #760, opened 2025-03-19
feat: APR tooltips

DanielSchiavini:
- Update tooltip component design
- Implement APR tooltips


packages/curve-ui-kit/src/themes/mui-tooltip.ts
```diff
@@ -0,0 +1,24 @@
...
+import type { DesignSystem } from './design'
+
+const { Spacing } = SizesAndSpaces
+
+export const defineMuiTooltip = (design: DesignSystem, typography: TypographyOptions): Components['MuiTooltip'] => ({
+  styleOverrides: {
+    tooltip: {
+      backgroundColor: design.Layer[3].Fill,
+      boxShadow: getShadow(design, 2),
+      minWidth: '27.5rem', // hardcoded 440px in the design
```
0xAlunara, line 15: `SizesAndSpaces.MinWidth.tooltip` in `1_sizes_spaces`?

apps/main/src/loan/components/PageLlamaMarkets/cells/cell.format.ts
```diff
@@ -0,0 +1,7 @@
+import { LlamaMarketType } from '@/loan/entities/llama-markets'
+import type { RewardsAction } from '@ui/CampaignRewards/types'
+
+export const formatPercent = (rate: number) => `${rate.toPrecision(4)}%`
+
+export const getRewardsAction = (marketType: LlamaMarketType, type: 'borrow' | 'lend'): RewardsAction =>
```
0xAlunara, line 6:
```suggestion
export const getRewardsAction = (marketType: LlamaMarketType, type: RateType): RewardsAction =>
```

apps/main/src/loan/components/PageLlamaMarkets/cells/index.ts
```diff
@@ -5,3 +5,5 @@ export * from './RateCell'
 export * from './PercentageCell'
 export * from './ErrorCell'
 export * from './PriceCell'
+export { RateTooltipContent } from '@/loan/components/PageLlamaMarkets/cells/RateCellTooltip'
```
0xAlunara, line 8: Relative imports for consistency?

packages/curve-ui-kit/src/shared/ui/TooltipContent.tsx
```diff
@@ -0,0 +1,25 @@
...
+import Typography from '@mui/material/Typography'
+import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
+import { InvertTheme } from './ThemeProvider'
+
+const { Spacing } = SizesAndSpaces
+
+// note: use this in the Tooltip component to remove default (inverted color) padding
+export const OmitTooltipPadding = { tooltip: { sx: { '&': { padding: 0 } } } }
+
+export const TooltipContent = ({ title, children }: { title: ReactNode; children: ReactNode }) => (
```
0xAlunara, line 12: Maybe make the `title` property optional? I assume we're going to need to use this for every tooltip we have due to the theme inversion, and sometimes tooltips don't have a title.

```suggestion
export const TooltipContent = ({ title, children }: { title?: ReactNode; children: ReactNode }) => (
```

packages/curve-ui-kit/src/themes/mui-tooltip.ts
DanielSchiavini: Good point!

packages/curve-ui-kit/src/shared/ui/TooltipContent.tsx
DanielSchiavini: For shorter tooltips I assumed the title would be the only thing

DanielSchiavini: We can always change if there is a use case

OnlyJousting:
https://github.com/user-attachments/assets/a1e1ecf7-8cd0-4999-aad8-f1d37c1f0097

points and favorites tooltips have this block of unintended color in all themes

0xAlunara:
image
Can't say I'm a fan of tooltips having this much unused space on the right if they don't contain a lot of text 🤔

0xAlunara:
image
Seems inversion of inversion isn't working for token icon tooltips. Should we do a pass on all MUI <Tooltip> usages and change it to use `@ui-kit/shared/ui/Tooltip` instead?

0xAlunara:
image
Almost there 👀

0xAlunara (changes requested):
image
Default behavior of the mui tooltip is that if the tooltip has no title or is an empty string, it wouldn't show the tooltip. This seems to break that behavior; if you for example go to https://curve-dapp-git-feat-tooltips-curvefi.vercel.app/dex/ethereum/pools/3pool/swap/ and open the token selection modal when swapping, the hover on the token icon now shows an empty tooltip.

0xAlunara (approved):

0xtutti (approved):

PR #768, opened 2025-03-20
feat: format gauge vote weight as percentage in GaugeVotesTable

OnlyJousting:
Changes:
- Formatted gauge vote weight to percentage instead of bps in gauge votes table on gauge page in DAO


0xAlunara (dismissed):
Not seeing any code that rounds to 2 decimals, but by checking some gauges it never exceeds 2 decimals regardless

apps/main/src/dao/components/PageGauge/GaugeVotesTable/index.tsx
```diff
@@ -17,6 +17,9 @@ interface GaugeVotesTableProps {
   tableMinWidth: number
 }
 
+// weight is recieved in bps, 10000 = 100%
```
DanielSchiavini, line 20: would be nice to call it `weightBps` or convert it when we receive it

DanielSchiavini (dismissed):

mimaklas (approved):

PR #775, opened 2025-03-21
feat: handle null values in ohlc data responses in order to be able to recieve oracle price data even when ohlc specific price data doesn't exist

OnlyJousting:
The API might only return values for oracle price (all ohlc data points might return null) in order to still display an accurate oracle price line for markets that doesn't have oracle ohlc data.

Changes:
- Added possibility of null to types for ohlc data in price-api package
- Handle null data in ohlc data fetches in lend and crvusd stores
- Bump prices-api package to 1.0.3

### DIFF
--- a/apps/main/src/lend/store/createOhlcChartSlice.ts
+++ b/apps/main/src/lend/store/createOhlcChartSlice.ts
@@ -452,15 +452,13 @@ const createOhlcChart = (set: SetState<State>, get: GetState<State>) => ({
             })
           }
 
-          if (item.open && item.close && item.high && item.low) {
-            ohlcDataArray.push({
-              time,
-              open: item.open,
-              close: item.close,
-              high: item.high,
-              low: item.low,
-            })
-          }
+          ohlcDataArray.push({
+            time,
+            open: item.open,
+            close: item.close,
+            high: item.high,
+            low: item.low,
+          })
         }
 
         set(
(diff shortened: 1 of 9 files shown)


apps/main/src/lend/store/createOhlcChartSlice.ts
```diff
@@ -233,29 +233,37 @@ const createOhlcChart = (set: SetState<State>, get: GetState<State>) => ({
...
-          volumeArray.push({
-            time,
-            value: item.volume,
-            color: item.open < item.close ? '#26a69982' : '#ef53507e',
-          })
+          if (item.volume && item.open && item.close) {
+            volumeArray.push({
+              time,
+              value: item.volume,
+              color: item.open < item.close ? '#26a69982' : '#ef53507e',
```
0xAlunara, line 240: could maybe opt for a neutral blue color if there's volume but no open and close?

apps/main/src/lend/store/createOhlcChartSlice.ts
```diff
@@ -292,7 +300,7 @@ const createOhlcChart = (set: SetState<State>, get: GetState<State>) => ({
             }
           }),
         )
-        console.log(error)
+        console.error(error)
```
0xAlunara, line 303: is this log actually still needed? 🤔

apps/main/src/loan/store/createOhlcChartSlice.ts
```diff
@@ -132,41 +132,34 @@ const createOhlcChart = (set: SetState<State>, get: GetState<State>) => ({
         for (const item of ohlc) {
           const time = item.time.getLocalTimestamp()
 
-          volumeArray = [
-            ...volumeArray,
-            {
+          if (item.volume && item.open && item.close) {
```
0xAlunara, line 135: same here, with the blue neutral color

apps/main/src/loan/store/createOhlcChartSlice.ts
```diff
@@ -225,41 +218,34 @@ const createOhlcChart = (set: SetState<State>, get: GetState<State>) => ({
...
```
0xAlunara, line 221: and here... lol a bit of a code smell we have the same logic 3 times 😅

0xAlunara (approved):

amfet42 (approved):

PR #777, opened 2025-03-22
feat: dudupe advanced settings and MUI-ify

0xAlunara:
Depends on: https://github.com/curvefi/curve-frontend/pull/760. This PR makes use of the default Mui tooltip, but once that PR is merged this PR should be updated to make use of the Tooltip wrapper instead.

_What this PR does_:
* Deduplicate AdvancedSettings and move to ui-kit
* Rename to SlippageSettings, as thats the only configuration offered
* Convert the modal into using Mui components
* Custom slippage no longer has a radio button. Click the text field deselects the radio buttons, which means you want to use the custom one. Of course, you can't leave it empty!
* Remove dependency on user profile feature and its store inside the settings modal. It now has an `onSave` callback property that can be used to pass on what needs to be done (usually just call `setMaxSlippage` from the user profile store).
* Add a storybook for the feature

_What this PR does **not**_:
* Style Mui input components. I know the input component itself lacks styling, and there's some one-off padding `sx` styling applied. The intention is to properly style input components in a different PR
* Deduplicate `DetailInfoSlippageTolerance`. I actually started out with this, but wanted to deduplicate `AdvancedSettings` first, so that's for a later PR
* Adjust card styling. The card footer (in this case the 'Save' button) has some incorrect padding. This again is for another PR
* Mess with the slippage back-end. I'm not touching any of the code that deals with setting slippage and its fallbacks. I'm merely using `maxSlippage` and `setMaxSlippage` as provided by the `useUserProfileStore`. This way I try to avoid breakig any of the default slippage behavior.

I also thought about adding a `compact` property to `ModalDialog`, but looking at `TokenSelector` it's slight different once again. I'll keep the manual `sx` for the `.MuiPaper-root` for now.

image


DanielSchiavini (dismissed):

OnlyJousting (dismissed):

DanielSchiavini (dismissed):

OnlyJousting (approved):

0xAlunara:
Sneaked in additional badge size changes as I was doing a live coding session with tutti.

0xtutti (approved):

0xtutti (approved):

DanielSchiavini (dismissed):

DanielSchiavini (approved):

0xtutti (approved):

PR #781, opened 2025-03-24
fix: set health to always be full

OnlyJousting:


apps/main/src/lend/lib/apiLending.ts
```diff
@@ -446,7 +446,7 @@ const user = {
       })
       .process(async (market) => {
         const userActiveKey = helpers.getUserActiveKey(api, market)
-        const [healthFull, healthNotFull] = await Promise.all([market.userHealth(), market.userHealth(false)])
+        const [healthFull, healthNotFull] = await Promise.all([market.userHealth(true), market.userHealth(true)])
```
DanielSchiavini, line 449: would be nice to have a constant, it's not readable when we pass `true` like this

OnlyJousting: this is not the final solution, was just testing if none full health was causing any side effects (it was)

PR #783, opened 2025-03-25
refactor: api store

0xAlunara:
A lot of supposedly generic components and (future generic) queries rely on the curve api. The problem is that the curve api is loaded and stored inside app-specific stores. That means if we want to move queries and components to curve-ui-kit, it's impossible to make them app-agnostic. This PR aims to extract curve apis into a single generic store, such that they are no longer coupled to apps.

An example would be the `tokenUsdRate` query. This query uses one of the APIs to call `getUsdRate` for a specific token address to gets its price. Very useful to make generic, but currently coupled to the lending app. After this PR, this query can then be moved to curve-ui-kit such that other places may use that same query (and its cache). Why it's currently using the lending api instead of the dex api is an unrelated question for another time :^)

When the apps initially load the APIs, the stores need to reset there state and fetch data. This now happens in so called `hydrate` functions, and still resides in the original app stores as they hydrate the various slices in the each app respectively. The api store itself is just a dumb container for API instances.

This PR aim is to just make the apis globally accessible. It does **not** aim to refactor or improve initialization speed. I want to keep the PR as small as possible.


apps/main/src/loan/hooks/usePageOnMount.ts
```diff
@@ -5,26 +5,32 @@ import { CONNECT_STAGE, ROUTE } from '@/loan/constants'
...
   const params = useParams() as UrlParams
   const { push } = useRouter()
   const { wallet, connect, disconnect, walletName, setWalletName } = useWallet()
   const [_, setChain] = useSetChain()
 
+  const curve = useApiStore((state) => state.stable)
+  const lending = useApiStore((state) => state.lending)
+  const updateStable = useApiStore((state) => state.updateStable)
+  const updateLending = useApiStore((state) => state.updateLending)
+  const setIsLoadingCurve = useApiStore((state) => state.setIsLoadingCurve)
```
DanielSchiavini, line 27: This app doesn't use curvejs? :thinking: 
```suggestion
```

apps/main/src/loan/hooks/usePageOnMount.ts
```diff
@@ -38,46 +44,61 @@ function usePageOnMount(chainIdNotRequired?: boolean) {
...
+          setIsLoadingLending(true)
+          setIsLoadingCurve(true)
 
-          updateConnectState(apiNew ? 'success' : '', '')
+          const api = await initLendApi(chainId, useWallet ? wallet : null)
+          if (api) {
+            updateLending(api)
+            updateConnectState('success', '')
+
+            hydrate(api, prevApi, wallet)
```
DanielSchiavini, line 88: seems that hydrate is called twice now. Shouldn't we merge the 2 init methods?

packages/curve-ui-kit/src/shared/useApiStore.ts
```diff
@@ -0,0 +1,58 @@
...
+export type CurveApi = typeof curveApi & { chainId: number }
+export type LendingApi = typeof lendingApi & { chainId: IChainId }
+export type StableApi = typeof stablecoinApi & { chainId: 1 }
+
+type State = {
+  curve: CurveApi | null
+  isLoadingCurve: boolean
+  lending: LendingApi | null
+  isLoadingLending: boolean
+  stable: StableApi | null
```
DanielSchiavini, line 17: We have been using crvUSD, loan and stablecoin as terms to refer to the crvUSD app - I've been trying to convert these references to `crvusd` as that's what is in the URL and the menus

packages/curve-ui-kit/src/lib/logging.ts
```diff
@@ -101,7 +101,7 @@ export function log(key: LogKey, status?: LogStatus | unknown, ...args: unknown[
 
   const hasDefinedStatus = status && Object.values(LogStatus).includes(status as LogStatus)
   const [formattedKeyString, keyStyles] = formatKeyArray(keyArray)
-  const restArgs = hasDefinedStatus ? args : [status, ...args]
+  const restArgs = (hasDefinedStatus ? args : [status, ...args]).filter(Boolean)
```
DanielSchiavini, line 104: I don't think we should filter out `0`
```suggestion
  const restArgs = (hasDefinedStatus ? args : [status, ...args]).filter(a => a != null)
```

DanielSchiavini (commented):
I am surprised by this change, I thought we agreed to keep the curve-ui-kit independent of any stores to avoid coupling. I'm not fully against this though :thinking: 
We need to have a chat as I was about to start refactoring to use the new merged [llamalend lib](https://github.com/curvefi/curve-llamalend.js)

apps/main/src/dao/store/createGaugesSlice.ts
```diff
@@ -361,7 +362,7 @@ const createGaugesSlice = (set: SetState<State>, get: GetState<State>): GaugesSl
     },
 
     castVote: async (userAddress: string, gaugeAddress: string, voteWeight: number) => {
-      const curve = get().curve
+      const curve = useApiStore.getState().curve
```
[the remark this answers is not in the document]
0xAlunara: Destructuring now when possible (aka where we don't need null assertion)

apps/main/src/loan/hooks/usePageOnMount.ts
0xAlunara: Good catch, actually have to use the (unused atm) setIsLoadingStable here

apps/main/src/lend/hooks/usePageOnMount.ts
```diff
@@ -33,23 +36,30 @@ function usePageOnMount(chainIdNotRequired?: boolean) {
       if (options) {
         try {
           const [chainId, useWallet] = options
-          const prevApi = api ?? null
-          updateGlobalStoreByKey('isLoadingApi', true)
-          updateGlobalStoreByKey('isLoadingCurve', true) // remove -> use connectState
+          const prevApi = lending ?? null
+
+          setIsLoadingLending(true)
+          setIsLoadingCurve(true)
```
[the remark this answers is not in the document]
0xAlunara: Aye, removed

DanielSchiavini (approved):

OnlyJousting (dismissed):

DanielSchiavini:
Import conflicts again :disappointed:

0xAlunara:
Do not merge: upon testing in prod the crvusd tab and lending can crash, need to fix that first.

DanielSchiavini (dismissed):

0xAlunara:
Had a few tests with mima and the crashes were relatively easy to fix. One was an incorrect chainId, the other a missing async. Should be all good now!

DanielSchiavini (approved):

DanielSchiavini (approved):

PR #792, opened 2025-03-26
fix: update slippage settings modal to reset form values on reopen

0xAlunara:
When you open the slippage settings modal, change it but discard and then re-open, the form values don't reset and it will show the discarded values even though you did not click 'Save'. This edge case was handled in the old modal but must've slipped my mind when re-implementing. It's only a cosmetic issue.

### DIFF
--- a/packages/curve-ui-kit/src/features/slippage-settings/ui/SlippageSettingsModal.tsx
+++ b/packages/curve-ui-kit/src/features/slippage-settings/ui/SlippageSettingsModal.tsx
@@ -128,9 +128,12 @@ export const SlippageSettingsModal = ({ isOpen, maxSlippage, onSave, onClose }: 
    */
   const [lastError, setLastError] = useState<Error | undefined>(undefined)
 
+  // Update form values when slippage changes or modal is closed without saving and re-opened
   useEffect(() => {
-    setFormValues(initFormValues(maxSlippage))
-  }, [maxSlippage])
+    if (isOpen) {
+      setFormValues(initFormValues(maxSlippage))
+    }
+  }, [isOpen, maxSlippage])
 
   useEffect(() => {
     if (error) {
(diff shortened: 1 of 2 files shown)


packages/curve-ui-kit/src/features/slippage-settings/ui/SlippageSettingsModal.tsx
```diff
@@ -128,9 +128,12 @@ export const SlippageSettingsModal = ({ isOpen, maxSlippage, onSave, onClose }:
    */
   const [lastError, setLastError] = useState<Error | undefined>(undefined)
 
+  // Update form values when slippage changes or modal is closed without saving and re-opened
   useEffect(() => {
-    setFormValues(initFormValues(maxSlippage))
-  }, [maxSlippage])
+    if (isOpen) {
```
DanielSchiavini, line 133: Maybe it makes more sense to do it when you close it? It might prevent a first incorrect render.
```suggestion
    if (!isOpen) {
```

packages/curve-ui-kit/src/features/slippage-settings/ui/SlippageSettingsModal.tsx
```diff
@@ -128,9 +128,12 @@ export const SlippageSettingsModal = ({ isOpen, maxSlippage, onSave, onClose }:
    */
   const [lastError, setLastError] = useState<Error | undefined>(undefined)
 
+  // Update form values when slippage changes or modal is closed without saving and re-opened
```
DanielSchiavini, line 131: hmm usually this would be handled automatically by de-mounting the modal. But mui likes to always render stuff :thinking:

packages/curve-ui-kit/src/features/slippage-settings/ui/SlippageSettingsModal.tsx
0xAlunara: I'm not against per se, but wouldn't the first render be covered by the initial state in
```const [formValues, setFormValues] = useState(initFormValues(maxSlippage))```?

DanielSchiavini: the first render **_ever_** would be covered, but not the 2nd one (unless the component is remounted)

0xtutti (dismissed):

packages/curve-ui-kit/src/features/slippage-settings/ui/SlippageSettingsModal.tsx
0xAlunara: Found this useful function we can use to reset the form only when the exit transition has been completed
image

OnlyJousting (approved):

PR #803, opened 2025-03-28
feat: add on-chain market rate fetching and validation suite

OnlyJousting:
Lend markets were using API data to display borrow and lending rates causing there to be some confusion after interaction with a position since the borrow/lend rates wouldn't update because of API delays.

I kept the API data as a fallback.

Changes:
- Added on chain rates query to get fresh data when a wallet is connected


0xtutti (approved):

mimaklas (approved):

Macket (dismissed):

apps/main/src/lend/components/DetailInfoRate.tsx
```diff
@@ -32,10 +35,15 @@ const DetailInfoRate = ({
           '?'
         ) : (
           <strong>
-            {formatNumber(isBorrow ? rates?.borrowApy : rates?.lendApr, {
-              ...FORMAT_OPTIONS.PERCENT,
-              defaultValue: '-',
-            })}
+            {formatNumber(
+              isBorrow
+                ? (onchainData?.rates?.borrowApy ?? rates?.borrowApy)
```
DanielSchiavini, line 40: this logic should be moved to a hook imo

apps/main/src/lend/entities/market-onchain-rate.ts
```diff
@@ -0,0 +1,32 @@
...
+
+type Params = ChainParams<ChainId> & {
+  marketId: string
+}
+
+const _fetchOnChainMarketRate = async ({ chainId, marketId }: ChainQuery<ChainId> & { marketId: string }) => {
+  const api = useApiStore.getState().lending
+  const market = api?.getOneWayMarket(marketId)
+
+  if (!market) return null
```
DanielSchiavini, line 15: I don't think this is reachable, api is validated in the suite and it will throw an error if the marketId is unknown

apps/main/src/lend/entities/market-onchain-rate.ts
```diff
@@ -0,0 +1,32 @@
+import { lendingJsValidationSuite } from '@/lend/entities/validation/lending-js-validation'
+import { ChainId } from '@/lend/types/lend.types'
+import type { ChainParams, ChainQuery } from '@ui-kit/lib/model/query'
+import { queryFactory } from '@ui-kit/lib/model/query'
+import { useApiStore } from '@ui-kit/shared/useApiStore'
+
+type Params = ChainParams<ChainId> & {
+  marketId: string
+}
```
DanielSchiavini, line 9: There should be a distinction between the query (after validation) and params (before validation)
```suggestion
type MarketQuery = ChainQuery<ChainId> & { marketId: string }
type MarketParams = FieldsOf<MarketQuery>
```

0xtutti (approved):

DanielSchiavini:
My comments are not blocking if someone is waiting for this feature it can be merged

apps/main/src/lend/components/DetailInfoRate.tsx
```diff
@@ -16,11 +18,22 @@ const DetailInfoRate = ({
...
   futureRates?: FutureRates | undefined | null
 }) => {
+  // uses onchain data if available
+  const { data: onChainData } = useMarketOnChainRates({ chainId: rChainId, marketId: rOwmId })
   const ratesResp = useStore((state) => state.markets.ratesMapper[rChainId]?.[rOwmId])
 
   const { rates, error } = ratesResp ?? {}
   const futureRate = isBorrow ? futureRates?.borrowApy : futureRates?.lendApr
 
+  const rate = useMemo(() => {
```
DanielSchiavini, line 28: Sorry for the confusion, what I meant here is that we should just call `useSupplyTotalApr` and let the hook decide where the data comes from. Basically, deciding where data comes from does not belong in an UI component but it's business logic

DanielSchiavini (dismissed):

DanielSchiavini (approved):

mimaklas (approved):

PR #815, opened 2025-03-31
chore: configure typescript-eslint

DanielSchiavini:


OnlyJousting (approved):

packages/eslint-config-custom/index.js
```diff
@@ -26,6 +26,7 @@ module.exports = {
         ],
       },
     ],
+    '@typescript-eslint/no-floating-promises': 'warn',
```
DanielSchiavini, line 29: Maybe info until they are fixed we might miss other important warns otherwise

0xtutti (approved):

packages/eslint-config-custom/index.js
0xAlunara: Sadly 'info' doesn't exist for eslint, only 'off', 'warn' and 'error' 😞

PR #818, opened 2025-04-01
Fix Badge Coloring

0xtutti:
Adds inverted colors for badges


0xAlunara (dismissed):

OnlyJousting (approved):

apps/lend/src/locales/en/messages.js
```diff
@@ -0,0 +1 @@
+/*eslint-disable*/module.exports={messages:JSON.parse("{}")};
```
DanielSchiavini, line 1: these translation files are super old they should be deleted :grin:

packages/curve-ui-kit/src/themes/chip/mui-chip.ts
```diff
@@ -7,6 +7,15 @@ import { Grays } from '@ui-kit/themes/design/0_primitives'
 import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
 import { TypographyVariantKey } from '@ui-kit/themes/typography'
```
DanielSchiavini, line 9: fyi the lint warn can be fixed with `yarn lint:fix`

0xtutti: should be fixed now.

DanielSchiavini (approved):

0xAlunara (approved):

PR #821, opened 2025-04-02
chore: remove unused i18n messages files

0xAlunara:
These aren't used anymore since we stopped supporting i18n


DanielSchiavini (approved):
I was sure these were removed in Tutti's PR :thinking:  :laughing:

OnlyJousting (approved):

PR #822, opened 2025-04-02
chore: fix floating promises linting errors

0xAlunara:
This PR mostly purposefully `voids` promises that are already not being awaited. This includes promises in `useEffect`, `useCallback` and the various slices. Not awaiting was already the default behavior and seems to be intended. You shouldn't await in `useEffect` and `useCallback`, and in the slices it seems data is being fetched haphazardly that doesn't have to be blocking. Only awaits added are to the `hydrate` function on init, and `async function getPools` in `apps\main\src\dex\lib\pools.ts`.

### DIFF
--- a/apps/main/src/app/crvusd/client.tsx
+++ b/apps/main/src/app/crvusd/client.tsx
@@ -69,8 +69,8 @@ export const App = ({ children }: { children: ReactNode }) => {
   usePageVisibleInterval(
     () => {
       if (isPageVisible && curve) {
-        fetchAllStoredUsdRates(curve)
-        fetchGasInfo(curve)
+        void fetchAllStoredUsdRates(curve)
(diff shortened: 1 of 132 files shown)


DanielSchiavini (commented):
imo we should _at least_ replace `void` by `.catch(console.error)` as ignoring is terrible for debugging

alternatively I'd prefer to ignore the eslint rule for existing code, the default behavior should at least show the error.

> You shouldn't await in useEffect and useCallback
but we need to handle the errors _somehow_

0xAlunara:
`void` does not mean errors aren't being handled anymore, it's merely for the linter to mark the promise is purposefully not being awaited
image
image

DanielSchiavini (approved):

mimaklas (approved):

PR #829, opened 2025-04-03
fix: delayed slippage experience by removing dependency on maxSlippage from _parseRoutesAndOutput

0xAlunara:
Max slippage was being passed all along the pipeline into `_parseRoutesAndOutput`, only for it to potentially modify it and set `isHighImpact`. If it modified slippage it would no longer be in sync with whatever the slippage was configured as in the store. The dependency on maxSlippage was also silly in that the slippage isn't being used when determining the route, as `const { route: routes, output } = await curve.router.getBestRouteAndOutput(fromAddress, toAddress, fromAmount)` has no dependency on `maxSlippage`.

This PR extracts that logic from the pipeline into a new function `getSlippageImpact()` that uses the output from `_parseRoutesAndOutput`. As a result, the slippage should now come from a single source of truth, and the quickswap page is the one that determines what slippage to use using

```ts
  const cryptoMaxSlippage = useUserProfileStore((state) => state.maxSlippage.crypto)
  const stableMaxSlippage = useUserProfileStore((state) => state.maxSlippage.stable)
  const isStableswapRoute = routesAndOutput?.isStableswapRoute
  const storeMaxSlippage = isStableswapRoute ? stableMaxSlippage : cryptoMaxSlippage
  const slippageImpact = routesAndOutput
    ? getSlippageImpact({ maxSlippage: storeMaxSlippage, ...routesAndOutput })
    : null
```

This also means that the quickswap page is the one that determines if `isHighImpact` is true or not.

### DIFF
--- a/apps/main/src/dex/components/PageRouterSwap/index.tsx
+++ b/apps/main/src/dex/components/PageRouterSwap/index.tsx
@@ -478,7 +478,7 @@ const QuickSwap = ({
         <DetailInfoPriceImpact
           loading={routesAndOutputLoading}
           priceImpact={routesAndOutput?.priceImpact}
-          isHighImpact={slippageImpact ? slippageImpact.isHighImpact : null}
+          isHighImpact={slippageImpact?.isHighImpact ?? null}
         />
         <DetailInfoTradeRoute
           params={params}


0xtutti (dismissed):

0xAlunara:
Needs extensive testing so don't merge willy nilly

apps/main/src/dex/components/PageRouterSwap/index.tsx
```diff
@@ -85,6 +86,9 @@ const QuickSwap = ({
   const stableMaxSlippage = useUserProfileStore((state) => state.maxSlippage.stable)
   const isStableswapRoute = routesAndOutput?.isStableswapRoute
   const storeMaxSlippage = isStableswapRoute ? stableMaxSlippage : cryptoMaxSlippage
+  const { isHighImpact } = routesAndOutput
+    ? getSlippageImpact(storeMaxSlippage, routesAndOutput)
+    : { isHighImpact: false }
```
DanielSchiavini, line 91: this looks like another case of showing incorrect data during loading or when errors occur. Maybe better to leave it undefined?
```suggestion
  const isHighImpact = routesAndOutput && getSlippageImpact(storeMaxSlippage, routesAndOutput).isHighImpact
```

0xAlunara: I don't think there's a visual difference between false and undefined. In both cases the message won't get shown

apps/main/src/dex/components/PagePool/Swap/index.tsx
```diff
@@ -527,7 +530,7 @@ const Swap = ({
         <DetailInfoPriceImpact
           loading={exchangeOutput.loading}
           priceImpact={exchangeOutput.priceImpact}
-          isHighImpact={exchangeOutput.isHighImpact}
+          isHighImpact={slippageImpact ? slippageImpact.isHighImpact : null}
```
DanielSchiavini, line 533:
```suggestion
          isHighImpact={slippageImpact?.isHighImpact ?? null}
```

apps/main/src/dex/components/PageRouterSwap/index.tsx
```diff
@@ -473,7 +478,7 @@ const QuickSwap = ({
         <DetailInfoPriceImpact
           loading={routesAndOutputLoading}
           priceImpact={routesAndOutput?.priceImpact}
-          isHighImpact={routesAndOutput?.isHighImpact}
+          isHighImpact={slippageImpact ? slippageImpact.isHighImpact : null}
```
DanielSchiavini, line 481:
```suggestion
          isHighImpact={slippageImpact?.isHighImpact ?? null}
```

DanielSchiavini (approved):

0xtutti (dismissed):

OnlyJousting (approved):

DanielSchiavini (approved):

PR #845, opened 2025-04-09
feat: filter small markets

DanielSchiavini:
- make the default liquidity filter be $10k based on the "small markets" option
- allow the user to change the filters afterwards


tests/cypress/e2e/loan/llamalend-markets.cy.ts
```diff
@@ -233,12 +232,11 @@ function selectChain(chain: string) {
 
 const selectCoin = (symbol: string, type: TokenType) => {
```
DanielSchiavini, line 233: this can still fail sometimes when the coin only exists in small markets

packages/curve-ui-kit/src/features/user-profile/store.ts
```diff
@@ -5,6 +5,8 @@ import { devtools, persist } from 'zustand/middleware'
 import type { PersistOptions } from 'zustand/middleware/persist'
 import type { ThemeKey } from '@ui-kit/themes/basic-theme'
 
+export const SMALL_POOL_VOLUME = 10000
```
0xAlunara, line 8: bit of a nitpick but I think it should be named `SMALL_POOL_TVL` as we use it in TVL comparisons, not volume

apps/main/src/loan/components/PageLlamaMarkets/Page.tsx
```diff
@@ -61,6 +63,8 @@ export const LlamaMarketsPage = (props: CrvUsdServerData) => {
   useInjectServerData(props)
   const { signerAddress } = useWallet()
   const { data, isError, isLoading } = useLlamaMarkets(signerAddress)
+  const minLiquidity = useUserProfileStore((s) => s.hideSmallPools) ? SMALL_POOL_VOLUME : 0
```
0xAlunara, line 66: would it be nicer to have a getter function in the store that returns minLiquidity directly?

`const minLiquidity = useUserProfileStore((s) => s.minLiquidity)` where minLiquidity is a small getter function that returns the value of `hideSmallPools ? SMALL_POOL_VOLUME : 0`.

Benefit would be not having to import and perform manual logic with `SMALL_POOL_VOLUME`
Downside would be added complexity to the store, even if it's such a simple getter.

I think the benefit outweighs the downside but it's not a blocker, just a question of preference.

apps/main/src/loan/components/PageLlamaMarkets/filters/RangeSliderFilter.tsx
```diff
@@ -16,7 +16,7 @@ import { cleanColumnId } from '@ui-kit/shared/ui/DataTable/TableVisibilitySettin
 const getMaxValueFromData = <T extends any, K extends DeepKeys<T>>(data: T[], field: K) =>
   data.reduce((acc, item) => Math.max(acc, get(item, field) as number), 0)
 
-type Range = [number, number]
+type Range<T = number> = [T, T]
```
0xAlunara, line 19: Curious to know why you made this generic. I don't see any case where you're using the range type where `T != number`

0xAlunara (approved):

0xAlunara:
Please be aware that the 'hide small pools' toggle inside the user profile isn't accessible yet on prod
image

It's accessible on the dex pools page, however, it's linked to the same boolean in the user profile store
image

However, it is **not** linked yet to this boolean toggle on the lend page.
image

Not too important for now but good to remember

apps/main/src/dex/store/createCacheSlice.ts
```diff
@@ -3,6 +3,7 @@ import type { State } from '@/dex/store/useStore'
 import { ChainId, PoolDataCacheMapper, type ValueMapperCached } from '@/dex/types/main.types'
 import { sleep } from '@/dex/utils'
 import { logSuccess } from '@ui-kit/lib'
+import { SMALL_POOL_VOLUME } from '@ui-kit/features/user-profile/store'
```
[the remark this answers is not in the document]
DanielSchiavini: I'll fix this in #849

packages/curve-ui-kit/src/features/user-profile/store.ts
DanielSchiavini: I'll fix this in https://github.com/curvefi/curve-frontend/pull/849

apps/main/src/loan/components/PageLlamaMarkets/Page.tsx
DanielSchiavini: Yeah I had the same thought now that they are based on the same constant. I'll fix this in https://github.com/curvefi/curve-frontend/pull/849

apps/main/src/loan/components/PageLlamaMarkets/filters/RangeSliderFilter.tsx
DanielSchiavini: Yeah it's not really used in this project yet. It's just to make it more reusable, but I'll revert this in https://github.com/curvefi/curve-frontend/pull/849

tests/cypress/e2e/loan/llamalend-markets.cy.ts
DanielSchiavini: I'll fix this in https://github.com/curvefi/curve-frontend/pull/849

apps/main/src/loan/components/PageLlamaMarkets/Page.tsx
DanielSchiavini: I remember now why I didn't do it :laughing: this constant is only used with the store here.
The other apps using the constant are adding it to the chain configuration

DanielSchiavini:
> Please be aware that the 'hide small pools' toggle inside the user profile isn't accessible yet on prod

Yes we knew that - in this page the user is able to change the filter to 0 manually

PR #870, opened 2025-04-16
refactor: get rid of usePageOnMount (dao)

DanielSchiavini:
- get rid of usePageOnMount
- move layout to [network] folder and access the network name via nextjs params
- get rid of unused router utils


apps/main/src/dao/utils/utilsRouter.ts
```diff
@@ -1,86 +1,9 @@
...
+import { type UrlParams } from '@/dao/types/dao.types'
+import { getInternalUrl } from '@ui-kit/shared/routes'
 
 /** Get the path for the given route in this app */
 export const getPath = ({ network }: UrlParams, route: string) => getInternalUrl('dao', network, route)
 
 export const getEthPath = (route: string) => getPath({ network: 'ethereum' }, route)
 
-export function parseParams(params: UrlParams, chainIdNotRequired?: boolean) {
-  const { proposalId, userAddress, gaugeAddress, formType } = params
```
0xAlunara, line 12: That's a lot of deletion lol

0xAlunara (dismissed):

apps/main/src/dao/utils/utilsRouter.ts
DanielSchiavini: So nice

apps/main/src/dao/utils/utilsRouter.ts
```diff
@@ -1,86 +1,9 @@
...
-    }
-  }
-}
-
-export function getRestFullPathname() {
-  const restPathnames = window.location.pathname.substring(1).split('/') ?? []
-  const { rNetworkIdx } = getNetworkFromUrl()
-  return restPathnames.slice(rNetworkIdx + 1, restPathnames.length).join('/')
-}
+export const getRestFullPathname = () => window.location.pathname.substring(1).split('/').slice(1).join('/')
```
DanielSchiavini, line 9:
```suggestion
/**
 * Get the part of a path after the network, removing the leading slash and the first two parts.
 * For example /:app/:network/:page/:id => `:page/:id`
 */
export const getRestFullPathname = () => window.location.pathname.substring(1).split('/').slice(2).join('/')
```

0xAlunara (approved):

0xtutti (dismissed):

0xAlunara (approved):

PR #866, opened 2025-04-16
refactor: connection logic (3/dex)

DanielSchiavini:
- use connection logic from https://github.com/curvefi/curve-frontend/pull/855 in the lend app
- depends on #859

PR #859, opened 2025-04-15
refactor: connection logic (2/lend)

DanielSchiavini:
- use connection logic from https://github.com/curvefi/curve-frontend/pull/855 in the lend app

PR #872, opened 2025-04-18
refactor: connection logic (2+3)

DanielSchiavini:
Includes:
- #870
- #859 
- #866
- #881
- #882 
- #884


packages/curve-ui-kit/src/features/connect-wallet/lib/ConnectionContext.tsx
```diff
@@ -118,34 +123,35 @@ export const ConnectionProvider = <
         if (walletChainId && walletChainId !== chainId) {
```
DanielSchiavini, line 123: We need to check if the document is visible to avoid fighting between tabs

```suggestion
        if (walletChainId && walletChainId !== chainId && !document.hidden) {
```

DanielSchiavini: We need to check if the document is visible to avoid fighting between tabs

```suggestion
        if (walletChainId && walletChainId !== chainId && !document.hidden) {
```

0xAlunara (approved):

mimaklas (approved):

PR #851, opened 2025-04-10
feat: refactor dao page fetching to react query

OnlyJousting:
Refactored DAO endpoints to react-query and the prices api package:
- Proposals list
- Proposal
- User gauge weight votes
- User gauge weigh vote next vote time to react-query
- User gauge votes
- User proposal vote
- User proposal votes
- User locks
- Analytics veCRV stats

Additional changes:
- Removed cache slice in DAO app

PR #887, opened 2025-04-30
fix: dao new queries with connection context

DanielSchiavini:
- fixes conflicts that were not detected properly in #884 due to new  code introduced in #851

### DIFF
--- a/apps/main/src/dao/entities/user-gauge-vote-next-time.ts
+++ b/apps/main/src/dao/entities/user-gauge-vote-next-time.ts
@@ -1,39 +1,33 @@
 import { enforce, group, test } from 'vest'
 import { chainValidationGroup, curvejsValidationGroup } from '@/dao/entities/validation/curvejs-validation'
-import type { ChainId } from '@/dao/types/dao.types'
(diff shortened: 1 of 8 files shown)


apps/main/src/dao/entities/user-gauge-vote-next-time.ts
```diff
@@ -1,39 +1,33 @@
...
 export const { useQuery: useUserGaugeVoteNextTimeQuery, invalidate: invalidateUserGaugeVoteNextTimeQuery } =
   queryFactory({
-    queryKey: (params: QueryKeyParams) =>
+    queryKey: (params: UserGaugeParams) =>
       [
         'user-gauge-vote-next-time',
         { chainId: params.chainId },
         { gaugeAddress: params.gaugeAddress },
         { userAddress: params.userAddress },
-        { enabled: params.enabled },
```
DanielSchiavini, line 31: @OnlyJousting it is not necessary to add `enabled` here. Every `useQuery` hook allows `enabled` as 2nd parameter.
However, in this case we should just validate `gaugeAddress` I believe

OnlyJousting (approved):

0xAlunara (approved):

PR #903, opened 2025-05-02
fix: crvUSD filters

DanielSchiavini:
- Move the filters from store to state.
- The filters were not refreshed until reload because they were not included in the useEffect.
- similar to https://github.com/curvefi/curve-frontend/pull/901


0xAlunara (approved):
dejavu lol

DanielSchiavini:
> dejavu lol

indeed so much duplicated code :sob:

0xtutti (approved):

PR #928, opened 2025-05-08
refactor: remove dependency of an available basepool list for pool page advanced tab

OnlyJousting:
The basepool list is fetched on chain, requiring access to an RPC. In the Advanced tab it is only used to verify if a token is a basepool or not. In this PR this dependency is removed and the basepool token validation will only happen when a wallet is connected, allowing a user not yet connected to browse pool parameters freely.

Changes:
- Removed dependency of a fully loaded list of basepool in order to show the advanced tab (for pool parameters) on pool pages

### DIFF
--- a/apps/main/src/dex/components/PagePool/PoolDetails/PoolParameters/index.tsx
+++ b/apps/main/src/dex/components/PagePool/PoolDetails/PoolParameters/index.tsx
@@ -24,11 +24,13 @@ type PoolParametersProps = {
 const PoolParameters = ({ pricesApi, poolData, rChainId }: PoolParametersProps) => {
   const poolAddress = poolData.pool.address
   const snapshotsMapper = useStore((state) => state.pools.snapshotsMapper)
+  const isBasePoolsLoading = useStore((state) => state.pools.basePoolsLoading)
   const basePools = useStore((state) => state.pools.basePools)
   const pricesApiPoolDataMapper = useStore((state) => state.pools.pricesApiPoolDataMapper)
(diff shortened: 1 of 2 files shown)


0xtutti (approved):

apps/main/src/dex/components/PagePool/PoolDetails/PoolParameters/index.tsx
```diff
@@ -24,11 +24,13 @@ type PoolParametersProps = {
 const PoolParameters = ({ pricesApi, poolData, rChainId }: PoolParametersProps) => {
   const poolAddress = poolData.pool.address
   const snapshotsMapper = useStore((state) => state.pools.snapshotsMapper)
+  const isBasePoolsLoading = useStore((state) => state.pools.basePoolsLoading)
   const basePools = useStore((state) => state.pools.basePools)
   const pricesApiPoolDataMapper = useStore((state) => state.pools.pricesApiPoolDataMapper)
   const network = useStore((state) => state.networks.networks[rChainId])
   const snapshotData = snapshotsMapper[poolAddress]
   const pricesData = pricesApiPoolDataMapper[poolAddress]
+  const basePoolList = isBasePoolsLoading ? [] : basePools[rChainId]
```
DanielSchiavini, line 33: why is this needed?

OnlyJousting: because of the structure of the basePools object, basePools[chainId] will not exist if the fetch has been stopped by isNoRPC

PR #961, opened 2025-05-15
fix: auto-reconnect on disconnecting injected wallet

0xAlunara:


OnlyJousting (approved):

0xtutti (approved):

packages/curve-ui-kit/src/features/connect-wallet/lib/useWallet.ts
```diff
@@ -38,14 +38,7 @@ const useWallet = () => {
...
   const { connectAsync } = useConnect()
-  const { disconnectAsync } = useDisconnect()
-
-  const disconnect = useCallback(async () => {
-    // Call disconnectAsync twice to ensure the useAccount hooks properly reset
-    // Bro I don't even know why, but if you call it once the useAccount states won't update...
-    await disconnectAsync()
-    await disconnectAsync()
-  }, [disconnectAsync])
+  const { disconnect } = useDisconnect()
```
DanielSchiavini, line 41: We should be displaying connect/disconnect errors

DanielSchiavini (approved):

packages/curve-ui-kit/src/features/connect-wallet/lib/useWallet.ts
0xAlunara: Connect errors are already shown, not sure where to place disconnect errors that isn't a snackbar / toast.

DanielSchiavini: Toast should be good enough

PR #962, opened 2025-05-16
feat: add balance widget

0xAlunara:
Adds a smol little 'balance' widget. We're going to use it in a few places, most notable user inputs. In short, you use it to show the balance of a specific token (symbol). The balance itself is optional as there's a chance we're waiting for the reply of an async (contract) call.

The 'max' button comes in three varieties:
1. 'off': There is no clickable button anywhere, it's read-only
2. 'balance': You can click the balance (if it's available)
3. 'button': There's a blue 'max' button all the way to the right

image
image
image

### DIFF
--- a/packages/curve-ui-kit/src/shared/ui/Balance.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/Balance.tsx
@@ -1,0 +1,93 @@
+import AccountBalanceWalletOutlinedIcon from '@mui/icons-material/AccountBalanceWalletOutlined'
+import Button from '@mui/material/Button'
+import Stack from '@mui/material/Stack'
+import Typography from '@mui/material/Typography'
+import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
+
+const { Spacing, IconSize } = SizesAndSpaces
+
+const formatNumber = (value: number): string =>
+  value.toLocaleString(undefined, {
+    minimumFractionDigits: 2,
+    maximumFractionDigits: 2,
+  })
+
+/**
+ * Props for the Balance component
+ */
+type Props = {
+  /** The token symbol to display */
+  symbol: string
+  /** Controls how the max value is displayed:
+   * - 'balance': Shows the balance as a clickable link
+   * - 'button': Shows a separate 'Max' button
+   * - 'off': No max functionality is shown
+   */
+  max: 'balance' | 'button' | 'off'
+  /** The token balance amount (optional, in case of loading) */
+  balance?: number
+  /** The USD value of the balance (optional) */
+  notionalValue?: number
+  /** Whether to hide the wallet icon */
+  hideIcon?: boolean
+  /** Callback function when max button/balance is clicked */
+  onMax?: (maxValue: number) => void
+}
+
+export const Balance = ({ symbol, max, balance, notionalValue, hideIcon, onMax }: Props) => {
+  const balanceText = (
+    <>
+      <Typography variant="highlightS" color={balance !== undefined ? 'textPrimary' : 'textTertiary'}>
+        {balance ? formatNumber(balance) : '?'}
+      </Typography>
+
+      <Typography variant="highlightS" color="textPrimary">
+        {symbol}
+      </Typography>
+    </>
+  )
+
+  return (
+    <Stack direction="row" gap={Spacing.xs} alignItems="center">
+      {!hideIcon && <AccountBalanceWalletOutlinedIcon sx={{ width: IconSize.xs, height: IconSize.xs }} />}
+
+      {max === 'balance' && balance !== undefined ? (
+        <Button
+          color="ghost"
+          variant="link"
+          size="extraSmall"
+          onClick={() => onMax?.(balance)}
+          sx={{
+            '&:hover': {
+              textDecoration: 'underline',
+            },
+          }}
+        >
+          {balanceText}
+        </Button>
+      ) : (
+        balanceText
+      )}
+
+      {notionalValue && (
+        <Typography variant="bodySRegular" color="textTertiary">
+          ${formatNumber(notionalValue)}
+        </Typography>
+      )}
+
+      {max === 'button' && balance !== undefined && (
+        <Button
+          color="ghost"
+          variant="link"
+          size="extraSmall"
+          onClick={() => onMax?.(balance)}
+          sx={{
+            minWidth: 'unset',
+          }}
+        >
+          Max
+        </Button>
+      )}
+    </Stack>
+  )
+}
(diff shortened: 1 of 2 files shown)


OnlyJousting (approved):

packages/curve-ui-kit/src/shared/ui/Balance.tsx
```diff
@@ -0,0 +1,96 @@
...
+        {symbol}
+      </Typography>
+    </Stack>
+  )
+
+  return (
+    <Stack direction="row" gap={Spacing.xs} alignItems="center">
+      {!hideIcon && <AccountBalanceWalletOutlinedIcon sx={{ width: IconSize.xs, height: IconSize.xs }} />}
+
+      {max === 'balance' && balance !== undefined ? (
```
DanielSchiavini, line 54: != null works more often 😁

packages/curve-ui-kit/src/shared/ui/Balance.tsx
```diff
@@ -0,0 +1,96 @@
...
+        <Button
+          color="ghost"
+          variant="link"
+          size="extraSmall"
+          onClick={() => onMax?.(balance)}
+          sx={{
+            minWidth: 'unset',
+            '&': { height: '0px !important' },
```
DanielSchiavini, line 62: Both & and important omg

DanielSchiavini (approved):

packages/curve-ui-kit/src/shared/ui/Balance.tsx
0xAlunara: Sadly no amount of & allows us to get rid of it. Added docs on why it's necessary

OnlyJousting (approved):

packages/curve-ui-kit/src/shared/ui/Balance.tsx
```diff
@@ -0,0 +1,106 @@
...
+/** Reusable Max button component with consistent styling */
+const MaxButton = ({ children, underline, onClick }: MaxButtonProps) => (
+  <Button
+    color="ghost"
+    variant="link"
+    size="extraSmall"
+    onClick={onClick}
+    sx={{
+      minWidth: 'unset',
+      // Sadly the !important is still necessary because of the !important in defineMuiButton for the Link variant :(
```
DanielSchiavini, line 30: Could we just remove the height from the link variant when size="extraSmall"?

packages/curve-ui-kit/src/shared/ui/Balance.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+type BalanceTextProps = {
+  symbol: string
+  balance?: number
+}
+
+const BalanceText = ({ symbol, balance }: BalanceTextProps) => (
+  <Stack direction="row" gap={Spacing.xs} alignItems="center">
+    <Typography className="balance" variant="highlightS" color={balance != null ? 'textPrimary' : 'textTertiary'}>
+      {balance != null ? (balance === 0 ? '0' : formatNumber(balance)) : '?'}
```
DanielSchiavini, line 51: Double negatives are hard to read
```suggestion
      {balance == null ? '?' : (balance === 0 ? 0 : formatNumber(balance))}
```

Can we move that to formatNumber directly for consistency?
```suggestion
      {formatNumber(balance)}
```

DanielSchiavini (dismissed):

0xtutti (dismissed):

packages/curve-ui-kit/src/shared/ui/Balance.tsx
0xAlunara: Sadly not, because the `ghost` variant [overrides](https://github.com/curvefi/curve-frontend/blob/main/packages/curve-ui-kit/src/themes/components/button/mui-button.ts#L77) the height set by the button size. At the same time we still need to have the `extraSmall` size because of other props like font size, line height, text transform etc.

DanielSchiavini: That inconsistency in the design would be a bug IMO - but I checked Figma and the variant should be `buttonXs` not "ghost"

That is a typography variant, not a button color though. We could either introduce that as a button color, wrap the children with `Typography` or add onClick to the `Typography` instead of `Button`

packages/curve-ui-kit/src/shared/ui/stories/Balance.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+    },
+  },
+  args: {
+    symbol: 'ETH',
+    balance: 1.234,
+    max: 'off',
+    hideIcon: false,
+    onMax: (maxValue: number) => {
+      // eslint-disable-next-line no-console
+      console.log('Max value:', maxValue)
```
DanielSchiavini, line 41: We can use console.log for testing and console.info for stuff that stays in the codebase
https://github.com/curvefi/curve-frontend/blob/532951c78308ab27606976d095a1437e97a6f10f/packages/eslint-config-custom/index.js#L31

Btw, [storybook has actions](https://storybook.js.org/docs/essentials/actions) for this purpose

```suggestion
      console.info('Max value:', maxValue)
```

packages/curve-ui-kit/src/shared/ui/Balance.tsx
DanielSchiavini: Btw I think the [override](https://github.com/curvefi/curve-frontend/blob/main/packages/curve-ui-kit/src/themes/components/button/mui-button.ts#L77) is just plain wrong - why should the ghost override the size

packages/curve-ui-kit/src/shared/ui/Balance.tsx
```diff
@@ -0,0 +1,106 @@
...
+    onClick={onClick}
+    sx={{
+      minWidth: 'unset',
+      // Sadly the !important is still necessary because of the !important in defineMuiButton for the Link variant :(
+      '&': { height: '0px !important' },
+      ...(underline && {
+        '&:hover .balance': {
+          textDecoration: 'underline',
+        },
+      }),
```
DanielSchiavini, line 40: This shouldn't be a link variant just a ghost color - then your overrides shouldn't be needed
```suggestion
    size="extraSmall"
    onClick={onClick}
```

packages/curve-ui-kit/src/shared/ui/Balance.tsx
DanielSchiavini: https://github.com/curvefi/curve-frontend/pull/962/files/f82dbbf7da69d549bb72d827e9b3445567aba4d1#r2097241632

0xtutti (approved):

packages/curve-ui-kit/src/shared/ui/stories/Balance.stories.tsx
0xAlunara: Oops, forgot to fix this. Will do it in the large input PR

PR #967, opened 2025-05-18
feat: add large token input widget

0xAlunara:
Just creates a simple, isolated component and storybook. Not being used in the app yet anywhere. Please read the docs on the `tokenSelector` property on why I've decided to use a slot instead of implementing one directly into the component.


OnlyJousting (approved):

packages/curve-ui-kit/src/shared/ui/stories/LargeTokenInput.stories.tsx
```diff
@@ -0,0 +1,234 @@
...
+  return (
+    <Select
+      value={value}
+      onChange={(e) => {
+        const selectedToken = e.target.value
+        setValue(selectedToken)
+        const selectedOption = TOKEN_OPTIONS.find((option) => option.name === selectedToken)
+
+        if (selectedOption) {
+          onTokenChange({
```
DanielSchiavini, line 29: wouldn't it be simpler to pass the whole option including the name?

DanielSchiavini (approved):
lgtm, left a couple style remarks

packages/curve-ui-kit/src/shared/ui/stories/LargeTokenInput.stories.tsx
0xAlunara: I on't think so, I only care about the sliders being configurable as args for the storybook. I want the name and symbol to come from the tokenselector itself.

OnlyJousting (approved):

PR #968, opened 2025-05-19
chore: update wagmi and viem (and fix wallet connect)

0xAlunara:
For unknown reason the link provided by the wallet connect modal isn't working anymore. By updating wagmi (and viem) we get the latest wallet connect modal, which has recently undergone a major revision. Perhaps the issue is fixed by merely updating the wagmi package. I've also removed the `@wagmi/connectors` and `@wagmi/core` packages as they are not needed and dependencies of wagmi itself. I think we only pinned it because of Onboard issues

Wallet Connect modal also has gotten a new look:
image
image

### DIFF
--- a/apps/main/src/app/.well-known/walletconnect.txt/route.ts
+++ b/apps/main/src/app/.well-known/walletconnect.txt/route.ts
@@ -2,8 +2,8 @@ import { headers } from 'next/headers'
 
 const WALLET_CONNECT_ACCOUNT = `c3fe8dd8-93df-44af-803f-83798aa1d440`
 
-// for curve-dapp-git-feat-connect-modal-curvefi.vercel.app, other domains can be added the dashboard
-const VERCEL_DOMAIN_VERIFICATION = '0c282d3503877d627a6a75095ebdb60530b900d3b194d046f6d8da3ae19fd48e'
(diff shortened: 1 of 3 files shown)


apps/main/src/app/.well-known/walletconnect.txt/route.ts
```diff
@@ -2,8 +2,8 @@ import { headers } from 'next/headers'
 
 const WALLET_CONNECT_ACCOUNT = `c3fe8dd8-93df-44af-803f-83798aa1d440`
 
-// for curve-dapp-git-feat-connect-modal-curvefi.vercel.app, other domains can be added the dashboard
-const VERCEL_DOMAIN_VERIFICATION = '0c282d3503877d627a6a75095ebdb60530b900d3b194d046f6d8da3ae19fd48e'
+// for curve-dapp-git-fix-wallet-connect-curvefi.vercel.app, other domains can be added the dashboard
```
DanielSchiavini, line 5: did you remove the old domains? usually the token stays the same between preview URLs

DanielSchiavini (approved):

0xtutti (approved):

apps/main/src/app/.well-known/walletconnect.txt/route.ts
0xAlunara: Oh, I did yes 🤔

PR #970, opened 2025-05-20
fix: ghost button height

DanielSchiavini:


DanielSchiavini:
This was a misunderstanding - see https://github.com/curvefi/curve-frontend/pull/962/files#r2097241632

PR #975, opened 2025-05-21
styling: round badges

0xAlunara:


0xtutti (approved):

packages/curve-ui-kit/src/themes/components/chip/mui-chip.ts
```diff
@@ -65,11 +65,14 @@ export const defineMuiChip = (
       borderRadius: Chips.BorderRadius.NonClickable,
       color: TextColors.Primary,
       backgroundColor: 'transparent',
-      ...handleBreakpoints({ paddingInline: Spacing.xs }),
+      ...handleBreakpoints({ paddingInline: Spacing.xxs }),
```
DanielSchiavini, line 68: was this a mistake or was it changed in the design?

DanielSchiavini (approved):

packages/curve-ui-kit/src/themes/components/chip/mui-chip.ts
0xAlunara: By design as per https://www.notion.so/Reduce-padding-in-icon-only-badges-such-that-they-become-nice-circles-1fa599aae06480928001cef5f1944d8c?v=1eb599aae064800d8d42000ca10fca61

OnlyJousting (approved):

PR #976, opened 2025-05-21
styling: action info padding

0xAlunara:
image

image
image
image

### DIFF
--- a/packages/curve-ui-kit/src/shared/ui/ActionInfo.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/ActionInfo.tsx
@@ -117,28 +117,31 @@ const ActionInfo = ({
           />
         )}
 
-        {valueLeft}
+        {/** Additional stack to add some space between left (icon), value and right (icon) */}
+        <Stack direction="row" alignItems="center" gap={Spacing.xxs}>
+          {valueLeft}
 
-        <WithSkeleton loading={!!loading}>
-          <Tooltip
-            title={valueTooltip}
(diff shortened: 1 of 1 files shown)


pilotvietnam (approved):

0xtutti (approved):

packages/curve-ui-kit/src/shared/ui/ActionInfo.tsx
```diff
@@ -100,45 +100,48 @@ const ActionInfo = ({
...
-              {!loading ? value : typeof loading === 'string' ? loading : MOCK_SKELETON}
-            </Typography>
-          </Tooltip>
-        </WithSkeleton>
-
-        {valueRight}
+              }}
+            >
+              <Typography variant={valueSize[size]} color={valueColor ?? 'textPrimary'}>
+                {!loading ? value : typeof loading === 'string' ? loading : MOCK_SKELETON}
```
DanielSchiavini, line 138: Should the MOCK_SKELETON be '00' since '0' is larger than '1'?

DanielSchiavini (approved):

packages/curve-ui-kit/src/shared/ui/ActionInfo.tsx
0xAlunara: We're not using monospace fonts so no digits has the same width. 10 might be too narrow, but 00 might be too wide depending on the real value, which we don't know. Probably a matter of pixels anyway.

PR #978, opened 2025-05-22
feat: calculate balanced values based on form

DanielSchiavini:
- update curvejs
  - https://github.com/curvefi/curve-js/pull/446
  - https://github.com/curvefi/curve-js/pull/453
  - https://github.com/curvefi/curve-js/pull/474
- allow balanced values based on the form values, besides being based on the wallet
  - when the user clicks on the checkbox, use the current calculation based on the wallet balances
    - however, ignore coins they have 0 of it (implemented in curvejs PRs above)
  - when the user changes the input value, ignore the wallet balances and calculate balanced proportions of all the coins on the pool


mimaklas (dismissed):

apps/main/src/dex/components/PagePool/Deposit/components/FieldsDeposit.tsx
```diff
@@ -1,13 +1,57 @@
...
+  )
+  const { reserveRatio: changedRatio, usdPrice: changedUsdPrice } = reserves[tokenAddresses[changedIndex]] ?? {}
+  return zip(oldAmounts, tokenAddresses).map((tuple, index) => {
+    const [amount, tokenAddress] = tuple as [Amount, string]
+    if (changedIndex === index) {
+      return { ...amount, value }
+    }
+    const { usdPrice, reserveRatio } = reserves[tokenAddress] ?? {}
+    if (usdPrice && reserveRatio && changedRatio) {
+      const valueUsd = BigNumber(value).times(changedUsdPrice)
```
0xAlunara, line 47: Is `BigNumber` really necessary or can we use javascript's native `bigint` here?

DanielSchiavini: These are floats

DanielSchiavini: tbh the calculation is probably precise enough with normal JS ints

PR #982, opened 2025-05-23
feat: hide columns without position

DanielSchiavini:
This pull request refactors the logic for determining column visibility in the Llama Markets table to use a new `hasPositions` parameter instead of relying on the `isConnected` state from the `useAccount` hook. This change improves clarity and ensures the visibility logic is tied to the user's actual positions rather than just their connection status.

### DIFF
--- a/apps/main/src/loan/components/PageLlamaMarkets/hooks/useLlamaMarketsColumnVisibility.tsx
+++ b/apps/main/src/loan/components/PageLlamaMarkets/hooks/useLlamaMarketsColumnVisibility.tsx
@@ -13,50 +13,53 @@ export const createLlamaMarketsMobileColumns = (sortBy: LlamaMarketColumnId) =>
 /**
  * Create a map of column visibility for the Llama markets table that can be customized by the user.
  * This is not used on mobile devices (see `createLlamaMarketsMobileColumns` above).
- * @param isConnected Whether there is a wallet connected.
+ * @param hasPositions Whether the user is connected and has positions. Undefined during loading.
  */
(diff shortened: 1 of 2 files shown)


apps/main/src/loan/components/PageLlamaMarkets/hooks/useLlamaMarketsColumnVisibility.tsx
```diff
@@ -13,50 +13,53 @@ export const createLlamaMarketsMobileColumns = (sortBy: LlamaMarketColumnId) =>
...
-        label: t`Lend Details`,
-        columns: [LlamaMarketColumnId.UserEarnings, LlamaMarketColumnId.UserDeposited],
-        active: true,
-        enabled: isConnected,
-      },
-    ],
-  },
-]
+export const createLlamaMarketsColumnOptions = (hasPositions: boolean | undefined) =>
+  hasPositions == null
```
0xAlunara, line 19: Is this ternary really needed? Previously, even when not connected wouldn't it still show the enabled columns (the first 2?)

0xAlunara (approved):

apps/main/src/loan/components/PageLlamaMarkets/hooks/useLlamaMarketsColumnVisibility.tsx
DanielSchiavini: That's right, previously we had the `isConnected` which came from a different data source than the rest of the table (wagmi).
Now, the `hasPositions` boolean is available at the same time as the table data is loaded, since it's part of the table query

PR #983, opened 2025-05-23
fix: hodgepodge

0xAlunara:
Hodgepodge of a handful of review comments and own small commits that I don't want to clog the soft liquidation PR with. Mainly concerns action infos and the large token input.

The reason why I want to use `SystemStyleObject` for `sx` props consistently instead of `SxProps`, is because `SxProps` doesn't work  with if the object contains a function like `color: t => t.design.Colors[200]`. Functionally this change doesn't change anything, it's just a typescript thing and I don't want to re figure out the solution when SxProps isn't working properly. Rather be consistent.

### DIFF
--- a/packages/curve-ui-kit/src/features/connect-wallet/ui/ConnectWalletIndicator.tsx
+++ b/packages/curve-ui-kit/src/features/connect-wallet/ui/ConnectWalletIndicator.tsx
@@ -1,10 +1,16 @@
 import { useAccount } from 'wagmi'
-import type { SxProps, Theme } from '@mui/material'
+import type { SystemStyleObject, Theme } from '@mui/system'
 import { CONNECT_STAGE, isLoading, useConnection, useWallet } from '@ui-kit/features/connect-wallet'
 import { ConnectedWalletLabel } from './ConnectedWalletLabel'
 import { ConnectWalletButton } from './ConnectWalletButton'
 
-export const ConnectWalletIndicator = ({ sx, onConnect }: { sx?: SxProps<Theme>; onConnect?: () => void }) => {
+export const ConnectWalletIndicator = ({
+  sx,
+  onConnect,
+}: {
+  sx?: SystemStyleObject<Theme>
+  onConnect?: () => void
+}) => {
   const { address } = useAccount()
   const { connect, disconnect } = useWallet()
   const { connectState } = useConnection()
--- a/packages/curve-ui-kit/src/features/connect-wallet/ui/ConnectWalletIndicator.tsx
+++ b/packages/curve-ui-kit/src/features/connect-wallet/ui/ConnectWalletIndicator.tsx
@@ -1,16 +1,10 @@
 import { useAccount } from 'wagmi'
-import type { SystemStyleObject, Theme } from '@mui/system'
 import { CONNECT_STAGE, isLoading, useConnection, useWallet } from '@ui-kit/features/connect-wallet'
+import type { SxProps } from '@ui-kit/utils'
 import { ConnectedWalletLabel } from './ConnectedWalletLabel'
 import { ConnectWalletButton } from './ConnectWalletButton'
 
-export const ConnectWalletIndicator = ({
-  sx,
-  onConnect,
-}: {
-  sx?: SystemStyleObject<Theme>
-  onConnect?: () => void
-}) => {
+export const ConnectWalletIndicator = ({ sx, onConnect }: { sx?: SxProps; onConnect?: () => void }) => {
   const { address } = useAccount()
   const { connect, disconnect } = useWallet()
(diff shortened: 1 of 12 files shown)


packages/curve-ui-kit/src/features/connect-wallet/ui/ConnectWalletIndicator.tsx
```diff
@@ -1,10 +1,16 @@
...
 import { CONNECT_STAGE, isLoading, useConnection, useWallet } from '@ui-kit/features/connect-wallet'
 import { ConnectedWalletLabel } from './ConnectedWalletLabel'
 import { ConnectWalletButton } from './ConnectWalletButton'
 
-export const ConnectWalletIndicator = ({ sx, onConnect }: { sx?: SxProps<Theme>; onConnect?: () => void }) => {
+export const ConnectWalletIndicator = ({
+  sx,
+  onConnect,
+}: {
+  sx?: SystemStyleObject<Theme>
```
DanielSchiavini, line 11: Why was it necessary to change `SxProps` to `SystemStyleObject`?

DanielSchiavini (dismissed):
Thanks for teaching me a new word :laughing: I've never heard of hodgepodges

packages/curve-ui-kit/src/features/connect-wallet/ui/ConnectWalletIndicator.tsx
0xAlunara: Not strictly necessary, but wanted to be consistent to avoid the typing issue if copy/pasted or looked at for inspiration on how to add `sx` support to future custom components.

DanielSchiavini: Basically
```
type SxProps<Theme extends object = {}> =
  | SystemStyleObject<Theme>
  | ((theme: Theme) => SystemStyleObject<Theme>)
  | ReadonlyArray<
      boolean | SystemStyleObject<Theme> | ((theme: Theme) => SystemStyleObject<Theme>)
    >;
```
We don't use the 3rd (array) option at all (afaik), but the 2nd (function) can be useful. Basically I used `SystemStyleObject` in cases where I wanted to add some props to `sx`.

0xAlunara: I'm very much confused that `SxProps` extends `SystemStyleObject`, because the following snippet from `TokenIcon.tsx` would generate a type error:
```ts
export interface TokenIconProps extends ImgHTMLAttributes<HTMLImageElement> {
  sx?: SxProps<Theme>
}

export const TokenIcon = ({
  sx,
}: TokenIconProps) => (
    <Box
      sx={(theme) => ({
        border: '1px solid transparent',
        ...sx,
      })}
    />
)
````

We want to use the theme in the Box's `sx` by using the function, and then append custom `sx` styling if the parent provides it. The type error is fixed if we use `sx?: SystemStyleObject<Theme>` instead...

0xAlunara: I've added a new commit that adds `SxProps` and `applySxProps` to `@ui-kit/utils`. If you use only use sx objects, you can keep using normally like:

```ts
const ActionInfo = ({
  sx,
}: ActionInfoProps) => {
  return (
    <Stack direction="row" alignItems="center" gap={Spacing.sm} sx={sx}></Stack>
  )
}```
or

```ts
const ActionInfo = ({
  sx,
}: ActionInfoProps) => {
  return (
    <Stack direction="row" alignItems="center" gap={Spacing.sm} sx={{ color: 'red', ...sx }}></Stack>
  )
}```

But if you use a function with a theem, you must use the `applySxProps` function:

```ts
const ActionInfo = ({
  sx,
}: ActionInfoProps) => {
  return (
    <Stack
      direction="row"
      alignItems="center"
      gap={Spacing.sm}
      sx={(theme) => ({
        color: theme.design.Color.Primary[100],
        ...applySxProps(sx, theme),
      })}
    >
  </Stack>
  )
}```

Maybe there's a better way of handling it, but that's the best I could come up without wasting too much time on it.

OnlyJousting (approved):

packages/curve-ui-kit/src/features/connect-wallet/ui/ConnectWalletIndicator.tsx
DanielSchiavini: Using just the system style object is fine as well, I asked out of curiosity.
The individual props can still be functions with it.

Instead of applySxProps we could also use the 3rd signature (an array of sxProps)

PR #985, opened 2025-05-23
fix: supplied amount for lending markets

DanielSchiavini:
The "Supplied Amount" column in the LlamaLend (BETA) interface needs to be clarified. Currently, it shows the collateral amount of the loan, but users expect it to show the amount of tokens and their value they have deposited into lending vaults. This causes confusion for users trying to track their lent out assets across different markets.

Changes:
- fix supplied amount column by calling API for all lend markets
- omit earnings column as the data isn't available yet
- hide the chart column by default
- migrate to new prices API multichain endpoints (bumping prices API version)
- share user validation between apps


packages/prices-api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@curvefi/prices-api",
-  "version": "1.1.6",
+  "version": "1.1.7",
```
OnlyJousting, line 3: the version number was just bumped to 1.1.7 in https://github.com/curvefi/curve-frontend/pull/1032

OnlyJousting (approved):

PR #986, opened 2025-05-23
Add token alert for zunETH and zunUSD tokens on Arbitrum network

mimaklas:


DanielSchiavini (approved):
We should refactor this to support a list of addresses

pilotvietnam (approved):

PR #999, opened 2025-05-26
feat: add button menu component

0xAlunara:
Button disabled state is not something we've implemented yet, must do that in another PR

image

### DIFF
--- a/packages/curve-ui-kit/src/shared/ui/ButtonMenu.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/ButtonMenu.tsx
@@ -4,6 +4,7 @@ import CircularProgress from '@mui/material/CircularProgress'
 import Menu from '@mui/material/Menu'
 import MenuItem from '@mui/material/MenuItem'
 import Stack from '@mui/material/Stack'
+import useResizeObserver from '@ui-kit/hooks/useResizeObserver'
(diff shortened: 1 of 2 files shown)


DanielSchiavini (approved):

packages/curve-ui-kit/src/shared/ui/ButtonMenu.tsx
```diff
@@ -0,0 +1,156 @@
...
+                  '--options-gap': Spacing.sm,
+                  transform: `translateY(calc(-1 * var(--options-gap))) !important`,
+                }),
+
+                /**
+                 * Sets the menu width to match the anchor element (the button stack).
+                 * By default, MUI Menu auto-sizes to fit content, but we want consistent
+                 * width alignment with the trigger buttons (see Figma).
+                 */
+                width: anchorEl.current?.clientWidth,
```
DanielSchiavini, line 125: Using a ref is not going to rerender at the right time, you might want to `useResizeObserver`

0xAlunara: Ah right, it didn't resize *at all*, but it does now with the observer.

OnlyJousting (approved):

PR #1000, opened 2025-05-26
feat: llamalend breakpoint improvements

DanielSchiavini:
- make adjustments so the table fits in very small screens
- adjustments to chips so they fit around the breakpoints
- autosize columns
- unify graph columns into a single one
- disable inverted theme on hover
- new metric unit (multiplier)

### DIFF
--- a/apps/main/src/loan/components/PageLlamaMarkets/Page.tsx
+++ b/apps/main/src/loan/components/PageLlamaMarkets/Page.tsx
@@ -14,6 +14,7 @@ import {
 import { useLlamaMarkets } from '@/loan/entities/llama-markets'
 import { invalidateAllUserMintMarkets, invalidateMintMarkets, setMintMarkets } from '@/loan/entities/mint-markets'
 import useStore from '@/loan/store/useStore'
+import { useMediaQuery } from '@mui/material'
 import Box from '@mui/material/Box'
 import Skeleton from '@mui/material/Skeleton'
 import { useUserProfileStore } from '@ui-kit/features/user-profile'
@@ -67,12 +68,13 @@ export const LlamaMarketsPage = (props: CrvUsdServerData) => {
   const { address } = useAccount()
   const { data, isError, isLoading } = useLlamaMarkets(address)
   const minLiquidity = useUserProfileStore((s) => s.hideSmallPools) ? SMALL_POOL_TVL : 0
+  const isTiny = useMediaQuery('(max-width:400px)')
 
   const bannerHeight = useStore((state) => state.layout.height.globalAlert)
   const headerHeight = useHeaderHeight(bannerHeight)
   const showSkeleton = !data && (!isError || isLoading) // on initial render isLoading is still false
   return (
-    <Box sx={{ marginBlockEnd: Spacing.xxl, marginInline: Spacing.md }}>
+    <Box sx={{ marginBlockEnd: Spacing.xxl, ...(!isTiny && { marginInline: Spacing.md }) }}>
       {showSkeleton ? (
         <Skeleton variant="rectangular" width={MaxWidth.table} height={ModalHeight.md.height} />
       ) : (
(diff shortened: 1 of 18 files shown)


packages/curve-ui-kit/src/shared/ui/InvertOnHover.tsx
```diff
@@ -37,36 +32,8 @@ type InvertOnHoverProps = {
   disabled?: boolean
 }
 
-const defaultHoverColor = (t: Theme) => t.design.Layer.TypeAction.Hover
-
-export const InvertOnHover = ({
```
0xAlunara, line 42: RIP 💀 

I'm not sure if we really want to remove it like that, given we also use it for the token selector. Maybe consider adding a stupid boolean instead?

image

apps/main/src/loan/components/PageLlamaMarkets/cells/RateCell.tsx
```diff
@@ -21,6 +22,7 @@ export const RateCell = ({ market, type }: { market: LlamaMarket; type: RateType
   const { rewards, type: marketType } = market
   const rewardsAction = getRewardsAction(marketType, type)
   const poolRewards = rewards.filter(({ action }) => action == rewardsAction)
+  const isMobile = useMediaQuery((t) => t.breakpoints.down('tablet'))
```
0xAlunara, line 25: Not too important, but having a hook in `curve-ui-kit/utils` would be great. Something like: `const { isMobile } = useBreakpoints()`. It's a tiny bit easier than having to remember `t.breakpoints.down` or `.up`.

apps/main/src/loan/components/PageLlamaMarkets/Page.tsx
```diff
@@ -67,12 +68,13 @@ export const LlamaMarketsPage = (props: CrvUsdServerData) => {
   const { address } = useAccount()
   const { data, isError, isLoading } = useLlamaMarkets(address)
   const minLiquidity = useUserProfileStore((s) => s.hideSmallPools) ? SMALL_POOL_TVL : 0
+  const isTiny = useMediaQuery('(max-width:400px)')
```
0xAlunara, line 71: This would be great to have in in aforementioned hook as well 😄

packages/curve-ui-kit/src/shared/ui/InvertOnHover.tsx
DanielSchiavini: Tutti asked me to remove there too for consistency

apps/main/src/loan/components/PageLlamaMarkets/cells/RateCell.tsx
DanielSchiavini: yeah I agree it's about time for that

apps/main/src/loan/components/PageLlamaMarkets/Page.tsx
DanielSchiavini: Hmm not sure, this is very specific to this component :thinking:

apps/main/src/loan/components/PageLlamaMarkets/LendingMarketsTable.tsx
```diff
@@ -89,9 +89,9 @@ export const LendingMarketsTable = ({
       <DataTable
         table={table}
         headerHeight={headerHeight}
-        rowSx={{ height: { ...Sizing['3xl'], mobile: 77 } }} // the 3xl is too small in mobile (64px)
         emptyText={isError ? t`Could not load markets` : t`No markets found`}
         expandedPanel={LlamaMarketExpandedPanel}
+        shouldStickFirstColumn={useIsTablet() && !!hasPositions}
```
DanielSchiavini, line 94: @0xtutti we can adjust this as wished or depending on the viewport

OnlyJousting (approved):

0xtutti (approved):

PR #1002, opened 2025-05-26
fix: large input decimals

0xAlunara:
When dragging the percentage slider (or just editing the value) you can get some annoying rounding. This PR adds a property that allows rounding the balance to a number of decimals (default 4)

image


packages/curve-ui-kit/src/shared/ui/LargeTokenInput.tsx
```diff
@@ -209,7 +226,14 @@ export const LargeTokenInput = ({ tokenSelector, maxBalance, message, isError =
    */
   useEffect(() => {
```
DanielSchiavini, line 227: @0xAlunara to answer your question I think the proper fix would be to lift the state out of the component and change the balance as soon as we get the balance instead of using an effect.

A fix without too much refactoring would be to add `useRef` with the balance or the balance input

DanielSchiavini (approved):

PR #1011, opened 2025-05-28
Add token alert for USDM

mimaklas:


0xtutti (changes requested):
Needs update in the wording. Should be "since May 12, 2025", not "from..."

pilotvietnam (approved):

0xtutti (approved):

PR #1027, opened 2025-06-02
fix: hide broken gauge manager

OnlyJousting:
Changes:
- Hides gauge manager from pool page

### DIFF
--- a/apps/main/src/dex/components/PagePool/index.tsx
+++ b/apps/main/src/dex/components/PagePool/index.tsx
@@ -1,7 +1,7 @@
 import { useRouter } from 'next/navigation'
 import { useCallback, useEffect, useMemo, useState } from 'react'
 import styled from 'styled-components'
-import { type Address, isAddressEqual } from 'viem'
+// import { type Address, isAddressEqual } from 'viem'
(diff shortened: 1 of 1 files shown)


apps/main/src/dex/components/PagePool/index.tsx
```diff
@@ -87,15 +87,15 @@ const Transfer = (pageTransferProps: PageTransferProps) => {
...
-  const basePoolsLoading = useStore((state) => state.pools.basePoolsLoading)
   const { initCampaignRewards, initiated } = useStore((state) => state.campaigns)
 
   const storeMaxSlippage = useUserProfileStore((state) => state.maxSlippage[chainIdPoolId])
 
-  const { data: gaugeManager, isPending: isPendingGaugeManager } = useGaugeManager({
-    chainId: rChainId,
-    poolId: poolData?.pool.id!,
-  })
+  // Hidden until gauge manager is fixed
```
DanielSchiavini, line 94: I think it's better to delete the code. Or otherwise just create a `const isEnabled = false` and pass it as 2nd argument to the query to disable it

DanielSchiavini (approved):

PR #1048, opened 2025-06-11
refactor: RPC fallbacks

DanielSchiavini:
- centralize RPC configuration so it's easy to adapt
- pass the RPC fallbacks for all injected providers
- wallet connect unfortunately only uses [the first one](https://github.com/wevm/wagmi/blob/wagmi%402.15.6/packages/connectors/src/walletConnect.ts#L258)

### DIFF
--- a/packages/curve-ui-kit/src/features/connect-wallet/lib/wagmi/wagmi-config.ts
+++ b/packages/curve-ui-kit/src/features/connect-wallet/lib/wagmi/wagmi-config.ts
@@ -29,33 +29,30 @@ const DECIMALS: Record<number, number> = {
  */
 export const createWagmiConfig = memoize(
   <ChainId extends number, NetworkConfig extends BaseConfig>(networks: Record<ChainId, NetworkConfig>) => {
-    const chains = Object.fromEntries(wagmiChains.map((chain) => [chain.id, chain]))
+    const wagmi = Object.fromEntries(wagmiChains.map((chain) => [chain.id, chain]))
     const networkEntries = Object.entries(networks).map(([id, config]) => [+id, config]) as [ChainId, NetworkConfig][]
 
     /**
      * Gets a list of the RPC URLs for a given chain.
      * First the hardcoded RPC URLs, then the CurveJS URL and then all default wagmi URLs
      */
-    const getRpcUrls = (id: ChainId, config: NetworkConfig, wagmiChain?: Chain) =>
-      uniq([...(RPC[id] ?? []), config.rpcUrl, ...(wagmiChain?.rpcUrls.default.http ?? [])])
+    const getRpcUrls = (id: ChainId) =>
+      uniq([...(RPC[id] ?? []), networks[id].rpcUrl, ...(wagmi[id]?.rpcUrls.default.http ?? [])])
 
     return createConfig({
       chains: networkEntries.map(
         ([id, config]): Chain =>
           // use the backend data to configure new chains, but use wagmi contract addresses and useful properties/RPCs
           defineChain({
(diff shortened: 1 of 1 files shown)


mimaklas (approved):

packages/curve-ui-kit/src/features/connect-wallet/lib/wagmi/wagmi-config.ts
```diff
@@ -63,14 +67,19 @@ export const createWagmiConfig = memoize(
...
-           * Note: WalletConnect ignores this http endpoint and uses chain.rpcUrls.default.http[0]
-           * regardless of the configuration here.
+           * Note: WalletConnect ignores this transport configuration and uses chain.rpcUrls.default.http
+           * in order, but having multiple transports helps with injected wallet resilience.
            */
-          fallback([unstable_connector(injected), http(undefined, { batch: { batchSize: 3 } })]),
+          fallback([
+            unstable_connector(injected),
+            http(undefined, { batch: { batchSize: 3 } }),
+            ...(RPC[id] ?? []).map((url) => http(url, { batch: { batchSize: 3 } })),
```
0xAlunara, line 81:
```suggestion
...([undefined, ...RPC[id] ?? []]).map((url) => http(url, { batch: { batchSize: 3 } })),
```
freestyling from github

packages/curve-ui-kit/src/utils/network.ts
```diff
@@ -22,6 +22,15 @@ export enum Chain {
   Mantle = 5000,
   Sonic = 146,
   Hyperliquid = 999,
+  Xdc = 50,
+  Tac = 2390,
+  Corn = 21000000,
+  Ink = 57073,
+  Taiko = 167000,
+  Plume = 161221135,
```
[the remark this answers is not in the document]
DanielSchiavini: Damn thanks for the alertness

packages/curve-ui-kit/src/features/connect-wallet/lib/wagmi/wagmi-config.ts
DanielSchiavini: undefined is weird, I've added all the different URLs instead and not as the first one

packages/curve-ui-kit/src/features/connect-wallet/lib/wagmi/wagmi-config.ts
```diff
@@ -31,31 +31,37 @@ export const createWagmiConfig = memoize(
   <ChainId extends number, NetworkConfig extends BaseConfig>(networks: Record<ChainId, NetworkConfig>) => {
     const chains = Object.fromEntries(wagmiChains.map((chain) => [chain.id, chain]))
     const networkEntries = Object.entries(networks).map(([id, config]) => [+id, config]) as [ChainId, NetworkConfig][]
+
+    /**
+     * Gets a list of the RPC URLs for a given chain.
+     * First the hardcoded RPC URLs, then the CurveJS URL and then all default wagmi URLs
+     */
+    const getRpcUrls = (id: ChainId, config: NetworkConfig, wagmiChain?: Chain) =>
+      uniq([...(RPC[id] ?? []), config.rpcUrl, ...(wagmiChain?.rpcUrls.default.http ?? [])])
```
0xAlunara: Just be aware that not using `undefined` means chains without configured RPCs work. `undefined` means to use wagmi's RPC, which I believe for Ethereum now is merkle.io.

Maybe add a fallback like `RPC[id] ?? [undefined]`, so `undefined` is used in case of no RPCs.

PR #1050, opened 2025-06-11
styling: fix select height, chevron icon, padding; like figma

0xAlunara:
image

### DIFF
--- a/packages/curve-ui-kit/src/themes/components/mui-select.ts
+++ b/packages/curve-ui-kit/src/themes/components/mui-select.ts
@@ -33,9 +33,16 @@ export const defineMuiSelect = (design: DesignSystem, typography: TypographyOpti
         paddingBlock: Spacing.xs,
         paddingInlineStart: Spacing.sm,
         height: Sizing.xl,
-        // The overflow hiding doesn't take into account the expansion chevron icon, so we have to deduct that from the width (100% by default).
+        /**
+         * The overflow hiding doesn't take into account the expansion chevron icon, so we need to deduct
+         * the icon width from the available text space (100% by default).
+         * Initially attempted to reduce the `width` property, but this also reduces the clickable area
+         * for opening the options menu. Instead, maintain 100% width but mask the right side using
+         * the responsive icon size to prevent text overlap with the chevron.
+         * Implementation uses a CSS mask with a linear gradient from opaque to transparent.
+         */
         '--icon-size': ICON_SIZE,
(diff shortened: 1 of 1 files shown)


packages/curve-ui-kit/src/themes/components/mui-select.ts
```diff
@@ -15,7 +25,41 @@ export const defineMuiSelect = (design: DesignSystem, typography: TypographyOpti
...
+    icon: {
+      // Not sure if there's a better way, and I don't want to use our custom TransitionFunction as it doesn't match
+      // the animation the MUI select option list uses when expanding.
+      // Use hardcoded transition values instead of MUI's theme function, which isn't accessible here.
+      transition: 'transform 225ms cubic-bezier(0.4, 0, 0.2, 1)',
+      ...handleBreakpoints({
+        width: ICON_SIZE,
+        height: ICON_SIZE,
+        // MUI default of `calc(50% - .5em)` doesn't really vertically center correctly given our responsive icon size
+        '--icon-size': ICON_SIZE,
```
DanielSchiavini, line 57: Do we really need this variable?

packages/curve-ui-kit/src/themes/components/mui-select.ts
```diff
@@ -15,7 +25,41 @@ export const defineMuiSelect = (design: DesignSystem, typography: TypographyOpti
...
+    },
+    icon: {
+      // Not sure if there's a better way, and I don't want to use our custom TransitionFunction as it doesn't match
+      // the animation the MUI select option list uses when expanding.
+      // Use hardcoded transition values instead of MUI's theme function, which isn't accessible here.
+      transition: 'transform 225ms cubic-bezier(0.4, 0, 0.2, 1)',
+      ...handleBreakpoints({
+        width: ICON_SIZE,
+        height: ICON_SIZE,
+        // MUI default of `calc(50% - .5em)` doesn't really vertically center correctly given our responsive icon size
```
DanielSchiavini, line 56: Is this comment in the right line?

packages/curve-ui-kit/src/themes/components/mui-select.ts
0xAlunara: Yeah as you can't use a responsive value like `IconSize.lg` directly in a `calc` like in the `translate` below. Need to put it in a var first.

DanielSchiavini (approved):

PR #1052, opened 2025-06-11
feat: borderless slider

0xAlunara:
MUI sliders don't accept custom props or variants so I'm kinda stuck adding support for a non-typed "borderless" class name. Not the way I like it but I've exhausted all options I can think of.

image


0xtutti (dismissed):

packages/curve-ui-kit/src/shared/ui/TradingSlider.tsx
```diff
@@ -38,6 +38,7 @@ export const TradingSlider = ({
     }}
   >
     <Slider
+      className="borderless"
```
DanielSchiavini, line 41: Maybe we can introduce some extra props, but className is fine. Can we use a constant though, so it's easy to find?

0xtutti (approved):

PR #1058, opened 2025-06-13
fix: manage soft liq review feedback

0xAlunara:
- Health action info first value color should be tertiary
- Add gap to estimated tx cost action info
- Move token selector to the right with a new label on top of the entire row
- remove "What are soft-liquidations" link, "the positions will be fully liquidated" points to https://resources.curve.finance/crvusd/loan-concepts
- Remove clamping of input from max balance, only negative vales should be clamped to 0.
- Reset balance to zero on token change (had to fix resetting to 0 when the selected token did not change from the old one).

The balance input field itself is still a bit buggy wrt decimals, but there's a different ticket (and thus PR) for that.

### DIFF
--- a/packages/curve-ui-kit/src/themes/components/tabs/mui-tabs.ts
+++ b/packages/curve-ui-kit/src/themes/components/tabs/mui-tabs.ts
@@ -107,7 +107,7 @@ const tabSizesNonContained = {
  *                             If false, applies to all tabs of the variant unconditionally.
  * @param variants - Tab variant class names
  */
-const inactiveTabSelector = (hideInactiveBorders: boolean, ...variants: string[]) =>
+const inactiveTabSelector = ({ hideInactiveBorders }: { hideInactiveBorders: boolean }, ...variants: string[]) =>
   variants
     .map(
       (variant) =>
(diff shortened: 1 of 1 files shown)


0xtutti (dismissed):

packages/curve-ui-kit/src/features/manage-soft-liquidation/ui/AlertCollateralAtRisk.tsx
```diff
@@ -24,7 +22,13 @@ export const AlertCollateralAtRisk = () => (
       </Typography>
 
       <Stack display="inline-flex">
-        <Button variant="inline" color="ghost" size="extraSmall">
+        <Button
+          href="https://resources.curve.finance/crvusd/loan-concepts"
```
DanielSchiavini, line 26: Don't we need a `component={Link}` here to make this into an `<a` in the HTML?

packages/curve-ui-kit/src/features/select-token/types.ts
```diff
@@ -7,3 +7,7 @@ export type TokenOption = {
   label?: string
   volume?: number
 }
+
+export function equals(a?: TokenOption, b?: TokenOption) {
```
DanielSchiavini, line 11: since this is exported, a more descriptive name might be better

packages/curve-ui-kit/src/features/select-token/types.ts
```diff
@@ -7,3 +7,7 @@ export type TokenOption = {
   label?: string
   volume?: number
 }
+
+export function tokenOptionEquals(a?: TokenOption, b?: TokenOption) {
```
DanielSchiavini, line 11: no arrow function? :grin: not important

DanielSchiavini (approved):

PR #1060, opened 2025-06-13
fix: localization and browser issues with regards to number inputs

0xAlunara:
Depends on #1058

This was a very tricky PR. Turns out that HTML input fields with type="number" are ill suitable for the fine control we desire. We don't want browsers to severely limit the input field and force default behavior on the user, especially if it's related to the user's localization.

The only way out of this mess is to introduce a new `NumericTextField` component that wraps around MUI's `TextField` component, with `type="text"`. We have to introduce our own sanitizing and clamping to make sure the user can only enter numbers, can enter both a comma and decimal (and only one!), and allows for only a minus sign at the start.

The tricky part is also to support partial user input, like "4,". When the user enters a comma, you don't want to immediately start sanitizing and clamping, only when the user 'submits' it by blur (when input loses focus, basically).

You also want to support an 'empty' input so we can show the placeholder. This is different from when a user manually enters a "0".

I've also taken extra care to make sure that there's no duplicate events. So if you click a text field with value "5", click outside, click the text field again and click outside again, I don't want two consecutive `onBlur` events for the same value when it's unchanged. You can argue that such change detection should be the parent's responsibility, but I foresee people forgetting about it, causing performance issues later on. Better to fix it at the root.

You can test the numeric text field in:

1. Its own dedicated storybook
2. Its usage in the `TradingSlider` component.
3. Its usage in the LargeTokenInput component for both the balance text field, and the trading slider percentage input.

I've spent like half a day trying to get this to work, it's super finnicky and very easy to break. So please have mercy upon my soul when suggesting changes 💀 

image

### DIFF
--- a/tests/cypress/support/commands.ts
+++ b/tests/cypress/support/commands.ts
@@ -1,4 +1,3 @@
-import { mount } from 'cypress/react'
 import { approveSpending } from './approval'
 import { inputMaxCollateral } from './collateral'
 import { inputMaxBorrow } from './debt'
@@ -33,6 +32,3 @@ Cypress.Commands.add('inputMaxCollateral', inputMaxCollateral)
 Cypress.Commands.add('inputMaxBorrow', inputMaxBorrow)
 Cypress.Commands.add('approveSpending', approveSpending)
 Cypress.Commands.add('createSofLiquidationLoan', createSofLiquidationLoan)
-
-// custom mounting of react components
-Cypress.Commands.add('mount', (component, options) => mount(component, options))


packages/curve-ui-kit/src/shared/ui/LargeTokenInput.tsx
```diff
@@ -37,16 +34,14 @@ const HelperMessage = ({ message, isError }: HelperMessageProps) => (
 )
 
 type BalanceTextFieldProps = {
-  balance: number
+  balance: number | undefined
   maxBalance?: number
-  label?: string
   isError: boolean
-  onChange: (balance: number) => void
+  onBlur: (balance: number | undefined) => void
```
DanielSchiavini, line 40: is onBlur the right name here? Maybe onChange is more correct, or onCommit as you used in the other component

packages/curve-ui-kit/src/shared/ui/NumericTextField.tsx
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * Validates and normalizes numeric input by replacing commas with dots
+ * and ensuring only valid numeric characters are allowed.
+ * Prevents multiple decimal points and ensures minus sign is only at the beginning.
+ *
+ * @param value - The new input value to validate
+ * @param current - The current input value to fall back to if validation fails
+ * @returns The validated and normalized input value, or the current value if invalid
+ */
+const sanitize = (value: string, current: string): string => {
```
DanielSchiavini, line 14: as you mentioned yourself, this kind of code is fragile and dependent on browsers and locales.
In general I'd say it's better to avoid such customization and rely on browser implementation

packages/curve-ui-kit/src/shared/ui/TradingSlider.tsx
```diff
@@ -1,33 +1,27 @@
...
 const { Spacing, FontSize, FontWeight, Sizing } = SizesAndSpaces
 
 type Props = {
   /** Current percentage value (0-100) */
-  percentage: number
+  percentage: number | undefined
   /** Callback when percentage changes on the slider */
-  onPercentageChange?: (percentage: number) => void
+  onChange?: (percentage: number | undefined) => void
   /** Callback when percentage changes by rleasing the slider or entering a number */
```
DanielSchiavini, line 16:
```suggestion
  /** Callback when percentage changes by releasing the slider or entering a number */
```

packages/curve-ui-kit/src/features/manage-soft-liquidation/ui/tabs/Withdraw.tsx
```diff
@@ -106,6 +106,8 @@ export const Withdraw = ({
           maxBalance={{ ...selectedDebtToken, showSlider: false }}
           message={t`Recover collateral by repaying debt.`}
           onBalance={(balance) => {
```
DanielSchiavini, line 108: Is it feasible to add (cypress) tests for the use cases you mentioned?
I can definitely see them being forgotten and breaking.

packages/curve-ui-kit/src/shared/ui/NumericTextField.tsx
0xAlunara: Sadly I've never seen default browser behavior behave as desired for number inputs, for any job I've ever worked. I hoped maybe MUI would take care of it, but it doesn't look like they've done anything wrt the behavior, only the styling 😞

packages/curve-ui-kit/src/shared/ui/LargeTokenInput.tsx
0xAlunara: I actually considered renaming the `onCommit` to `onBlur` as well 🙈 
But I'll change the blur to commit for consistency

tests/cypress/component/numeric-text-field.cy.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  return (
+    <>
+      <NumericTextField value={value} onChange={setValue} max={max} />
+      <div data-testid="state-value">{value}</div>
+    </>
+  )
+}
+
+describe('NumericTextField', () => {
```
DanielSchiavini, line 15: Thanks for setting this up :pray: :fire:

DanielSchiavini (approved):

PR #1062, opened 2025-06-14
refactor: loan user details hooks

0xAlunara:
There's a discrepancy on how user loan details (and the loan status which I need later on) are loaded. Therefore, I introduced similar hooks for both lend & loan apps that aim to streamline and make fetching user loan details more simple. No more direct store usage, but just use a hook to fetch user loan data instead without caring about the backend.

### DIFF
--- a/apps/main/src/lend/hooks/useUserLoanDetails.ts
+++ b/apps/main/src/lend/hooks/useUserLoanDetails.ts
@@ -1,0 +1,28 @@
+import useStore from '@/lend/store/useStore'
+import type { HealthColorKey, UserLoanDetails } from '@/lend/types/lend.types'
+
+type Details = UserLoanDetails['details'] & { error: string }
+
+/**
+ * Retrieves user loan details for a specific user active key.
+ * @param userActiveKey - The unique identifier for the user's active loan
+ * @returns User loan details object with error property merged for easier destructuring
+ */
+export function useUserLoanDetails(userActiveKey: string): Details {
+  const loanDetails = useStore((state) => state.user.loansDetailsMapper[userActiveKey])
+
+  return !loanDetails || loanDetails.details == null
+    ? ({} as Details)
+    : { ...loanDetails.details, error: loanDetails.error }
+}
+
+/**
+ * Gets the health status color key for a user's loan.
+ * @param userActiveKey - The unique identifier for the user's active loan
+ * @returns Health color key indicating loan status, or empty string if no status available
+ */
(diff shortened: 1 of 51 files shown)


apps/main/src/lend/hooks/useUserLoanDetails.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Retrieves user loan details for a specific user active key.
+ * @param userActiveKey - The unique identifier for the user's active loan
+ * @returns User loan details object with error property merged for easier destructuring
+ */
+export function useUserLoanDetails(userActiveKey: string): Details {
+  const loanDetails = useStore((state) => state.user.loansDetailsMapper[userActiveKey])
+
+  return !loanDetails || loanDetails.details == null
+    ? ({} as Details)
```
DanielSchiavini, line 15: this is a bad cast, if the details can be empty we should change the type to have optional fields

apps/main/src/loan/hooks/useUserLoanDetails.ts
```diff
@@ -0,0 +1,22 @@
+import useStore from '@/loan/store/useStore'
+import type { HealthColorKey } from '@/loan/types/loan.types'
+
+/**
+ * Retrieves user loan details for a specific llamma market
+ * @param llammaId - The llamma market identifier
+ * @returns User loan details object or empty object if not found
+ */
+export function useUserLoanDetails(llammaId: string) {
```
DanielSchiavini, line 9: these 2 functions could be one-line arrows

DanielSchiavini (dismissed):

apps/main/src/lend/hooks/useUserLoanDetails.ts
```diff
@@ -0,0 +1,28 @@
+import useStore from '@/lend/store/useStore'
+import type { HealthColorKey, UserLoanDetails } from '@/lend/types/lend.types'
+
+type Details = UserLoanDetails['details'] & { error: string }
+
+/**
+ * Retrieves user loan details for a specific user active key.
+ * @param userActiveKey - The unique identifier for the user's active loan
```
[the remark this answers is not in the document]
0xAlunara: I've actually tried to place that logic inside the hook, so the component using the hook wouldn't have to worry about those store keys. But it turns out we use it in other places as well, which meant you still had to use keys all over the place, hence I had to remove it again leaving just a simple 'useStore()` call with a store key as param 😢

DanielSchiavini (approved):

PR #1069, opened 2025-06-18
fix: disable deleverage when in soft liquidation

0xAlunara:
I couldn't find a user in soft liq, so how I tested it instead:

Use mich's address which has a had liquidation in the wsteth crvusd market, then locally change:
```const isSoftLiquidation = useUserLoanStatus(llammaId) === 'soft_liquidation'```
into
```const isSoftLiquidation = useUserLoanStatus(llammaId) === 'hard_liquidation'```

Do not merge yet: I'm awaiting a reply on my question whether deleverage should also be disabled when hard liquidated

EDIT: Turns out a **full** deleverage should still be possible while in SL, while a **partial** deleverage is not. In that case, hiding the tab would not be a good idea as we'd take away functionality.


0xAlunara:
After consideration it's chosen to not touch the legacy UI and rather properly take care of the various scenarios in the redesign

PR #1077, opened 2025-06-19
feat: multiple metric notionals

0xAlunara:
For the manage soft liquidation card we need to support multiple notionals underneath a metric:
image

By default, notionals are joined with ` + `. For the moment there's no support for further customization like custom joining and/or wrapping. That's a fight only worth fighting when it's an actual problem.

In order to support this, I've also flattened the options in both the main value and notionals. There's now a `ValueOptions` type, and
```typescript
/** A unit can be a currency symbol or percentage, prefix or suffix */
unit?: Unit | undefined
/** The number of decimals the value should contain */
decimals?: number
/** If the value should be abbreviated to 1.23k or 3.45m */
abbreviate?: boolean
/** Optional formatter for metric value */
formatter?: (value: number) => string
```
has been consolidated into a single property:
`valueOptions: ValueOptions`

Likewise,
```typescript
/** Notional values give extra context to the metric, like underlying value */
notional?: number
notionalFormatter?: (value: number) => string
notionalAbbreviate?: boolean
notionalDecimals?: number
notionalUnit?: Unit
```
has been consolidated into `notional?: number | Notional | Notional[]`

**PS:**
You can argue `abbreviate` shouldn't be part of the unit type. Maybe, but I want to keep this PR small and focused so that's a discussion and refactor for another day.

### DIFF
--- a/packages/curve-ui-kit/src/shared/ui/Metric.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/Metric.tsx
@@ -46,6 +46,7 @@ export const SIZES = Object.keys(MetricSize) as (keyof typeof MetricSize)[]
 export type UnitOptions = {
   symbol: string
   position: 'prefix' | 'suffix'
+  /** If the value should be abbreviated to 1.23k or 3.45m */
   abbreviate: boolean
 }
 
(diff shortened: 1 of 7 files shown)


OnlyJousting (approved):

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -286,12 +289,29 @@ export const Metric = ({
...
-      {notional !== undefined && (
+      {notionals.length > 0 && (
         <Typography variant="highlightXsNotional" color="textTertiary">
-          {notionalUnit?.position === 'prefix' && notionalUnit.symbol}
-          {notionalFormatter(notionalAbbreviate ? abbreviateNumber(notional) : notional)}
-          {notionalAbbreviate && scaleSuffix(notional)}
-          {notionalUnit?.position === 'suffix' && notionalUnit.symbol}
+          {notionals
+            .map((notional) => {
+              const {
```
DanielSchiavini, line 296: Maybe it's worth to useMemo here

DanielSchiavini (approved):

PR #1080, opened 2025-06-21
refactor: use useMemo in Metric and improve readability

0xAlunara:
ok ok usememo might not be super needed but still improves readability a bit in the TSX

### DIFF
--- a/packages/curve-ui-kit/src/shared/ui/Metric.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/Metric.tsx
@@ -243,13 +243,33 @@ export const Metric = ({
   const { decimals = 1, formatter = (value: number) => formatValue(value, decimals) } = valueOptions
   const unit = typeof valueOptions.unit === 'string' ? UNIT_MAP[valueOptions.unit] : valueOptions.unit
 
-  // Converge the various notional types to an array of Notional
-  const notionals =
-    typeof notional === 'number'
-      ? [{ value: notional }]
-      : notional && !Array.isArray(notional)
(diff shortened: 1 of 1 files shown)


packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -243,13 +243,33 @@ export const Metric = ({
...
 
-  // Converge the various notional types to an array of Notional
-  const notionals =
-    typeof notional === 'number'
-      ? [{ value: notional }]
-      : notional && !Array.isArray(notional)
-        ? [notional]
-        : (notional ?? [])
+  const notionals = useMemo(() => {
+    // Converge the various notional types to an array of Notional
```
DanielSchiavini, line 247: this comment wouldn't be necessary if we used a helper function

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -243,13 +243,33 @@ export const Metric = ({
...
+    const ns =
+      typeof notional === 'number'
+        ? [{ value: notional }]
+        : notional && !Array.isArray(notional)
+          ? [notional]
+          : (notional ?? [])
+
+    return ns
+      .map((notional) => {
+        const { value, decimals = 1, formatter = (value: number) => formatNotionalValue(value, decimals) } = notional
```
DanielSchiavini, line 257: this should also be extracted to the global scope imo

OnlyJousting (approved):

PR #1081, opened 2025-06-23
External rewards: Add Yieldoor campaign

michaeldim:


mimaklas:
hey, external rewards is for external points campaigns, or additional external tokens campaign (token rewards that are claimable from an external app, e.g. USUAL token campaign). what kind of campaign the "ycvxcrv rewards" is? can you provide some more informations pls? (e.g. is there an exteral app where these "ycvxcrv rewards" is claimable?

mimaklas:
> Ah ok sorry. They are permissionless rewards sent to the gauge of the vault. I just thought I needed to set this up for them to become visible in the UI as they are not currently shown.

i see you added permisionless rewards on gauge right?
well once the API has pricing for ycvxCRV token, then APR will show up on UI
External rewards campaign is for something else.

PR #1089, opened 2025-06-24
feat: loading page

DanielSchiavini:
- add a [loading file](https://nextjs.org/docs/app/api-reference/file-conventions/loading) that is used for transitions


DanielSchiavini:
This isn't loaded between page navigations. We might wanna use [`useLinkStatus`](https://nextjs.org/docs/app/api-reference/functions/use-link-status).

PR #1100, opened 2025-06-25
feat: add support for custom value font color in Metric

OnlyJousting:
Changes:
- Adds support for setting a custom font color for value in Metric component

### DIFF
--- a/packages/curve-ui-kit/src/shared/ui/Metric.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/Metric.tsx
@@ -157,12 +157,10 @@ type MetricValueProps = Pick<Props, 'value' | 'valueOptions'> & {
 }
 
 const MetricValue = ({ value, valueOptions, change, size, copyValue, tooltip }: MetricValueProps) => {
-  const numberValue: number | null = useMemo(() => {
-    if (typeof value === 'number' && isFinite(value)) {
-      return value
-    }
-    return null
-  }, [value])
+  const numberValue: number | null = useMemo(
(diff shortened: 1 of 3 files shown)


0xtutti (dismissed):

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -83,6 +83,7 @@ type ValueOptions = {
   unit?: Unit | undefined
   /** The number of decimals the value should contain */
   decimals?: number
+  fontColor?: TypographyProps['color']
```
0xAlunara, line 86: I wouldn't place it here actually, because these are common types between both the metric value itself and the notionals. And notional don't support a font color. Would rather make it a prop of `MetricValueProps` directly.

Also personally I'd just call it `color` myself.

And I just noticed `/** Options for any being used, whether` is missing the word 'value' behind 'any', but that was already missing originally.

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -212,7 +214,7 @@ const MetricValue = ({
             </Typography>
           )}
 
-          <Typography variant={fontVariant} color="textPrimary">
+          <Typography variant={fontVariant} color={fontColor}>
```
0xAlunara, line 217: I'd put the default value here with `fontColor ?? 'textPrimary`, as technically the parameter is still optional and can be `undefined`

0xAlunara (approved):

0xAlunara (approved):

0xAlunara (approved):

PR #1103, opened 2025-06-26
fix: double initialization

DanielSchiavini:


packages/curve-ui-kit/src/features/connect-wallet/lib/ConnectionContext.tsx
```diff
@@ -39,7 +39,7 @@ function useWagmiWallet() {
   const { isReconnecting, isConnected } = useAccount()
   return {
     // wagmi flips to isReconnecting when switching pages, but leaves isConnected as true
-    isReconnecting: isReconnecting && !isConnected,
+    isReconnecting: isReconnecting && !(isConnected && !!address),
```
0xAlunara, line 42: Triple negation! I trust you there's a valid reason for this.

0xAlunara (approved):

packages/curve-ui-kit/src/features/connect-wallet/lib/ConnectionContext.tsx
DanielSchiavini: you can trust me but I won't be here forever (and wouldn't remember forever even if I was). Documented in the following commit!

mimaklas (approved):

PR #1106, opened 2025-06-27
feat: replace tooltips with mui

DanielSchiavini:
- use the mui tooltip instead of react-aria


0xtutti (approved):

0xAlunara (approved):

0xAlunara (changes requested):
Getting an application error when hovering over "-0.02%" borrow rate of the 3pool market
image

0xAlunara:
Tooltips with clickable content aren't clickable anymore, like the Ethena points tooltip on the Pool page
image

mimaklas (dismissed):

0xAlunara (changes requested):
Quickly going through a few tooltips which should be clickable but aren't
image
image

0xAlunara (approved):

mimaklas (approved):

PR #1117, opened 2025-07-01
feat: reward badges overlapping

DanielSchiavini:
Screenshot from 2025-07-01 12-43-30


packages/curve-ui-kit/src/themes/components/chip/mui-chip.ts
```diff
@@ -69,9 +69,9 @@ export const defineMuiChip = (
         '& .MuiChip-label': {
           paddingInlineStart: 0,
           paddingInlineEnd: Spacing.xs.desktop,
-          '&:empty': { display: 'none' },
```
DanielSchiavini, line 72: Note: This was only working when using Chip with MUI icons (due to `.MuiChip-icon` check), but we can use Chip with any image

packages/curve-ui-kit/src/themes/components/chip/mui-chip.ts
```diff
@@ -69,9 +69,9 @@ export const defineMuiChip = (
...
         },
       },
+      '& .MuiChip-label:empty': { display: 'none' },
```
0xAlunara, line 74: This diff doesn't change anything functionally right? It's still the same selector as far as I can tell

apps/main/src/llamalend/PageLlamaMarkets/cells/RateCell/RewardsTooltipItems.tsx
```diff
@@ -29,7 +30,10 @@ export const RewardsTooltipItems = ({
       {hasIncentives && <TooltipItem title={title}>{percentage}</TooltipItem>}
       {extraIncentives?.map(({ percentage, title }, i) => (
         <TooltipItem key={i} subitem title={title}>
-          {formatPercent(percentage)}
+          <Stack direction="row">
+            <RewardsImg src={RCCrvLogoMD} alt={t`CRV`} width={20} />
```
0xAlunara, line 34: I had to double check if these were really only CRV rewards, and not possibly some other token.
So the only place where these `extraIncentives` are being used, is in `LendRateTooltip.tsx`
image
Here, we're taking just `lendCrvAprUnboosted` and putting it in an array of size 1.

Imo we need to make a choice here:

a) Either we get rid of the `extraIncentives` array property and replace it with a single more descriptive number, like for example `extraCrvIncentive`.
b) Or we don't assume that the rewards are CRV rewards and keep it as it is, allowing other tokens to be rewards in the future.

apps/main/src/llamalend/PageLlamaMarkets/cells/RateCell/RewardsIcons.tsx
```diff
@@ -1,30 +1,46 @@
...
+import { RCCrvLogoMD } from '@ui/images'
+import { t } from '@ui-kit/lib/i18n'
+import { RewardIcon, RewardsImg } from '@ui-kit/shared/ui/RewardIcon'
 import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
 
-const { Spacing } = SizesAndSpaces
+const { IconSize } = SizesAndSpaces
+
+const RewardChip = ({ isLast, icon }: { isLast: boolean; icon: ReactElement }) => (
+  <Chip icon={icon} size="small" color="highlight" sx={{ ...(!isLast && { marginInline: '-8px' }) }} />
```
0xAlunara, line 16: Can get rid of the need for `isLast` by relying purely on CSS:
```suggestion
<Chip icon={icon} size="small" color="highlight" sx={{ '&:not(:last-child)': { marginInline: '-8px' } }} />
```

packages/curve-ui-kit/src/themes/components/chip/mui-chip.ts
DanielSchiavini: https://github.com/curvefi/curve-frontend/pull/1117#discussion_r2177229127

apps/main/src/llamalend/PageLlamaMarkets/cells/RateCell/RewardsTooltipItems.tsx
DanielSchiavini: We will have more rewards in the future when the information is available regarding yield bearing tokens.
So, I'll just make sure to use the same hook here.

0xAlunara (approved):

0xAlunara (approved):

PR #1120, opened 2025-07-02
refactor: fix and simplify header height

DanielSchiavini:
- get rid of the layout heights, keep only the main nav height as we didn't need the rest
  - that includes the global banners
- make sure that the ref is remounted when switching between desktop and mobile
- stop prop drilling for the header height, access store directly


apps/main/src/app/GlobalLayout.tsx
```diff
@@ -60,43 +58,23 @@ export const GlobalLayout = <TId extends string, TChainId extends number>({
...
   return (
-    <Stack sx={{ minHeight: `calc(100vh - ${layoutHeight?.globalAlert}px)` }}>
+    <Stack>
       <Header
         currentApp={currentApp}
         chainId={network.chainId}
         networkId={network.id}
-        mainNavRef={mainNavRef}
         currentMenu={useAppMenu(currentApp)}
         supportedNetworks={supportedNetworks}
```
0xAlunara, line 69: If you inline this you can use an arrow expression :nerd_face:

apps/main/next-env.d.ts
```diff
@@ -1,6 +1,5 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-/// <reference types="next/navigation-types/compat/navigation" />
```
0xAlunara, line 3: What about the NOTE ser

packages/curve-ui-kit/src/features/layout/store.ts
```diff
@@ -27,22 +25,14 @@ interface LayoutState {
...
-  globalAlert: 0,
-  mainNav: 0,
-  secondaryNav: 0,
-  footer: 0,
-}
-
 const DEFAULT_STATE: LayoutState = {
-  height: DEFAULT_LAYOUT_HEIGHT,
-  navHeight: 0,
+  navHeight: 96, // Default height for desktop
```
0xAlunara, line 35: I guess we're sacrifing some clarity on how the building blocks of the navHeight with a simple 'magic' number. I don't mind either way tbh, simplifies everything by a lot

packages/curve-ui-kit/src/widgets/Header/Header.tsx
```diff
@@ -1,19 +1,23 @@
...
 import { useBetaFlag } from '@ui-kit/hooks/useLocalStorage'
 import { isChinese, t } from '@ui-kit/lib/i18n'
 import { type AppName, getInternalUrl, PAGE_DISCLAIMER, PAGE_INTEGRATIONS, routeToPage } from '@ui-kit/shared/routes'
-import { DESKTOP_HEADER_HEIGHT, DesktopHeader } from './DesktopHeader'
-import { calcMobileHeaderHeight, MobileHeader } from './MobileHeader'
+import { DesktopHeader } from './DesktopHeader'
+import { MobileHeader } from './MobileHeader'
 import { HeaderProps, NavigationSection } from './types'
 
+/* empty box to take the place behind the header (given it's sticky) */
```
0xAlunara, line 15: I was happy for a moment and thought you deleted it, but now it's back here. This is such an ugly hack.

Why not add `position: sticky` to both `DesktopHeader` and `MobileHeader` (in both cases on probably the `AppBar`) and delete this spacer. Works fine if I quickly try it with Inspect Element

0xAlunara (changes requested):

apps/main/next-env.d.ts
DanielSchiavini: This happened automatically, tried to revert but nextjs automatically does this again

packages/curve-ui-kit/src/features/layout/store.ts
DanielSchiavini: This is just the initial value, will add a comment

mimaklas (approved):

PR #1122, opened 2025-07-02
styling: lending markets

0xAlunara:
1. Fix incorrect skeleton size
image

2. Fix incorrect market title text alignment
image

3. Use • instead of - for market titles as per Figma


apps/main/src/llamalend/PageLlamaMarkets/cells/MarketTitleCell/MarketTitleCell.tsx
```diff
@@ -41,7 +47,7 @@ export const MarketTitleCell = ({ row: { original: market } }: CellContext<Llama
               paddingBlock: { mobile: '5px', tablet: 0 },
             }}
           >
-            {market.assets.collateral.symbol} - {market.assets.borrowed.symbol}
+            {market.assets.collateral.symbol} &bull; {market.assets.borrowed.symbol}
```
DanielSchiavini, line 50: I think we can use the UTF symbols in tsx?

DanielSchiavini (approved):

apps/main/src/llamalend/PageLlamaMarkets/cells/MarketTitleCell/MarketTitleCell.tsx
DanielSchiavini: :rofl: afaik `{'•'}` should render `&bull;`

PR #1127, opened 2025-07-03
fix: scroll-up button was not visible yet clickable

0xAlunara:
Was broken with the refactor to `ClientWrapper.tsx`. Removed `scrollY` from the layout store as it was literally only used for the scroll button logic. When you haven't scrolled and the button is not visible it's still clickable, but didn't bother to fix that as I suspect this entire thing gets replaced in the future anyway, and it's non-breaking.

image

### DIFF
--- a/apps/main/src/lend/layout/Settings.tsx
+++ b/apps/main/src/lend/layout/Settings.tsx
@@ -8,7 +8,7 @@ type Props = {
 }
 
 const Settings = ({ showScrollButton }: Props) => {
-  const scrollY = useLayoutStore((state) => state.scrollY)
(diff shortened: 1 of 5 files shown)


apps/main/src/lend/layout/Settings.tsx
```diff
@@ -8,7 +8,7 @@ type Props = {
 }
 
 const Settings = ({ showScrollButton }: Props) => {
-  const scrollY = useLayoutStore((state) => state.scrollY)
+  const isShowScrollButton = useLayoutStore((state) => state.showScrollButton)
```
DanielSchiavini, line 11: Is and show are a bit redundant

DanielSchiavini (approved):

PR #1134, opened 2025-07-05
feat: manage soft liq improvements

0xAlunara:
Various fixes and changes I had to make while trying to hook up the feature to the loan app itself. Made it a standalone PR so that the actual hooking up PR is smaller.

Major changes:

1. Add a 'Get crvUSD' button that opens a new tab to the quickswap page with crvUSD pre-selected as 'to' token.
2. Debt token in both tabs are now optional, because of the fact that the data may not yet be present when initially loading the page (async data loading)
3. Add an alert in the close position tab that tells the user he needs an additional crvusd balance in order to close the position.
4. When repaying the debt in the Improve Health tab, I incorrectly set the debt amount as the user's wallet balance, oops! Now shows the user balance instead, but clamped to the debt so that if you press "Max" you basically pay off the entire debt. Also renamed `balance` in the `Token` type to `amount` to more clearly indicate it's not the user balance!
5. Health action info only supported a change in value, not a single static value (in case you just opened the page and haven't changed a thing for your loan yet)
6. All action infos have been made optional

Following this PR, I'll create another PR that splits the action infos into a component each. There's a few more changes I need to make, and while justified at the start, a single `ActionInfos.tsx` component is getting unyieldy.

### DIFF
--- a/packages/curve-ui-kit/src/features/manage-soft-liquidation/ui/ActionInfos.tsx
+++ b/packages/curve-ui-kit/src/features/manage-soft-liquidation/ui/ActionInfos.tsx
@@ -95,7 +95,7 @@ export type Props = {
   }
   collateral: {
     /** Borrowed collateral token information */
-    borrowed?: TokenAmount
+    borrowed?: TokenAmount & { new?: number }
     /** The leverage multiplier if present, like 9x or 10x */
     leverage?: Delta
     /** Assets the user gets when withdrawing or closing the position */
@@ -187,7 +187,13 @@ export const ActionInfos = ({
           />
         )}
 
-        {borrowed && <ActionInfo label="Collateral" value={`${formatTokens(borrowed)}`} />}
+        {borrowed && (
+          <ActionInfo
+            label="Collateral"
+            value={`${formatTokens({ symbol: borrowed.symbol, amount: borrowed.new ?? borrowed.amount })}`}
+            prevValue={`${formatTokens({ symbol: borrowed.symbol, amount: borrowed.amount })}`}
+          />
+        )}
 
         {assetsToWithdraw && <ActionInfo label="Assets to withdraw" value={`${formatTokens(assetsToWithdraw)}`} />}
       </Stack>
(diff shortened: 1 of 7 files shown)


mimaklas (dismissed):

mimaklas (approved):

packages/curve-ui-kit/src/features/manage-soft-liquidation/ui/ActionInfos.tsx
```diff
@@ -53,6 +53,9 @@ const formatTokens = (tokens: TokenAmount | TokenAmount[]) =>
     .map((x) => `${abbreviateNumber(x.amount)}${scaleSuffix(x.amount)} ${x.symbol}`)
     .join(' + ')
 
+/** Short-hand type for MUI Typography color */
+type TextColor = React.ComponentProps<typeof Typography>['color']
```
DanielSchiavini, line 57:
```suggestion
import { TypographyProps } from '@mui/material/Typography';
type TextColor = TypographyProps['color']
```

packages/curve-ui-kit/src/features/manage-soft-liquidation/ui/ButtonGetCrvUsd.tsx
```diff
@@ -0,0 +1,33 @@
...
+ * Opens in a new tab and dynamically determines the network based on the current pathname.
+ */
+export const ButtonGetCrvUsd = () => {
+  const pathname = usePathname()
+  const networkId = getCurrentNetwork(pathname)
+  const href = `${getInternalUrl('dex', networkId, DEX_ROUTES.PAGE_SWAP)}?to=${CRVUSD_ADDRESS}`
+
+  return (
+    <Button
+      component="a"
```
DanielSchiavini, line 22: its more correct to use a next link here, but that doesn't matter much due to `target="_blank"`
```suggestion
      component={Link}
```

DanielSchiavini (approved):

DanielSchiavini (approved):

PR #1135, opened 2025-07-06
feat: move action infos to own component files and add idle states

0xAlunara:
Depends on #1134

I didn't want to copy paste the word `ActionInfo` a million times so I've chosen for component (file)names such as `Health` instead of `HealthActionInfo` or `ActionInfoHealth`. The meaning be obvious from the context and the folder they're in.


packages/curve-ui-kit/src/shared/ui/ActionInfo.tsx
```diff
@@ -120,31 +120,31 @@ const ActionInfo = ({
...
-                    pointerEvents: 'none',
-                  },
-                },
-              }}
-            >
+        <Tooltip
+          title={valueTooltip}
+          placement="top"
+          slotProps={{
+            popper: {
```
DanielSchiavini, line 127: This is already in `packages/curve-ui-kit/src/shared/ui/Tooltip.tsx` we should use that for style consistency

packages/curve-ui-kit/src/features/manage-soft-liquidation/ui/action-infos/types.ts
```diff
@@ -0,0 +1,15 @@
...
+ * @remarks Property names used to be 'old' and 'new', but you can't destructure 'new' as it's a reserved keyword.
+ * Sometimes there's also no 'next', so having just 'old' made no sense either, hence it's called 'current' instead.
+ */
+export type Delta = { current: number; next?: number }
+
+/** Simple combination of a token symbol with a certain amount */
+export type TokenAmount = { symbol: string; amount: number }
+
+/** Short-hand type for MUI Typography color */
+export type TextColor = React.ComponentProps<typeof Typography>['color']
```
DanielSchiavini, line 15:
```suggestion
export type TextColor = TypographyProps['color']
```

PR #1139, opened 2025-07-07
feat: show different crv emission label with tooltip for sidechains

0xAlunara:
Adjusts the gauge emissions label and shows a tooltip explaining the situation for side-chains with regards to the 1-week delay.
However, it needs to be confirmed that the data returned for side-chain emissions really is *always* on a 1-week delay. The data is sometimes wrong / inconsistent for some side-chains, so it needs to be fixed and/or confirmed that this holds for all side-chains before merging.

PS: I didn't bother using mui elements. Not the goal here.

image
image

### DIFF
--- a/apps/main/src/dao/components/PageGauges/GaugeListItem/GaugeDetails.tsx
+++ b/apps/main/src/dao/components/PageGauges/GaugeListItem/GaugeDetails.tsx
@@ -72,7 +72,7 @@ const GaugeDetails = ({ gaugeData, className }: { gaugeData: GaugeFormattedData;
       <Box flex flexColumn>
         <StatsTitleRow>
           <h6>{t`Gauge`}</h6>
-          <h6>{isSideChain ? t`Next Week Emissions (CRV)` : t`Current Week Emissions (CRV)`}</h6>
+          <h6>{t`Emissions (CRV)`}</h6>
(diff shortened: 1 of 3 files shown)


apps/main/src/dao/components/PageGauges/GaugeListItem/GaugeDetails.tsx
```diff
@@ -60,7 +62,7 @@ const GaugeDetails = ({ gaugeData, className }: { gaugeData: GaugeFormattedData;
       <Box flex flexColumn>
         <StatsTitleRow>
           <h6>{t`Gauge`}</h6>
-          <h6>{t`Current Week Emissions (CRV)`}</h6>
+          <h6>{chainId === 1 ? t`Current Week Emissions (CRV)` : t`Next Week Emissions (CRV)`}</h6>
```
DanielSchiavini, line 65: Please use a constant, we have the `Chain` enum and `ETHEREUM_CHAIN_ID`

DanielSchiavini (changes requested):

DanielSchiavini (dismissed):

mimaklas (dismissed):

mimaklas (approved):

OnlyJousting (approved):

PR #1155, opened 2025-07-12
refactor: simplify query combination typing

0xAlunara:
My mind is drained after this session and I no longer wish to spend further time on this nor the convoluted typing 😭 

The `useQueryMapping` function and corresponding `combineQueriesToObject` typing were all kinds of fucked up and broken anyway; straight up not working. I've tried to keep the `useQueryMapping` function, but trying to do so gets you into type hell.

In short, instead of using
```typescript
return useQueryMapping(
  uniqueAddresses.map((tokenAddress) => getTokenUsdRateQueryOptions({ chainId, tokenAddress })),
  uniqueAddresses,
)
```
you can use
```typescript
return useQueries({
  queries: uniqueAddresses.map((tokenAddress) => getTokenUsdRateQueryOptions({ chainId, tokenAddress })),
  combine: (results) => combineQueriesToObject(results, uniqueAddresses),
})
```

In the case of token queries it correctly deduces the type now as `Record<string, number>`, and not `Record<string, any>`, so you dont need these ugly hacks like
```typescript
const { data: usdRatesRaw } = useTokenUsdRates({ chainId, tokenAddresses: collateralTokenAddresses })
const usdRates = useMemo(() => usdRatesRaw ?? ({} as Record<string, number>), [usdRatesRaw])
```
but can simply use
```typescript
const { data: usdRates } = useTokenUsdRates({ chainId, tokenAddresses })
```

Very useful for other cases where you want to execute multiple non-batched queries and group them back into a record (like user balances, which is why I started working on this)

### DIFF
--- a/packages/curve-ui-kit/src/lib/queries/combine.ts
+++ b/packages/curve-ui-kit/src/lib/queries/combine.ts
@@ -14,6 +14,7 @@ export const combineQueriesToObject = <TData, K extends string[]>(
   results: QueryResultsArray<UseQueryOptions<TData, any, any, any>[]>,
   keys: K,
 ) => ({
+  // Using flatMap instead of map + filter(Boolean), because it's not correctly erasing | undefined from the Record value type
   data: Object.fromEntries(results.flatMap(({ data }, index) => (data !== undefined ? [[keys[index], data]] : []))),
   ...combineQueriesMeta(results),
 })


packages/curve-ui-kit/src/lib/queries/combine.ts
```diff
@@ -1,48 +1,20 @@
...
   ...combineQueriesMeta(results),
 })
-
-/**
- * Combines multiple queries into a single object with keys for each query
- * @param queries The query options to combine
- * @param keys The keys to use for each query
- * @returns The combined queries in an object
- */
-export const useQueryMapping = <TOptions extends Array<any>, TKey extends string[]>(
```
DanielSchiavini, line 40: not sure what was wrong with this one, but the replacement is just as simple :+1:

DanielSchiavini (approved):

packages/curve-ui-kit/src/lib/queries/combine.ts
0xAlunara: My bad I typed too hastily. The function itself was fine, it was rather the Typescript typing that wasn't working at all. I would've preferred being able to keep the function, but wasn't able to get the typing correct. Rather when using `useQueries` directly with inline function it all works fine.

PR #1166, opened 2025-07-16
fix: reset search text when resetting filters

DanielSchiavini:
- make the search input controlled so it listens properly to the column filters changing
- remove unused `extraMargin` prop


packages/curve-ui-kit/src/shared/ui/DataTable/TableFilters.tsx
```diff
@@ -70,6 +71,7 @@ export const TableFilters = <ColumnIds extends string>({
   collapsible: ReactNode // filters that may be collapsed
   chips: ReactNode // buttons that are part of the collapsible (on mobile) or always visible (on larger screens)
   sort: ReactNode // sorting options, only used for mobile (larger screens can use the table header)
+  searchText: string // text to search for, only used for mobile
```
0xAlunara, line 74: > only used for mobile

I'm not sure what's meant by this as the search also works on desktop?

0xAlunara (approved):

PR #1240, opened 2025-08-07
test: tenderly pegkeeper

0xAlunara:
For now using a personal Tenderly account, and the tests are skipped on CI/CD because the environment variables aren't set.

[Video.webm](https://github.com/user-attachments/assets/609b256e-7037-45a8-9667-b57ca58e0789)


mimaklas (dismissed):

apps/main/src/loan/components/PagePegKeepers/hooks/usePegkeeper.ts
```diff
@@ -102,7 +113,9 @@ export function usePegkeeper({ address, pool: { address: poolAddress } }: PegKee
     rate,
     debt: debt !== undefined ? formatEther(debt) : undefined,
     estCallerProfit: estCallerProfitError
-      ? '0'
+      ? estCallerProfitFallback !== undefined
+        ? formatEther(estCallerProfitFallback)
+        : '0'
```
DanielSchiavini, line 118: IMO we shouldn't use 0 when values are undefined

packages/curve-ui-kit/src/features/connect-wallet/lib/types.ts
```diff
@@ -38,3 +38,12 @@ export type Libs = {
 export type LibKey = keyof Libs
 
 export { CurveChainId, LlamaChainId, CurveNetworkId, LlamaNetworkId }
+
+export type Network = {
```
DanielSchiavini, line 42: How is this related to the other network definitions elsewhere? :skull:

packages/curve-ui-kit/src/features/connect-wallet/lib/useWallet.ts
```diff
@@ -28,23 +29,23 @@ export const useWallet = () => {
   const { disconnect } = useDisconnect()
 
   const connect = useCallback(
-    async (connector?: (typeof supportedWallets)[number]['connector']) => {
-      if (!connector) {
+    async (selectedConnector?: Connector) => {
+      // When using Cypress, we want to use the one and only (test) connector without blocking modal
```
DanielSchiavini, line 33: Can't cypress use the modal? Although auto-connect would be nice

packages/curve-ui-kit/src/features/connect-wallet/lib/wagmi/wagmi-config.ts
```diff
@@ -17,84 +10,45 @@ declare module 'wagmi' {
...
        * Solves the same issue as discussed on https://github.com/wevm/wagmi/discussions/3537.
        * We don't use wallet specific injected connectors anyway and prefer to use the single global one.
        */
       multiInjectedProviderDiscovery: false,
-    })
-  },
+    }),
   {
     max: 1, // only memoize the last call
+    normalizer: ([{ chains, transports }]) =>
```
DanielSchiavini, line 47: I think this is only necessary due to the missing useMemo above?

tests/cypress/component/llamalend/pegkeeper.cy.tsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react'
+import { PegKeeper } from '@/loan/components/PagePegKeepers/components/PegKeeper'
+import { PEG_KEEPERS } from '@/loan/components/PagePegKeepers/constants'
+import { ClientWrapper, type Config } from '@/support/helpers/ClientWrapper'
+import { createTestWagmiConfigFromVNet, withVirtualTestnet } from '@/support/helpers/vnet'
+
+const getVnet = withVirtualTestnet((uuid) => ({
```
DanielSchiavini, line 7:
```suggestion
const getVirtualNetwork = withVirtualTestnet((uuid) => ({
```

tests/cypress/fixtures/next-font-local.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * NextJS local font isn't available in Cypress component tests.
```
DanielSchiavini, line 2: This can be removed after #1151

tests/cypress/support/helpers/ClientWrapper.tsx
```diff
@@ -0,0 +1,36 @@
...
+export type Config = ResolvedRegister['config']
+
+type Props = {
+  config: Config
+  children: ReactNode
+  autoConnect: boolean
+}
+
+/**
+ * Client wrapper for Cypress component testing.
```
DanielSchiavini, line 16: nice

tests/cypress/support/helpers/test-connector.ts
```diff
@@ -0,0 +1,46 @@
...
+  /** The testnet chain configuration */
+  chain: Chain
+}
+
+/**
+ * Cypress test connector for Wagmi.
+ *
+ * This connector is designed for use in test environments (e.g., Cypress) with a testnet chain.
+ * It creates a wallet using a custom seed (private key) to allow testing contract write calls
+ * without relying on third-party browser extensions like MetaMask.
```
DanielSchiavini, line 17: Did you try just injecting `window.ethereum` instead? It sounds simpler to use the injected provider

tests/cypress/support/helpers/wagmi.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+/**
+ * Creates a Wagmi configuration for testing with a private key connector to a custom (testnet) RPC
+ *
+ * @param options - Configuration options
+ * @returns Configured Wagmi config instance for testing
+ */
+export function createTestWagmiConfig({ privateKey, rpcUrl, explorerUrl }: Options) {
+  // Tenderly docs recommends using chain ID 73571 to prevent replay attack, but a lot of our code relies on `if chain === Chain.Ethereum`.
```
DanielSchiavini, line 23: That's why I always argue we should stop checking specific chains and check for features instead

tests/cypress/support/helpers/wagmi.ts
```diff
@@ -0,0 +1,38 @@
...
+  // Should be okay, since we're not using real life wallets. A browser wallet extension isn't even available.
+  const chain = defineChain({
+    ...mainnet,
+    rpcUrls: { default: { http: [rpcUrl] } },
+    blockExplorers: explorerUrl ? { default: { name: 'Tenderly Explorer', url: explorerUrl } } : undefined,
+  })
+
+  const config = createWagmiConfig({
```
DanielSchiavini, line 31:
```suggestion
  return createWagmiConfig({
```

tests/cypress/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "jsx": "react",
     "baseUrl": "./",
     "paths": {
-      "@/*": ["*"],
+      "@/*": ["*", "../../apps/main/src/*"],
```
DanielSchiavini, line 7: Please don't make the imports ambiguous

```suggestion
      "@/*": ["*"],
      "@main/*": ["../../apps/main/src/*"],
```

tests/cypress/support/tasks/tenderly/vnet-delete.ts
```diff
@@ -0,0 +1,31 @@
...
+    headers: {
+      'Content-Type': 'application/json',
+      Accept: 'application/json',
+      'X-Access-Key': accessKey,
+    },
+    body: JSON.stringify(deleteOptions),
+  })
+
+  if (!response.ok) {
+    throw new Error(
```
DanielSchiavini, line 25: We can reuse `fetchJson` from the prices API

tests/cypress/types/tasks.d.ts
```diff
@@ -0,0 +1,17 @@
...
+  CreateVirtualTestnetOptions,
+  CreateVirtualTestnetResponse,
+  DeleteVirtualTestnetsOptions,
+} from 'support/helpers/tenderly'
+
+/** Doesn't work properly, because arg is still inferred as `any`. Not sure if this is solvable, leaving it here for now */
+declare global {
+  namespace Cypress {
+    interface Chainable {
+      task(event: 'createVirtualTestnet', arg: CreateVirtualTestnetOptions): Chainable<CreateVirtualTestnetResponse>
```
DanielSchiavini, line 13: Does this really need be a task that runs in nodejs? I think it can run in the browser just fine, it's just fetch requests

turbo.json
```diff
@@ -30,7 +30,11 @@
     "ARBITRUM_ALCHEMY_API_KEY",
     "HARDHAT_CHAIN_ID",
     "HARDHAT_MNEMONIC",
+    "CYPRESS_COMPONENT_TEST",
     "CYPRESS_DAPP",
+    "CYPRESS_TENDERLY_ACCOUNT_SLUG",
+    "CYPRESS_TENDERLY_PROJECT_SLUG",
+    "CYPRESS_TENDERLY_ACCESS_KEY",
```
DanielSchiavini, line 37: Please add these to [the workflow](https://github.com/curvefi/curve-frontend/blob/e6ea3446ab98287323f38533f1811529530a477f/.github/workflows/ci.yaml#L29) and send an admin the actual secrets

apps/main/src/app/ClientWrapper.tsx
```diff
@@ -95,15 +99,20 @@ function useThemeAfterSsr(preferredScheme: 'light' | 'dark' | null) {
...
   preferredScheme: 'light' | 'dark' | null
 }) => {
-  const theme = useThemeAfterSsr(preferredScheme)
-  const config = useMemo(() => createWagmiConfig(networks), [networks])
+  const networks = recordValues(networkDefs)
+  const chains = networks.map((network) => createChainFromNetwork(network, defaultGetRpcUrls)) as [Chain, ...Chain[]]
+  const transports = Object.fromEntries(
+    networks.map((network) => [network.chainId, createTransportFromNetwork(network, defaultGetRpcUrls)]),
+  )
+  const config = createWagmiConfig({ chains, transports })
```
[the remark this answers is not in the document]
0xAlunara: I actually ran into some issues ([fixed in this commit](https://github.com/curvefi/curve-frontend/pull/1240/commits/a30308c2f11dae9689f7c618b80caa202c1e3348)) as I made the chains, connectors and transports parameters of the `createWagmiConfig` function. Turns out that, because these two parameters are arrays and objects, I had to add the `normalizer` function to the `memoize` function, as otherwise I'd indeed end up with different configs.

However, as it turns out, since we're already using `memoize` we don't need to use `useMemo` at all.

packages/curve-ui-kit/src/features/connect-wallet/lib/wagmi/wagmi-config.ts
0xAlunara: Even with the useMemo it's necessary. Took me a lot of debugging to find out I could get rid of all the memos and this normalizer function was literally all that's needed to avoid multiple config instances.

tests/cypress/support/helpers/test-connector.ts
0xAlunara: Sounds like a really roundabout way of achieving the same result. I think using a connector is the way to go. Wagmi also provides a [mock](https://wagmi.sh/core/api/connectors/mock) connector but that's of course not what we want here. They used to have docs that suggest creating connectors but it seems to have vanished...

tests/cypress/support/tasks/tenderly/vnet-delete.ts
0xAlunara: Nope, because I need to add the X-Access-Key HTTP header and also need to change the `method`. Now I *could* add header options to `fetchJson` from prices api, but honestly, that function was precisely made for prices api and I don't want to needlessly make it more complex. Nothing wrong with just raw dogging `fetch` imo

0xAlunara: By the way I kinda need to make this a cy.request() anyway if we want to get rid of tasks

apps/main/src/app/ClientWrapper.tsx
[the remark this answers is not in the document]
DanielSchiavini: I do think we want the useMemo for performance at least, as otherwise the whole configuration will be still recreated and compared in every render, just to be thrown away by the memoization

tests/cypress/support/helpers/test-connector.ts
DanielSchiavini: No problem. FYI window.ethereum interface is [super simple](https://docs.metamask.io/wallet/reference/provider-api/)

tests/cypress/tsconfig.json
0xAlunara: This doesn't work, because the app code imported from main doesn't use @main itself, so if there's further imports it can't find them.

1. You import @main/foo.ts
2. But foo imports @lend/bla

Now step 2 fails because it can't find @lend (as you just reconfigured it to @main/lend)

packages/curve-ui-kit/src/features/connect-wallet/lib/types.ts
0xAlunara: Cheers, got lost in the many types; using that now

apps/main/src/loan/components/PagePegKeepers/hooks/usePegkeeper.ts
DanielSchiavini: btw this could be possibly moved to the format function

OnlyJousting (approved):

tests/cypress/component/llamalend/pegkeeper.cy.tsx
```diff
@@ -0,0 +1,82 @@
...
+  },
+}))
+
+const TestComponent = ({ config }: { config: Config }) => (
+  <ClientWrapper config={config} autoConnect={false}>
+    <PegKeeper {...PEG_KEEPERS[0]} sx={{ maxWidth: '30rem' }} testId="pegkeeper" />
+  </ClientWrapper>
+)
+
+describe('PegKeeper', () => {
```
DanielSchiavini, line 21: This should be in the `loan` folder not `llamalend` also should it not be called "peg stability reserves"?

tests/cypress/e2e/loan/peg-stability-reserves.ts
```diff
@@ -15,10 +16,9 @@ describe(`Peg Stability Reserves`, () => {
     cy.get('[data-testid="pegkeepers"]', LOAD_TIMEOUT).should('be.visible')
   })
 
-  // Ideally I would match it with the number of entries from PEG_KEEPERS, but that import from @/loan isn't available in tests
   it('should render all pegkeepers', () => {
-    cy.get('[data-testid^="pegkeeper-"]').should('have.length.greaterThan', 0)
-    cy.get('[data-testid^="pegkeeper-"]').each(($el) => {
+    cy.get('[data-testid^="pegkeeper-card"]').should('have.length', PEG_KEEPERS.length)
```
DanielSchiavini, line 20: The tests are failing here
image

tests/cypress/component/llamalend/pegkeeper.cy.tsx
0xAlunara: At one point we renamed pegkeepers to peg stability reserves and I didn't want a giant PR renaming every single file, so there's some intermingling. Renamed it regardless

Technically it's not loan specific but it does live in the loan app because that's where it originated. Still looking forward to merge both apps in a single llamalend app 😅

tests/cypress/e2e/loan/peg-stability-reserves.ts
0xAlunara: Should be fixed now, I changed some testids for the other tests

tests/cypress/e2e/loan/peg-stability-reserves.ts
```diff
@@ -1,5 +1,15 @@
+import { PEG_KEEPERS } from '@/loan/components/PagePegKeepers/constants'
 import { LOAD_TIMEOUT, oneViewport } from '@/support/ui'
 
+/**
+ * Gets pegkeeper cards by test ID suffix.
+ * The complete test ID format is: pegkeeper-card-{contractAddress}-{suffix}
+ * where contractAddress makes each pegkeeper's test IDs unique.
+ */
+function getPegkeeperCards(suffix: string) {
+  return cy.get(`[data-testid^="pegkeeper-card"][data-testid$="-${suffix}"]`)
```
DanielSchiavini, line 10: I usually create nested testids instead, instead of concatenating - then you don't need to filter out code "in between"

DanielSchiavini (approved):

PR #1241, opened 2025-08-08
feat: top page progress bar loader

pierreVinet:
**Description:**
When navigating between pages, the current UI does not indicate that the next page is loading on the server.
This lack of feedback can cause users to think their click didn’t register, leading to multiple clicks or leaving the page, especially noticeable on the /llamalend page.

**Solution:**
Implemented a top-page progress bar to visually indicate loading state during page transitions. Using the project’s theme colors.

**Preview:**

https://github.com/user-attachments/assets/c3d6ab98-b11a-4c30-bb65-e6eca2514898

### DIFF
--- a/apps/main/package.json
+++ b/apps/main/package.json
@@ -36,6 +36,7 @@
     "memoizee": "^0.4.17",
     "next": "15.3.5",
     "next-images": "^1.8.5",
+    "nextjs-toploader": "^3.8.16",
     "react": "*",
     "react-dom": "*",
     "react-hook-form": "7.59.0",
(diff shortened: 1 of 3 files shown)


0xAlunara (approved):
Note that we're working on removing NextJS so this might be short lived. Looks cool nonetheless.

pierreVinet:
> Note that we're working on removing NextJS so this might be short lived. Looks cool nonetheless.

Is there already a working branch without nextjs? Happy to work on a free nextjs version with nprogress if doesn't exist already

OnlyJousting (approved):

DanielSchiavini (approved):

PR #1270, opened 2025-08-20
feat: user position tables

DanielSchiavini:
**User Positions**
- Create user positions tables for borrow and supply; show them in a 
- Simplify the tables so they don't show subtitle or any buttons (reload, filter, column settings)
- Refactor filter chips so different options can be shown in each page, depending on the data and table

_Other changes_
- Fix line height for the headingS typography variant (mistake in design system)
- Move the vite plugin dependency to the right project (main)
- Remove unused code
- Make the `href` prop of the tabs switcher optional, removing '' where passed


apps/main/src/llamalend/PageLlamaMarkets/LlamaMarketExpandedPanel.tsx
```diff
@@ -62,7 +62,7 @@ const RateItem = ({ market, title, type }: { market: LlamaMarket; title: string;
 export const LlamaMarketExpandedPanel: ExpandedPanel<LlamaMarket> = ({ row: { original: market } }) => {
   const { data: earnings, error: earningsError } = useUserMarketStats(market, LlamaMarketColumnId.UserEarnings)
   const { data: deposited, error: depositedError } = useUserMarketStats(market, LlamaMarketColumnId.UserDeposited)
```
0xAlunara, line 64: Unrelated perhaps, but `earningsError` and `depositedError` aren't being used

0xAlunara:
Perhaps should disable the token icon tooltip
image

0xAlunara:
The empty table is quite large 😄 
image

0xAlunara:
If a user only has supply positions and no borrow positions, does the table automatically go to the "Supply" tab? Don't have a user to test that for atm
image

DanielSchiavini:
@0xAlunara [the rename](https://github.com/curvefi/curve-frontend/pull/1270/commits/179e7ac34d3044276f337fd04927620c488af209) broke the table if you had the old name in local storage. So just remove the local storage data to test again please

0xAlunara (dismissed):

0xAlunara (approved):

OnlyJousting (approved):

PR #1277, opened 2025-08-22
feat: user position indicator

DanielSchiavini:
- add llama position indicator for user positions

### DIFF
--- a/packages/curve-ui-kit/src/shared/ui/DataTable/DataCell.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/DataTable/DataCell.tsx
@@ -19,7 +19,6 @@ export function getCellSx<T extends TableItem>({
   showCollapseIcon?: boolean
   isSticky: boolean
 }) {
-  const { borderRight } = column.columnDef.meta ?? {}
   // with the collapse icon there is an extra wrapper, so keep the sx separate
   const wrapperSx = {
     textAlign: getAlignment(column),


apps/main/src/llamalend/PageLlamaMarkets/cells/MarketTitleCell/MarketTitleCell.tsx
```diff
@@ -12,13 +13,18 @@ import { RouterLink } from '@ui-kit/shared/ui/RouterLink'
 import { TokenPair } from '@ui-kit/shared/ui/TokenPair'
 import { Sizing } from '@ui-kit/themes/design/0_primitives'
 import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
+import { UserPositionIndicator } from './UserPositionIndicator'
 
 const { Spacing } = SizesAndSpaces
 
+// the padding is removed from this cell, so we can show the indicator on the border, the spacer returns it back
+const PaddingSpacer = () => <Box sx={{ width: Spacing.md }} />
```
0xAlunara, line 21: What I dislike about this hack, is that when you remove all padding from a cell and add in this spacer, you suddenly have twice the padding on the left side, but the padding on the right side is still gone.

What I think is a better solution, is to remove the `noPadding` cell property and just give the `<UserPositionIndicator>` a `margin-inline-start` of negative `Spacing.sm` to place it back inside the padding. (An alternative would be absolute positioning but that's a bit meh as well)

Here's a screenshot where I added `padding-inline: 2rem` to the cell, followed by `margin-inline-start` of `-2rem` applied to the `<UserPositionIndicator>` as a test, so the idea should work
image

DanielSchiavini: I did try that, to make the code cleaner, but that changes the layout of the other columns as well somehow. The right padding should definitely be fixed!

0xAlunara (commented):
<snip, comment for wrong PR>

packages/curve-ui-kit/src/shared/ui/DataTable/DataTable.d.ts
```diff
@@ -12,6 +12,7 @@ declare module '@tanstack/react-table' {
     variant?: TypographyVariantKey
     borderRight?: boolean
     hideZero?: boolean
+    noPadding?: boolean
```
0xAlunara, line 15: This is no longer being used, is it?

DanielSchiavini: yes sorry, I was supposed to remove this

0xtutti (approved):

PR #1278, opened 2025-08-22
refactor: use usd balance instead of usd price

DanielSchiavini:
- We currently use the API data to calculate the USD rate of a specific token
- However, we use some complicated querying to retrieve it at query time, while that's not necessary
- Instead, we can use react-query to retrieve the usd rates as wished
- Also we can add some proper loading skeletons

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ node_modules
 # testing
 coverage
 cypress.env.json
+tests/cypress/screenshots
 
 # next.js
 .next/
--- a/apps/main/src/dex/components/PageRouterSwap/components/DetailInfoTradeRouteRoute.tsx
+++ b/apps/main/src/dex/components/PageRouterSwap/components/DetailInfoTradeRouteRoute.tsx
@@ -4,7 +4,7 @@ import { styled } from 'styled-components'
 import type { Route } from '@/dex/components/PageRouterSwap/types'
 import { ROUTE } from '@/dex/constants'
 import useStore from '@/dex/store/useStore'
-import { ChainId, NetworkEnum, type UrlParams } from '@/dex/types/main.types'
+import { type UrlParams } from '@/dex/types/main.types'
 import { getPath } from '@/dex/utils/utilsRouter'
 import Icon from '@ui/Icon'
 import { ExternalLink } from '@ui/Link'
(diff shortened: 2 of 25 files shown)


apps/main/src/llamalend/entities/usd-prices.ts
```diff
@@ -1,9 +1,7 @@
-import type { Chain } from '@curvefi/prices-api'
 import { ContractParams, queryFactory, rootKeys } from '@ui-kit/lib/model'
 import { contractValidationSuite } from '@ui-kit/lib/model/query/contract-validation'
-import type { Address } from '@ui-kit/utils'
 
-const { fetchQuery: fetchCoinPrice } = queryFactory({
+export const { useQuery: useTokenUsdPrice } = queryFactory({
```
0xAlunara, line 4: Wait is this entity still required? Can't we use `useTokenUsdRate` from `curve-ui-kit\src\lib\model\entities\token-usd-rate.ts`? Having two token usd queries is a bit ... odd, prob best if we consolidate it? Afaik that query doesn't need the library to be hydrated and queries coingecko too, and there's a PR for Denis to check Prices API as well

apps/main/src/llamalend/PageLlamaMarkets/cells/UtilizationCell.tsx
```diff
@@ -12,7 +12,6 @@ import { Tooltip } from '@ui-kit/shared/ui/Tooltip'
 import { TooltipItem, TooltipItems } from '@ui-kit/shared/ui/TooltipComponents'
 import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
 import { LlamaMarketType } from '@ui-kit/types/market'
-import { PercentageCell } from './PercentageCell'
```
0xAlunara, line 15: Byebye lint warning 😄

apps/main/src/llamalend/entities/usd-prices.ts
0xAlunara: I don't even know why we have 2 libraries to begin with...

DanielSchiavini: we used to have 3, of which 2 have been merged :see_no_evil: 
they are mostly copy paste

OnlyJousting (approved):

pilotvietnam (approved):

0xtutti (approved):

0xAlunara (approved):

OnlyJousting (approved):

0xtutti (approved):

PR #1281, opened 2025-08-22
fix: loading indicator

DanielSchiavini:
- IconButton sizes change the size of all svg, but the loading indicator gets larger than its container


0xAlunara:
Not sure how you tested, but if throttle network speed it's still rotating with an incorrect transformation origin for me.
Bonus point: I just noticed there's a spinner in the connect wallet button too 🤣 
image

Fwiw, I throttle speed to 3G and disable cache
image

PR #1295, opened 2025-08-27
fix: typechecking

0xAlunara:
Refactor TypeScript configurations for better type checking and remove unused files. There's now a single tsconfig.base.json of which all other tsconfigs derive from; usually they only need to add some paths and includes.

Address various type errors in snapshot queries, imports, and component props. These were already broken, but never popped up for unknown reason when we were using next.js


pilotvietnam (approved):

apps/main/src/llamalend/PageLlamaMarkets/cells/HealthCell.tsx
```diff
@@ -20,7 +20,7 @@ export const HealthCell = ({ row }: CellContext<LlamaMarket, number>) => {
   ) : (
     <Tooltip
       title={softLiquidation ? 'Liquidation Protection On' : 'Position active'}
-      body={<HealthTooltipContent softLiquidation={softLiquidation} />}
+      body={<HealthTooltipContent softLiquidation={!!softLiquidation} />}
```
DanielSchiavini, line 23: wasn't softliquidation a boolean? :thinking:

DanielSchiavini (approved):

apps/main/src/llamalend/PageLlamaMarkets/cells/HealthCell.tsx
0xAlunara: `const { health, softLiquidation } = data ?? {}`
means the type of `softLiquidated` is `boolean | undefined`, the prop expects only `boolean`, no `undefined`.

PR #1304, opened 2025-08-29
feat: local storage migration

DanielSchiavini:
- add migrations for old local storage filters and visibility settings
  - filters remove the old 10k liquidity filter, keeping the new TVL filter
  - visibility gets replaced, old value is only used if it already contains TVL
- add test to make sure the old fields work properly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ node_modules
 # testing
 coverage
 cypress.env.json
+tests/cypress/screenshots
 
 # next.js
 .next/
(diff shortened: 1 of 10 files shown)


0xAlunara (dismissed):
I would've personally picked the lazy option and just clear the localstorage and reset to default settings instead of cooking up migration logic 🙈

OnlyJousting (approved):

OnlyJousting (approved):

0xtutti (approved):

0xAlunara (approved):

PR #1309, opened 2025-08-30
feat: use new number formatter in old formatter

0xAlunara:
Initially I started replacing all calls of the old `formatNumber` to the new `formatNumber`, but quickly figured that would make the PR super large and unwieldy. Instead, I've decided to turn the old `formatNumber` into a wrapper function to the new one.

Over time as we replace the old UI with the new one, usage of the old `formatNumber` will be replaced with the new one. Alternatively, we could create a ticket & PR at a later point to remove the wrapper.

In addition to the wrapper, I've also:
* Moved date formatting to `utilsDate`
* Added support for `showDecimalIfSmallNumberOnly` in the new number formatter

I've tried to compare every page on the front-end with prod, and found no big issues. Formatting might've changed here and there (namely decimal amount), but in its entirety everything should be more consistent now.


packages/ui/src/utils/utilsDate.ts
```diff
@@ -39,3 +39,61 @@ export const convertTimeAgo = (input: Date | number | string) => {
...
+    return new Intl.DateTimeFormat(undefined, {
+      month: 'short',
+      day: 'numeric',
+      year: 'numeric',
+      hour: 'numeric',
+      minute: '2-digit',
+      hour12: true,
+    }).format(date)
+  }
+}
```
DanielSchiavini, line 77:
```suggestion
export const formatDate = (date: Date | string, variant: 'short' | 'long' = 'short') => new Intl.DateTimeFormat(undefined,  variant === 'short' ? {
      // short - example: 31/01/25 adjusted for local timezone
      day: '2-digit',
      month: 'short',
      year: 'numeric',
    } : {
      // long - example: 31 January, 12:00 AM adjusted for local timezone
      month: 'short',
      day: 'numeric',
      year: 'numeric',
      hour: 'numeric',
      minute: '2-digit',
      hour12: true,
    }).format(typeof date === 'string' ? new Date(date) : date)
```

packages/ui/src/utils/utilsDate.ts
```diff
@@ -39,3 +39,61 @@ export const convertTimeAgo = (input: Date | number | string) => {
...
+ *
+ * @param unixTime - The Unix timestamp in seconds (not milliseconds)
+ * @returns A formatted date string adjusted for the local timezone.
+ *
+ * @see {@link formatDate} for formatting the resulting timestamp into human-readable strings
+ */
+export const formatDateFromTimestamp = (unixTime: number) => {
+  const date = new Date(unixTime * 1000)
+  return formatDate(date)
+}
```
DanielSchiavini, line 90:
```suggestion
export const formatDateFromTimestamp = (unixTime: number) => formatDate(new Date(unixTime * 1000))
```

packages/ui/src/utils/utilsFormat.ts
```diff
@@ -55,169 +33,25 @@ export function getFractionDigitsOptions(val: number | string | undefined | null
...
-    console.error(error, { params: { val, options } })
-    if (typeof val === 'undefined' || val === null || val === '') {
-      return defaultValue || ''
-    } else if (lodash.isNaN(Number(val))) {
-      return '?'
-    } else {
-      return val.toString()
-    }
-  }
+  if (val === undefined) return options?.defaultValue ?? '-'
```
DanielSchiavini, line 38: The `-` should be moved to the destruct statement imo
We should use `==` so it also catches `null`
```suggestion
  if (val == null) return defaultValue ?? '-'
```

packages/ui/src/utils/utilsDate.ts
DanielSchiavini: alternatively the objects could be in a constant (`Record<Variant, DateOptions>`)

DanielSchiavini (approved):

DanielSchiavini (commented):
Please update the llamalend test, it expects the old format

apps/main/src/lend/components/DetailsUser/components/UserInfoLeverage.tsx
```diff
@@ -1,11 +1,16 @@
...
 
 export const UserInfoLeverage = ({ userActiveKey }: { userActiveKey: string }) => {
   const { error, ...details } = useUserLoanDetails(userActiveKey)
 
   if (error) return '?'
 
   return (
-    <>{!details?.leverage || !isFinite(Number(details.leverage)) ? '-' : `${Number(details.leverage).toFixed(2)}x`}</>
+    <>
+      {!details?.leverage || !isFinite(Number(details.leverage))
```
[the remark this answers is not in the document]
0xAlunara: Remainder from when I started replacing the calls before writing a wrapper 😄

PR #1310, opened 2025-08-31
feat: add max leverage metric for v1 and v2 mint markets

OnlyJousting:
Changes:
- Show max leverage metric in v1 and v2 mint markets


DanielSchiavini (approved):
This should be added to the backend and the table too

OnlyJousting:
> This should be added to the backend and the table too

I can ask them to handle this on the API too (not sure what the issue has been there)

apps/main/src/loan/entities/mint-market-max-leverage.ts
```diff
@@ -15,8 +15,15 @@ type MarketParams = FieldsOf<MarketQuery>
...
-  return { value: maxLeverage }
+  const hasMaxLeverage = market.leverageV2.hasLeverage()
+
+  if (hasMaxLeverage) {
+    const v2MaxLeverage = await market.leverageV2.maxLeverage(market?.minBands)
+    return { value: v2MaxLeverage }
+  }
+
+  const v1MaxLeverage = await market.leverage.maxLeverage(market?.minBands)
+  return { value: v1MaxLeverage }
```
DanielSchiavini, line 26: The `value` field is not necessary
```suggestion
  return hasMaxLeverage ? market.leverageV2.maxLeverage(minBands) : market.leverage.maxLeverage(minBands)
```

apps/main/src/loan/hooks/useMarketDetails.ts
```diff
@@ -87,6 +91,10 @@ export const useMarketDetails = ({ chainId, llamma, llammaId }: UseMarketDetails
         : null,
       loading: isSnapshotsLoading || (loanDetails?.loading ?? true),
     },
+    maxLeverage: {
+      value: maxLeverage ? Number(maxLeverage?.value) : null,
```
DanielSchiavini, line 95: converting to number should happen in the query not in the hook imo

DanielSchiavini (dismissed):

DanielSchiavini (approved):

PR #1324, opened 2025-09-02
refactor: route details component

DanielSchiavini:
- rename the `DetailInfoLeverageAdvancedExpected` component to `RouteDetails`
- move it to the llamalend app
- allows imports from the lend and loan apps
- this will be used in #1315


apps/main/src/llamalend/widgets/RouteDetails/utils.ts
```diff
@@ -0,0 +1,5 @@
+import { formatNumber, NumberFormatOptions } from '@ui/utils'
```
0xAlunara, line 1: Preferably use the one from ui-kit directly, from ui is just a wrapper now.

0xAlunara (commented):
Forgot to commit the deletion of `DetailInfoLeverageAdvancedExpected`? I was expecting a rename, not only new code 😄

pilotvietnam (dismissed):

OnlyJousting (approved):

PR #1326, opened 2025-09-02
test: tenderly endpoints

0xAlunara:
* Expands tenderly functionality with getting and forking existing vnets.
* Call the correct endpoint for deleting a single vnet (was ioncorrectly calling the plural variant with just 1 vnet).
* Re-uses common tenderly API types for shorter code.
* Splits component tests into normal component tests and special component tests that will require a tenderly account. These are more expensive to call, so in the future we probably want to call them only once as a requirement before merging, and not every commit. This PR does not yet bother with that, though. For now, if no Tenderly account is configured, the .rpc.cy.tsx tests will still pass.
* Splits off component tests from the e2e tests as well.


pilotvietnam (approved):

.github/workflows/ci.yaml
```diff
@@ -33,7 +34,50 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: yarn
+      - run: yarn set version stable
+      - run: yarn install --immutable
+
+      - run: yarn run cy:run:rpc --browser electron # rpc test are expensive, so we stick to electron
```
DanielSchiavini, line 74:
```suggestion
      - run: yarn run cy:run:rpc --browser electron # rpc tests are expensive, stick to electron
```

.github/workflows/ci.yaml
```diff
@@ -33,7 +34,50 @@ jobs:
...
+        working-directory: tests
+        env:
+          ELECTRON_ENABLE_LOGGING: true # send console logs to stdout in electron browser
+          NODE_ENV: test
```
DanielSchiavini, line 78: for RPC I guess we need the secrets being passed here

tests/cypress/support/helpers/tenderly/account.ts
```diff
@@ -0,0 +1,12 @@
...
+  accountSlug: string
+  projectSlug: string
+  accessKey: string
+}
+
+const accountSlug = Cypress.env('TENDERLY_ACCOUNT_SLUG')
+const projectSlug = Cypress.env('TENDERLY_PROJECT_SLUG')
+const accessKey = Cypress.env('TENDERLY_ACCESS_KEY')
+
+export const tenderlyAccount =
```
DanielSchiavini, line 11: tbh I'd rather throw an error if you import this file without the proper environment set.
If needed move it to a function. It makes much easier to find problems in your local/ci setup.

tests/cypress/support/helpers/tenderly/types.ts
```diff
@@ -1,5 +1,126 @@
...
+ * // Make 'email' optional
+ * type UserWithOptionalEmail = MakeOptional<User, 'email'>;
+ * // Result: { id: string; name: string; email?: string }
+ *
+ * // Make multiple properties optional
+ * type UserWithOptionalFields = MakeOptional<User, 'name' | 'email'>;
+ * // Result: { id: string; name?: string; email?: string }
+ * ```
+ */
+export type MakeOptional<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>
```
DanielSchiavini, line 126: pls make this available everywhere, I've missed this!

tests/cypress/support/helpers/tenderly/vnet-fork.ts
```diff
@@ -0,0 +1,57 @@
+import type { TenderlyAccount } from './account'
+import type { TestnetProps } from './types'
+
+/** Implemented as per https://docs.tenderly.co/reference/api#/operations/forkVnet */
+export type ForkVirtualTestnetOptions = Partial<Pick<TestnetProps, 'slug' | 'display_name' | 'description'>> & {
+  /** @see {@link TestnetProps.id} for complete documentation */
+  vnet_id: TestnetProps['id']
```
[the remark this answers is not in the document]
DanielSchiavini: Just add this comment to the code pls, it'll get lost here

tests/cypress/support/helpers/tenderly/vnet-fork.ts
```diff
@@ -0,0 +1,57 @@
...
+  | 'created_at'
+  | 'fork_config'
+  | 'virtual_network_config'
+  | 'sync_state_config'
+  | 'explorer_page_config'
+  | 'status'
+  | 'rpcs'
+>
+
+export function forkVirtualTestnet({
```
DanielSchiavini, line 31: These would be a bit simpler as arrow functions.. not important

tests/cypress/support/helpers/tenderly/vnet.ts
```diff
@@ -1,39 +1,77 @@
...
+  privateKey?: Hex
+}) {
   const rpcUrl = vnet.rpcs.find((rpc) => rpc.name === 'Admin RPC')?.url
   const explorerUrl = vnet.rpcs.find((rpc) => rpc.name === 'Public RPC')?.url
 
   if (!rpcUrl) throw new Error('RPC URL is undefined')
 
   return createTestWagmiConfig({
-    privateKey: generatePrivateKey(),
+    privateKey: privateKey ?? generatePrivateKey(),
```
DanielSchiavini, line 36: generatePrivateKey could be moved to default parameter value

tests/cypress/support/helpers/tenderly/vnet.ts
```diff
@@ -76,17 +114,71 @@ export function withVirtualTestnet(opts: (uuid: number) => DeepPartial<CreateVir
...
+  })
+
+  after(() => {
+    if (!vnet) return
+
+    const deleteOpts: DeleteVirtualTestnetOptions = {
+      vnetId: vnet.id,
+    }
+
+    deleteVirtualTestnetRequest({ ...tenderlyAccount!, ...deleteOpts })
```
DanielSchiavini, line 127: if deleteVirtualTestnetRequest is properly typed, DeleteVirtualTestnetOptions should be inferred, right?
```suggestion
    deleteVirtualTestnetRequest({ ...tenderlyAccount!, vnetId: vnet.id })
```

tests/cypress/support/helpers/tenderly/vnet.ts
```diff
@@ -51,7 +89,7 @@ export type DeepPartial<T> = {
  * })
  * ```
  */
-export function withVirtualTestnet(opts: (uuid: number) => DeepPartial<CreateVirtualTestnetOptions>) {
+export function createVirtualTestnet(opts: (uuid: number) => DeepPartial<CreateVirtualTestnetOptions>) {
   let vnet: CreateVirtualTestnetResponse
 
   before(() => {
```
DanielSchiavini, line 95: also here, imo the `if (!tenderlyAccount)` makes it hard to find setup issues

DanielSchiavini (approved):
Nice stuff!

.github/workflows/ci.yaml
0xAlunara: I don't think anything needs to be passed. You can check `prices-api.yaml` where we use the NPM secret for publishing the package: `yarn config set npmAuthToken "${{ secrets.NPM_TOKEN }}"`

tests/cypress/support/helpers/tenderly/account.ts
0xAlunara: Sure. For the time being the tests will fail on Github Actions until we get an account set up on the Github repository

tests/cypress/support/helpers/tenderly/vnet.ts
0xAlunara: Ah yeah, kept the explicit typing out of consistency, but can remove

tests/cypress/support/helpers/tenderly/types.ts
0xAlunara: Moved it to `ui-kit/types/util.ts` for the alck of a better place. Could add more useful types, but figured it's best to keep it small as possible with types that we actually use.

OnlyJousting (dismissed):

OnlyJousting (approved):

PR #1329, opened 2025-09-02
refactor: inline query factory method

DanielSchiavini:
- inline `createQueryHook` function, it doesn't deserve its own file
- make `getParamsFromQueryKey` an arrow function
- add `gcTime` option to `queryFactory`


pilotvietnam (dismissed):

packages/curve-ui-kit/src/lib/model/query/factory.ts
```diff
@@ -75,7 +65,7 @@ export function queryFactory<
     prefetchQuery: (params, staleTime = 0) =>
       queryClient.prefetchQuery({ queryKey: queryKey(params), queryFn, staleTime }),
     fetchQuery: (params, options) => queryClient.fetchQuery({ ...getQueryOptions(params), ...options }),
-    useQuery: createQueryHook(getQueryOptions),
+    useQuery: (params, condition) => useQuery(getQueryOptions(params, condition)),
```
0xAlunara, line 68: Excuse me for adding a commit to your PR, but imo TanStack shines when you inline stuff. The moment you are creating wrappers you're stuck in type-and-generic hell. We can get rid of `createQueryHook` in its entirety, together with all of the complex typing, by just directly inline here. This was literally the only usage of that function anyway.

packages/curve-ui-kit/src/lib/model/query/factory.ts
```diff
@@ -1,21 +1,13 @@
...
-export function getParamsFromQueryKey<TKey extends readonly unknown[], TParams, TQuery, TField>(
-  queryKey: TKey,
-  assertValidity: (data: TParams, fields?: TField[]) => TQuery,
-) {
-  const queryParams = Object.fromEntries(
-    queryKey.flatMap((i) => (i && typeof i === 'object' ? Object.entries(i) : [])),
-  ) as TParams
-  return assertValidity(queryParams)
-}
+const getParamsFromQueryKey = <TKey extends readonly unknown[], TParams, TQuery, TField>(queryKey: TKey) =>
```
0xAlunara: Doesn't have to be exported as there's only one usage (down below), so I removed it

packages/curve-ui-kit/src/lib/model/query/factory.ts
```diff
@@ -1,21 +1,13 @@
...
-export function getParamsFromQueryKey<TKey extends readonly unknown[], TParams, TQuery, TField>(
-  queryKey: TKey,
-  assertValidity: (data: TParams, fields?: TField[]) => TQuery,
-) {
-  const queryParams = Object.fromEntries(
-    queryKey.flatMap((i) => (i && typeof i === 'object' ? Object.entries(i) : [])),
-  ) as TParams
-  return assertValidity(queryParams)
-}
+const getParamsFromQueryKey = <TKey extends readonly unknown[], TParams>(queryKey: TKey) =>
```
0xAlunara, line 9: I was also able to remove some unused generic types

packages/curve-ui-kit/src/lib/model/query/factory.ts
```diff
@@ -65,7 +54,8 @@ export function queryFactory<
     })
 
   return {
-    assertValidity,
+    assertValidity: (data: TParams, fields?: TField[]) =>
```
0xAlunara, line 57: A query's `assertValidity` is now literally never used. Only in that broken ass gauge reward feature that was barely completed is `assertValidity` used, and i nthat case it's the global version from `import { assertValidity } from '@ui-kit/lib/validation'` where you pass the validation suite as the parameter

Imo we can remove this exported prop entirely.

packages/curve-ui-kit/src/lib/model/query/factory.ts
```diff
@@ -49,14 +50,14 @@ export function queryFactory<
 
   const queryFn = ({ queryKey }: QueryFunctionContext<TKey>) => {
     logQuery(queryKey)
-    const params = getParamsFromQueryKey(queryKey, assertValidity)
-    return runQuery(params)
+    return runQuery(getParamsFromQueryKey(queryKey))
```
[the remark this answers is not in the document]
0xAlunara: That issue's been a pain in my butt quite a few times. One example I remember was that I wanted to validate a llama mint market wasn't null, but then it wanted me to have the entire llama mint market also as a param. Yesh, no bueno. Not sure if I gave up or came up with a hacky way around it.

packages/curve-ui-kit/src/lib/model/query/factory.ts
DanielSchiavini: Good catch, thanks, that did not make any sense anymore after this PR indeed

packages/curve-ui-kit/src/lib/model/query/factory.ts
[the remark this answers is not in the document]
DanielSchiavini: For queries you cannot have complex keys because they gotta be compared and cached.
But yeah hmm you could validate them :thinking:

PR #1332, opened 2025-09-03
fix: formatting round two

0xAlunara:
* Remove some dead code (two cell components)
* Remove a bunch of duplicate unit tests
* Remove `showAllFractionDigits`. Primarily only used in the DAO anyway, just set decimals to 5 and call it a day.
* Remove `showDecimalIfSmallNumberOnly`. In graph ticks we can set the decimals property conditionally to 0 if it's bigger than 10
* Allow all of Intl.NumberFormatOptions to be used as part of our custom number formatter options
* When formatted output would show "0.00", "1.00", or just "0"/"1" but value has more precision, switches to 6 significant digits
* True integer numbers will have no decimals, unless trailingZeroDisplay is set to 'auto'


pilotvietnam (approved):

mimaklas (approved):

apps/main/src/dao/components/PageAnalytics/VeCrvFeesChart/FeesBarChart.tsx
```diff
@@ -36,7 +36,7 @@ const FeesBarChart = ({ data, height = 500 }: FeesBarChartProps) => (
         tick={{ fill: 'var(--page--text-color)', fontSize: 'var(--font-size-1)' }}
         tickLine={{ opacity: 0.3, strokeWidth: 0.3 }}
         axisLine={{ opacity: 0.3, strokeWidth: 0.3 }}
-        tickFormatter={(value) => `${formatNumber(value, { showDecimalIfSmallNumberOnly: true })}`}
+        tickFormatter={(value) => `${formatNumber(value, { ...(value > 10 && { decimals: 0 }) })}`}
```
DanielSchiavini, line 39: The `>10` is quite arbitrary and repetitive.
Why do we need it? What's special about these numbers?
 Can't we just abstract that away

0xAlunara: Hmm, perhaps we can just set decimals to 0 regardless of the size of value. This is btw how the `showDecimalIfSmallNumberOnly` property worked under the hood. But with proper abbreviations I think thos logic is no longer needed

PR #1339, opened 2025-09-04
refactor: move manage-soft-liquidation feature to llamalend

0xAlunara:
Not being used yet in prod, so an easy move


packages/curve-ui-kit/.storybook/main.ts
```diff
@@ -11,7 +11,11 @@ function getAbsolutePath(packageName: string) {
 }
 
 const config: StorybookConfig = {
-  stories: ['../src/**/*.mdx', '../src/**/*.stories.@(js|jsx|mjs|mts|ts|tsx)'],
+  stories: [
+    '../src/**/*.mdx',
+    '../src/**/*.stories.@(js|jsx|mjs|mts|ts|tsx)',
+    '../../../apps/**/*.stories.@(js|jsx|mjs|mts|ts|tsx)',
```
DanielSchiavini, line 17: bit weird to import from apps, tbh storybook should be in `/apps` imo
good solution for the time being tho

PR #1344, opened 2025-09-05
refactor: share liquidation range chart

DanielSchiavini:
- share the implementation of the liquidation chart between mint and lend markets, moving it to the llamalend app


DanielSchiavini:
Duplicate of #1350 :see_no_evil:

PR #1347, opened 2025-09-06
feat: replace subtabs by mui component

0xAlunara:
TODO: need to check disabled states of each tab, and also the initial tab on loan. I might have accidentally removed required "loanExists" usages for the disable checks.


apps/main/src/dao/components/PageAnalytics/index.tsx
```diff
@@ -35,12 +34,14 @@ const Analytics = () => {
...
-          <SubNav
-            activeKey={navSelection}
-            navItems={navItems}
-            setNavChange={setNavSelection as (key: string) => void}
-            nested
+          <TabsSwitcher
+            variant="contained"
+            size="medium"
+            value={navSelection}
+            onChange={setNavSelection as (key: string) => void}
```
DanielSchiavini, line 41: TabSwitcher is generic so this shouldn't be needed

```suggestion
            onChange={setNavSelection}
```

apps/main/src/dex/components/PagePool/index.tsx
```diff
@@ -191,11 +190,20 @@ const Transfer = (pageTransferProps: PageTransferProps) => {
...
+      { value: 'withdraw', label: t`Withdraw` },
+      { value: 'swap', label: t`Swap` },
+    ]
+
+    if (isAvailableManageGauge) {
+      tabs.push({ value: 'manage-gauge', label: t`Gauge` })
+    }
+
+    return tabs
+  }, [isAvailableManageGauge])
```
DanielSchiavini, line 206:
```suggestion
  const ACTION_TABS: Tab[] = useMemo((): Tab[] => [
    { value: 'deposit', label: t`Deposit` },
    { value: 'withdraw', label: t`Withdraw` },
    { value: 'swap', label: t`Swap` },
    ...isAvailableManageGauge ? [{ value: 'manage-gauge', label: t`Gauge` }] : [],
  ], [isAvailableManageGauge])
```

PR #1351, opened 2025-09-09
feat: liquidation range chart legend

DanielSchiavini:
- Add legend to the liquidation range chart of #1315
- Customize colors according to MUI


apps/main/src/llamalend/widgets/ChartLiquidationRange.tsx
```diff
@@ -321,3 +352,54 @@ const TipIcon = styled(Icon)`
   position: relative;
   left: -2px;
 `
+
+const LegendContainer = styled.div`
```
0xAlunara, line 356: Oh no, new styled components!

0xAlunara (approved):

PR #1373, opened 2025-09-12
fix: unable to get wallet balances

0xAlunara:
By the light this shit is nasty.

So the problem is, when you visit the website for the first time the CurveJS library gets hydrated without an RPC. But to fetch user walles balances, you need a CurveJS instance *with* an RPC. After a short time, a second hydration occurs *with* an RPC. However, there's a brief moment in time where you have a CurveJS instance *with* RPC, but it's not yet hydrated. Only the old `NoRPC` version is hydrated. There's no easy way to detect if the new RPC'd version of CurveJS is hydrated, so out of caution if you want to call `fetchUserPoolInfo` you have to make absolutely sure the pool we're fetching data from is hydrated, hence the preceding calls of `fetchUserPoolList`.

Ugh, this one ugly fix. Can't wait until we get rid of this whole library and can rely on just TanStack queries, which would be a proper fix.

EDIT: it's still not 100% fixed but it's better. Burn this shit with fire ugh

If you go to a pool page in prod and refresh with F5 you can reproduce it
image

### DIFF
--- a/apps/main/src/dex/store/createUserSlice.ts
+++ b/apps/main/src/dex/store/createUserSlice.ts
@@ -151,6 +151,7 @@ const createUserSlice = (set: SetState<State>, get: GetState<State>): UserSlice 
         return fetchedWalletBalances
       } catch (error) {
         get()[sliceKey].setStateByKey('error', 'error-get-pool-wallet-balances')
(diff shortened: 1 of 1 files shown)


0xAlunara:
I cannot reproduce this anymore, so probably fixed with the updates to the hydration logic that have been made over time

PR #1391, opened 2025-09-16
refactor: remove hacky filler box in sidebar footer

0xAlunara:
Having an empty filler box in the mobile footer component felt super hacky and a source of maintenance hell, so figured we could get away with it being absolutely positioned. Basically it sits at the bottom of the drawer content, always visible as the sections above are now just part of a scrollable area.

### DIFF
--- a/packages/curve-ui-kit/src/widgets/Header/MobileHeader.tsx
+++ b/packages/curve-ui-kit/src/widgets/Header/MobileHeader.tsx
@@ -1,6 +1,5 @@
 import { useCallback, useEffect, useMemo, useState } from 'react'
 import AppBar from '@mui/material/AppBar'
-import Box from '@mui/material/Box'
 import Drawer from '@mui/material/Drawer'
 import Stack from '@mui/material/Stack'
 import Toolbar from '@mui/material/Toolbar'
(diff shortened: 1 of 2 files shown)


packages/curve-ui-kit/src/widgets/Header/MobileHeader.tsx
```diff
@@ -83,13 +82,22 @@ export const MobileHeader = ({
...
+              },
+            },
+          }}
           sx={{ top }}
           variant="temporary"
           hideBackdrop
           data-testid="mobile-drawer"
         >
-          <Box>
+          <Stack sx={{ overflowY: 'auto', ...HIDE_SCROLLBAR }}>
```
DanielSchiavini, line 100: The paper component has already overflow, and so does the drawer. Why do we need another one?

DanielSchiavini (approved):
Awesome to get rid of this

packages/curve-ui-kit/src/widgets/Header/MobileHeader.tsx
0xAlunara: Yeah, I removed it initially at first, but notice how the footer sits outside of this stack. That footer must not be in the scrollable area.

PR #1392, opened 2025-09-16
feat: legacy mode

DanielSchiavini:
- add legacy mode switch
- for now, I didn't remove the beta flag from the user profile menu and the llama click
- depends on https://github.com/curvefi/curve-frontend/pull/1387

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -83,6 +83,12 @@ jobs:
           CYPRESS_TENDERLY_PROJECT_SLUG: ${{ vars.TENDERLY_PROJECT_SLUG }}
           CYPRESS_TENDERLY_ACCESS_KEY: ${{ secrets.TENDERLY_ACCESS_KEY }}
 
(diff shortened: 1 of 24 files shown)


0xAlunara (commented):
Can I opt for putting the new user profile menu out of beta in this PR? To enable legacy mode you must first enable beta mode through the llama icon at the bottom; nobody will be able to find this 😄 

Then, if you enable legacy mode the user profile is gone so you need to enable beta mode again to go back to stable 🥴 
(Although you can disable it through the global banner as well I guess)

BTW: the toast notifications don't seem to work when toggling modes through the user profile menu 🤔

DanielSchiavini:
> Can I opt for putting the new user profile menu out of beta in this PR?

We'll do that when #1315 goes out of beta, too

> BTW: the toast notifications don't seem to work when toggling modes through the user profile menu 🤔

I'll have a look

0xAlunara (approved):

PR #1400, opened 2025-09-18
feat: market id validation

OnlyJousting:
Changes:
- Add marketId validation to relevant queries
- Rename scrvusd queries to match query naming convention


DanielSchiavini (dismissed):
It would be nice to get rid of the '' defaults

DanielSchiavini (approved):

PR #1401, opened 2025-09-18
refactor: remove unused chain parameter from getTokens

0xAlunara:
It's not part of the market object, so you're passing in data you already had to fetch elsewhere and is readily available.

Real reason is I want to use this function too without having to pass a chain name for no reason 😄 
Want to make it a pure `LlamaMarketTemplate` function

### DIFF
--- a/apps/main/src/llamalend/features/borrow/useBorrowForm.tsx
+++ b/apps/main/src/llamalend/features/borrow/useBorrowForm.tsx
@@ -63,7 +63,7 @@ export function useBorrowForm<ChainId extends IChainId>({
     reset: resetCreation,
   } = useCreateLoanMutation({ chainId, poolId: market?.id, reset: form.reset })
 
-  const { borrowToken, collateralToken } = useMemo(() => market && getTokens(market, chain), [market, chain]) ?? {}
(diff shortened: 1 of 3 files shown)


pilotvietnam (approved):

apps/main/src/llamalend/features/borrow/useBorrowForm.tsx
```diff
@@ -63,7 +63,18 @@ export function useBorrowForm<ChainId extends IChainId>({
...
   } = useCreateLoanMutation({ chainId, poolId: market?.id, reset: form.reset })
 
-  const { borrowToken, collateralToken } = useMemo(() => market && getTokens(market, chain), [market, chain]) ?? {}
+  const { borrowToken, collateralToken } =
+    useMemo(() => {
+      const tokens = market && getTokens(market)
+
+      return (
+        tokens && {
+          borrowToken: { ...tokens.borrowToken, chain },
```
DanielSchiavini, line 72: Should we consider the possibility of tokens in different chains?

0xAlunara: Yeah, sometimes tokens have the same address across chains

PR #1404, opened 2025-09-18
refactor: remove old markets lists

0xAlunara:


pilotvietnam (dismissed):

0xtutti (approved):

tests/cypress/e2e/lend/basic.cy.ts
```diff
@@ -3,13 +3,12 @@ import { LOAD_TIMEOUT } from '@cy/support/ui'
...
   it('should open the Lend DApp successfully', () => {
     cy.visit('/lend/')
-    cy.title(LOAD_TIMEOUT).should('include', 'Lend')
+    cy.url(LOAD_TIMEOUT).should('match', /http:\/\/localhost:\d+\/llamalend\/ethereum\/markets\/?$/)
   })
 
   it('should redirect from the old root URL successfully', () => {
     cy.visit('/lend/#/ethereum')
-    cy.url(LOAD_TIMEOUT).should('match', /http:\/\/localhost:\d+\/lend\/ethereum\/markets\/?$/)
-    cy.title().should('equal', 'Markets - Curve Lend')
```
DanielSchiavini, line 12: I think it's nice to test the title, why was it removed?

DanielSchiavini (changes requested):
`CRVUSD_ROUTES.PAGE_MARKETS` and `LEND_ROUTES.PAGE_MARKETS` are still being used and will result in a 404 now. We should update the existing code to use `LLAMALEND_ROUTES` but we also need to set up redirects to the new page. We cannot have all the existing links break just like that.

DanielSchiavini (approved):

DanielSchiavini (approved):

tests/cypress/e2e/lend/basic.cy.ts
0xAlunara: Didn't think much of it tbh 🤔

PR #1408, opened 2025-09-20
fix: leverage input grouping lend

0xAlunara:
I know not too important but it bothered me. It's only temporary as cards get replaced in the future.
Gets rid of the stupid double borders

New
image
image

Old
image


0xAlunara:
Added the same for leverage subtab

New
image
image

Old
image

DanielSchiavini (approved):

0xtutti (approved):

DanielSchiavini (approved):

0xtutti (approved):

PR #1409, opened 2025-09-20
fix: remove lend position details rates data api fallback

OnlyJousting:
Falling back on the API data is redundant and prone to confusion on the market page when a user has a wallet connected and is viewing a position.

Changes:
- Removes lend position details rates data api fallback

### DIFF
--- a/apps/main/src/lend/hooks/useBorrowPositionDetails.ts
+++ b/apps/main/src/lend/hooks/useBorrowPositionDetails.ts
@@ -44,7 +44,6 @@ export const useBorrowPositionDetails = ({
     status,
     state: { collateral, borrowed, debt } = {},
   } = userLoanDetails ?? {}
-  const marketRate = useStore((state) => state.markets.ratesMapper[chainId]?.[marketId])
   const prices = useStore((state) => state.markets.pricesMapper[chainId]?.[marketId])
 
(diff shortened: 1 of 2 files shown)


apps/main/src/lend/hooks/useBorrowPositionDetails.ts
```diff
@@ -44,7 +44,6 @@ export const useBorrowPositionDetails = ({
     status,
     state: { collateral, borrowed, debt } = {},
   } = userLoanDetails ?? {}
-  const marketRate = useStore((state) => state.markets.ratesMapper[chainId]?.[marketId])
   const prices = useStore((state) => state.markets.pricesMapper[chainId]?.[marketId])
```
0xAlunara, line 47: Should we also remove reliance on `pricesMapper`? That would elimate all Zustand usages here

DanielSchiavini (commented):
What is this actually fixing?

apps/main/src/lend/hooks/useBorrowPositionDetails.ts
OnlyJousting: yea why not, just need to replace it with a query

0xtutti (approved):

PR #1421, opened 2025-09-23
refactor: rename useOneWayMarket to useLendMarket and simplify hooks

0xAlunara:
Preparation work for future `useMintMarket` hook, which will replace the mint market collateral Zustand store which in turn is needed to add support for unified support for addresses in URL.

* Rename 'useOneWayMarket' stuff to 'useLendMarket'
* Simplify the mapping creation, add some docs as well
* Remove unnecessary usage of `hydratedChainId`. Just fetch & store in Tanstack cache.
* `useLendMarkets` no longer just returns a string array of market ids, it now also filters out bad markets and converts it to proper lend market instances, given in all cases we want to access the market data anyway.


apps/main/src/lend/entities/chain/chain-tvl.ts
```diff
@@ -3,20 +3,16 @@ import useStore from '@/lend/store/useStore'
...
   const marketsCollateralMapper = useStore((state) => chainId && state.markets.statsAmmBalancesMapper[chainId])
   const marketsTotalSupplyMapper = useStore((state) => chainId && state.markets.totalLiquidityMapper[chainId])
   const marketsTotalDebtMapper = useStore((state) => chainId && state.markets.statsTotalsMapper[chainId])
   const tokenAddresses = useMemo(
-    () =>
-      Object.values(marketMapping ?? {})
-        // note: include the borrowed tokens here, to be used in `filterSmallMarkets`
-        .flatMap((market) => [market.borrowed_token, market.collateral_token])
-        .map((t) => t.address),
+    () => marketMapping.flatMap((market) => [market.borrowed_token, market.collateral_token]).map((t) => t.address),
```
DanielSchiavini, line 15: the comment is deleted

apps/main/src/lend/entities/lend-markets.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Hook to get a specific lending market by its id or controller address.
+ * @param chainId The chain for which to get the lending market of.
+ * @param marketId Lend market id or controller address
+ * @returns The market instance, if found.
+ */
+export const useLendMarket = ({ chainId, marketId }: { chainId: ChainId; marketId: string }) => {
+  const { data: markets, isSuccess, error } = useLendMarkets({ chainId })
+
+  /** Create mappings from market name or controller id to market instance. */
+  const mapping = useMemo(
```
DanielSchiavini, line 34: this mapping is never reused, might as well use find

PR #1423, opened 2025-09-24
refactor: remove collateral zustand

0xAlunara:
Focus is on removing stores in favor of ~~tanstack queries &~~ hooks, not whether legacy component make sense. In a few cases I've renamed `llama` or `llamma` to `market` for simplicity.

EDIT: this stuff is all sync so doesn't really need tanstack at all!


apps/main/src/loan/components/PageLoanCreate/Page.tsx
```diff
@@ -94,49 +94,41 @@ const Page = () => {
   )
 
   useEffect(() => {
-    if (pageLoaded && curve) {
-      if (llamma) {
-        resetUserDetailsState(llamma)
-        fetchInitial(curve, isLeverage, llamma)
-        void fetchLoanDetails(curve, llamma)
-        setLoaded(true)
-      } else if (collateralDatasMapper) {
```
0xAlunara, line 103: I accidentally removed the redirect if market does not exist / could not be found. Do I really need to bring that back? We don't do this for lending markets either fwiw...

I personally dislike redirects

apps/main/src/loan/components/PageLoanCreate/Page.tsx
```diff
@@ -151,7 +143,7 @@ const Page = () => {
 
   const TitleComp = () => (
     <AppPageFormTitleWrapper>
-      <Title>{displayName || getTokenName(llamma).collateral}</Title>
+      <Title>{market?.id === 'sfrxeth2' ? 'sfrxETH v2' : getTokenName(market).collateral}</Title>
```
0xAlunara, line 146: This logic came from the old collateral store, this was the only exception. Couldn't be bothered to make a helper function just for this.

apps/main/src/loan/components/PageLoanCreate/Page.tsx
```diff
@@ -184,8 +176,8 @@ const Page = () => {
               isReady={isReady}
               isLeverage={isLeverage}
               loanExists={loanExists}
-              llamma={llamma}
-              llammaId={llammaId}
+              llamma={market ?? null}
```
0xAlunara, line 179: Markets can be undefined if they're not yet loaded. The old components however rely on `null` for whatever reason. Didn't want to touch those to keep the PR small.

apps/main/src/loan/components/PageLoanCreate/Page.tsx
DanielSchiavini: This case is also handled here more generically
https://github.com/curvefi/curve-frontend/blob/75c6ce5eab9f9c2d5cce023f952b740d046df25f/apps/main/src/llamalend/entities/llama-markets.ts#L256-L258

apps/main/src/loan/components/PageLoanCreate/Page.tsx
DanielSchiavini: Either we add redirects or we show a proper 404 error. Currently it just loads forever

apps/main/src/loan/components/PageLoanCreate/Page.tsx
DanielSchiavini: I feel like we should add ` vN` for markets ending with a nr

apps/main/src/loan/components/PageLoanCreate/Page.tsx
```diff
@@ -202,13 +194,15 @@ const Page = () => {
...
           <Stack>
             <MarketDetails {...marketDetails} />
-            <MarketInformationComp
-              llamma={llamma}
-              llammaId={llammaId}
-              chainId={rChainId}
-              chartExpanded={chartExpanded}
-              page="create"
-            />
+            {market && (
```
DanielSchiavini, line 197: Why is this added? It results in the page jumping during load

apps/main/src/loan/entities/mint-markets.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Hook to get a specific mint market by its id or controller address.
+ * @param chainId The chain for which to get the mint market of.
+ * @param marketId Mint market id or controller address
+ * @returns The market instance, if found.
+ */
+export const useMintMarket = ({ chainId, marketId }: { chainId: ChainId; marketId: string }) => {
+  const { data: markets, isSuccess, error } = useMintMarkets({ chainId })
+
+  /** Create mappings from market name or controller id to market instance. */
+  const mapping = useMemo(
```
DanielSchiavini, line 28: I think I wrote this, but it doesn't make sense to create a mapping, it doesn't get shared with other calls.
We could just use `find`

apps/main/src/loan/components/PageLoanCreate/Page.tsx
0xAlunara: For some reason the old code type asserted that llamma couldn't be `undefined`. But I'm not sure if that assertion is valid. How can it be asserted to not be null when stuff is being loaded? Either the assertion is somehow true and this won't jump, or stuff was being shown prematurely in the old verison. I'd rather keep this check to make sure there's no errors. If people never complained about the page jumping it's probably a safe assumption?

apps/main/src/loan/components/PageLoanCreate/Page.tsx
0xAlunara: I agree, but the name generation is a bit too internal in llama-markets to re-use outside. I gave it a quick try but quickly ended up having refactor even more stuff, out of scope. Added a TODO for generalization later, not the point of this PR anyway.

0xAlunara: We don't even have this for lending markets and its only for the title when expanding charts 😭 
Super not important

apps/main/src/loan/components/PageLoanCreate/Page.tsx
DanielSchiavini: People dont complain because it works, it just looks unprofessional

0xtutti (approved):

apps/main/src/loan/components/PageLoanCreate/Page.tsx
```diff
@@ -119,47 +120,44 @@ const Page = () => {
...
-        fetchInitial(curve, isLeverage, llamma)
-        void fetchLoanDetails(curve, llamma)
+      if (market) {
+        resetUserDetailsState(market)
+        fetchInitial(curve, isLeverage, market)
+        void fetchLoanDetails(curve, market)
         setLoaded(true)
-      } else if (collateralDatasMapper) {
-        console.warn(
-          `Collateral ${rCollateralId} not found for chain ${rChainId}. Redirecting to market list.`,
```
0xAlunara: I made a choice here to remove the redirect, given that `api.getMintMarket` will throw an error if the market could not be found. This means the user now gets an error shown via the error boundary. I think explicitly failing like this is better than silently redirecting without telling the user what just happened. (And, I'm not sure how to 'catch' errors thrown in hooks 🙈 )

apps/main/src/loan/hooks/useLoanAppStats.tsx
```diff
@@ -8,9 +8,6 @@ import type { ChainId } from '@/loan/types/loan.types'
 import { formatNumber } from '@ui/utils'
 import { t } from '@ui-kit/lib/i18n'
 
-const hasKeys = <K extends keyof any, V>(obj: Record<K, V> | undefined | null): obj is Record<K, V> =>
```
0xAlunara: Unrelated to the PR but I did it in an older commit anyway and decided to keep it.

apps/main/src/loan/entities/mint-markets.ts
```diff
@@ -0,0 +1,39 @@
...
+import { ChainId } from '@/loan/types/loan.types'
+import { fromEntries } from '@curvefi/prices-api/objects.util'
+import { useConnection } from '@ui-kit/features/connect-wallet'
+import type { Address } from '@ui-kit/utils'
+
+/**
+ * Hook to get a mapping of mint market controller address to market name for all mint markets on a specific chain.
+ * Primarily useful fetching mint markets via URL.
+ */
+export const useMintMarketMapping = ({ chainId }: { chainId: ChainId | undefined }) => {
```
0xAlunara: Note how all these things are not async, hence not bothering with Tanstack. I'm afraid it won't be this easy for `lend` markets.

apps/main/src/loan/store/createAppSlice.ts
```diff
@@ -59,9 +59,8 @@ const createAppSlice = (set: StoreApi<State>['setState'], get: StoreApi<State>['
       loans.setStateByKey('userDetailsMapper', {})
     }
 
-    // Check if curveApi is actually a Curve instance and not a LendingApi
-    const { collateralDatas } = await collaterals.fetchCollaterals(curveApi)
-    await loans.fetchLoansDetails(curveApi, collateralDatas)
+    const markets = curveApi.mintMarkets.getMarketList().map((name) => curveApi.getMintMarket(name))
```
0xAlunara: could probably inline but meh, readability

apps/main/src/loan/components/PageLoanCreate/Page.tsx
```diff
@@ -54,11 +55,11 @@ const Page = () => {
   const { address } = useAccount()
   const [loaded, setLoaded] = useState(false)
 
-  const collateralDatasMapper = useStore((state) => state.collaterals.collateralDatasMapper[rChainId])
-  const { llamma, llamma: { id: llammaId = '' } = {}, displayName } = collateralDatasMapper?.[rCollateralId] ?? {}
+  const market = useMintMarket({ chainId: rChainId, marketId: rCollateralId })
+  const marketId = market?.id ?? ''
```
DanielSchiavini: The queries should all support undefined, '' is just cheating the type system

```suggestion
  const marketId = market?.id
```

apps/main/src/loan/components/PageLoanCreate/Page.tsx
DanielSchiavini: This breaks the chain switcher, if you try to switch from any market page it will go to a non-existing market

apps/main/src/loan/components/PageLoanManage/Page.tsx
```diff
@@ -67,13 +63,13 @@ const Page = () => {
...
+  const isReady = !!curve?.signerAddress && !!market
 
-  const marketDetails = useMarketDetails({ chainId: rChainId, llamma, llammaId })
+  const marketDetails = useMarketDetails({ chainId: rChainId, llamma: market, llammaId: marketId })
   const positionDetails = useLoanPositionDetails({
     chainId: rChainId,
-    llamma,
-    llammaId,
+    llamma: market,
+    llammaId: marketId,
```
DanielSchiavini: we should decide how to name these parameters, you are refactoring to marketId which is fine, but we also use poolId, llammaId and whatever else

DanielSchiavini (dismissed):
We should add a test changing markets from the detail page, to make sure it works. The loading forever issue is probably due to using `loading = market==null`

0xAlunara:
Test is not possible because the loan app is only visible with a connected wallet, which is not something we can do yet in E2E tests. And a component test does not make sense; testing the redirect on root level when the entire page is loaded and hdyrated can hardly be called a component test

I've created a Notion ticket to see if we can re-use the wagmi test connector for e2e tests. There's also two tickets to get rid of the wallet dependency on both loan and lend apps. Finally there's also two tickets for the ubiquitous language regarding markets / llamma / pools, and one that investigates if `llammaId` props are still needed for components that already have a `llamma` prop (which basically contains `llamma.id`)

PR #1433, opened 2025-09-24
fix: do not show NaN balance

DanielSchiavini:
Due to [this line](https://github.com/curvefi/curve-frontend/blob/9593dec9440d94fc8352a8da2d23e3384315f1f3/apps/main/src/dex/store/createPoolsSlice.ts#L364) the balance can be `NaN` during loading. Show the default value instead.

### DIFF
--- a/apps/main/src/dex/components/PagePool/PoolDetails/CurrencyReserves/CurrencyReservesContent.tsx
+++ b/apps/main/src/dex/components/PagePool/PoolDetails/CurrencyReserves/CurrencyReservesContent.tsx
@@ -64,7 +64,7 @@ const CurrencyReservesContent = ({
 
     <Box className={'right'} flex flexDirection="column">
       <Chip size="md" isBold>
-        {!cr || isNaN(cr.balance) ? '-' : formatNumber(cr.balance, { defaultValue: '-' })}{' '}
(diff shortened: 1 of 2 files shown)


pilotvietnam (approved):

apps/main/src/dex/components/PagePool/PoolDetails/CurrencyReserves/CurrencyReservesContent.tsx
```diff
@@ -64,10 +64,10 @@ const CurrencyReservesContent = ({
 
     <Box className={'right'} flex flexDirection="column">
       <Chip size="md" isBold>
-        {formatNumber(cr?.balance, { defaultValue: '-' })}{' '}
+        {!cr || isNaN(cr.balance) ? '-' : formatNumber(cr.balance, { defaultValue: '-' })}{' '}
```
0xAlunara, line 67: Should we not just also show the `defaultValue` as configured when the number is `NaN`?

Currently the code only works with `null`: `if (val == null) return options?.defaultValue ?? '-'`

DanielSchiavini: image

PR #1434, opened 2025-09-24
refactor: lend markets query refactoring part 2

0xAlunara:
Preparation work for future `useMintMarket` hook, which will replace the mint market collateral Zustand store which in turn is needed to add support for unified support for addresses in URL.

* Rename 'useOneWayMarket' stuff to 'useLendMarket'
* Simplify the mapping creation, add some docs as well
* Remove unnecessary usage of `hydratedChainId`.
* `useLendMarkets` is no more. There's only a `useLendMarketMapping`, which maps lend market controller addresses to the lend market ids. This hook is used internally for `useLendMarket`, and also `useTvl`. It is knows that `useTvl` could use prices API instead, but that's out of the scope of this ticket and I want to keep it small; that's for later.
* I've tried solving some of the routing issues where if you click "Supply" it routs the URL to the version with the lend market. This, however, opened a can of worms outside of the scope of this ticket. It seems to work even though it's inconsistent. I'm keeping that as it for for later to deal with


apps/main/src/lend/components/PageLoanCreate/Page.tsx
```diff
@@ -61,11 +61,11 @@ const Page = () => {
   })
 
   useEffect(() => {
-    if (isSuccess && !market) {
+    if (!market) {
```
0xAlunara, line 64: Hmm, I wonder if we should do this? Technically there's no `loading` state anymore right? Or is this page still shown when not yet hydrated? There's either a market or not, there's no in between state anymore.

Regardless I'd rather show either a loading state or error state rather than redirect

pilotvietnam (approved):

apps/main/src/lend/components/PageLoanCreate/Page.tsx
0xAlunara: Okay I need to check if hydrated otherwise it instantly redirects back lol

pilotvietnam (approved):

apps/main/src/lend/entities/lend-markets.ts
```diff
@@ -0,0 +1,63 @@
...
+  const { llamaApi: api } = useConnection()
+  const mapping = useMemo(
+    () =>
+      api?.hydrated &&
+      chainId &&
+      fromEntries(
+        api.lendMarkets
+          .getMarketList()
+          .filter((marketName) => !networks[chainId].hideMarketsInUI[marketName])
+          .map((name) => [api.getLendMarket(name).addresses.controller as Address, name]),
```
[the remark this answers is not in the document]
DanielSchiavini: Why not a query, though? We only need the mapping for redirecting.

0xAlunara:
Might put this PR back to draft until hydration issues are better resolved

PR #1438, opened 2025-09-25
Disables deposits for yieldbasis pools and adds information message

0xtutti:


apps/main/src/dex/hooks/usePoolAlert.tsx
```diff
@@ -134,6 +135,24 @@ const usePoolAlert = (poolData?: PoolData | PoolDataCache) => {
         </MessageWrapper>
       ),
     })
+    const yieldbasisAlert = (): PoolAlert => ({
+      alertType: 'info',
+      isDisableDeposit: true,
+      isInformationOnly: true,
+      isCloseOnTooltipOnly: true,
+      message: (
+        <MessageWrapper>
```
DanielSchiavini, line 144: officially we want to enable translation for these strings, but it seems that's not happening in this file at all

0xtutti: We can keep it as is for now and do it in a future PR?

PR #1441, opened 2025-09-25
fixes uppercase in yb-cbBTC pool for warning 

0xtutti:
An uppercase "A" was missed in the pool address for the yb-cbBTC pool, which prevented it from being properly flagged as withdraw only.


0xAlunara (approved):
Lmao did you convert the casing by hand 🤣

PR #1444, opened 2025-09-25
refactor: reapply campaigns refactor

0xAlunara:
Re-applies #1437 but with 2 fixes:

1. Campaigns from the `useCampaigns` hook are now filtered by a required blockchain id (network)
2. Multipliers aren't guaranteed to be numbers. They can be just tokens too.

When reviewing it's prob easier to ignore `Files Changes` and just review the two additional commits https://github.com/curvefi/curve-frontend/pull/1444/commits/a07b3eb4b61caafc4507b17f65049cf8cbc3fcb0 and https://github.com/curvefi/curve-frontend/pull/1444/commits/f07af677882b87544f177ee59edac4cf7f1b2bcb


0xAlunara:
There's a bug I need to fix first (llamalend table does *not* need filtering ugh), and Vercel is giving me issues wrt deploying

PR #1449, opened 2025-09-26
chore: delete unused code

DanielSchiavini:
- inspect unused components and delete everything that's unused :boom:


apps/main/src/llamalend/queries/user-balances.query.ts
```diff
@@ -1,30 +0,0 @@
-import { getLlamaMarket } from '@/llamalend/llama.utils'
```
0xAlunara, line 1: I actually need this for manage soft liq, but I'll just copy and add it back in that PR

apps/main/src/llamalend/features/market-list/cells/RateCell/RewardsTooltipItems.tsx
```diff
@@ -1,59 +0,0 @@
-import lodash from 'lodash'
```
0xAlunara, line 1: Darnit I refactored this in #1446, could've just deleted it 😭

packages/curve-ui-kit/src/utils/BigDecimal.ts
```diff
@@ -1,311 +0,0 @@
-import { BigDecimalTypes } from './BigDecimal.d'
```
0xAlunara, line 1: Oh finally!

packages/prices-api/src/proposal/util.ts
```diff
@@ -1,30 +0,0 @@
-import type { Proposal, ProposalStatus } from './models'
```
0xAlunara, line 1: Is there a way for your tools to exclude projects? This project is an NPM package, so unused exported functions may still be used by consumers of said package. (I know I use them in curvemonitor 🤓 )

apps/main/src/llamalend/features/market-list/cells/RateCell/RewardsTooltipItems.tsx
DanielSchiavini: :laughing: too bad

packages/curve-ui-kit/src/utils/BigDecimal.ts
DanielSchiavini: lol it wasnt used at all

PR #1457, opened 2025-09-29
feat: migrate DetailInfo to mui

DanielSchiavini:
- replace usages of `DetailInfo` component to use the new `ActionInfo` component in beta mode
- update `ActionInfo` to:
  - support `className` 
  - `ReactNode` as `value` 
  - `loading` as `[number, number]`
- get rid of some unused properties


packages/ui/src/DetailInfo/DetailInfo.tsx
```diff
@@ -1,66 +1,99 @@
...
+        <>
+          {children || '-'} {!!tooltip && tooltip}
+        </>
+      )}
+    </DetailValue>
+  </Wrapper>
+)
+
+const NewDetailInfo = ({
+  className,
```
0xAlunara, line 71: Is this something we want to keep, or just temporary because we're migrating? Imo we should add a todo once this feature goes out of beta so only the advised `sx` becomes available.

packages/curve-ui-kit/src/shared/ui/ActionInfo.tsx
```diff
@@ -98,18 +104,18 @@ const ActionInfo = ({
   loading = false,
   error = false,
   testId = 'action-info',
-  sx,
+  ...styleProps // sx, className
 }: ActionInfoProps) => {
   const [isSnackbarOpen, openSnackbar, closeSnackbar] = useSwitch(false)
-  const copyValue = (givenCopyValue ?? value).trim()
+  const copyValue = (givenCopyValue ?? (typeof value === 'string' ? value : '')).trim()
```
0xAlunara, line 110: Should we really fail silently if value is a `ReactNode`? Perhaps copying should be disabled altogether if it's not a string?

Debatable because you've already used `copy={typeof children === 'string'}` in the `DetailInfo` component, but maybe it's better to put that guard internally into the component so it can't be forgotten.

packages/ui/src/DetailInfo/DetailInfo.tsx
```diff
@@ -1,66 +1,99 @@
...
+}: Props) => (
+  <>
+    {isDivider && <Divider sx={{ marginBlock: Spacing.sm }} />}
+    <ActionInfo
+      className={className}
+      label={label}
+      value={children || '-'}
+      valueColor={VariantToColorMap[variant || '']}
+      valueTooltip={tooltip}
+      copy={typeof children === 'string'}
```
0xAlunara, line 89: I think this is too general, I don't think there's a need to copy these values

image
image

packages/ui/src/DetailInfo/DetailInfo.tsx
DanielSchiavini: It's used here only but I'm not sure how to deal with it :thinking: I don't see matching css rules
https://github.com/curvefi/curve-frontend/blob/3898365f0c760726669e5d4bc9c3d4fc34b84b7a/apps/main/src/dex/components/PagePool/components/DetailInfoEstLpTokens.tsx#L37

packages/ui/src/DetailInfo/DetailInfo.tsx
DanielSchiavini: Just checked, they indeed did not have buttons

packages/ui/src/DetailInfo/DetailInfo.tsx
0xAlunara: Prob can be removed

packages/ui/src/DetailInfo/DetailInfo.tsx
```diff
@@ -1,66 +1,99 @@
...
+      error={variant === 'error'}
+      loading={loading && (loadingSkeleton || true)}
+      copyValue={typeof children === 'string' ? children : ''}
+      size={isBold ? 'large' : 'medium'}
```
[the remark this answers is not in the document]
DanielSchiavini: I think we should keep them bold until we have some feedback from tutti tbh

0xAlunara (approved):

packages/curve-ui-kit/src/shared/ui/WithWrapper.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { ReactNode } from 'react'
```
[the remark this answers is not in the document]
DanielSchiavini: Yeah it was too easy not to do

packages/curve-ui-kit/src/shared/ui/WithTooltip.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
```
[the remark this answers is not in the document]
DanielSchiavini: We usually get warnings in typecheck (and in my IDE) for unused imports, but the `React` global isn't detected :disappointed:

packages/ui/src/DetailInfo/DetailInfo.tsx
DanielSchiavini: Indeed, sorry I'll fix in a follow up PR

packages/curve-ui-kit/src/shared/ui/WithTooltip.tsx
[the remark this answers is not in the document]
DanielSchiavini: https://github.com/curvefi/curve-frontend/pull/1469

PR #1466, opened 2025-10-01
feat: implement outlier-resistant price scaling for candle charts

OnlyJousting:
Candle charts were being flattened when data contained anomalous price points (extreme outliers). These outlier candles forced the Y-axis to accommodate extreme values, compressing the normal price range and making it difficult to read the meaningful price movements.

Changes:
- Implement outlier-resistant price scaling for candle charts

### DIFF
--- a/packages/ui/src/Chart/CandleChart.tsx
+++ b/packages/ui/src/Chart/CandleChart.tsx
@@ -20,7 +20,7 @@ import type {
   LiquidationRanges,
   ChartColors,
 } from './types'
-import { hslaToRgb } from './utils'
+import { hslaToRgb, calculateRobustPriceRange } from './utils'
 
 /**
  * @param totalDecimalPlacesRef - A ref to track the total decimal places for consistent formatting
@@ -108,6 +108,7 @@ const CandleChart = ({
   const volumeSeriesRef = useRef<ISeriesApi<'Histogram'> | null>(null)
   const oraclePriceSeriesRef = useRef<ISeriesApi<'Line'> | null>(null)
   const lastFetchEndTimeRef = useRef(lastFetchEndTime)
+  const ohlcDataRef = useRef(ohlcData)
 
   const isMounted = useRef(true)
   const totalDecimalPlacesRef = useRef(4)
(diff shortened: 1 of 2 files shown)


packages/ui/src/Chart/utils.ts
```diff
@@ -41,3 +41,45 @@ export function hslaToRgb(hsla: string) {
...
+ * const range = calculateRobustPriceRange(prices) // excludes 1000
+ */
+export function calculateRobustPriceRange(
+  prices: number[],
+  lowerPercentile = 0.02,
+  upperPercentile = 0.98,
+  padding = 0.05,
+): { minValue: number; maxValue: number } | null {
+  if (!prices || prices.length === 0) {
+    return null
```
0xAlunara, line 64: Why return `null` and not work with `[1]` in sortedPrices such that it continues as if it's perfectly pegged at 1 and the algorithm works with the configured padding and no branching?

packages/ui/src/Chart/CandleChart.tsx
```diff
@@ -478,18 +525,6 @@ const CandleChart = ({
     })
   }, [liquidationRange])
 
-  useEffect(() => {
-    if (!candlestickSeriesRef.current || !ohlcData) return
-
-    candlestickSeriesRef.current.setData(ohlcData)
-  }, [ohlcData])
-
-  useEffect(() => {
```
0xAlunara, line 487: lol was wondering why you deleted it, but it seems it was there twice 😭

0xAlunara (commented):
Works pretty well. One exception we could make however, is to only look for outliers inside the inner 80% of the candles, or exclude the latest 5 datapoints.

One chart I found is https://curve-dapp-git-feat-handle-anomalous-candles-in-60405e-curvefi.vercel.app/dex/ethereum/pools/factory-crvusd-12/deposit, where the latest candle isn't fully visible even though it's the current valid price. An outlier by technical definition, but pretty important regardless 😅 
image

packages/ui/src/Chart/utils.ts
OnlyJousting: after further consideration this was reverted as it was problematic for volatile pools

PR #1467, opened 2025-10-01
Fix/pools disable colon

0xPearce:
Characters added to the `INVALID_POOLS_NAME_CHARACTERS` array are not allowed in pool names.
If the user inserts an invalid character from this list, an error message is shown and the "Create Pool" button is disabled.

Screenshot 2025-10-01 at 18 35 23


apps/main/src/dex/components/PageCreatePool/PoolInfo/index.tsx
```diff
@@ -10,11 +12,23 @@ const PoolInfo = () => {
   const updatePoolName = useStore((state) => state.createPool.updatePoolName)
   const updatePoolSymbol = useStore((state) => state.createPool.updatePoolSymbol)
 
+  const invalidCharsFound = getInvalidCharactersInString(poolName, INVALID_POOLS_NAME_CHARACTERS)
+  const poolNameError =
+    invalidCharsFound.length > 0
+      ? t`Pool name cannot contain the following character${INVALID_POOLS_NAME_CHARACTERS.length > 1 ? 's' : ''}: (${INVALID_POOLS_NAME_CHARACTERS.join('), (')})`
```
0xPearce, line 18: Can also be shown as ":" instead of (:) let me know what you think

OnlyJousting: I'd prefer ":" personally

OnlyJousting (dismissed):

0xAlunara (dismissed):
Comments are me just nitpicking; feature does what it's supposed to and works good! 👍

apps/main/src/dex/utils/index.ts
```diff
@@ -85,3 +85,7 @@ export const toTokenOption =
     symbol: token.symbol,
     volume: token.volume,
   })
+
+/** Check for invalid characters in string */
+export const getInvalidCharactersInPoolName = (str: string, invalidCharacters: string[]): string[] =>
```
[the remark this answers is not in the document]
0xPearce: I made a function because we reuse it twice (in the slice + PoolInfo)

OnlyJousting (approved):

0xAlunara (approved):

PR #1468, opened 2025-10-01
feat: max ltv column

0xAlunara:
Adds max ltv column from prices api to market list. Off by default unless some UX designer tells me otherwise
Also fixes prices-api package typescript compilation

image


OnlyJousting (approved):

DanielSchiavini:
> Also fixes prices-api package typescript compilation

But it breaks the cypress compilation :sweat_smile: 

image

apps/main/src/llamalend/features/market-list/cells/MaxLtvCell.tsx
```diff
@@ -0,0 +1,17 @@
...
+import type { CellContext } from '@tanstack/react-table'
+
+/**
+ * Looked into expanding the LtvCell component to support the max LTV value,
+ * but that cell uses a user position hook which complicates the whole thing.
+ *
+ * The max LTV value comes from prices API, so figured a smaller separate
+ * cell component was the way to go here.
+ */
+export const MaxLtvCell = ({ getValue }: CellContext<LlamaMarket, number>) => (
```
DanielSchiavini, line 13: This is not specific to LTV at all. It's just a `TypographyCell` or smth
We need to stop creating cells for everything unless necessary

apps/main/src/llamalend/entities/llama-markets.ts
```diff
@@ -38,6 +38,7 @@ export type LlamaMarket = {
   address: Address
   controllerAddress: Address
   assets: Assets
+  maxLtv: number
```
DanielSchiavini, line 41: feels like we're doing something wrong when adding a single attribute takes 56 lines 😵

DanielSchiavini (approved):

apps/main/src/llamalend/entities/llama-markets.ts
0xAlunara: Modifying 4 endpoints, adding the column itself, setting visibility, sorting behavior, a new cell component. I don't think it can be shortened 🤔

apps/main/src/llamalend/features/market-list/cells/MaxLtvCell.tsx
DanielSchiavini: Hmm also something to be said. But if we go that way, such a simple one could just be inlined

DanielSchiavini (approved):

PR #1475, opened 2025-10-03
feat: legal page with terms/conditions, privacy policy and disclaimers (refactored)

0xPearce:
- Refactored the `/disclaimer` page to include three main tabs: `Terms`, `Privacy` and `Risk Disclaimers`
- Refactored the disclaimer tabs into sub tabs
- url search params now are `tab=terms` for the main tabs and `subtab=dex` for the risk disclaimers subtabs (only available when in the `tab=disclaimer`)
- New `/legal` route now includes all main tabs, rerouted `/disclaimer` to the new `/legal` route
- Privacy Notice and Terms & Condition content
- new `widgets/legal` directory with refactored and organized components from old `widgets/Disclaimer` directory in the `@ui-kit/src/widgets`
- tests functions for redirection and default behavior


packages/curve-ui-kit/src/widgets/Footer/Sections.tsx
```diff
@@ -22,15 +22,15 @@ import { AppName } from '@ui-kit/shared/routes'
 
 function getDisclaimersTab(appName: AppName) {
```
DanielSchiavini, line 23: The name isn't very precise.. Maybe something like this is more clear?
```suggestion
function getDisclaimersQueryString(appName: AppName) {
```

packages/curve-ui-kit/src/widgets/Legal/LegalPage.tsx
```diff
@@ -0,0 +1,139 @@
...
+      ? (subtabParam as DisclaimerTab)
+      : DEFAULT_DISCLAIMERS_TABS[currentApp]
+
+  const tabs = useMemo(
+    () => [
+      ...TABS.map(({ value, ...props }) => ({
+        ...props,
+        value,
+        href: { query: { tab: value }, pathname },
+        onClick: (e: MouseEvent<HTMLAnchorElement>) => pushSearchParams(e, { tab: value }),
```
DanielSchiavini, line 54: I think we don't need this anymore
``` suggestion
```

packages/curve-ui-kit/src/widgets/Legal/LegalPage.tsx
```diff
@@ -0,0 +1,139 @@
...
+    [pathname],
+  )
+
+  const disclaimerTabs = useMemo(
+    () => [
+      ...DISCLAIMER_TABS.map(({ value, ...props }) => ({
+        ...props,
+        value,
+        href: { query: { tab: 'disclaimers', subtab: value }, pathname },
+        onClick: (e: MouseEvent<HTMLAnchorElement>) => pushSearchParams(e, { subtab: value }),
```
DanielSchiavini, line 66: Same here, it was necessary for nextjs when we used to have it
```suggestion
```

packages/curve-ui-kit/src/widgets/Legal/components/general/LastUpdated.tsx
```diff
@@ -5,6 +5,6 @@ import { t } from '@ui-kit/lib/i18n'
 export const LastUpdated = () => (
   <Stack justifyContent="space-evenly" alignItems="end">
     <Typography variant="bodyXsRegular" color="textTertiary">{t`Last Updated`}</Typography>
-    <Typography variant="highlightM">12.08.2024</Typography>
+    <Typography variant="highlightM">06.10.2025</Typography>
```
DanielSchiavini, line 8: I'm not sure if we should mark the risk disclaimer as updated on the same date

packages/curve-ui-kit/src/widgets/Legal/components/tabs/Terms.tsx
```diff
@@ -0,0 +1,338 @@
...
+import { pushSearchParams } from '@ui-kit/utils/urls'
+import { List } from '../general/List'
+import { Section, Header, Title, Paragraph, Bold, SubTitle } from '../general/Section'
+
+export const Terms = ({ currentApp, network }: { currentApp: AppName; network: CurveNetworkId | LlamaNetworkId }) => (
+  <>
+    <Header>{t`Preamble`}</Header>
+    <Section>
+      <Paragraph>
+        {t`These terms and conditions (`}
```
DanielSchiavini, line 16: FYI it's not really important rn as our translation is turned off, but usually it's better to use the [`Trans` placeholder component](https://github.com/curvefi/curve-frontend/blob/de2dace158df548344ccf78e820d76c9b884d958/packages/curve-ui-kit/src/lib/i18n.ts#L16) instead of breaking a sentence into multiple `t` calls

tests/cypress/e2e/all/legal.cy.ts
```diff
@@ -0,0 +1,71 @@
...
+      cy.visit(`/${oneAppPath() || 'dex'}/ethereum/legal`, {
+        onBeforeLoad: (win) => win.localStorage.setItem('phishing-warning-dismissed', `"${new Date().toISOString()}"`),
+      })
+
+      // Make sure there's tabs available and click one.
+      cy.get(`[data-testid='legal-page']`, LOAD_TIMEOUT).should('be.visible')
+      const tabSelector = "[data-testid='legal-page'] [role='tablist'] [role='tab']"
+      cy.get(tabSelector).should('have.length', 3)
+
+      // Navigate to the Disclaimer tab and check if the other tabs load. Then click the scrvusd tab
```
DanielSchiavini, line 41: do we need a test for the new tabs?

packages/curve-ui-kit/src/widgets/Footer/Sections.tsx
0xPearce: it's not used anymore, I removed it. Thanks

packages/curve-ui-kit/src/widgets/Legal/LegalPage.tsx
0xPearce: Doesn't work without

tests/cypress/e2e/all/legal.cy.ts
0xPearce: We now have 2 rows of tabs, the second one is rendered only when the `Risk Disclaimer` tab is active. The current tests handle both tabs

OnlyJousting:
SCR-20251007-mckf

Is it right that these links have padding: 0?

PR #1479, opened 2025-10-06
feat: allow llamalend use while connected to any dex network

DanielSchiavini:
- simplify chain switcher so it only uses the network ID, the chain ID and name aren't really necessary
- make sure any unnecessary queries are disabled
  - enhance validation for `useTokenUsdRate` so it doesn't try to load prices from the wrong chain
- depends on ~~#1464~~ #1498


packages/curve-ui-kit/src/features/switch-chain/ui/ChainSwitcher.tsx
```diff
@@ -42,11 +45,10 @@ export const ChainSwitcher = ({ networks, chainId }: ChainSwitcherProps) => {
 
   const onClick = options.length > 1 ? toggle : openSnackbar
   const top = useLayoutStore((state) => state.navHeight)
-  const network: NetworkDef | undefined = networks[chainId] // the network may not be defined when switching apps
   return (
     <>
       <IconButton size="small" onClick={onClick} data-testid="btn-change-chain">
-        {network && <ChainSwitcherIcon network={network} />}
+        {networkId && <ChainSwitcherIcon networkId={networkId} />}
```
0xAlunara, line 51: Is there a case where this is false, now that it's retrieved from the URL which always has a network defined?

apps/main/src/dex/components/PageIntegrations/index.tsx
```diff
@@ -144,14 +144,16 @@ const IntegrationsComp = ({
               integrationsAppNetworks={
                 !rChainId && (
                   <Box margin="0.25rem 0 0 0">
-                    {Object.keys(app.networks).map((networkId) => {
-                      if (networkId in networksIdMapper) {
```
0xAlunara, line 148: You can delete line 40 in the new file now, as networksIdMapper is now unused

`const networksIdMapper = useStore((state) => state.networks.networksIdMapper)`

0xAlunara (approved):

packages/curve-ui-kit/src/features/switch-chain/ui/ChainSwitcher.tsx
DanielSchiavini: If you navigate to the root it can be undefined

packages/curve-ui-kit/src/widgets/Header/MobileTopBar.tsx
```diff
@@ -9,14 +10,14 @@ export type MobileTopBarProps = {
   isSidebarOpen: boolean
   isLite: boolean
   currentMenu: AppMenuOption
-  ChainProps: ChainSwitcherProps
+  networks: NetworkMapping
```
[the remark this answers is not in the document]
DanielSchiavini: For just one prop? Not sure if it's more clear 🤔

PR #1480, opened 2025-10-06
feat: instant scroll to the top for every route

0xPearce:
Previously, when navigating between routes, the scroll position of the new page would match the scroll position of the previous page.
This ensures that each new page loaded is scrolled to the top.
Only in Beta for now.


apps/main/src/routes/RootLayout.tsx
```diff
@@ -36,8 +38,18 @@ export const RootLayout = () => {
   const theme = useUserProfileStore((state) => state.theme)
   const currentApp = getCurrentApp(usePathname())
   const onChainUnavailable = useOnChainUnavailable(networks)
+  const { location } = useRouterState()
+  const [releaseChannel] = useReleaseChannel()
   useLayoutStoreResponsive()
 
+  useEffect(() => {
+    // If there's a hash (#section-id), let the browser handle scrolling
```
DanielSchiavini, line 46: How about https://tanstack.com/router/v1/docs/framework/react/guide/scroll-restoration#scroll-restoration

PR #1482, opened 2025-10-07
feat: action info for pool parameters

DanielSchiavini:
- convert the pool parameters to action info components in beta mode
- align contract elements to the center of the line

### DIFF
--- a/apps/main/package.json
+++ b/apps/main/package.json
@@ -17,7 +17,7 @@
     "postpublish": "pinst --enable"
   },
   "dependencies": {
-    "@curvefi/api": "2.68.9",
+    "@curvefi/api": "2.68.10",
     "@curvefi/llamalend-api": "1.0.33",
     "@ethersproject/abi": "^5.8.0",
     "@hookform/error-message": "^2.0.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -677,6 +677,18 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@curvefi/api@npm:2.68.10":
+  version: 2.68.10
+  resolution: "@curvefi/api@npm:2.68.10"
+  dependencies:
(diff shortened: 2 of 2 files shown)


apps/main/src/dex/components/PagePool/PoolDetails/PoolStats/PoolParameters.tsx
```diff
@@ -36,20 +47,162 @@ const PoolParameters = ({
...
+            {rampADetails && !rampADetails.isFutureATimePassedToday && (
+              <>
+                <ActionInfo
+                  label={t`Ramping A`}
+                  valueTooltip={t`Slowly changing up A so that it doesn't negatively change virtual price growth of shares`}
+                  prevValue={formatNumber(initial_A, { useGrouping: false })}
+                  value={formatNumber(future_A, { useGrouping: false })}
+                />
+                <ActionInfo
+                  label=" "
```
DanielSchiavini, line 195: not sure how to deal with this one, just keeping it as it was for now

0xAlunara: Could set the type of `label: ReactNode` in the `ActionInfo` component to `label: ReactNode | undefined`. That way the label can be set to be empty, but it the property is still required and setting as empty has to be done so explicitly as it's an edge case.

0xPearce (approved):

apps/main/src/dex/components/PagePool/PoolDetails/PoolStats/PoolParameters.tsx
DanielSchiavini: I don't think it's a good practice to have an actionInfo without a title, I prefer to keep this as an edge case for now

PR #1488, opened 2025-10-07
fix: deleverage chart bug

OnlyJousting:
Changes:
- Ensure deleveragePrices is not empty before processing in ChartOhlcWrapper
- Ensure unused areas series are cleared out after data changes


apps/main/src/loan/components/ChartOhlcWrapper/index.tsx
```diff
@@ -127,7 +127,7 @@ const ChartOhlcWrapper = ({ rChainId, llamma, llammaId, betaBackgroundColor }: C
     }
 
     // deleverage prices
-    if (deleveragePrices && chartOhlcObject.data) {
+    if (deleveragePrices && deleveragePrices.length !== 0 && chartOhlcObject.data) {
```
DanielSchiavini, line 130:
```suggestion
    if (deleveragePrices?.length && chartOhlcObject.data) {
```

DanielSchiavini (approved):

packages/ui/src/Chart/CandleChart.tsx
```diff
@@ -491,12 +491,19 @@ const CandleChart = ({
...
   // Update liquidation range data when it changes
   useEffect(() => {
-    if (!liquidationRange) return
-
     const setSeriesData = (series: ISeriesApi<'Area'> | null, data: any) => {
       if (series) series.setData(data)
     }
 
+    // Clear all series if liquidationRange is undefined or has no data
+    if (!liquidationRange || (!liquidationRange.current && !liquidationRange.new)) {
```
0xAlunara, line 499: It's a bit fucked up how `liquidationRange` itself can be non-null while at the same time all the data inside can be🤔

0xAlunara (approved):

PR #1489, opened 2025-10-07
feat: add merkl rewards

0xAlunara:
Waiting for a response by the Merkl team with respect to pagination, but should be ready for review & test regardless.

The following external campaigns have been deleted as they are provided by Merkl instead:
1. Plasma
2. Apple
3. MezoUSD
4. Carrot
5. YieldNest

The APR shown is the APR for the entire campaign, as individual tokens don't really have APR data on their own.

image


apps/main/src/dex/components/PagePool/PoolDetails/PoolStats/Rewards.tsx
```diff
@@ -31,8 +31,10 @@ const Rewards = ({ chainId, poolData, rewardsApy }: RewardsProps) => {
   const { base, other } = rewardsApy ?? {}
   const { haveBase, haveOther, haveCrv } = haveRewardsApy(rewardsApy ?? {})
   const network = useStore((state) => state.networks.networks[chainId])
-  const { data: campaigns } = useCampaignsByNetwork(network.networkId as Chain)
-  const campaignRewardsPool = campaigns?.[poolData.pool.address]
+  const { data: campaigns } = useCampaignsByAddress({
+    blockchainId: network.networkId as Chain,
+    address: poolData.pool.address as Address,
```
0xAlunara, line 36: `as Address` every is annoying as hell to me, but keeping it as `string` is bad practice too imo

packages/curve-ui-kit/src/entities/campaigns/campaigns.ts
```diff
@@ -0,0 +1,118 @@
...
+export const combineCampaigns = ({
+  campaigns,
+  filter,
+}: {
+  campaigns: (Campaigns | undefined)[]
+  filter?: (campaign: CampaignPoolRewards) => boolean
+}): Campaigns => {
+  // Get all unique addresses from all campaign sources
+  const allAddresses = new Set(
+    campaigns.filter((record): record is Campaigns => record !== undefined).flatMap((record) => Object.keys(record)),
```
DanielSchiavini, line 39:
```suggestion
    notFalsy(...campaigns).flatMap(recordKeys),
```

packages/curve-ui-kit/src/entities/campaigns/campaigns.ts
```diff
@@ -0,0 +1,118 @@
...
+  )
+
+  // Combine campaigns by address, applying optional filter
+  return fromEntries(
+    [...allAddresses]
```
DanielSchiavini, line 44: No reason to copy if you're calling `map`
```suggestion
    allAddresses
```

0xAlunara: allAddresses is a set which has no `.map`, it needs to be converted to an array first

DanielSchiavini (approved):

PR #1491, opened 2025-10-07
fix: Lend market page title 

0xPearce:
- fix: lend markets page updates to the collateral token symbol instead of the address
- refactor: consistent page titles across all LlamaLend routes


apps/main/src/lend/hooks/useLendPageTitle.ts
```diff
@@ -0,0 +1,15 @@
+import { useEffect } from 'react'
+
+/**
+ * Custom hook to set the document title for markets pages
+ * @param collateralSymbol - The symbol of the collateral token
+ * @param prefix - The prefix part of the title (e.g., "Create - ...")
+ * @param suffix - The suffix part of the title (e.g., ".. - Curve Llamalend")
```
DanielSchiavini, line 7: It's actually always the same, if the hook is specific for lend we don't need this arg

0xPearce:
Sometimes the `document.title` is updated, but the change isn’t reflected in the tab title. I guess it depends on the browser. It happened only a few times

0xtutti (approved):

PR #1493, opened 2025-10-08
refactor: get rid of multiple useChainId variations, rely mostly on wagmi

0xAlunara:
We have three different implementations of `useChainId` for the dex, lend and loan apps. In most cases, the selected chain from the wallet should be leading. Just rely on wagmi to get that chainId for simplicity. Relying on the URL is silly, because the wallet chain is being changed to whatever the URL is. Single source of truth is nice.

Stepping stone to rely less on complicated network objects and network id mappers.

Also removes manual bookkeeping of the `chainId` in our wallet singleton.

This change even works if you don't have a wallet extension installed and have no wallet connected, as `switchChainAsync` works even without a wallet. So this PR should reduce complexity a bit wrt chain ids and networks.


packages/curve-ui-kit/src/features/connect-wallet/lib/ConnectionProvider.tsx
```diff
@@ -27,6 +27,7 @@ export const ConnectionProvider = <TChainId extends number, NetworkConfig extend
   children: ReactNode
 }) => {
   const [connectState, setConnectState] = useState<ConnectState>(LOADING)
+  const chainId = useChainId()
```
DanielSchiavini, line 30:
```suggestion
  const walletChainId = useChainId()
```

PR #1495, opened 2025-10-08
feat: enhance balance input precision

0xAlunara:
Introduce a high precision flag for balance values to prevent rounding errors, add ellipsis support to the input field for better text handling, and ensure the input retains full precision when selecting the maximum balance. Also adds tooltip on balance hover.

### DIFF
--- a/packages/curve-ui-kit/src/shared/ui/Balance.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/Balance.tsx
@@ -1,4 +1,5 @@
 import type { ReactNode } from 'react'
+import { notFalsy } from '@curvefi/prices-api/objects.util'
 import AccountBalanceWalletOutlinedIcon from '@mui/icons-material/AccountBalanceWalletOutlined'
 import Button from '@mui/material/Button'
 import Stack from '@mui/material/Stack'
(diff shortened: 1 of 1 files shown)


packages/curve-ui-kit/src/shared/ui/LargeTokenInput.tsx
```diff
@@ -189,6 +189,9 @@ type Props<T> = DataType<T> & {
  * Calculate the new balance based on max balance and percentage, rounding to specified decimals
  */
 function calculateNewBalance<T extends Amount>(max: T, newPercentage: Decimal, balanceDecimals: number | undefined): T {
+  // Avoid loss of precision when clicking 'Max'
```
DanielSchiavini, line 192: does that happen even with `BigNumber` :skull:

0xtutti (approved):

packages/curve-ui-kit/src/shared/ui/Balance.tsx
```diff
@@ -58,20 +59,22 @@ type BalanceTextProps<T> = {
...
-    <Stack direction="row" gap={Spacing.xs} alignItems="center">
-      <Typography
-        className="balance"
-        variant="highlightS"
-        color={balance != null ? 'textPrimary' : 'textTertiary'}
-        data-testid="balance-value"
-      >
-        {balance == null ? '-' : formatNumber(balance, { abbreviate: true })}
-      </Typography>
+    <Tooltip title={t`Wallet balance`} body={`${balance} ${symbol}`} clickable>
```
DanielSchiavini, line 62: It seems that `balance` can be `null`
```suggestion
    <Tooltip title={t`Wallet balance`} body={notFalsy(balance, symbol).join(' ')} clickable>
```

DanielSchiavini (dismissed):

packages/curve-ui-kit/src/shared/ui/LargeTokenInput.tsx
0xAlunara: Well we're rounding below with `newBalance = BigNumber.min(newBalance.toFixed(balanceDecimals), max)`, which is fine for any percentage except 100%

DanielSchiavini (approved):

0xtutti (approved):

PR #2028, opened 2026-02-12
feat(llamalend): onchain-first multicall table data for markets list

0xtutti:
## What this PR does
- Moves LlamaLend market-list tables to an onchain-first data path.
- Keeps API for market discovery/metadata and field-level fallback only.
- Batches chain reads via multicall per chain for market rates and borrow-side user stats.
- Applies merged table data through one canonical query hook (`useLlamaMarketsTableData`).
- Preserves net-borrow table semantics by recomputing net values from onchain base + API spread delta.
- Prevents one slow chain from blocking the full table via per-chain timeout degradation.

## How it works
1. Fetch market list from API (`useLlamaMarkets`).
2. Group markets by chain and execute multicalls for `rate`, `total_debt`, `totalAssets(controller)` and user borrow reads (`user_state`, `health`) when wallet is connected.
3. Merge results into table rows with onchain-first precedence for available fields.
4. Fall back to API only for missing/failed onchain fields.

## Scope
- Tables only (`/llamalend/*/markets` list and user-position cells).
- No market-page data plumbing changes.
- No backend/API contract changes.

## Why this helps
- Reduces UX lag from API staleness for table-critical values.
- Keeps rendering resilient under partial RPC/chain failures.
- Centralizes table merge logic in one query path to simplify maintenance/review.

## Quick verification
1. Open `/llamalend/ethereum/markets` with a wallet that has positions.
2. Confirm borrow/net-borrow columns refresh from onchain values without long stalls.
3. Confirm user borrow columns populate and do not stay in skeleton state.
4. Temporarily degrading one chain should not block other chain rows from updating.

## Review size
- `5 files changed, +466/-24`


apps/main/src/llamalend/queries/market-list/llama-markets.ts
```diff
@@ -325,6 +330,7 @@ const convertMintMarket = (
       borrowed: {
         symbol: stablecoinToken.symbol,
         address: stablecoinToken.address,
+        decimals: Number((stablecoinToken as { decimals?: number }).decimals ?? 18),
```
0xAlunara: This data comes from the API which doesn't return decimals, so essentially it's always 18 and may be wrong (like for USDC where they should be 6)

I don't think we need decimals though, given this is only for the on-chain data it should be able to just fetch the decimals on demand if it really needs it with the given address

apps/main/src/llamalend/queries/market-list/llama-markets-table-data.ts
```diff
@@ -0,0 +1,111 @@
+import { useMemo } from 'react'
+import { type Address } from '@ui-kit/utils'
```
0xAlunara: Lint error due to import ordering

apps/main/src/llamalend/queries/market-list/llama-markets-table-data.ts
```diff
@@ -0,0 +1,111 @@
...
+  onchainBase: number | null | undefined
+  apiBase: number | null | undefined
+  apiNet: number | null | undefined
+}) => {
+  if (!isFiniteNumber(onchainBase)) return undefined
+  const spread = isFiniteNumber(apiBase) && isFiniteNumber(apiNet) ? apiNet - apiBase : undefined
+  return isFiniteNumber(spread) ? onchainBase + spread : onchainBase
+}
+
+const getOnchainSignature = (markets: LlamaMarket[]) =>
```
0xAlunara: This entire function is unnecessary, we don't need all this data in the query key

apps/main/src/llamalend/queries/market-list/llama-markets-table-data.ts
```diff
@@ -0,0 +1,111 @@
...
+    enabled: apiMarkets.length > 0,
+    staleTime: REFRESH_INTERVAL['10s'],
+    refetchInterval: getRefetchInterval,
+  })
+
+  return {
+    ...apiQuery,
+    data:
+      apiQuery.data && onchainQuery.data
+        ? { ...apiQuery.data, markets: applyOnchainData(apiQuery.data.markets, onchainQuery.data) }
```
0xAlunara: Prob easier if we return `apiQuery` directly wrapped in a `mapQuery` whose selector maps over the markets if there's `onchainQuery.data`

apps/main/src/llamalend/queries/market-list/llama-markets-table-data.ts
```diff
@@ -0,0 +1,111 @@
...
+        `${chain}:${controllerAddress.toLowerCase()}:${ammAddress.toLowerCase()}:${(vaultAddress ?? '').toLowerCase()}:${type}:${userHasPositions?.Borrow ? 1 : 0}`,
+    )
+    .join('|')
+
+const getRefetchInterval = () =>
+  typeof document === 'undefined' || document.visibilityState === 'visible' ? REFRESH_INTERVAL['15s'] : false
+
+const applyOnchainData = (
+  apiMarkets: LlamaMarket[],
+  onchainData: Awaited<ReturnType<typeof fetchLlamaMarketsOnchainData>>,
```
0xAlunara: Type should be `Query<T>` for whatever `fetchLlamaMarketsOnchainData` would return if it were a `queryFactory`.

0xAlunara (commented):
Whelp. See Slack.

PR #2032, opened 2026-02-14
fix: update net borrow apr tooltip description

0xAlunara:
1. 'Borrow APR' becomes the default visible column, 'Net borrow APR' column becomes optional.
2. All 'Net borrow APR' metrics on market page become 'Borrow APR' metrics instead.
3. Remove separate 'Net borrow APR' and 'Borrow APR' tooltips for just a single 'Borrow APR' tooltip. If there's extra rewords or rebasing yields it'll show the extra data.
4. Update tooltip descriptions according to Notion ticket & Figma.

With yield bearing tokens
image

With no yield bearing tokens
image

'Borrow APR' is on, 'Net borrow APR' is off and optional
image


0xtutti (approved):

apps/main/src/llamalend/features/market-position-details/BorrowInformation.tsx
```diff
@@ -73,7 +73,7 @@ export const BorrowInformation = ({
         {!showPageHeader && (
           <Metric
             size="medium"
-            label={t`Net borrow APR`}
+            label={t`Borrow APR`}
             value={borrowRate?.totalBorrowRate}
```
0xPearce: Here we are still rendering the Net borrow rate

Pearce542 (commented):
Ugh this will have conflicts with this PR #2031 
Borrow APR / Net borrow APR it's such a mess we have defined it everywhere, that what's I am trying to do in my PR: remove the duplicate logic and only have two places where it is defined/computed, one for the prices api and the other for llamalend

apps/main/src/llamalend/features/market-list/cells/RateCell/BorrowRateTooltip.tsx
```diff
@@ -1,44 +1,14 @@
...
-      title={title}
-      body={<MarketBorrowRateTooltipWrapper market={market} borrowRateType={type} />}
-      placement="top"
-    >
-      {children}
-    </Tooltip>
-  )
-}
+export const BorrowRateTooltip = ({ market, children }: BorrowRateTooltipProps) => (
+  <Tooltip clickable title={t`Borrow APR`} body={<MarketBorrowRateTooltipWrapper market={market} />} placement="top">
```
0xPearce: Oh and now we have hardcoded "Borrow APR" title for the "Net borrow APR" column as well, not a huge problem though

PR #2145, opened 2026-03-04
feat: forms tabs on the right layout

Pearce542:
New detail page layout:
- forms tabs on the right
- page header inside the left context in llamalend


0xPearce:
replaced by: #2159

PR #2193, opened 2026-03-12
refactor: invalidate instead of refetching during hydration

DanielSchiavini:
- if we now properly fixed the hydration we can simplify the volume and tvl fetching so it invalidates the queries instead of manually fetching them
- depends on #2192


DanielSchiavini:
OK I'll close this for now (hard to fully test), hopefully we can get rid of the slices altogether soon :tm:

PR #2334, opened 2026-04-09
fix: add multicall for monad and arc when using wagmi multicall

0xAlunara:
This fixes the lack of multicall support for when we use Wagmi, but it doesn't solve the lack of multicall for these networks in curve-js which in turn uses curvefi/ethcall.

If that's still needed we need to add some chains to https://github.com/curvefi/ethcall/blob/master/src/multicall.ts most likely. This is primarily an issue for Arc, as the API only returns the first 500 pools out of 2853, and curve-js tries to fetch the rest by non-existing multicall when calling `curve.stableNgFactory.fetchNewPools()` in `fetchPoolIds`. I've asked phil if that 500 cut-off is really necessary. If it is we need to update the multicall mapping and curve-js respectively.


DanielSchiavini (approved):
Can we have some typescript or runtime check to validate the chains in this list match the ones given by curve-js?

OnlyJousting (approved):

0xPearce (approved):

0xtutti (approved):

0xAlunara:
I've created a Notion ticket to add an alert of some kind when a Curve network definition is being used without corresponding wagmi chain definition.

PR #2336, opened 2026-04-10
chore: omit liquidation prices on vault page

DanielSchiavini:
- do not show liquidation prices on the vault page


DanielSchiavini:
replaced by #2351

PR #2338, opened 2026-04-10
fix: leverage loading after transaction

DanielSchiavini:
- Fixes an issue where the leverage checkbox would be loading forever after a transaction
  - when a transaction is done we reset all the form values. However, data received via `useFormSync` might never receive a value again, since the form is reset but the queries aren't
  - instead of resetting all fields, we separate now the user-filled fields and the other ones
  - after updating tests, I had to introduce a new hasLeverageManagement field for test data, to identify markets with leverage only for creating loans but not for managing them


DanielSchiavini:
I'll try to handle this issue via invalidation instead

0xtutti (dismissed):

apps/main/src/llamalend/queries/borrow-more/borrow-more-future-leverage.query.ts
```diff
@@ -64,3 +66,9 @@ export const {
   category: 'llamalend.borrowMore',
   validationSuite: borrowMoreLeverageValidationSuite,
 })
+
+export function useBorrowMoreLeverage(params: BorrowMoreParams) {
+  const current = useUserCurrentLeverage(params)
+  const future = useBorrowMoreFutureLeverage(params)
+  return q(future.data == null ? current : future)
```
Pearce542: what happens if the `useBorrowMoreFutureLeverage` fails, we render the current leverage

0xAlunara (dismissed):

Pearce542 (dismissed):

0xAlunara (approved):

Pearce542 (approved):

PR #2340, opened 2026-04-10
chore: update vite to v8

0xAlunara:
NOTE: this PR exists merely as a reminder to see what might needs fixing when we finally do upgrade to Vite v8

Cypress component tests don't work yet with Vite v8, [but it's being worked on](https://github.com/cypress-io/cypress/issues/33078). In general it seems Cypress doesn't fully support it yet.


DanielSchiavini:
@0xAlunara check out what got merged: https://github.com/cypress-io/cypress/issues/33078

PR #2342, opened 2026-04-10
feat: default markets collateral and borrow token invalidations

0xPearce:
- default `useLlammaMutation` token balance invalidation to the market collateral and borrow tokens
- add explicit invalidation for the claim mutations to invalidate the actual claimable token addresses

### DIFF
--- a/apps/main/src/llamalend/mutations/claim.mutation.ts
+++ b/apps/main/src/llamalend/mutations/claim.mutation.ts
@@ -10,7 +10,7 @@ import {
 } from '@/llamalend/queries/validation/supply.validation'
 import type { IChainId as LlamaChainId, INetworkName as LlamaNetworkId } from '@curvefi/llamalend-api/lib/interfaces'
 import { type Address, type Hex } from '@primitives/address.utils'
-import { assert } from '@primitives/objects.utils'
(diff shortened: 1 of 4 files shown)


DanielSchiavini (approved):

apps/main/src/llamalend/mutations/claim.mutation.ts
```diff
@@ -64,6 +79,7 @@ export const useClaimRewardsMutation = ({ network, network: { chainId }, marketI
     validationSuite: claimableRewardsValidationSuite,
     pendingMessage: () => t`Claiming rewards...`,
     successMessage: () => t`Claimed rewards!`,
+    mutationTokenAddresses: () => rewardTokenAddresses,
```
0xAlunara, line 82: You sure you there's no need to invoke `notFalsy` here like with `crvTokenAddress` above?

0xAlunara (approved):

apps/main/src/llamalend/mutations/claim.mutation.ts
0xPearce: Yes the array is empty when no rewards

PR #2346, opened 2026-04-11
fix: balance values in loan form message were wrapping incorrectly

0xAlunara:
Turns out that for all our `<LoanFormTokenInput>` usages we passed on a direct `<Balance>` component to the `message` prop, rather than relying on the internal `buildClickableMessage` of the `<HelperMessage>` component it uses.

While the `<Balance>` component works, it wasn't really made for inline usage. I've tried various things to make it inline (with `inline-flex` and `component='span' on the `<Typography>`), none of it really worked out well.

Instead, I've decided to just rely on the `buildClickableMessage` of `<MessageHelper`>. The only downside is that there's no loading skeleton, but honestly that's not a big loss.

The only downside / exception I had to build was to add a `onMessageNumberClick` prop to `<LoanFormTokenInput>`. The default behavior of clickable numbers is to use the given numbers directly, but in the case of repaying inside `<RepayForm`> we want to show something like "Repay 5 crvUSD", but when you click it the input might actually have to be 4.5 sreUSD or whatever. Hence:
```tsx
onMessageNumberClick={() => updateForm(form, { [selectedField]: max[selectedField].data })}
```
Technically there can be more than one clickable number, but I didn't want to built complex code around this special case where we have to identify what number exactly got clicked.

image

### DIFF
--- a/apps/main/src/llamalend/features/borrow/components/CreateLoanForm.tsx
+++ b/apps/main/src/llamalend/features/borrow/components/CreateLoanForm.tsx
@@ -14,7 +14,6 @@ import { AlertBox } from '@ui/AlertBox'
 import type { AlertType } from '@ui/AlertBox/types'
 import { useCreateLoanPreset } from '@ui-kit/hooks/useLocalStorage'
 import { t } from '@ui-kit/lib/i18n'
-import { Balance } from '@ui-kit/shared/ui/LargeTokenInput/Balance'
 import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
 import { q, type Range } from '@ui-kit/types/util'
 import { updateForm } from '@ui-kit/utils/react-form.utils'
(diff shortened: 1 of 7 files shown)


OnlyJousting (approved):

0xPearce (approved):

apps/main/src/llamalend/features/borrow/components/CreateLoanForm.tsx
```diff
@@ -121,17 +120,7 @@ export const CreateLoanForm = <ChainId extends IChainId>({
...
-              prefix={t`Max borrow:`}
-              tooltip={t`Max borrow`}
-              symbol={borrowToken?.symbol}
-              balance={values.maxDebt}
-              loading={maxDebt.isLoading}
-              onClick={useCallback(() => updateForm(form, { debt: values.maxDebt }), [form, values.maxDebt])}
-              buttonTestId="borrow-set-debt-to-max"
-            />
-          }
+          message={`${t`Max borrow:`} ${values.maxDebt ?? '-'} ${borrowToken?.symbol}`}
```
DanielSchiavini, line 123: this is repeated a lot of times, should we create a helper so they are easy to find?
I do not like to format the token symbol directly on the components, that will lead to inconsistencies
```suggestion
          message={formatMax(t`Max borrow:`, values.maxDebt, borrowToken?.symbol)}
```

0xtutti (approved):

apps/main/src/llamalend/features/borrow/components/CreateLoanForm.tsx
0xAlunara: Only the fallback value of `-` is being handled manually, the actual formatting is already being done internally in the helper component. Not sure if it's worth generalizing just the `-` though 🤔

PR #2347, opened 2026-04-11
chore: replace eslint-plugin-react with @eslint-react

0xAlunara:
Pre-requisite for upgrading eslint from v9 to v10.

Replaces https://www.npmjs.com/package/eslint-plugin-react and https://www.npmjs.com/package/eslint-plugin-react-hooks with https://www.eslint-react.xyz/, which is designed as a modern replacement for both with improved performance, better TypeScript support, and more accurate rule implementations, and ESLint v10 support.


apps/main/src/dex/components/PagePool/Deposit/components/FormStake.tsx
```diff
@@ -148,15 +148,15 @@ export const FormStake = ({ curve, poolData, poolDataCacheOrApi, routerParams, s
     if (poolId) {
       resetState(poolData, 'STAKE')
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
+    // eslint-disable-next-line @eslint-react/exhaustive-deps
```
0xAlunara, line 151: Biggest chunk of the diff is basically just updating this pre-existing disabling of the lint rule name

apps/main/src/loan/components/DetailInfoHealth.tsx
```diff
@@ -85,7 +85,6 @@ export const DetailInfoHealth = ({
   useEffect(() => {
     if (typeof oraclePriceBand !== 'undefined' && userLoanDetails) {
       const { healthFull, healthNotFull, userBands } = userLoanDetails
-      // eslint-disable-next-line react-hooks/set-state-in-effect
```
0xAlunara, line 88: For some reason the old eslint rule didn't check for this consistently and correctly all the time, and if it did we'd disable it anyway. Hence I've disabled this on the config level for now. Not sure if we want to re-enable it in the future.

packages/curve-ui-kit/src/hooks/useLocalStorage.ts
```diff
@@ -85,7 +85,6 @@ export const useDismissBanner = (bannerKey: string, expirationTime: number) => {
   const [dismissedAt, setDismissedAt] = useLocalStorage<number | null>(bannerKey, null)
 
   const shouldShowBanner = useMemo(
-    // eslint-disable-next-line react-hooks/purity
```
0xAlunara, line 93: Disabled this on the config level, we can always re-enable it in a future refactor PR

packages/curve-ui-kit/src/shared/ui/stories/Accordion.stories.tsx
```diff
@@ -168,6 +168,7 @@ export const ComplexContent: Story = {
 
 export const Controlled: Story = {
   render: () => {
+    // eslint-disable-next-line @eslint-react/rules-of-hooks
```
0xAlunara, line 171: In story files it complays we're using hooks in stories that don't begin their name with `use`, like `useControlled`. But we don't really care though; not sure how to disable that on a config level just for storybooks.

packages/eslint-config-custom/index.js
```diff
@@ -3,17 +3,43 @@ module.exports = {
...
+    '@eslint-react/no-children-to-array': 'off',
+    '@eslint-react/no-children-map': 'off',
+    '@eslint-react/no-nested-component-definitions': 'off',
+    '@eslint-react/web-api-no-leaked-timeout': 'off',
+    '@eslint-react/dom-no-flush-sync': 'off',
+    '@eslint-react/web-api-no-leaked-event-listener': 'off',
+    '@eslint-react/use-state': [
+      'warn',
+      {
+        enforceSetterName: false, // we're not really sticking to the 'setX' naming convention in a lot of cases
```
0xAlunara, line 35: I want to keep the PR diff small so we can enable this rule and rename incorrectly named setters in a future PR.

0xPearce (dismissed):

DanielSchiavini (dismissed):

0xPearce (approved):

DanielSchiavini (approved):

PR #2354, opened 2026-04-14
fix: useLlamaMarket params

0xPearce:
- fixed `useLlamaMarket` with updated Llv2 params


apps/main/src/llamalend/queries/market-list/llama-markets.ts
```diff
@@ -557,7 +558,10 @@ export const useLlamaMarket = (
   },
   enabled: boolean | undefined = true,
 ) => {
-  const llamaMarketsQuery = useLlamaMarkets(undefined, enabled && !!blockchainId && !!controllerAddress)
+  const llamaMarketsQuery = useLlamaMarkets(
+    { userAddress: undefined, enableLLv2: useLLv2() },
```
DanielSchiavini: Why are we not passing the user address? That will lead to incorrect data in the market object!

DanielSchiavini (changes requested):
I fixed the problem and did a small refactor in #2355

DanielSchiavini (approved):

PR #2366, opened 2026-04-15
fix: test addresses comparison

0xPearce:
- token addresses coming from prices api are checksummed. We should always use `isAddressEqual`

this was causing the user position summary error


packages/curve-ui-kit/src/lib/model/entities/token-usd-rate.ts
```diff
@@ -22,10 +22,11 @@ import { readContract } from '@wagmi/core'
 
 export const QUERY_KEY_IDENTIFIER = 'usdRate' as const
 
-const getTestTokenPrice = async (chainId: number, tokenAddress: string): Promise<number | null> => {
+const getTestTokenPrice = async (chainId: number, tokenAddress: Address): Promise<number | null> => {
   if (chainId === Chain.Optimism) {
-    if (tokenAddress === '0x79cc1c5c0171ff25ef391055e529a56a12bf3d39') return await fetchUsdRate(chainId, ethAddress)
-    if (tokenAddress === '0xe8e9cd957dc2b5a32ea822a3799f65940cf51f19') return 1
+    if (isAddressEqual(tokenAddress, '0x79cc1c5c0171ff25ef391055e529a56a12bf3d39'))
```
DanielSchiavini, line 26: argh I think I wrote this, but we need some constants or comments to be able to understand these

DanielSchiavini (dismissed):
We should apply this to the whole app at some point
Image

packages/curve-ui-kit/src/lib/model/entities/token-usd-rate.ts
0xPearce: You mean the test token addresses? We should but it's temporary so ...

0xPearce:
> We should apply this to the whole app at some point

Created a task

0xAlunara (dismissed):
`isAddressEqual` everywhere would be ideal like dani said, but having to cast to `Address` is annoying. We prob need to properly type addresses across the app first.

0xPearce:
> `isAddressEqual` everywhere would be ideal like dani said, but having to cast to `Address` is annoying. We prob need to properly type addresses across the app first.

Agree, we probably need to address the validation of `prices-api` entries correctly first using zod or similar.

DanielSchiavini (approved):

0xAlunara (approved):

0xtutti (approved):

PR #2367, opened 2026-04-15
fix: circular import breaking rpc tests

DanielSchiavini:
- fix error `can't access lexical declaration 'USD_RATE_KEY_IDENTIFER' before initialization` in rpc tests


0xPearce (approved):

0xAlunara (approved):
I'm too dumb to see where the actual circular import happens but if it fixes it it's good I suppose.

DanielSchiavini:
> I'm too dumb to see where the actual circular import happens but if it fixes it it's good I suppose.

cache.ts is imported from a lot of places, I don't have the exact import chain rn but I can share later

PR #2368, opened 2026-04-15
refactor: move logging to query factory

DanielSchiavini:
- move logging to query factory
- depends on #2367


0xtutti (approved):

packages/curve-ui-kit/src/lib/api/cache.ts
```diff
@@ -1,21 +1,7 @@
-import { type Mutation, MutationCache, QueryCache } from '@tanstack/react-query'
+import { type Mutation, MutationCache } from '@tanstack/react-query'
 import { addBreadcrumb, captureError } from '@ui-kit/features/sentry'
 import { logError, logMutation, logSuccess } from '@ui-kit/lib/logging'
 
-export const USD_RATE_KEY_IDENTIFER = 'usdRate' as const
-
-// disable logging for queries that are too verbose
-const disableCacheQueryKeys = new Set<unknown>(['readContracts', USD_RATE_KEY_IDENTIFER])
```
0xAlunara, line 8: Did we lose this functionality in the new `runQuery`? You might find that the console now gets spammed on the Swap page I think.

EDIT: just checked and not seeing any usd-rate logs at all, so the spamming isn't happening. But now I'm wondering if we're really logging successful query runs properly? 🤔 

EDIT2: I guess we're now only logging queries made through `queryFactory`. We're no longer querying calls made by wagmi itself, like when we use `readContract` natively. Is this ok?

0xAlunara (approved):

packages/curve-ui-kit/src/lib/api/cache.ts
DanielSchiavini: Yes that was my idea. I don't think we need to know about those

DanielSchiavini: Pls merge If you agree

PR #2369, opened 2026-04-15
feat: add sorting by crv and incentives for mobile

0xAlunara:
I split `RewardsOtherCell` up in `RewardsCrvCell` and `RewardsIncentivesCell` so that the mobile table can show the columns separated.

I removed the original fallback of `-` when no campaigns, crv or other incentives. Not sure why but it never showed in production anyway. Removing it simplifies the code quite a bit.

image
image


apps/main/src/dex/features/pool-list/cells/RewardsIncentivesCell.tsx
```diff
@@ -0,0 +1,36 @@
...
+type Prop = CellContext<PoolListItem, RewardsApy | undefined>
+
+export const RewardsIncentivesCell = ({ getValue, table, row: { original: poolData } }: Prop) => {
+  const { data: campaigns } = useCampaignsByAddress({
+    blockchainId: poolData.network as Chain,
+    address: poolData?.pool?.address as Address,
+  })
+  const { other } = getValue() ?? {}
+
+  return other?.length || campaigns.length ? (
```
DanielSchiavini: is it really a problem to render an empty stack?
```suggestion
  return (
```

apps/main/src/dex/features/pool-list/cells/RewardsOtherCell.tsx
```diff
@@ -1,44 +1,12 @@
...
-      {other?.map((o) => (
-        <Typography
-          fontWeight={isSortedBy(table, PoolColumnId.RewardsIncentives) ? 'bold' : 'normal'}
-          key={o.tokenAddress}
-        >
-          {formatNumber(o.apy, FORMAT_OPTIONS.PERCENT)} {o.symbol}
-        </Typography>
-      ))}
-      {campaigns.length > 0 && <CampaignRewardsRow rewardItems={campaigns} />}
-      {!crv && !other?.length && !campaigns.length && '-'}
```
DanielSchiavini, line 41: I don't think we should remove the empty state

0xtutti (dismissed):

0xAlunara:
In order to fix the missing `-` placeholder as requested, the code has gotten a bit more complex, as the desktop column is a composite of the two mobile columns, and in all cases we want to show a single '-'.

The difficulty lies in the fact that the placeholder should work for both Crv and Inventices cells in isolation on mobile, but we also don't want two - placeholders on desktop, which re-uses those very two cell components.

I've done some back and forth with Claude to find the best solution that avoids code duplication as much as possible while also not introducing React anti-patterns.

**Desktop**
image

**Mobile CRV**
image

**Mobile incentives**
image

0xPearce (approved):

apps/main/src/dex/features/pool-list/cells/RewardsOtherCell.tsx
```diff
@@ -1,12 +1,22 @@
...
+
+export const RewardsOtherCell = (props: Prop) => {
+  const rewards = props.getValue()
+  const { hasCrv, hasIncentives } = useHasPoolRewards(rewards, props.row.original)
+
+  if (!hasCrv && !hasIncentives) return <>-</>
+
+  return (
+    <>
+      <RewardsCrvCell placeholder={null} {...props} />
```
DanielSchiavini, line 18: I think making this the default would make more sense

DanielSchiavini (approved):

apps/main/src/dex/features/pool-list/cells/RewardsOtherCell.tsx
0xAlunara: Had to make a choice here; figured it'd be easier to have the exception in a single place rather than two:

image

Otherwise I had to add a custom placeholder in these two column definitions (for mobile)

Pick your poison 😄

PR #2373, opened 2026-04-15
refactor: add error state to metric and use it in user positions summary

OnlyJousting:
Changes:
- Add error state to `Metric`, if error exists `value` and `rightAdornment` won't be shown
- Use new Metric error state in `UserPositionSummary` and pass `Error` object instead of isError boolean


packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -206,6 +219,7 @@ export const Metric = ({
       showToast({ title: copyText, message: value, severity: 'info' })
     }
   }, [value, copyText])
+  const hasError = !!error
```
0xAlunara: It's used only once in a single place, so might as well inline it there?

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -222,17 +236,28 @@ export const Metric = ({
...
-            value={value}
-            valueOptions={valueOptions}
-            change={change}
-            size={size}
-            copyValue={value || value === 0 ? copyValue : undefined}
-            tooltip={valueTooltip}
-            testId={testId}
-          />
-          {rightAdornment}
+        <Stack direction="row" alignItems="baseline" gap={Spacing.xxs}>
```
0xAlunara: In essence you've just added a bit of spacing between the metric value and optional right adornment, right?

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -172,6 +178,11 @@ export type MetricProps = {
   /** Optional content to display to the right of the value */
   rightAdornment?: ReactNode
 
+  /** Shows an error triangle icon on the metric value row. */
+  error?: Error
```
0xAlunara: You've made the type `Error` but you're not really using it other than as a boolean check 🤔 
Might as well just check if `errorTooltip` down below is undefined or not if a check is all you need.

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -222,17 +236,28 @@ export const Metric = ({
...
-            size={size}
-            copyValue={value || value === 0 ? copyValue : undefined}
-            tooltip={valueTooltip}
-            testId={testId}
-          />
-          {rightAdornment}
+        <Stack direction="row" alignItems="baseline" gap={Spacing.xxs}>
+          {/* Keep error state vertical rhythm aligned with regular metric values by inheriting metric typography sizing. */}
+          {hasError ? (
+            <Tooltip arrow placement="bottom" title={errorTooltip?.title} body={errorTooltip?.body} {...errorTooltip}>
```
0xAlunara, line 241: I'm super puzzled by this, because `title` on `<Tooltip>` is non-optional, yet `errorTooltip?.title` makes it optional. How is typescript not complaining? 😂 

And then `title` and `body` are part of `errorTooltip`, so I hoped `<Tooltip arrow placement="bottom"  {...errorTooltip}>` would suffice, but I'm getting type errors if I try that. I'm just puzzled is all.

0xAlunara (commented):
Beetiful

Image

packages/curve-ui-kit/src/shared/ui/Metric.tsx
OnlyJousting: ah, I should have cleaned that up

packages/curve-ui-kit/src/shared/ui/Metric.tsx
OnlyJousting: yea I was trying to get to spread it all but was running into these issues

packages/curve-ui-kit/src/shared/ui/Metric.tsx
OnlyJousting: yea I think our preference is to pass the error objects rather then using isError, the tooltip is optional but maybe it shouldn't be? I guess it's nice to always populate

apps/main/src/llamalend/features/market-list/hooks/useUserPositionsSummary.tsx
```diff
@@ -18,7 +18,7 @@ import { type QueryOptionsData } from '@ui-kit/lib/queries/types'
 import { LlamaMarketType } from '@ui-kit/types/market'
 import { requireChainId } from '@ui-kit/utils'
 
-export type UserPositionSummaryMetric = { label: string; data: number; isLoading: boolean; error?: unknown }
+export type UserPositionSummaryMetric = { label: string; data: number; isLoading: boolean; error?: Error }
```
DanielSchiavini: We should be using the Query type
```suggestion
export type UserPositionSummaryMetric = { label: ReactNode; metric: Query<number> }
```

OnlyJousting: I'll update to use it

0xtutti (approved):

0xAlunara (approved):

PR #2359, opened 2026-04-14
feat: allow improve health with other tokens

DanielSchiavini:
- reuse the repay form for soft liquidation cases
- this will allow users to use the collateral token to repay their debt when in soft liquidation
- depends on #2372

PR #2374, opened 2026-04-16
feat: llamalend manage forms in stable mode

DanielSchiavini:
- upgrades the new loan management forms to stable mode :rocket: 
- ~~depends on #2359~~

image

### DIFF
--- a/packages/curve-ui-kit/src/hooks/useFeatureFlags.ts
+++ b/packages/curve-ui-kit/src/hooks/useFeatureFlags.ts
@@ -19,17 +19,20 @@ const isAlpha = () => getReleaseChannel() === ReleaseChannel.Beta && defaultRele
 /** New unified create loan form */
 export const useCreateLoanMuiForm = useStableChannel
 
-/** New unified supply/vault forms (deposit/withdraw/claim) */
-export const useLendingMuiForm = useBetaChannel
-
 /** New manage loan forms (add/remove/repay) */
-export const useManageLoanMuiForm = useBetaChannel
+export const useManageLoanMuiForm = useStableChannel
+
+/** New unified supply/vault forms (deposit/withdraw/claim) */
+export const useLendingMuiForm = useManageLoanMuiForm
+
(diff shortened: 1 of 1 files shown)


mimaklas (approved):

0xtutti (approved):

packages/curve-ui-kit/src/hooks/useFeatureFlags.ts
```diff
@@ -19,18 +19,21 @@ const isAlpha = () => getReleaseChannel() === ReleaseChannel.Beta && defaultRele
...
+export const useManageLoanMuiForm = useStableChannel
+
 /** New unified supply/vault forms (deposit/withdraw/claim) */
-export const useLendingMuiForm = useBetaChannel
+export const useLendingMuiForm = useManageLoanMuiForm
 
-/** New manage loan forms (add/remove/repay) */
-export const useManageLoanMuiForm = useBetaChannel
+/** New card for managing soft liquidations */
+export const useManageSoftLiquidation = useManageLoanMuiForm
```
0xAlunara, line 29: There's currently a ticket in progress to modify manage soft liquidation presentation. Is it okay if we move this into stable at a later stage?

0xAlunara (changes requested):

packages/curve-ui-kit/src/hooks/useFeatureFlags.ts
DanielSchiavini: I guess so but we'd have to test if the flow works fine. I'm not sure you'd see the old forms when on soft liquidation

0xAlunara: Aight, could release it then as it is. The improvements to the manage soft liq form can be added later, they shouldn't block the entirety of the new forms.

0xAlunara (approved):

DanielSchiavini:
I discussed with tutti we wanna put this behind a modal first, next week!

PR #2380, opened 2026-04-16
fix: transparent background color disabled ghost button

0xPearce:
- transparent background color for ghost button when disabled. 

Needs to be fixed in the design system on Figma as well I think. Because the color exists on Figma but it's not being used, and has been imported automatically.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
Screenshot 2026-04-16 at 13 58 27
</td>
    <td>
Screenshot 2026-04-16 at 13 57 54
</td>
  </tr>
</table>


DanielSchiavini (approved):
Could you create a task for @0xtutti to fix figma?

OnlyJousting (approved):

0xPearce:
> Could you create a task for @0xtutti to fix figma?

Already done!

PR #2381, opened 2026-04-16
fix: exchange rate precision

DanielSchiavini:
- fix exchange rate in the swap page and llamalend borrow form with higher precision


apps/main/src/dex/components/PageRouterSwap/index.tsx
```diff
@@ -53,7 +53,7 @@ import { SlippageToleranceActionInfo } from '@ui-kit/widgets/SlippageSettings'
 const { Spacing } = SizesAndSpaces
 
 const formatExchangeRate = ({ from, to, value }: ExchangeRate) =>
-  ['1', from, '=', +value ? formatNumber(value, { abbreviate: true }) : '-', to].join(' ')
+  ['1', from, '=', +value ? formatNumber(value, { abbreviate: true, highPrecision: true }) : '-', to].join(' ')
```
DanielSchiavini, line 56: Abbreviate with high precision is resulting in weird formatting but thats out of scope. Otherwise lgtm

OnlyJousting (approved):

apps/main/src/dex/components/PageRouterSwap/index.tsx
DanielSchiavini: @0xPearce I cannot approve since I created the PR

0xAlunara (approved):
Not really an improvement imo but otoh it's also not worse? 

Image

PR #2384, opened 2026-04-17
feat: loan manage invitation

DanielSchiavini:
- add a new modal that invites uses to try the new forms
- after the modal is dismissed, a banner is displayed to switch between the new and the old forms


apps/main/src/llamalend/widgets/banners/MarketBanners.tsx
```diff
@@ -19,11 +20,13 @@ export const MarketBanners = <ChainId extends IChainId>({
   chainId,
   market,
   rewardsBanner,
+  loanExists,
```
0xPearce, line 23: Never used

0xPearce (commented):
Nice! just the unused var

DanielSchiavini:
https://github.com/curvefi/curve-frontend/pull/2403

PR #2386, opened 2026-04-17
feat: add more market user position alert types

OnlyJousting:
Changes:
- Adds "Fully converted" and "Incomplete conversion" alerts to user positions
- Adds new `getLiquidationStatus` method while marking old one as depreciated (still used in legacy forms)
- Syncs user position market page alerts with market-list user position tooltips


apps/main/src/llamalend/llamalend.types.ts
```diff
@@ -11,6 +11,15 @@ export type NetworkDict<ChainId extends IChainId = IChainId> = Record<ChainId, L
 export type LlamaMarketTemplate = MintMarketTemplate | LendMarketTemplate
 
 export type HealthColorKey = 'healthy' | 'close_to_liquidation' | 'soft_liquidation' | 'hard_liquidation' | ''
+
+export type UserPositionStatus =
```
0xAlunara, line 15: Not that it matters I guess but could've deduced implicitly with `Required<ReturnType<typeof getLiquidationStatus>>`

apps/main/src/llamalend/position-status-content.tsx
```diff
@@ -0,0 +1,80 @@
...
+    severity: 'info',
+  },
+  softLiquidation: {
+    title: t`Liquidation protection active`,
+    description: (
+      <Trans>
+        The collateral price is inside your liquidation range. Within the range, collateral is being converted between{' '}
+        {collateralSymbol} and {borrowSymbol} as the price goes up or down. These conversions incur losses that reduce
+        the position's health.
+        <br />
```
0xAlunara, line 36: Reminder that you need a double <br /> for an actual whiteline / paragraph spacing. Personally I try to avoid br's and prefer gapped stacks with typography (or a span if you dont give a shit about typography)
I believe we've removed the default margins for <p> so that's not really an option I think

apps/main/src/llamalend/position-status-content.tsx
```diff
@@ -0,0 +1,80 @@
...
+): Record<UserPositionStatus, PositionStatusContent> => ({
+  healthy: {
+    title: t`Position is healthy`,
+    description: (
+      <Trans>The oracle price is above your liquidation range. No soft-liquidation conversions are occurring.</Trans>
+    ),
+    severity: 'info',
+  },
+  closeToLiquidation: {
+    title: t`Position is healthy`,
```
0xAlunara, line 23: Double checking if this is really intended to equal `healthy` or a copy-paste error

apps/main/src/llamalend/position-status-content.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <br />
+        <br />
+        If price recovers into your range, conversions resume and losses will further reduce your health. Consider
+        closing your position or repaying debt to improve health before that happens.
+      </Trans>
+    ),
+    severity: 'warning',
+  },
+  incompleteConversion: {
+    title: t`Position at risk — incomplete conversion`,
```
0xAlunara, line 70: No em-dash? 🤓

apps/main/src/llamalend/features/market-list/cells/MarketTitleCell/UserMarketPositionIndicator.tsx
```diff
@@ -32,55 +24,29 @@ const flickerEffect = (
   return () => clearTimeout(timeoutId)
 }
 
-const Paragraph = ({ children }: { children?: ReactNode }) => (
-  <Typography variant="bodySRegular" fontWeight={FontWeight.Medium}>
```
0xAlunara, line 36: Note that you're no longer using `bodySRegular` in the new tooltips, which means the (description) text will be a bit bigger.

0xAlunara (dismissed):

apps/main/src/llamalend/features/market-position-details/hooks/useBorrowPositionDetails.ts
```diff
@@ -112,16 +112,23 @@ export const useBorrowPositionDetails = ({
...
-    return getLiquidationStatus(healthNotFullValue, isClose, borrowed ?? '0')
-  }, [hasLoan, healthNotFullValue, userBandsValue, liquidationBand, oraclePriceBand, borrowed])
+  const positionStatus = useMemo(() => {
+    if (!hasLoan || !healthNotFullValue || !userBandsValue) return undefined
+    const isCloseToSoftLiquidation = getIsUserCloseToSoftLiquidation(
+      userBandsValue[0],
+      liquidationBand ?? null,
+      oraclePriceBand,
+    )
+    // userBandsValue[1] is the lower price boundary of a users range. Band numbers go up in numbers when going down in prices
```
DanielSchiavini: seems like an array destructuring like `const [upperBoundary, lowerBoundary] = userBandsValue`

apps/main/src/llamalend/features/market-position-details/BorrowPositionDetails.tsx
```diff
@@ -80,8 +78,15 @@ export const BorrowPositionDetails = ({
   totalDebt,
 }: BorrowPositionDetailsProps) => (
   <Stack padding={Spacing.md} gap={Spacing.xs}>
-    {liquidationAlert.softLiquidation && <LiquidationAlert type="soft" />}
-    {liquidationAlert.hardLiquidation && <LiquidationAlert type="hard" />}
+    {liquidationAlert.status &&
+      liquidationAlert.status !== 'healthy' &&
```
DanielSchiavini: this will break when we add more statuses, status that need alerts should be defined together with the status enum imo

apps/main/src/llamalend/queries/market-list/llama-market-stats.ts
```diff
@@ -70,11 +70,20 @@ export function useUserMarketStats(market: LlamaMarket, column?: LlamaMarketColu
 
   const borrowedAmount = stats ? ('borrowed' in stats ? stats.borrowed : stats.stablecoin) : 0
 
+  const status = stats
+    ? getLiquidationStatus(
+        decimal(stats.health) ?? '0',
```
DanielSchiavini: no position != 0 health. Users will be alerted for liquidation incorrectly

apps/main/src/llamalend/llama.utils.ts
```diff
@@ -242,7 +260,11 @@ export function getLiquidationStatus(
   return userStatus
 }
 
-export function getIsUserCloseToLiquidation(
+// is user fully converted
```
DanielSchiavini: what is this about?

apps/main/src/llamalend/features/market-position-details/BorrowPositionDetails.tsx
OnlyJousting: I was trying to solve it with types alone at first but adding a hasMarketAlert field to the object worked out better 🤝

apps/main/src/llamalend/llama.utils.ts
OnlyJousting: I forgot to clean them up, was considering making helpers

apps/main/src/llamalend/position-status-content.tsx
OnlyJousting: Br seems a bit more declarative to me when there's a need for line breaks too

apps/main/src/llamalend/position-status-content.tsx
OnlyJousting: ey I just copied the copy from the notion ticket, props for noticing though

apps/main/src/llamalend/position-status-content.tsx
OnlyJousting: ah this is actually not a status we have designed for. it's used in the legacy use cases but not the new ones all though maybe it would be nice to have. I'll remove it for now though

apps/main/src/llamalend/llamalend.types.ts
OnlyJousting: it's cleaner 👌

apps/main/src/llamalend/queries/market-list/llama-market-stats.ts
OnlyJousting: omg they weren't even needed anymore, thanks for pointing it out

apps/main/src/llamalend/position-status-content.tsx
OnlyJousting: actually, they are — not -?

apps/main/src/llamalend/features/market-list/cells/MarketTitleCell/UserMarketPositionIndicator.tsx
OnlyJousting: nice, I restored the typography variant

apps/main/src/llamalend/position-status-content.tsx
```diff
@@ -0,0 +1,80 @@
...
+    description: (
+      <Trans>
+        The collateral price is inside your liquidation range. Within the range, collateral is being converted between{' '}
+        {collateralSymbol} and {borrowSymbol} as the price goes up or down. These conversions incur losses that reduce
+        the position's health.
+        <br />
+        You cannot add collateral while in the range. Close your position to recover current collateral. You can also
+        monitor health closely and repay debt to keep the position open.
+        <br />
+        If health reaches 0, your position is hard-liquidated.
```
0xPearce: Inconsistent with the description of hard liquidation 

```suggestion
        If health reaches 0, your position can be hard-liquidated.
```

apps/main/src/llamalend/llama.utils.ts
```diff
@@ -212,6 +212,23 @@ const getSoftLiquidationThreshold = (userIsCloseToLiquidation: boolean) => (user
...
+  userIsCloseToSoftLiquidation: boolean,
+  userIsBelowRange: boolean,
+  userStateCollateral: Decimal | undefined,
+  userStateBorrowed: Decimal | undefined,
+) {
+  if (healthNotFull == null || userStateCollateral == null || userStateBorrowed == null) return undefined
+  const threshold = getSoftLiquidationThreshold(userIsCloseToSoftLiquidation)
+  if (+healthNotFull < 0) return 'hardLiquidation' as const
+  if (userIsBelowRange && +userStateCollateral > threshold) return 'incompleteConversion' as const
+  if (userIsBelowRange && +userStateCollateral <= threshold) return 'fullyConverted' as const
```
0xPearce: we shouldn't use the same threshold for collateral !! we should use the usd value of the collateral? If the collateral is WBTC here, 0.05 WBTC would be considered as dust and the position as fully converted, even though it represent +3k $

apps/main/src/llamalend/queries/market-list/llama-market-stats.ts
```diff
@@ -70,11 +70,20 @@ export function useUserMarketStats(market: LlamaMarket, column?: LlamaMarketColu
 
   const borrowedAmount = stats ? ('borrowed' in stats ? stats.borrowed : stats.stablecoin) : 0
 
+  const status = stats
+    ? getLiquidationStatus(
+        decimal(stats.health),
+        stats.softLiquidation,
+        !stats.softLiquidation && borrowedAmount > 0,
```
0xPearce: for the `userIsBelowRange` how about the case that user can retain a small converted borrow token after they already exited soft liquidation? In that case positions that are back above range will be shown as fully converted

apps/main/src/llamalend/features/market-position-details/hooks/useBorrowPositionDetails.ts
```diff
@@ -119,8 +119,13 @@ export const useBorrowPositionDetails = ({
...
     )
-    // userBandsValue[1] is the lower price boundary of a users range. Band numbers go up in numbers when going down in prices
-    const isUserBelowRange = oraclePriceBand && userBandsValue[1] ? oraclePriceBand > userBandsValue[0] : false
+    // Band numbers go up as prices go down (e.g. band 20 = 100-200, band 21 = 50-100), so the
+    // higher band number (userBandsValue[1]) is the lower price boundary of the user's range.
+    const [upperBoundary, lowerBoundary] = userBandsValue
+    const isUserBelowRange =
+      oraclePriceBand != null && upperBoundary != null && lowerBoundary != null
+        ? oraclePriceBand > lowerBoundary
+        : false
```
DanielSchiavini:
```suggestion
      oraclePriceBand != null && lowerBoundary != null && oraclePriceBand > lowerBoundary
```

DanielSchiavini (dismissed):

apps/main/src/llamalend/llama.utils.ts
OnlyJousting: ah ofc, I wasn't sure if threshold made sense to use anyways but this makes it clear that it's not

apps/main/src/llamalend/queries/market-list/llama-market-stats.ts
OnlyJousting: the best would be to be able to look at current oracle band and user range band numbers like on the market page, maybe we can request this 🤔

apps/main/src/llamalend/position-status-content.tsx
DanielSchiavini: We should be consistent, we use "-" everywhere

OnlyJousting: I'll change them to - then

apps/main/src/llamalend/queries/market-list/llama-market-stats.ts
OnlyJousting: yea Benny is adding current oracle band to the response so we can simply check if we've passed the range

DanielSchiavini (dismissed):

apps/main/src/llamalend/queries/market-list/llama-market-stats.ts
OnlyJousting: added now

0xtutti (dismissed):

0xAlunara (approved):

apps/main/src/llamalend/queries/market-list/llama-market-stats.ts
```diff
@@ -70,11 +70,20 @@ export function useUserMarketStats(market: LlamaMarket, column?: LlamaMarketColu
...
+      )
+    : undefined
+
   return {
     ...(stats && {
       data: {
-        softLiquidation: stats.softLiquidation,
-        liquidated: stats.healthFull < 0 && borrowedAmount === 0,
+        status,
         health: stats.healthFull,
```
Pearce542: we use `stats.health` to calculate the status, but then we display the `stats.healthFull` which can create a mismatch

apps/main/src/llamalend/features/market-list/cells/HealthCell.tsx
```diff
@@ -15,13 +15,14 @@ const { Spacing } = SizesAndSpaces
...
+  const { health, status } = data ?? {}
+  const softLiquidation = status === 'softLiquidation'
   return health == null ? (
     error && <ErrorCell error={error} />
   ) : (
     <Tooltip
-      title={softLiquidation ? 'Liquidation Protection On' : 'Position active'}
-      body={<HealthTooltipContent softLiquidation={!!softLiquidation} />}
+      title={softLiquidation ? t`Liquidation Protection On` : t`Position active`}
+      body={<HealthTooltipContent softLiquidation={softLiquidation} />}
```
Pearce542: We should handle the other statuses as well (fully converted, hard liquidation, incompleteConversion etc) otherwise the tooltip content is wrong. Or at least invert `const healthy = status === 'healthy'`

Pearce542 (commented):
I don't know how you tested all the edge cases, or whether you're confident this will display the correct content in every case. We should absolutely add mocked tests like we did for the borrow and supply forms otherwise, we'll keep fixing one thing and breaking another.

apps/main/src/llamalend/features/market-list/cells/HealthCell.tsx
OnlyJousting: "Position active" is universal enough to not be incorrect but I guess we can just match the alerts and it would be even more accurate

OnlyJousting:
> I don't know how you tested all the edge cases, or whether you're confident this will display the correct content in every case. We should absolutely add mocked tests like we did for the borrow and supply forms otherwise, we'll keep fixing one thing and breaking another.

I think now that we use current band to determine if the position is below together with a soft liquidation flag that it's pretty straight forward to determine what the correct state is but I can add mock tests too

apps/main/src/llamalend/queries/market-list/llama-market-stats.ts
OnlyJousting: ah, stats.health is healthNotFull?

OnlyJousting: I'll double check all health things

OnlyJousting: nvm this is correct, you use healthNotFull to check if there's hard liquidation risk. the user table isn't utilizing both values to display health like we do in the market page though which might cause some confusion 🤔

0xtutti (dismissed):

0xAlunara (dismissed):

0xAlunara (approved):

0xtutti (approved):

PR #2388, opened 2026-04-17
refactor: redundant marketId params

Pearce542:
- refactor: avoid passing redundant `marketId` when passing `market` object


apps/main/src/lend/components/PageLendMarket/LendMarketPage.tsx
```diff
@@ -116,7 +115,7 @@ export const LendMarketPage = () => {
   const pageProps = {
     params,
     rChainId: chainId,
-    rOwmId: marketId,
+    rOwmId: marketId ?? '',
```
0xPearce, line 118: we could go deeper I think and remove it from the `pageProps` as well?
This wasn't a task assigned to me, I just found myself doing it because I was frustrated of seeing the `marketId` and the `market` together. But will create a proper task to remove it deeper if needed

0xAlunara: Have yet to do a proper review, but last time I tried this a few months ago (I thought having both `market` and `marketId` was silly too), I found out the hard way that they might deviate.

I believe that `marketId`, from the URL param, might differ from `market.id` as markets are loading asynchronously. That means that if you decide to rely on `market?.id` rather than `marketId` you could encounter issues regarding `undefined`, even though there's an id (or controller address) given in the URL.

Not saying that will be the case here, just putting it out before I do my review.

apps/main/src/llamalend/widgets/page-header/hooks/usePageHeader.ts
```diff
@@ -93,15 +93,14 @@ function buildSupplyRate({
...
-  marketId,
   market,
   blockchainId,
 }: {
   chainId: number
-  marketId: string
   market: LlamaMarketTemplate | null | undefined
   blockchainId: Chain | undefined
 }) => {
+  const marketId = market?.id
```
DanielSchiavini, line 103: This is not the same! MarketId is not redundant, most definitely if we want to implement the market pages without llamalend.js

DanielSchiavini (changes requested):

Pearce542:
closed following [this](https://github.com/curvefi/curve-frontend/pull/2388#discussion_r3109881969) comment

PR #2394, opened 2026-04-19
styling: update liquidation range bottom line color

OnlyJousting:
The variables for bottom lines in liquidation zones had drifted with changes in the Line color variables at the same time as the lines have expanded usage in other charts. The design sets `Yellow[500]` in all themes for the bottom line while the corresponding line yellow line variable has some different shades between the themes. I added specific variables for the liquidation zone range lines in order to have a variable (instead of setting Yellow[500]) in order to easier maintain consistency in case of future changes, it also made sense to add variables for "top" and "future" lines.

Changes:
- Added `CurrentTopLine`, `CurrentBottomLine` and `FutureLine` (same is used for top and bottom) variables for liquidation zone border lines in bands and candle charts
- Update bands and candle chart palettes to use new variables


0xPearce (approved):

DanielSchiavini (approved):
This should be applied in the design system too for consistency! @0xtutti

OnlyJousting:
> This should be applied in the design system too for consistency! @0xtutti

I made a comment in the notion ticket tagging tutti about this too so we can get everything in sync

PR #2399, opened 2026-04-20
feat: derive borrow token symbol instead of hard coding crvusd

OnlyJousting:
Changes:
- Derive borrow token symbol from data in places that was still assuming crvUSD would be the borrow token

### DIFF
--- a/apps/main/src/llamalend/widgets/tooltips/CollateralMetricTooltipContent.tsx
+++ b/apps/main/src/llamalend/widgets/tooltips/CollateralMetricTooltipContent.tsx
@@ -22,8 +22,8 @@ export const CollateralMetricTooltipContent = ({ collateralValue }: CollateralMe
     collateralValue?.collateral?.usdRate,
   )
 
-  const crvUSDValueFormatted = formatMetricValue(collateralValue?.borrow?.value)
-  const crvUSDPercentage = formatPercentage(
+  const borrowValueFormatted = formatMetricValue(collateralValue?.borrow?.value)
+  const borrowPercentage = formatPercentage(
     collateralValue?.borrow?.value,
     collateralValue?.totalValue,
     collateralValue?.borrow?.usdRate,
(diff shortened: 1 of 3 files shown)


apps/main/src/llamalend/widgets/tooltips/CollateralMetricTooltipContent.tsx
```diff
@@ -37,7 +37,7 @@ export const CollateralMetricTooltipContent = ({ collateralValue }: CollateralMe
   return (
     <TooltipWrapper>
       <TooltipDescription
-        text={t`Collateral value is taken by multiplying tokens in collateral by the oracle price. In soft liquidation, it may include crvUSD due to liquidation protection.`}
+        text={t`Collateral value is taken by multiplying tokens in collateral by the oracle price. In soft liquidation, it may include ${collateralValue?.borrow.symbol} due to liquidation protection.`}
```
0xAlunara: No nullable fallback?

0xAlunara (dismissed):

apps/main/src/llamalend/widgets/tooltips/CollateralMetricTooltipContent.tsx
OnlyJousting: fixed it 🤝

0xtutti (approved):

0xAlunara (approved):

apps/main/src/llamalend/widgets/tooltips/CollateralMetricTooltipContent.tsx
```diff
@@ -34,21 +34,24 @@ export const CollateralMetricTooltipContent = ({ collateralValue }: CollateralMe
       ? formatUsd(collateralValue.totalValue, { abbreviate: false })
       : UnavailableNotation
 
+  const collateralTokenSymbol = collateralValue?.collateral?.symbol ?? '?'
```
DanielSchiavini, line 37: we shouldn't be hardcoding '?' in components, it should be in a shared formatting function imo

PR #2400, opened 2026-04-20
fix: borrow more with zapv2

DanielSchiavini:
- pass the maxRecv parameter in formatted units instead of raw wei
- make the route ID more stable across routers
- simplify invalidation, no need for refetching at all (in both cases stale-while-invalidate is applied anyway)
- make sure slippage is available in the repay arguments


apps/main/src/llamalend/features/borrow/hooks/useCreateLoanForm.tsx
```diff
@@ -148,8 +148,8 @@ export function useCreateLoanForm<ChainId extends LlamaChainId>({
       ...pick(params, 'slippage', 'routeId'),
       enabled: params.leverageEnabled && !!market && hasZapV2(market),
       onChange: async (route: RouteResponse | undefined) => {
+        await invalidateCreateLoanRouteQueries(route, params)
```
0xPearce, line 151: why did you changed the invalidation order? It can take a while before the action can be seen in the UI

DanielSchiavini: I thought the update form might miss some new data, but I don't think it was necessary in the end

0xAlunara (approved):

0xtutti (approved):

PR #2407, opened 2026-04-22
refactor: market list layout (part 1)

Pearce542:
As part of improving the search and filtering experience in the market list, this work will be delivered in multiple steps. This PR represents the first step and focuses on moving the existing components to their final structure. Subsequent PRs will introduce the new filters within a popover, along with additional refinements to fully match the updated design. You will notice that during this transition, chain chips will be temporarily unavailable (Beta only).

While refactoring the market table, and keeping the current implementation intact for stable and legacy channels, I chose to duplicate the relevant components rather than introduce multiple `if`, as the number of changes is significant.

The new components follow the same naming as the originals, with a New prefix (e.g., `DataTable` → `NewDataTable`). When the refactor is complete across the user position table and the pool list table as well, the original components will be removed and the new ones will be renamed.

Next PR: [refactor Chip component](https://github.com/curvefi/curve-frontend/pull/2411)

Screenshot 2026-04-24 at 17 03 08

### DIFF
--- a/apps/main/src/llamalend/features/market-list/chips/NewFilterChip.tsx
+++ b/apps/main/src/llamalend/features/market-list/chips/NewFilterChip.tsx
@@ -1,0 +1,53 @@
+import { Dispatch, ReactNode, SetStateAction } from 'react'
+import { LlamaMarket } from '@/llamalend/queries/market-list/llama-markets'
+import { useIsMobile } from '@ui-kit/hooks/useBreakpoints'
+import { t } from '@ui-kit/lib/i18n'
+import { FilterIcon } from '@ui-kit/shared/icons/FilterIcon'
+import { GridChip } from '@ui-kit/shared/ui/DataTable/chips/GridChip'
+import type { FilterProps } from '@ui-kit/shared/ui/DataTable/data-table.utils'
+import { MarketRateType } from '@ui-kit/types/market'
+import { LlamaMarketColumnId } from '../columns'
+import { NewMarketListFilterDrawer } from '../drawers/NewMarketListFilterDrawer'
+
+type Props = {
+  filterExpanded: boolean
+  setFilterExpanded: Dispatch<SetStateAction<boolean>>
+  hiddenCount: number
+  resetFilters: () => void
+  children?: ReactNode
+  hasFavorites?: boolean
+  data: LlamaMarket[]
+  userPositionsTab?: MarketRateType
+} & FilterProps<LlamaMarketColumnId>
+
+export const NewFilterChip = ({
+  filterExpanded,
+  setFilterExpanded,
+  hiddenCount,
+  resetFilters,
+  hasFavorites,
+  data,
+  userPositionsTab,
+  ...filterProps
+}: Props) => {
+  const isMobile = useIsMobile()
+  return isMobile ? (
+    <NewMarketListFilterDrawer
+      hasFavorites={hasFavorites}
+      data={data}
+      hiddenCount={hiddenCount}
+      resetFilters={resetFilters}
+      userPositionsTab={userPositionsTab}
+      {...filterProps}
+    />
+  ) : (
+    <GridChip
+      label={t`Filters`}
+      selectableChipSize="large"
+      selected={filterExpanded}
+      icon={<FilterIcon />}
+      toggle={() => setFilterExpanded((prev) => !prev)}
+      data-testid="btn-expand-filters"
+    />
+  )
+}
--- a/apps/main/src/llamalend/features/market-list/chips/NewLlamaListChips.tsx
+++ b/apps/main/src/llamalend/features/market-list/chips/NewLlamaListChips.tsx
@@ -1,0 +1,76 @@
+import { type ReactNode } from 'react'
+import { LlamaMarket } from '@/llamalend/queries/market-list/llama-markets'
+import Grid from '@mui/material/Grid'
+import { OnChangeFn, SortingState } from '@tanstack/react-table'
+import { useIsMobile } from '@ui-kit/hooks/useBreakpoints'
+import { t } from '@ui-kit/lib/i18n'
+import { HeartIcon } from '@ui-kit/shared/icons/HeartIcon'
+import { GridChip } from '@ui-kit/shared/ui/DataTable/chips/GridChip'
+import type { FilterProps } from '@ui-kit/shared/ui/DataTable/data-table.utils'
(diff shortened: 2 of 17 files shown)


apps/main/src/llamalend/features/market-list/chips/NewLlamaListChips.tsx
```diff
@@ -0,0 +1,31 @@
...
+const { Spacing } = SizesAndSpaces
+
+type LlamaListChipsProps = {
+  hasFavorites: boolean | undefined
+} & FilterProps<LlamaMarketColumnId>
+
+export const NewLlamaListChips = ({ hasFavorites, ...filterProps }: LlamaListChipsProps) => {
+  const [favorites, toggleFavorites] = useToggleFilter(LlamaMarketColumnId.IsFavorite, filterProps)
+  return (
+    <Grid container spacing={Spacing.sm} size={{ mobile: 12, tablet: 'auto' }}>
```
Pearce542, line 19: I left the grid container because there will be another chip in here soon

OnlyJousting:
When looking through the changes I noticed the filters are a little wonky, this is present in prod so not introduced in this PR but maybe it's a quick fix the new components can have? 

SCR-20260428-mzkt
SCR-20260428-mzhp
SCR-20260428-mzny

OnlyJousting (approved):

0xtutti (approved):

packages/curve-ui-kit/src/shared/ui/DataTable/NewTableSearchField.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+type Props = {
+  value: string
+  placeholder?: string
+  onChange: (value: string) => void
+  testId?: string
+  disableAutoFocus?: boolean
+}
+
+export const NewTableSearchField = ({ value, onChange, testId, disableAutoFocus }: Props) => (
```
0xAlunara, line 12: It's used only once, no? Not sure if that warrants its own custom component vs just inlining the `SearchField` directly.

Alternatively we could have it as a non-exported local component. Similar case for `NewTableFiltersHeader`. Although this is really just a personal preference of mine, maybe I'm the only one. Just less lines of code in general that way, no need to define types or props or whatever.

packages/curve-ui-kit/src/shared/ui/DataTable/NewTableSearchField.tsx
```diff
@@ -0,0 +1,21 @@
...
+  <SearchField
+    placeholder="Search"
```
0xAlunara, line 14: t`Search`

packages/curve-ui-kit/src/shared/ui/DataTable/NewTableFiltersHeader.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
+
+const { Spacing } = SizesAndSpaces
+
+export const NewTableFiltersHeader = ({ title, rightChildren }: { title: string; rightChildren?: ReactNode }) => (
+  <Stack
+    direction="row"
+    justifyContent="space-between"
+    alignItems="end"
+    // background needed because table head has a dfault color, and transparent is not possible because table head is sticky
```
0xAlunara, line 13: `dfault` -> `default`

packages/curve-ui-kit/src/shared/ui/DataTable/NewTableFiltersHeader.tsx
```diff
@@ -0,0 +1,21 @@
...
+    sx={{ backgroundColor: (t) => t.design.Layer.App.Background }}
+    // cannot use Stack + gap because the background color need to be this one and not the default
```
0xAlunara, line 15: I'm confused by this comment. How exactly does gap vs padding have an impact on background color?

0xAlunara (approved):

packages/curve-ui-kit/src/shared/ui/DataTable/NewTableSearchField.tsx
Pearce542: Agree with the search field. The header I prefer to make it a seperate component, because it is very likely different for each table (markets, pools, user position)

packages/curve-ui-kit/src/shared/ui/DataTable/NewTableFiltersHeader.tsx
Pearce542: Instead of using `paddingBlockEnd`, you could add a Stack with a gap to the parent component. The difference is that when using padding, the extra space inherits the child’s background color (in this case, `t.design.Layer.App.Background`). With the Stack + gap approach, the gap instead takes on the parent’s background color, and we don't want that.

Pearce542:
thanks @0xAlunara I will implement the changes in the follow up PR

apps/main/src/llamalend/features/market-list/chips/NewFilterChip.tsx
```diff
@@ -0,0 +1,53 @@
...
+  setFilterExpanded: Dispatch<SetStateAction<boolean>>
+  hiddenCount: number
+  resetFilters: () => void
+  children?: ReactNode
+  hasFavorites?: boolean
+  data: LlamaMarket[]
+  userPositionsTab?: MarketRateType
+} & FilterProps<LlamaMarketColumnId>
+
+export const NewFilterChip = ({
```
DanielSchiavini, line 23: using the 'New' prefix will just end up creating extra diffs later on

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
```diff
@@ -228,21 +216,7 @@ testCases.forEach(([width, height, breakpoint]) => {
...
-
-      cy.get(`[data-testid="data-table-cell-assets"]:first [data-testid="chain-icon-${chain}"]`).should('be.visible')
-
-      const otherChain = oneOf(...chains.filter((c) => c !== chain))
-      cy.get(`[data-testid="chip-chain-${otherChain}"]`).click()
-      ;[chain, otherChain].forEach((c) => cy.get(`[data-testid="chain-icon-${c}"]`).should('be.visible'))
-      cy.url().should('include', chain)
-      cy.url().should('include', otherChain)
-    })
+    // todo: filter by chain
```
DanielSchiavini, line 219: tests were deleted, do we have tasks to restore them?

PR #2411, opened 2026-04-23
styling: chip component

Pearce542:
Depends on #2407 

We are using the MUI Chip component as a Chip and Badge. This PR creates a `Badge` component and updates the `SelectableChip` in sync with Figma. Added also eslint restricted-imports so we use either the `Badge` or `SelectableChip` component

- Sync SelectableChip component with Figma
- Create Badge component (wrapper of non selectable MUI Chip)
- split and update stories

Screenshot 2026-04-24 at 17 08 40


OnlyJousting (dismissed):

packages/curve-ui-kit/src/shared/ui/SelectableChip.tsx
```diff
@@ -1,32 +1,31 @@
...
+  outlined: undefined,
+  ghost: 'ghost',
+}
 /**
  * Renders a chip that can be selected or deselected.
  * This customizes the MUI Chip component to change color and icon based on selection state.
  * The delete icon is always visible, but hidden when the chip is not selected with some transition.
  */
-export const SelectableChip = ({ selected, toggle, onDelete, ...props }: SelectableChipProps) => (
+export const SelectableChip = ({ selected, toggle, variant = 'outlined', ...props }: SelectableChipProps) => (
```
0xAlunara, line 22: Would love to just call this 'Chip' if possible (I'm not aware we have unselectable chips now that `Badge` is its own component)

0xAlunara (dismissed):

packages/curve-ui-kit/src/shared/ui/SelectableChip.tsx
Pearce542: Same but we already have another one called `Chip` in the ui kit we should rename / refactor first

0xAlunara (approved):

OnlyJousting (approved):

packages/curve-ui-kit/src/features/integrations/components/IntegrationsList.tsx
```diff
@@ -109,8 +112,8 @@ export const IntegrationsList = ({ networkId, searchText }: { networkId?: string
         <Stack gap={Spacing.xs} direction="row" flexWrap="wrap">
           {Object.values(tags).map((tag) => (
             <SelectableChip
+              size={getDefaultSelectableChipSize(isMobile)}
```
DanielSchiavini, line 115: looks like we need a size='default'

packages/curve-ui-kit/src/shared/ui/Chart/ToggleBandsChartButton.tsx
```diff
@@ -1,29 +1,29 @@
...
   label: string
   isVisible: boolean
-  onClick: () => void
+  toggle: () => void
 }) => (
-  <Chip
-    color="unselected"
-    size="small"
-    onClick={onClick}
+  // todo: refactor chip to button
```
DanielSchiavini, line 18: why?

packages/eslint-config-custom/index.js
```diff
@@ -72,10 +72,25 @@ module.exports = {
...
+              "Use `import { SelectableChip } from '@ui-kit/shared/ui/SelectableChip'` or `import { Badge } from '@ui-kit/shared/ui/Badge'` instead.",
+          },
+        ].map((component) =>
+          typeof component === 'string'
+            ? {
+                name: `@mui/material/${component}`,
+                message: `Use \`import { ${component} } from '@ui-kit/shared/ui/${component}'\` instead.`,
+              }
+            : component,
+        ),
```
DanielSchiavini, line 93:
```suggestion
        paths: [
          ...['Accordion', 'Tabs', 'Slider', 'Tooltip', 'Select'].map((component) => ({
            name: `@mui/material/${component}`,
            message: `Use \`import { ${component} } from '@ui-kit/shared/ui/${component}'\` instead.`,
          })),
          {
            name: '@mui/material/Chip',
            message:
              "Use `import { SelectableChip } from '@ui-kit/shared/ui/SelectableChip'` or `import { Badge } from '@ui-kit/shared/ui/Badge'` instead.",
          },
        ]
```

packages/curve-ui-kit/src/shared/ui/Chart/ToggleBandsChartButton.tsx
Pearce542: it's not being used as a chip but as a button. I have activated the selected property now, so it acts like a chip

packages/curve-ui-kit/src/themes/basic-theme/basic-theme.ts
```diff
@@ -70,3 +70,9 @@ export const mapBreakpoints = (
   values: Responsive,
   callback: (value: string, breakpoint: Breakpoint) => string,
 ): CSSObject => mapRecord(values, (breakpoint, value) => callback(value, breakpoint))
+
+export const fixedResponsive = <T extends string>(value: T): Responsive<T> => ({
```
DanielSchiavini, line 74: Why would this be needed at all? We can just pass `height: value` instead of this mapping

PR #2415, opened 2026-04-23
fear: soft liq close position sheet

0xAlunara:
The close position tab of the manage soft liquidation form is getting a presentational revamp: it's now a datatable, explaining the calculations behind its singular button.

PS: Numbers don't have the desired precision yet, I'm postponing that to a different PR

**Bad debt**
image

**Some dude in soft liq**
image

**Component test**
image


0xtutti (dismissed):

apps/main/src/llamalend/features/manage-soft-liquidation/hooks/useClosePositionForm.ts
```diff
@@ -33,8 +39,43 @@ export function useClosePositionForm({
...
+    error: userStateError,
+    isLoading: userStateLoading,
+  } = useUserState({ chainId, marketId, userAddress }, enabled)
+
+  const { collateral, debt, stablecoin } = userState ?? {}
+
+  // Loading and error state, which are derived from token data hooks above
+  const isLoading =
+    userStateLoading || userBalancesLoading || borrowTokenUsdRateLoading || collateralTokenUsdRateLoading
+  const error = userStateError || userBalancesError || borrowTokenUsdRateError || collateralTokenUsdRateError
```
Pearce542: IMO we shouldn't disable the form if the token balances are loading or there is an error. It's just additional information, to close the position you should only need loan balances and approval state. I would show the loading and error of the token usd rate in the table for example

apps/main/src/llamalend/features/manage-soft-liquidation/hooks/useClosePositionForm.ts
```diff
@@ -49,28 +90,146 @@ export function useClosePositionForm({
...
+      },
+    [borrowToken, borrowTokenUsdRate, excess],
+  )
+
+  // Table creation
+  const tableData: ClosePositionRow[] = useMemo(
+    () =>
+      notFalsy(
+        { label: t`Collateral`, value: notFalsy(collateralAmount, stablecoinAmount) },
+        { label: t`Outstanding debt`, value: notFalsy(debtAmount), testId: 'outstanding-debt' },
```
Pearce542, line 172: Not sure about these `notFlasy`, if the value of `collateralAmount` and `stablecoinAmount` is undefined we show it as 0$? We should display "-" imo if either one is undefined

apps/main/src/llamalend/features/manage-soft-liquidation/hooks/useClosePositionForm.ts
0xAlunara: Ah yeah fair enough, I've removed the disabling of the form while loading.

apps/main/src/llamalend/features/manage-soft-liquidation/hooks/useClosePositionForm.ts
0xAlunara: In the case of the `value` prop, I've only used it as a shorthand way of converting it both to an array and also getting rid of `undefined`. In practise the table won't show until `isLoading` is false, which means it'll never really be `undefined` anyway. I'd like to keep the table skeleton with `isLoading` for the time being. I can make a new ticket for getting rid of the global table skeleton in favor of individual loading states of the rows, although I'm not particulary in favor as the rows might pop up with a delay.

0xtutti (approved):

apps/main/src/llamalend/features/manage-soft-liquidation/hooks/useClosePositionForm.ts
```diff
@@ -33,8 +39,40 @@ export function useClosePositionForm({
...
+
+  const {
+    data: userState,
+    error: userStateError,
+    isLoading: userStateLoading,
+  } = useUserState({ chainId, marketId, userAddress }, enabled)
+
+  const { collateral, debt, stablecoin } = userState ?? {}
+
+  const error = userStateError || userBalancesError || borrowTokenUsdRateError || collateralTokenUsdRateError
```
Pearce542, line 73: We still disable the form if one of the token USD query fails which we shouldn't do imo, but can be refactored in another PR if you want

Pearce542 (approved):

apps/main/src/llamalend/features/manage-soft-liquidation/hooks/useClosePositionForm.ts
DanielSchiavini: @0xAlunara the problem is the other  way around, notFalsy will filter out 0's which seems incorrect?!

apps/main/src/llamalend/features/manage-soft-liquidation/ui/cells/LabelCell.tsx
```diff
@@ -0,0 +1,31 @@
...
+    {label}
+  </Typography>
+)
+
+export const LabelCell = ({ getValue }: CellContext<ClosePositionRow, ClosePositionRow['label'] | undefined>) => {
+  const value = getValue()
+  return (
+    value != null && (
+      // Min width of 16ch so labels don't wrap (biggest one is 'Paid from collateral')
+      <BaseCell sx={{ minWidth: '16ch' }}>
```
DanielSchiavini, line 26: this looks super fragile, we should at least derive the 16

apps/main/src/llamalend/features/manage-soft-liquidation/hooks/useClosePositionForm.ts
0xAlunara: I think that's fine, if a row is 0 it doesn't apply and we dont want to show it. Most of the undefineds are only there because of decimal arithmetic,  even if there is a number available originally.

apps/main/src/llamalend/features/manage-soft-liquidation/ui/cells/BaseCell.tsx
```diff
@@ -0,0 +1,12 @@
+import type { ReactNode } from 'react'
+import Stack from '@mui/material/Stack'
+import type { SxProps, Theme } from '@mui/material/styles'
+import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
+
+const { Spacing } = SizesAndSpaces
+
+export const BaseCell = ({ sx, children }: { sx?: SxProps<Theme>; children: ReactNode }) => (
+  <Stack paddingBlock={Spacing.md} sx={{ cursor: 'pointer', ...sx }}>
```
DanielSchiavini, line 9: why do we have cursor:pointer without an event handler?

apps/main/src/llamalend/features/manage-soft-liquidation/ui/cells/BaseCell.tsx
```diff
@@ -0,0 +1,12 @@
+import type { ReactNode } from 'react'
+import Stack from '@mui/material/Stack'
+import type { SxProps, Theme } from '@mui/material/styles'
+import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
+
+const { Spacing } = SizesAndSpaces
+
+export const BaseCell = ({ sx, children }: { sx?: SxProps<Theme>; children: ReactNode }) => (
```
DanielSchiavini, line 8: BaseCell doesn't explain why we need this component - and why we don't have it on the generic rows.
Can we think of any names related to the actual requirements? And maybe we want to include Stack in the name?

apps/main/src/llamalend/features/manage-soft-liquidation/ui/cells/LabelCell.tsx
0xAlunara: Found a different solution that doesn't involve min width (`white-space: nowrap`)

apps/main/src/llamalend/features/manage-soft-liquidation/ui/cells/BaseCell.tsx
0xAlunara: Pointer was a left over from copying the implementation of the market supply incentives table. Now that it's removed the only thing that remains was a Stack with padding, so decided to just inline it to get rid of the base cell altogether.

PR #2423, opened 2026-04-25
feat: add support for market merkl campaigns

0xAlunara:
Technically should add support for Merkl campaigns for lending markets for both supplying and borrowing, but it's a bit hard to test because their test infrastructure doesn't return `rewardsRecord.breakdowns` which we rely on. Merkl team said this is only the case for their testing infrastructre, and the data should look like pools in prod.

But this means that we can't properly test everything works until there's a live lending campaign.

[Test borrow campaign](https://app.merkl.xyz/opportunities/optimism/LLAMALEND_V2_BORROW/0xA8d25cAc55f1453900c38746751E83a37481630b?test=true)
[Test supply campaign](https://app.merkl.xyz/opportunities/optimism/LLAMALEND_V2_SUPPLY/0x3410ad9C785A32c489feB5973edC3b167DaC6Ff8?test=true)
[Test API data](https://api.merkl.xyz/v4/opportunities?mainProtocolId=llamalend&test=true)


packages/external-rewards/src/campaigns/Etherfi.json
```diff
@@ -127,7 +127,7 @@
     },
     {
       "id": "null",
-      "action": "loan",
+      "action": "borrow",
```
0xAlunara, line 130: This was the only external campaign with a `loan` action. I've changed it to `borrow` so there's no distinction from borrowing from mint or lend markets, which mimics Merkl more closely as they lack that distinction too.

Hence, the `loan` option has been removed from `RewardsAction` as well

0xtutti (dismissed):

Pearce542 (approved):

packages/ui/src/CampaignRewards/CampaignRewardsComp.tsx
```diff
@@ -2,10 +2,10 @@ import { styled } from 'styled-components'
 import { TooltipMessage } from '@ui/CampaignRewards/TooltipMessage'
 import { Icon } from '@ui/Icon'
 import { TooltipButton as Tooltip } from '@ui/Tooltip/TooltipButton'
-import type { CampaignPoolRewards } from '@ui-kit/entities/campaigns'
+import type { CampaignRewards } from '@ui-kit/entities/campaigns'
 
 type CampaignRewardsCompProps = {
-  rewardsPool: CampaignPoolRewards
+  rewardsPool: CampaignRewards
```
OnlyJousting, line 8: could have gotten rid of `Pool` here too

OnlyJousting (approved):

PR #2432, opened 2026-04-28
styling: shorten low crv emissions aprs to just < 0.01%

0xAlunara:
This component is still using the old school number formatter, so I just opted for this quick custom behavior instead of adjusting fornatting. There's another ticket for coming up with categories and such.

Opted comparing base value rather than boosted, because otherwise we'd still end up with stuff like `0.0000004% -> 0.1%`

image

### DIFF
--- a/apps/main/src/dex/components/PoolRewardsCrv.tsx
+++ b/apps/main/src/dex/components/PoolRewardsCrv.tsx
@@ -29,6 +29,7 @@ export const PoolRewardsCrv = ({
     } else if (rewardsApy?.crv && (rewardsApy?.crv[0] !== 0 || rewardsApy?.crv[1] !== 0)) {
       const [base, boosted] = rewardsApy.crv
       if (!base && !boosted) return null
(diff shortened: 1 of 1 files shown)


mimaklas (approved):

Pearce542 (approved):

apps/main/src/dex/components/PoolRewardsCrv.tsx
```diff
@@ -29,6 +29,7 @@ export const PoolRewardsCrv = ({
     } else if (rewardsApy?.crv && (rewardsApy?.crv[0] !== 0 || rewardsApy?.crv[1] !== 0)) {
       const [base, boosted] = rewardsApy.crv
       if (!base && !boosted) return null
+      if (base < 0.01) return '< 0.01% CRV' // Anything less is basically not worth the time
```
DanielSchiavini, line 32: imo this should be sharing the format with the code below, and the 0.01 shouldn't be hardcoded

PR #2433, opened 2026-04-28
fix: always use usd rates for percentages in currency reserves if available

0xAlunara:
Turns out the USD rates for currency reserve percentages were only used if the pool was a crypto pool, but I think we should always use the USD rates if they're available, even if it's a stable pool (so that depegs are also more accurately presented?)

PS: I tried to refactor using queries and prices API, but the tech debt makes it impossible (due to how pool data is not querified yet).

**New**
image

**Old**
image

### DIFF
--- a/apps/main/src/dex/store/createPoolsSlice.ts
+++ b/apps/main/src/dex/store/createPoolsSlice.ts
@@ -235,19 +235,16 @@ export const createPoolsSlice = (set: StoreApi<State>['setState'], get: StoreApi
         crTokens.push(crToken)
       }
 
-      // update percentShareInPool
-      crTokens.map((cr: CurrencyReservesToken) => {
+      for (const cr of crTokens) {
         if (isEmpty) {
           cr.percentShareInPool = '0'
-        } else if (poolData.pool.isCrypto && isNaN(cr.usdRate)) {
-          cr.percentShareInPool = 'NaN'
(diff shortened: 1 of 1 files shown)


Pearce542 (approved):

0xtutti (approved):

apps/main/src/dex/store/createPoolsSlice.ts
```diff
@@ -235,19 +235,16 @@ export const createPoolsSlice = (set: StoreApi<State>['setState'], get: StoreApi
...
-      // update percentShareInPool
-      crTokens.map((cr: CurrencyReservesToken) => {
+      for (const cr of crTokens) {
         if (isEmpty) {
           cr.percentShareInPool = '0'
-        } else if (poolData.pool.isCrypto && isNaN(cr.usdRate)) {
-          cr.percentShareInPool = 'NaN'
-        } else if (poolData.pool.isCrypto) {
+          // Only use USD balances for currency reserves if all tokens have a usd balance (and pool isn't empty)
+        } else if (crTokens.every((cr) => cr.balanceUsd > 0)) {
```
DanielSchiavini, line 242: Could this ever be negative?
```suggestion
          // Only use USD balances for currency reserves if all tokens have a usd balance (and pool isn't empty)
        } else if (crTokens.every((cr) => cr.balanceUsd)) {
```

PR #2436, opened 2026-04-28
feat: verify wagmi chains being loaded for supported networks

0xAlunara:
Verifies (in a non-breaking way) that we have loaded and are using the corresponding wagmi configs of all non-testnet chains we support in the chain list.

1. We were missing a few, I've added them
2. Apparently the TAC chain id should be 239, not 2390

PS: I figured I could turn the chain list into a Stack with gap so it looks nicer, but it breaks the scroll behavior, so I kept it gapless.

image


mimaklas (approved):

Pearce542 (approved):

packages/curve-ui-kit/src/features/switch-chain/ui/ChainList.tsx
```diff
@@ -47,8 +48,17 @@ export function ChainList({
   }
 
   const entries = Object.entries(groupedOptions)
+  const missingWagmiChains = options.filter(({ isTestnet, chainId }) => !isTestnet && !wagmiChainsMap[chainId])
+
   return (
     <>
+      {missingWagmiChains.length > 0 && (
```
DanielSchiavini, line 55: I don't see why we want to check this in a 'non-breaking' way - imo we should just break everything so it's never missed.

packages/curve-ui-kit/src/features/switch-chain/ui/ChainList.tsx
```diff
@@ -47,8 +48,17 @@ export function ChainList({
   }
 
   const entries = Object.entries(groupedOptions)
+  const missingWagmiChains = options.filter(({ isTestnet, chainId }) => !isTestnet && !wagmiChainsMap[chainId])
```
DanielSchiavini, line 51: Why are we excluding testnets? They also need config.

0xAlunara: There's no special wagmi config for them

packages/curve-ui-kit/src/features/switch-chain/ui/ChainList.tsx
0xAlunara: We dont need to break the entire website just because some lite network has the wrong defaults or missing multicall

PR #2438, opened 2026-04-29
refactor: dani welcome-back comments

0xAlunara:
A variety of small time comments by dani for various PRs, nothing super big

### DIFF
--- a/apps/main/src/llamalend/features/manage-soft-liquidation/ui/cells/LabelCell.tsx
+++ b/apps/main/src/llamalend/features/manage-soft-liquidation/ui/cells/LabelCell.tsx
@@ -13,6 +13,7 @@ export const LabelCellDisplay = ({
   <Typography
     variant={isFooter ? 'tableCellMBold' : 'tableCellMRegular'}
     color={isFooter ? 'textPrimary' : 'textSecondary'}
+    whiteSpace="nowrap" // Aesthetics; it looks ugly if this column wraps
   >
     {label}
   </Typography>
@@ -22,8 +23,7 @@ export const LabelCell = ({ getValue }: CellContext<ClosePositionRow, ClosePosit
   const value = getValue()
   return (
     value != null && (
-      // Min width of 16ch so labels don't wrap (biggest one is 'Paid from collateral')
-      <BaseCell sx={{ minWidth: '16ch' }}>
+      <BaseCell>
         <LabelCellDisplay label={value} />
       </BaseCell>
     )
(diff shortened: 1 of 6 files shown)


apps/main/src/llamalend/features/manage-soft-liquidation/ui/cells/LabelCell.tsx
```diff
@@ -22,10 +26,9 @@ export const LabelCell = ({ getValue }: CellContext<ClosePositionRow, ClosePosit
   const value = getValue()
   return (
     value != null && (
-      // Min width of 16ch so labels don't wrap (biggest one is 'Paid from collateral')
-      <BaseCell sx={{ minWidth: '16ch' }}>
+      <Box paddingBlock={Spacing.md}>
```
0xAlunara, line 29: I was thinking that maybe we could add some default padding to `<TableCell>`, but this case of an inline table might be too niche as it'd be applied everywhere. Probably better for now if the default padding is zero and can be set as desired by the user of the table component.

0xtutti (dismissed):

packages/curve-ui-kit/src/shared/ui/DataTable/hooks/useFilters.tsx
```diff
@@ -102,8 +102,8 @@ export const useFilters = <TColumnId extends string>({
   searchKey = DEFAULT_SEARCH_KEY,
   ...columnFilterOptions
 }: Parameters<typeof useColumnFilters<TColumnId>>[0] & { searchKey?: string }) => {
-  const { resetGlobalFilter, ...globalFilter } = useGlobalFilter(searchKey)
-  const { resetFilters: resetColumnFilters, ...columnFilters } = useColumnFilters(columnFilterOptions)
+  const globalFilter = useGlobalFilter(searchKey)
```
DanielSchiavini, line 99: the reset functions are still unused 😁

0xAlunara: Oh, I figured they might come in handy later, but I can remove them at the root level as well

DanielSchiavini (approved):

packages/curve-ui-kit/src/shared/ui/DataTable/hooks/useFilters.tsx
DanielSchiavini: I don't see why we'd reset them separately now, that could lead to the same bug again. Thanks!

0xtutti (approved):

PR #2450, opened 2026-04-29
chore: update eslint to v10

0xAlunara:
Aaaaaaaaah it finally works! Giant diff, because of course I had to fix a load of new lint errors

1. Updates eslint to v10
2. Updates a few related linting packages like `@eslint-react/eslint-plugin` and `eslint-plugin-unused-imports`
3. Migrates from `eslint-plugin-import` to `eslint-plugin-import-x`
4. We now lint with both `recommendedTypeChecked` and `stylisticTypeChecked` presets from `typescript-eslint`

Linting is now a parallel yarn workspace script via `yarn workspaces foreach --parallel --all --exclude curve-frontend run lint`. This is because a single eslint command to lint everything at once would cause heap overflow errors. Apparantly Node.js memory usage would exceed 4GB, yikes!


OnlyJousting (dismissed):

packages/curve-ui-kit/src/utils/dom.ts
```diff
@@ -11,7 +11,7 @@ export function hasParentWithClass(target: EventTarget, className: string, { unt
     if (element.classList.contains(className)) {
       return true
     }
-    element = element.parentElement as HTMLElement
+    element = element.parentElement!
   }
```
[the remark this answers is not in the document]
0xAlunara: Yes but this is old code, it was already null asserted.

eslint.config.mjs
```diff
@@ -158,6 +176,30 @@ const config = [
...
+      '@typescript-eslint/no-duplicate-type-constituents': 'off',
+      '@typescript-eslint/no-misused-promises': 'off',
+      '@typescript-eslint/no-unsafe-argument': 'off',
+      '@typescript-eslint/no-unsafe-assignment': 'off',
+      '@typescript-eslint/no-unsafe-member-access': 'off',
+      '@typescript-eslint/no-unsafe-return': 'off',
+      '@typescript-eslint/no-unsafe-call': 'off',
+      '@typescript-eslint/no-unsafe-enum-comparison': 'off',
+      '@typescript-eslint/no-unsafe-enum-comparison': 'off',
+      '@typescript-eslint/no-unnecessary-type-assertion': 'off',
```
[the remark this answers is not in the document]
0xAlunara: Oops, too much copy paste!

packages/ui/package.json
```diff
@@ -4,8 +4,8 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "lint": "eslint .",
-    "lint:fix": "eslint . --fix",
+    "lint": "eslint --cache --cache-location ../../.cache/eslint --cache-strategy content .",
+    "lint:fix": "eslint --cache --cache-location ../../.cache/eslint --cache-strategy content . --fix",
```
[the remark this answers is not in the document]
0xAlunara: Hmmm, Claude may have lied to me. Is this true?

OnlyJousting (dismissed):

package.json
```diff
@@ -16,8 +16,8 @@
     "start": "yarn workspace main start",
     "dev": "yarn workspace main dev",
     "dev-router-api": "yarn workspace router-api dev",
-    "lint": "eslint --cache --cache-location .cache/eslint --cache-strategy content \"{apps,packages,tests}/**/*.{ts,tsx}\"",
-    "lint:fix": "eslint --cache --cache-location .cache/eslint --cache-strategy content \"{apps,packages,tests}/**/*.{ts,tsx}\" --fix",
+    "lint": "yarn workspaces foreach --parallel --all --exclude curve-frontend run lint",
+    "lint:fix": "yarn workspaces foreach --parallel --all --exclude curve-frontend run lint:fix",
```
Pearce542, line 20: the tests `package.json` is missing the lint script

packages/curve-ui-kit/package.json
```diff
@@ -7,8 +7,8 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "lint": "eslint .",
-    "lint:fix": "eslint . --fix",
+    "lint": "eslint --cache --cache-location ../../.cache/eslint/curve-ui-kit --cache-strategy content .",
+    "lint:fix": "eslint --cache --cache-location ../../.cache/eslint /curve-ui-kit --cache-strategy content . --fix",
```
Pearce542: Additional whitespace

```suggestion
    "lint:fix": "eslint --cache --cache-location ../../.cache/eslint/curve-ui-kit --cache-strategy content . --fix",
```

apps/main/src/dex/hooks/useChainId.ts
```diff
@@ -10,4 +11,4 @@ const useNetwork = (networkId: string) => {
 
 export const useNetworkFromUrl = () => useNetwork(useParams<NetworkUrlParams>().network)
 
-export const useChainId = (networkId: NetworkEnum): ChainId => useNetwork(networkId)?.chainId as ChainId
+export const useChainId = (networkId: NetworkEnum): ChainId => useNetwork(networkId)?.chainId ?? Chain.Ethereum
```
Pearce542: Why the fallback to Ethereum?

package.json
0xAlunara: Thought I could get away with it, but now you've spotted it 😞

apps/main/src/dex/hooks/useChainId.ts
0xAlunara: It was being null asserted when in reality it cannot be, and I had multiple lint rules conflicting with it. However, on second thought defaulting to Ethereum might be worse than letting it be undefined but typing it as if it can't be.

I'll just ignore the lint errors for now

0xtutti (approved):

Pearce542 (approved):

apps/main/src/dex/components/ChipPool.tsx
```diff
@@ -47,7 +46,7 @@ interface ChipPoolProps {
 }
 
 export const ChipPool = ({ poolId, poolName, poolAddress }: ChipPoolProps) => {
-  const network = useCurve().network?.id as NetworkEnum
+  const network = useCurve().network?.id ?? 'ethereum'
```
DanielSchiavini, line 49: doesn't seem correct

apps/main/src/dex/utils/index.ts
```diff
@@ -37,9 +37,9 @@ export function getCurvefiUrl(poolId: string, host: string) {
   const v2Key = 'factory-v2-'
   const cryptoKey = 'factory-crypto-'
 
-  if (poolId.match(v2Key)) {
+  if (RegExp(v2Key).exec(poolId)) {
     return `${host}/factory/${poolId.split(v2Key)[1]}`
-  } else if (poolId.match(cryptoKey)) {
+  } else if (RegExp(cryptoKey).exec(poolId)) {
```
DanielSchiavini, line 42: what is wrong with match btw?

apps/main/src/lend/hooks/useUserLoanDetails.ts
```diff
@@ -11,5 +11,5 @@ type Details = UserLoanDetails['details'] & { error: string }
 export function useUserLoanDetails(userActiveKey: string): Partial<Details> {
   const loanDetails = useStore(state => state.user.loansDetailsMapper[userActiveKey])
 
-  return !loanDetails || loanDetails.details == null ? {} : { ...loanDetails.details, error: loanDetails.error }
+  return loanDetails?.details == null ? {} : { ...loanDetails.details, error: loanDetails.error }
```
DanielSchiavini, line 14: Existing problem, but why is there no error being passed when there is no data?
```suggestion
  return { ...loanDetails.details, error: loanDetails.error }
```

DanielSchiavini (approved):

apps/main/src/dex/utils/index.ts
0xAlunara: Not much in particular, but it's one of the default stylistic rules: https://typescript-eslint.io/rules/prefer-regexp-exec/

PR #2451, opened 2026-04-30
refactor: introduce branded Timestamp type for Prices API

0xAlunara:
Introduces a branded `Timestamp` type for the Prices API, which has to be explicitly parsed to a `Date` using `toDate` in a lazy fashion. Prices API response models no longer contain actual Date objects, such that they're completely serializable and work properly with TanStack Query.


0xtutti (dismissed):

apps/main/src/dao/components/PageVeCrv/components/FieldDatePicker.tsx
```diff
@@ -169,7 +169,7 @@ export const FieldDatePicker = ({
         <Chip size="xs">
           {t`Current unlock date:`}{' '}
           {vecrvInfo.lockedAmountAndUnlockTime.unlockTime
-            ? formatDate(vecrvInfo.lockedAmountAndUnlockTime.unlockTime)
+            ? formatDate(new Date(vecrvInfo.lockedAmountAndUnlockTime.unlockTime))
```
DanielSchiavini: why do we use `new Date` in some places? It could lead to mistakes i'm afraid

packages/prices-api/src/timestamp.ts
```diff
@@ -1,3 +1,29 @@
+declare const TimestampBrand: unique symbol
+
+/**
+ * A serialized timestamp that can indicate many of the timestamp variants
+ * as returned by the Prices API.
```
DanielSchiavini: 80 characters in jsdoc again 😭 
"many of the timestamp variants as returned by the Prices API" sounds pretty bad - is the API that inconsistent?

Does it mean there are even more timestamp variants that aren't handled?

packages/prices-api/src/timestamp.ts
```diff
@@ -1,3 +1,29 @@
...
+ * a single, consistent parsing path and prevents accidental arithmetic
+ * (e.g. `timestamp * 1000`) or string concatenation on values whose unit
+ * (seconds vs. ms) and format (unix vs. ISO) are not guaranteed.
+ *
+ * Underlying shape is one of:
+ * - unix seconds as `number` (e.g. `1704067200`)
+ * - unix seconds as `string` (e.g. `"1704067200"`)
+ * - ISO 8601 date string (e.g. `"2024-01-01T00:00:00.000Z"`)
+ */
+export type Timestamp = (number | `${number}` | `${number}-${number}-${number}T${string}`) & {
```
DanielSchiavini: imo the prices-api package should be converting all timestamps to a single format. This implementation leaks the API inconsistencies into our components
```suggestion
export type SerializedDate = (`${number}-${number}-${number}T${string}`)
```

DanielSchiavini (dismissed):
I don't totally agree with the approach but I think this is a great improvement anyway

apps/main/src/dao/components/PageVeCrv/components/FieldDatePicker.tsx
0xAlunara: I'm forcing it `Date` casting everywhere by removing ` | number | string` from the parameter type of `formatDate`.

Branded types allows us to prohibit casting from `number` to `Timestamp`, but not from `Timestamp` to `number`. The `Date` constructor accepts a unix timestamp in milliseconds, while the prices api returns it in seconds, so you first need to multiply by 1000. Sometimes, like in this case with the on-chain call, we get it correctly in seconds.

Rather than figuring out which one needs `* 1000` and which one does not, I force users to either cast the `number` to `Date`, or use `toDate()` to cast `Timestamp` to `Date`.

Hence the entire thing is to prevent mistakes.

packages/prices-api/src/timestamp.ts
0xAlunara: No, just these 3 forms as described by the docs.

packages/prices-api/src/timestamp.ts
0xAlunara: Hmm not bad, that prevents accidental downcasting from `Timestamp` to `number` without multiplying by 1000 because of the seconds/milliseconds thing 🤔

apps/main/src/dao/components/PageVeCrv/components/FieldDatePicker.tsx
DanielSchiavini: I was just hoping we could keep the conversion part in the prices API client

0xAlunara: Just did that: the parsers now parse `TimestampResponse` (which can be 1 of 3 variants) into a `Timestamp` which can be passed into `new Date()`.

apps/main/src/dao/components/PageGauge/GaugeMetrics/index.tsx
```diff
@@ -60,7 +61,11 @@ export const GaugeMetrics = ({ gaugeData, dataLoading }: GaugeMetricsProps) => {
         <MetricsComp
           loading={dataLoading}
           title={t`Created`}
-          data={<StyledMetricsColumnData>{gaugeData && formatDate(gaugeData.creation_date)}</StyledMetricsColumnData>}
+          data={
+            <StyledMetricsColumnData>
+              {gaugeData && formatDate(parseTimestamp(gaugeData.creation_date))}
```
0xAlunara, line 66: There's like 3 usages of `parseTimestamp` outside the actual parsers for the gauge stuff. These pages aren't using the `prices-api` library properly, we need to fix that later.

0xAlunara:
1. API responses now expect the timestamps to be of type `(number | `${number}` | IsoDateString)`
2. The parsers now convert it to a (branded and serializable) `Timestamp`, which is a correctly parsed `number` that can be passed to `new Date`. No longer will `Date` objects be stored inside TanStack queries.

apps/main/src/dao/components/PageGauges/GaugeListItem/GaugeDetails.tsx
```diff
@@ -99,7 +100,12 @@ export const GaugeDetails = ({ gaugeData, className }: { gaugeData: GaugeFormatt
               {isSideChain && <StyledInformationSquare16 name="InformationSquare" size={16} className="svg-tooltip" />}
             </h5>
           </Chip>
-          <h5>{formatDate(new Date(convertToLocaleTimestamp(new Date(gaugeData.creation_date).getTime())), 'long')}</h5>
+          <h5>
+            {formatDate(
+              new Date(convertToLocaleTimestamp(new Date(parseTimestamp(gaugeData.creation_date)).getTime())),
```
DanielSchiavini: this is crazy hard to read (was it even changed?) 🤔 
imo we should never have new Date or getTime directly in components, that should be handled in the price API and date helpers

apps/main/src/dao/components/PageUser/UserGaugeVotesTable/index.tsx
```diff
@@ -59,7 +56,7 @@ export const UserGaugeVotesTable = ({ userAddress, tableMinWidth }: UserGaugeVot
         <TableRowWrapper key={index} columns={GAUGE_VOTES_LABELS.length} gridTemplateColumns={gridTemplateColumns}>
           <TableData className="align-left">{gaugeVote.gaugeName}</TableData>
           <TableData className={`right-padding ${userGaugeVotesSortBy.key === 'timestamp' ? 'sortby-active' : ''}`}>
-            {formatLocaleDateFromTimestamp(gaugeVote.timestamp)}
+            {formatLocaleDateFromTimestamp(gaugeVote.timestamp / 1000)}
```
DanielSchiavini: I thought the /1000 division would be handled by the prices API client? 🤔 
If really necessary it should be in a helper, repeating 1000 in a component doesn't seem the right place

packages/prices-api/src/timestamp.ts
```diff
@@ -1,17 +1,29 @@
...
+ * Underlying shape is one of:
+ * - unix seconds as `number` (e.g. `1704067200`)
+ * - unix seconds as `string` (e.g. `"1704067200"`)
+ * - ISO 8601 date string (e.g. `"2024-01-01T00:00:00.000Z"`)
  */
-export function toDate(timestamp: string | number): Date {
+export type TimestampResponse = number | `${number}` | IsoDateString
+
+/** Unix epoch milliseconds. Pass to `new Date(t)` to get a `Date`. Branded to enforce being parsed. */
+export type Timestamp = number & { readonly [TimestampBrand]: true }
```
DanielSchiavini, line 18: As I mentioned before, imo the correct format for dates is an ISO string (which includes the timezone)

DanielSchiavini:
```suggestion
export type Timestamp = IsoDateString & { readonly [TimestampBrand]: true }
```

apps/main/src/dao/components/PageUser/UserGaugeVotesTable/index.tsx
DanielSchiavini: is this related to [your comment](https://github.com/curvefi/curve-frontend/pull/2451#discussion_r3174542166)?

apps/main/src/dao/components/PageGauges/GaugeListItem/GaugeDetails.tsx
DanielSchiavini: might be related to [your comment](https://github.com/curvefi/curve-frontend/pull/2451#discussion_r3174542166)

0xAlunara: Might need to rethink how `convertToLocaleTimestamp` works at a later stage as it only works on numbers atm...

apps/main/src/dao/components/PageUser/UserGaugeVotesTable/index.tsx
0xAlunara: The prices api client multiplies timestamp by 1000 as that's what `new Date()` would expect, except we're now using it with `formatLocaleDateFromTimestamp` which accepts a timestamp number in seconds rather than milliseconds. I'd rather fix `formatLocaleDateFromTimestamp` at a later stage.

packages/prices-api/src/timestamp.ts
0xAlunara: I can revert the change to `number`, but I figured `number` would be a better serialization than parsing strings. In the end, we went from `date.toISOString()` to `date.getTime()`, and both outputs are accepted as arguments for `new Date()`. I don't think we need the timezone information, as `date.getTime()` returns the number of milliseconds since 1 Jan 1970 in UTC. Wdyt?

DanielSchiavini: Imo the date string is less ambiguous, more clear. I won't block the PR for that tho

packages/prices-api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@curvefi/prices-api",
-  "version": "1.1.22",
+  "version": "1.2.0",
```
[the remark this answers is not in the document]
0xAlunara: Ok fine lol

apps/main/src/dao/components/PageGauges/GaugeListItem/GaugeDetails.tsx
0xAlunara: Been able to remove it as it had no effect, `formatDate` with the response achieves the same result

apps/main/src/loan/entities/scrvusd-statistics.ts
```diff
@@ -48,7 +49,7 @@ async function _fetchSavingsStatistics(): Promise<Statistics> {
     const apy = (1 + apr / 100 / 365.25) ** 365.25 - 1
 
     return {
-      lastUpdated: new Date(block?.timestamp ?? 0),
+      lastUpdated: fromDate(new Date(block?.timestamp ?? 0)),
```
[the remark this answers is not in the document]
0xAlunara: Removed this altogether as it wasn't even being used at all

DanielSchiavini (dismissed):

tests/cypress/support/helpers/lending-mocks.ts
```diff
@@ -81,7 +82,7 @@ const oneLendingPool = (
       rebasing_yield_apr: null,
     },
     extra_reward_apr: [],
-    created_at: oneDate().toISOString(),
+    created_at: (fromDate(oneDate()) * 1000) as unknown as TimestampResponse,
```
Pearce542: do we need to multiply by 1000 here?

Pearce542 (dismissed):

tests/cypress/support/helpers/lending-mocks.ts
0xAlunara: Argh, I got jebaited by Github code review x.x

I'll just get the unix timestamp the old fashioned way rather than using `fromDate`
I even need to divide by 1000 because `getTime()` returns milliseconds, while the API responses are unix times in seconds.

Pearce542 (dismissed):

DanielSchiavini (dismissed):

DanielSchiavini (approved):

0xtutti (approved):

PR #2452, opened 2026-04-30
feat: update card sizes and paddings

0xAlunara:
1. `<Card>` component now can have `size='small'` and you no longer need to apply it separately to its child `<CardHeader>` and `<CardContent>` (although still possible if you want).
2. Homoginized the card sizes across the board to be more consistent.
3. Adjusted some other paddings accordingly together with tutti for a nice consistent look.
4. Fix empty gap between borrow details and market page header.
image


0xtutti (approved):

OnlyJousting (approved):

apps/main/src/llamalend/widgets/banners/MarketBanners.tsx
```diff
@@ -16,6 +16,12 @@ import { MarketAlertBanner } from './MarketAlertBanner'
 
 const isHighSeverityAlert = (type: AlertType | undefined) => type && ['warning', 'danger'].includes(type)
 
+/**
+ * When the stack is empty, we don't want to display it at all,
+ * because an empty div influences a parent flex's gap.
```
DanielSchiavini, line 21: LLMs keep on adding needless linebreaks
```suggestion
 * When the stack is empty, we don't want to display it at all, because an empty div influences a parent flex's gap.
```
But we can make it shorter
```suggestion
 * Hide empty stacks, empty div influences parents' flex gap.
```

packages/curve-ui-kit/src/themes/components/card/index.ts
```diff
@@ -0,0 +1 @@
+export * from './mui-card'
```
DanielSchiavini, line 1: this is not necessary, we always keep the d.ts files next to the component

apps/main/src/llamalend/widgets/banners/MarketBanners.tsx
0xAlunara: I wrote this comment myself 😭

PR #2456, opened 2026-05-01
fix: max balance loading state

DanielSchiavini:
- revert #2346
- allow the balance component to be used inline
- add spaces between each balance component (they are ignored in flexboxes)

### DIFF
--- a/apps/main/src/llamalend/features/borrow/components/CreateLoanForm.tsx
+++ b/apps/main/src/llamalend/features/borrow/components/CreateLoanForm.tsx
@@ -11,8 +11,8 @@ import type { Decimal } from '@primitives/decimal.utils'
 import { joinButtonText } from '@primitives/string.utils'
 import { useCreateLoanPreset } from '@ui-kit/hooks/useLocalStorage'
 import { t } from '@ui-kit/lib/i18n'
-import { Balance } from '@ui-kit/shared/ui/LargeTokenInput/Balance'
 import { AlertDisableForm } from '@ui-kit/shared/ui/AlertDisableForm'
+import { Balance } from '@ui-kit/shared/ui/LargeTokenInput/Balance'
 import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
 import { q, type Range } from '@ui-kit/types/util'
 import { updateForm } from '@ui-kit/utils/react-form.utils'
@@ -113,10 +113,11 @@ export const CreateLoanForm = <ChainId extends IChainId>({
           network={network}
           message={
             <Balance
+              inline
               prefix={t`Max borrow:`}
               tooltip={t`Max borrow`}
               symbol={borrowToken?.symbol}
-              balance={values.maxDebt}
+              balance={maxDebt.data}
               loading={maxDebt.isLoading}
               onClick={useCallback(() => updateForm(form, { debt: values.maxDebt }), [form, values.maxDebt])}
               buttonTestId="borrow-set-debt-to-max"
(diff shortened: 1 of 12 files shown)


packages/curve-ui-kit/src/shared/ui/LargeTokenInput/Balance.tsx
```diff
@@ -36,8 +38,19 @@ export type Props<T> = {
...
   onClick?: () => void
+  /** Whether the balance should be displayed inline */
+  inline?: boolean
 }
 
+const BalanceContent = ({ children, inline }: { children: ReactNode; inline: boolean }) =>
+  inline ? (
+    <Box>{children}</Box>
+  ) : (
+    <Stack direction="row" gap={Spacing.xs} alignItems="center">
```
DanielSchiavini: I consider to make this also a box, and just change the it to flexbox via sx

apps/main/src/llamalend/features/borrow/components/CreateLoanForm.tsx
```diff
@@ -110,7 +111,18 @@ export const CreateLoanForm = <ChainId extends IChainId>({
...
-          message={`${t`Max borrow:`} ${values.maxDebt ?? '-'} ${borrowToken?.symbol}`}
+          message={
+            <Balance
+              inline
+              prefix={t`Max borrow:`}
+              tooltip={t`Max borrow`}
+              symbol={borrowToken?.symbol}
+              balance={maxDebt.data}
+              loading={maxDebt.isLoading}
+              onClick={useCallback(() => updateForm(form, { debt: values.maxDebt }), [form, values.maxDebt])}
```
0xAlunara, line 122: `maxDebt.data != values.maxDebt`
Maybe it does but want to double check with you

packages/curve-ui-kit/src/shared/ui/LargeTokenInput/Balance.tsx
```diff
@@ -74,26 +88,29 @@ export const Balance = <T extends Amount>({
...
+            variant="bodyXsRegular"
+            color="textTertiary"
+            sx={{ ...VERTICAL_CENTER_TEXT }}
+            {...(inline && { component: 'span' })}
+          >
             {formatNumber(notionalValueUsd, { unit: 'dollar', abbreviate: true })}
           </Typography>
-        )}
-      </Stack>
+        )}{' '}
```
0xAlunara, line 101: You even have these spaces without notional value, but I guess multiple lose spaces don't really matter. Just a remark

packages/curve-ui-kit/src/shared/ui/LargeTokenInput/Balance.tsx
0xAlunara: Would be preferable, then you don't need a separate component and/or ternary.

0xAlunara:
I noticed that during loading the message isn't inline anymore and stacks vertically
image

packages/curve-ui-kit/src/shared/ui/LargeTokenInput/Balance.tsx
DanielSchiavini: Extra spaces are ignored anyway, I've added it here just in case. I didn't think it was necessary to add an extra React fragment

apps/main/src/llamalend/features/borrow/components/CreateLoanForm.tsx
DanielSchiavini: they get synchronized [here](https://github.com/curvefi/curve-frontend/blob/2f24df19b4dbc9c47f4e97b2664e4517249d071d/apps/main/src/llamalend/features/borrow/hooks/useMaxTokenValues.tsx#L55)

Pearce542 (approved):

packages/curve-ui-kit/src/shared/ui/LargeTokenInput/Balance.tsx
DanielSchiavini: Maybe we could even get rid of the flex version, if we just take care of the vertical alignment

0xtutti (approved):

PR #2458, opened 2026-05-01
refactor: react hook form

DanielSchiavini:
- encapsulate react hook form types and hooks in preparation to remove the dependency
- besides typing changes, we also implement a new `FormProvider`
- after this PR we can start refactoring the arguments and result of `useForm` to fit our needs better
- then, once we have the interface that we want, we can choose another form backend to be used such as tanstack or formik


0xtutti (dismissed):

packages/curve-ui-kit/src/features/forms/useForm.ts
```diff
@@ -0,0 +1,53 @@
+// eslint-disable-next-line no-restricted-imports
+import { DefaultValues, useForm as _useForm, type ResolverResult, type ResolverOptions } from 'react-hook-form'
```
Pearce542, line 2: shouldn't the `react-hook-form` be declared in the ui-kit `package.json`?

apps/main/src/llamalend/widgets/action-card/StakeTokenLabel.tsx
```diff
@@ -1,4 +1,4 @@
-import type { FieldValues } from 'react-hook-form'
+import type { FieldValues } from '@ui-kit/features/forms'
 import { Address } from '@primitives/address.utils'
```
Pearce542, line 1: import warning

packages/curve-ui-kit/src/features/forms/useForm.ts
DanielSchiavini: I'll move it but seriously maybe we should just get rid of the monorepo structure, this is really just a single app now, the dependencies should be centralized

Pearce542 (approved):

0xAlunara (approved):

PR #2460, opened 2026-05-02
fix: double slider clamping

DanielSchiavini:
- the NumericTextField already handles the clamping. By clamping twice, the internal state gets out-of-date.
- setRange would be called with the already clamped value, resetting the maxDebt even if there is no range change.
- without a range change, queries would not get triggered, so the max debt was never loaded


0xtutti (approved):

packages/curve-ui-kit/src/shared/ui/SliderInput.tsx
Pearce542, line 80: can be removed

OnlyJousting (approved):

packages/curve-ui-kit/src/shared/ui/SliderInput.tsx
DanielSchiavini: It seems I forgot to push it. I'll put it in another PR

PR #2461, opened 2026-05-02
refactor: use DetailPageLayout for all dao pages

0xAlunara:
Little front-work I had to do for a consistent look across all DAO pages. The focus isn't to convert everything to MUI, just the root pages so I can safely create a new one.

All pages should look full width now with consistent margins and paddings, even across mobile.

image


apps/main/src/dao/components/PageAnalytics/DailyLocksChart/index.tsx
```diff
@@ -22,22 +22,17 @@ export const DailyLocks = () => {
   }, [getVeCrvLocks, veCrvLocks.locks.length, veCrvLocks.fetchStatus])
 
   return (
-    <Wrapper>
+    <Box sx={{ backgroundColor: t => t.design.Layer[1].Fill }}>
```
0xAlunara, line 25: You'll see me apply `sx={{ backgroundColor: t => t.design.Layer[1].Fill }}` in a lot of cases to either a `<Box>` or `<Stack>`. That is to bring back the background lost when removing all these old `<Wrapper>` components.

The proper fix is to later convert them to `<Card>` which applies the correct colors and paddings. I didn't do that in this PR as it's not the focus.

apps/main/src/dao/components/PageVeCrv/index.tsx
```diff
@@ -1,75 +1,67 @@
...
+    // eslint-disable-next-line @eslint-react/exhaustive-deps
+    [],
+  )
 
-  return tab === 'adjust_crv' || tab === 'adjust_date' || tab === 'withdraw' ? (
-    <>
-      <TabsSwitcher variant="underlined" value={tab} onChange={setTab} options={tabs} overflow="fullWidth" />
+  return (
+    <DetailPageLayout formTabs={null} testId="vecrv-page">
+      <Stack margin="auto" maxWidth="30rem" sx={{ backgroundColor: t => t.design.Layer[1].Fill }}>
```
0xAlunara: Didn't bother making a constant for this `30 rem`, it's temporary. In a future design this form will be moved to `formTabs` of `DetailPageLayout`.

DanielSchiavini: A comment would be appreciated

apps/main/src/dao/components/PageAnalytics/DailyLocksChart/index.tsx
DanielSchiavini: Isn't it easy to put the box or stack inside the card without changing the rest?

apps/main/src/dao/components/PageVeCrv/components/FormCrvLocker.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const isPageVisible = useLayoutStore(state => state.isPageVisible)
+  const setFormValues = useStore(state => state.lockedCrv.setFormValues)
+  const signerAddress = curve?.signerAddress
+  const { chainId } = curve ?? {}
+  const canUnlock = getIsLockExpired(
+    vecrvInfo.lockedAmountAndUnlockTime.lockedAmount,
+    vecrvInfo.lockedAmountAndUnlockTime.unlockTime,
+  )
+
+  const tabs: TabOption<FormType>[] = [
```
DanielSchiavini, line 32: Can't we use FormTabs?

apps/main/src/dao/components/PageAnalytics/DailyLocksChart/index.tsx
0xAlunara: Thought it would be a good moment to get rid of some legacy styled component wrappers in favor of a few Mui components. Halfway through I thought the same but figured I'd just continue my path.

apps/main/src/dao/components/PageVeCrv/index.tsx
0xAlunara: but then i might as well add a const 😭

apps/main/src/dao/components/PageVeCrv/components/FormCrvLocker.tsx
0xAlunara: Yeah but wasn't really the goal here, keeping the diff as small as possible

DanielSchiavini (dismissed):

apps/main/src/dao/components/PageAnalytics/DailyLocksChart/index.tsx
DanielSchiavini: that's nice, we could get confused by this in the future

0xtutti (dismissed):

apps/main/src/dao/components/PageAnalytics/index.tsx
```diff
@@ -29,46 +30,19 @@ export const Analytics = () => {
...
-                <TopHolders />
-                <HoldersTable />
-              </Box>
-            )}
-            {tab === 'locks' && <DailyLocks />}
-          </Container>
-        </Box>
-      </Box>
-    </Wrapper>
+    <DetailPageLayout formTabs={null} testId="gauges-page">
```
Pearce542:
```suggestion
    <DetailPageLayout formTabs={null} testId="analytics-page">
```

apps/main/src/dao/components/PageUser/index.tsx
```diff
@@ -56,49 +54,18 @@ export const UserPage = ({ routerParams: { userAddress: rUserAddress } }: UserPa
   const { data: userEnsName } = useEnsName({ address: userAddress as Address })
 
   return (
-    <Wrapper>
-      <UserPageContainer variant="secondary">
+    <DetailPageLayout formTabs={null}>
```
Pearce542:
```suggestion
    <DetailPageLayout formTabs={null}  testId="user-page">
```

Pearce542 (changes requested):
I think tests are failing because of these `testId`

apps/main/src/dao/components/PageAnalytics/index.tsx
0xAlunara: Oh cheers, copy-paste error! Because it didn't fail for firefox or electron I figured they were just failing randomly x.x

Pearce542 (approved):

0xtutti (approved):

PR #2466, opened 2026-05-04
feat: zapV2 in beta mode

DanielSchiavini:
- enable zapv2 on beta mode
- add a key to loan management tabs so they get remounted

### DIFF
--- a/packages/curve-ui-kit/src/hooks/useFeatureFlags.ts
+++ b/packages/curve-ui-kit/src/hooks/useFeatureFlags.ts
@@ -3,8 +3,8 @@
  * These return booleans indicating whether a new experience is enabled.
  */
 
-import { defaultReleaseChannel, ReleaseChannel } from '@ui-kit/utils'
-import { getReleaseChannel, useReleaseChannel } from './useLocalStorage'
+import { ReleaseChannel } from '@ui-kit/utils'
+import { getReleaseChannel, isZapV2Disabled, useDisableZapV2, useReleaseChannel } from './useLocalStorage'
 
 const useBetaChannel = () => useReleaseChannel()[0] === ReleaseChannel.Beta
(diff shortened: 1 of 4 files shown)


OnlyJousting (approved):

0xtutti (approved):

packages/curve-ui-kit/src/hooks/useFeatureFlags.ts
```diff
@@ -35,7 +34,12 @@ export const useManageSoftLiquidation = useStableChannel
 export const useAnalyticsApp = useStableChannel
 
 /** New ZapV2 leverage implementation for LlamaLend markets */
-export const isZapV2Enabled = () => isAlpha() && localStorage.getItem('disableZapV2') != 'true'
+export const isZapV2Enabled = () => getReleaseChannel() === ReleaseChannel.Beta && !isZapV2Disabled()
+
+const useZapV2 = () => [useBetaChannel(), !useDisableZapV2()].every(Boolean)
+
+/** gets a key to remount components when ZapV2 is toggled, forcing calls to non-reactive isZapV2Enabled */
```
DanielSchiavini, line 41: I'm not sure this is needed, it might work fine because all queries get different arguments when using zapv2

0xAlunara: Yeah, diff key means queries will refetch and thus trigger a redraw anyway?

0xAlunara (approved):

packages/curve-ui-kit/src/hooks/useFeatureFlags.ts
DanielSchiavini: actually it doesn't even mean that, because react-query will just happily give the previous data unless the parameters changed. I'll leave this to another PR tho.

PR #2467, opened 2026-05-04
feat: llamalend popover filters

Pearce542:
Part 3 of improving the search and filtering experience in the market list. 
In this PR I just moved the filters from the collapsible to the popover. 
Next step is to match the popover look and feel of the design

- Part 2: https://github.com/curvefi/curve-frontend/pull/2465
- Part 1: https://github.com/curvefi/curve-frontend/pull/2407


tests/cypress/e2e/all/error-boundary.cy.ts
```diff
@@ -10,11 +10,11 @@ const visitErrorBoundary = () => {
       chains: {
         ethereum: {
           data: [
-            // make an error occur on purpose by returning nonsense data (`symbol` should be string)
+            // make an error occur on purpose by returning nonsense data (`address` should be string)
             // note that the error only triggers the boundary if the query succeeds, but it fails during rendering
             {
-              collateral_token: { symbol: 1 },
-              borrowed_token: { symbol: 1 },
+              collateral_token: { symbol: 'ETH', address: 1 },
```
Pearce542, line 16: Had to change the malformed field from `symbol` to `address` because the `symbol` was only crashing inside filters, which are now loaded only when the popover renders. The `address` is used in the first load

apps/main/src/llamalend/features/market-list/chips/FilterChip.tsx
```diff
@@ -10,8 +10,9 @@ import { LlamaMarketColumnId } from '../columns'
 import { MarketListFilterDrawer } from '../drawers/MarketListFilterDrawer'
 
 type Props = {
-  filterExpanded: boolean
-  setFilterExpanded: Dispatch<SetStateAction<boolean>>
+  refFilters: RefObject<HTMLDivElement | null>
```
DanielSchiavini: refFilters seems like the incorrect name. It should be `filterChipRef` imo. Since this isn't a ref to the filters but to the filter chip and only on desktop. So it's a bit easy to make mistakes here, I wonder if we could make it more fool-proof.

apps/main/src/llamalend/features/market-list/LlamaMarketsTable.tsx
```diff
@@ -41,7 +41,8 @@ export const LlamaMarketsTable = ({
 }) => {
   const { markets, userHasPositions, hasFavorites } = result ?? {}
   const data = useMemo(() => markets ?? [], [markets])
-  const [filterExpanded, setFilterExpanded] = useFilterExpanded(LOCAL_STORAGE_KEY)
+  const [filtersOpen, , closeFilters, toggleFilters] = useSwitch(false)
```
DanielSchiavini: similarly here, `filtersOpen` applies only to desktop. We should probably call it `filterPopoverOpen`

DanielSchiavini (commented):
Looks good, just a couple naming remarks

DanielSchiavini (approved):

0xtutti (approved):

PR #2476, opened 2026-05-06
refactor: improve solvency query performance

Pearce542:
Instead of fetching all queries from the `useLlamaMarkets` just to calculate the market solvency, we only fetch the necessary queries: lend vaults and bad debt endpoints.

### DIFF
--- a/apps/main/src/lend/entities/chain/chain-info.ts
+++ b/apps/main/src/lend/entities/chain/chain-info.ts
@@ -1,10 +1,10 @@
 import { useMemo } from 'react'
 import { ChainId } from '@/lend/types/lend.types'
 import { getLib } from '@ui-kit/features/connect-wallet'
-import { FETCHING, PartialQueryResult, READY } from '@ui-kit/lib/queries'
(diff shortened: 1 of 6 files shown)


DanielSchiavini (approved):
I was actually going to do this, since I'm working on improving the page performance, but didn't have the chance to do it yet. So thanks!

0xtutti (approved):

Pearce542:
> I was actually going to do this, since I'm working on improving the page performance, but didn't have the chance to do it yet. So thanks!

Yeah, I wanted to ask you first, but then I forgot. I wanted to do this because some form tests were failing due to timeout issues on my branch. I suspected that `useLlamaMarkets` was sometimes taking more than 30 seconds.

DanielSchiavini:
> Yeah, I wanted to ask you first, but then I forgot. I wanted to do this because some form tests were failing due to timeout issues on my branch. I suspected that `useLlamaMarkets` was sometimes taking more than 30 seconds.

That would be way too long (though we have API_LOAD_TIMEOUT of 2 minutes) - but in tests you'd need a lot more mocking.

Btw, shouldn't we remove unused mocks?

PR #2485, opened 2026-05-07
feat: append on chain rate to historical rate charts

OnlyJousting:
Changes:
- Adds on-chain borrow/supply rate as last datapoints in historical rate charts


apps/main/src/llamalend/widgets/MarketHistoricalRatesChart.tsx
```diff
@@ -67,14 +77,18 @@ export const MarketHistoricalRatesChart = ({ market, blockchainId, rateMode }: M
     isLoading,
     error,
   } = useLlamaSnapshot(market, blockchainId, Boolean(market && blockchainId), { kind: 'timeRange', timeOption })
+  const { data: marketRates } = useMarketRates({ chainId, marketId }, Boolean(market))
```
DanielSchiavini: We generally should only need to pass the `enabled` flag when the UI element is hidden
```suggestion
  const { data: marketRates } = useMarketRates({ chainId, marketId })
```

apps/main/src/llamalend/widgets/MarketHistoricalRatesChart.tsx
```diff
@@ -67,14 +77,18 @@ export const MarketHistoricalRatesChart = ({ market, blockchainId, rateMode }: M
...
-        // timestamp is typed as Date but may be a string after JSON serialization (e.g. React Query cache)
-        timestamp: new Date(snapshot.timestamp).getTime(),
-        rate: Number(rateMode === 'borrow' ? snapshot.borrowApr : 'lendApy' in snapshot ? snapshot.lendApy * 100 : 0),
-      })),
+      notFalsy(
+        ...snapshots.map(snapshot => ({
+          timestamp: snapshot.timestamp,
+          rate: Number(rateMode === 'borrow' ? snapshot.borrowApr : 'lendApy' in snapshot ? snapshot.lendApy * 100 : 0),
+        })),
+        onChainRate != null && { timestamp: Date.now(), rate: Number(onChainRate) },
```
DanielSchiavini, line 90: using the block timestamp would be more correct 🤓 
why do we need to cast this to number?
btw, `onChainRate` sounds like a callback name because of the `on` prefix

DanielSchiavini (dismissed):

0xtutti (dismissed):

apps/main/src/llamalend/widgets/MarketHistoricalRatesChart.tsx
OnlyJousting: I'll update the variable name but I think Date.now() is fine for this timestamp. Llamlend-js would need to pass along block timestamp which it doesn't right now

0xtutti (approved):

DanielSchiavini (approved):
you said you would update the const name, but it's good enough tbh

OnlyJousting:
> you said you would update the const name, but it's good enough tbh

ah shit, I forgot to save the file 😭

PR #2490, opened 2026-05-08
feat: show max balance even when errors occur

DanielSchiavini:
- when the user has set a max balance that is too high, either manually or due to market conditions changing, the max isn't displayed anymore. The user will only see a transaction and/or max balance error. Removing the value completely also doesn't help as you then see the required field error.
- Therefore, I think we should be showing the max balance even when errors are visible


0xAlunara (approved):
Stacks nicely

Image

0xtutti (approved):

PR #2492, opened 2026-05-09
refactor: single number formatter

0xAlunara:
This is one nasty big PR that drained my mind and time, because I want to get rid of the legacy `formatNumber` which wraps calls to the new `formatNumber` in order to more easily enforce categories later on.

Started out with Codex doing codemods, but had to manually edit each diff because it kept fucking it up :(

**Note**: The diff is already as big as it is, so I the focus is really on removing the old legacy wrapper, and not simplifying ternaries or other low hanging fruit.


apps/main/src/dao/components/Charts/GaugesBarChartCustomTooltip.tsx
```diff
@@ -22,7 +22,7 @@ export const GaugesBarChartCustomTooltip = ({ active, payload }: TooltipProps<Va
             <TooltipDataTitle>{t`Gauge weight 7d delta`}</TooltipDataTitle>
             {sevenDayDelta ? (
               <TooltipData className={sevenDayDelta > 0 ? 'positive' : 'negative'}>
-                {formatNumber(sevenDayDelta, { notation: 'compact' })}%
+                {formatNumber(amount(sevenDayDelta), { abbreviate: true }) ?? '-'}%
```
DanielSchiavini: I dont think we should be repeating the fallback values everywhere. This should be handled by formatNumber and formatPercentage and soon by the number categories

0xAlunara:
`formatNumber` now has an optional `fallback` property, replacing all the `?? '-'`. I tried to enforce that if the return type is `string` and not `string | undefined`, you can't enter a fallback as it wouldn't make sense, but alas there's no simple solution to that problem. So there is a chance that a caller might invoke `formatNumber(5, { fallback: '-' })`, even though the fallback makes no sense since `5` can never be nullish. Asked codex to make an eslint rule, which worked, but was unmaintainably complex.

apps/main/src/dao/components/Charts/GaugesBarChartCustomTooltip.tsx
```diff
@@ -22,7 +22,7 @@ export const GaugesBarChartCustomTooltip = ({ active, payload }: TooltipProps<Va
             <TooltipDataTitle>{t`Gauge weight 7d delta`}</TooltipDataTitle>
             {sevenDayDelta ? (
               <TooltipData className={sevenDayDelta > 0 ? 'positive' : 'negative'}>
-                {formatNumber(sevenDayDelta, { notation: 'compact' })}%
+                {formatNumber(amount(sevenDayDelta), { abbreviate: true, fallback: '-' })}%
```
0xAlunara, line 25: For the moment we're very explicit with `fallback: '-'` everywhere, but that could be centralized and factored away behind the future categories.

0xtutti (dismissed):

apps/main/src/dao/components/PageProposal/Voters/index.tsx
```diff
@@ -62,8 +64,9 @@ export const Voters = ({ totalVotes, voteId, proposalType, className }: Props) =
                 </Box>
                 <StyledExternalLink href={scanTxPath(networks[Chain.Ethereum], vote.txHash)}>
                   <Data>
-                    {formatNumber(+vote.stake, { notation: 'compact' })} (
-                    {formatNumber(vote.relativePower, { notation: 'compact' })}%)
+                    {formatNumber(+vote.stake, { abbreviate: true })} (
+                    {formatNumber(vote.relativePower, { abbreviate: true })}
```
DanielSchiavini, line 68: formatNumber+% should be refactored to formatPercentage later on

packages/curve-ui-kit/src/utils/decimal.ts
```diff
@@ -1,6 +1,14 @@
...
 import { formatUnits, parseUnits } from 'viem'
-import type { Decimal } from '@primitives/decimal.utils'
+import type { Amount, Decimal } from '@primitives/decimal.utils'
+
+/** Converts loose numeric input to an Amount for formatting, returning undefined for empty or non-numeric values. */
+export const amount = (value: number | string | BigNumber | bigint | null | undefined): Amount | undefined => {
+  if (value == null || value === '') return undefined
+  if (typeof value === 'number') return Number.isNaN(value) ? undefined : value
+
+  return decimal(value)
```
DanielSchiavini:
```suggestion
  return value == null || value === '' || Number.isNaN(value) ? undefined : typeof value === 'number' ? value : decimal(value)
```

packages/curve-ui-kit/src/utils/number.ts
```diff
@@ -260,8 +266,24 @@ export const decomposeNumber = (value: Amount, options: NumberFormatOptions): De
...
+ *
+ * formatNumber(NaN, { abbreviate: false, fallback: '-' })
+ * // Returns "-"
+ *
+ * formatNumber(NaN, { abbreviate: false })
+ * // Returns "NaN"
  */
-export const formatNumber = (value: Amount, options: NumberFormatOptions) => {
+export function formatNumber(value: Amount, options: NumberFormatOptions): string
+export function formatNumber(value: Amount | MissingAmount, options: NumberFormatOptions & { fallback: string }): string
```
DanielSchiavini, line 280: Nice! I feel like it would be easier to do the categories and then refactor all usages 🤔

apps/main/src/dex/components/PageDashboard/components/FormClaimFees.tsx
```diff
@@ -35,8 +35,14 @@ export const FormClaimFees = () => {
           <Tooltip>{t`Trading fees distributed to CRV lockers`}</Tooltip>:{' '}
         </div>
         <div>
-          <strong>{formatNumber(claimFeesAmounts?.[claimButtonsKey['3CRV']], { defaultValue: '-' })}</strong> 3CRV,{' '}
-          <strong>{formatNumber(claimFeesAmounts?.[claimButtonsKey.crvUSD], { defaultValue: '-' })}</strong> crvUSD
+          <strong>
+            {formatNumber(amount(claimFeesAmounts?.[claimButtonsKey['3CRV']]), { abbreviate: false, fallback: '-' })}
+          </strong>{' '}
+          3CRV,{' '}
```
DanielSchiavini, line 41: We definitely need a `formatToken` helper in the future

apps/main/src/loan/lib/apiCrvusd.ts
```diff
@@ -85,13 +85,13 @@ async function fetchChartBandBalancesData(
...
-    const pUpDownMedian = new BN(p_up).plus(p_down).dividedBy(2).toFixed(5)
-    const collateralUsd = new BN(collateral).multipliedBy(sqrt)
+    const sqrt = new BigNumber(p_up).multipliedBy(p_down).squareRoot()
+    const pUpDownMedian = new BigNumber(p_up).plus(p_down).dividedBy(2).toFixed(5)
+    const collateralUsd = new BigNumber(collateral).multipliedBy(sqrt)
 
     return {
       collateral,
-      collateralUsd: collateralUsd.toString(),
+      collateralUsd: collateralUsd.toFixed(),
```
DanielSchiavini, line 94: good one, toString is messed up

DanielSchiavini (dismissed):

apps/main/src/llamalend/features/borrow/components/LeverageInput.tsx
```diff
@@ -30,7 +30,7 @@ export const LeverageInput = ({
       endContent={
         <ActionInfo
           label={t`Leverage`}
-          value={leverage == null ? '–' : `${formatNumber(leverage, { maximumFractionDigits: 2 })}x`}
+          value={`${formatNumber(leverage, { maximumFractionDigits: 2, abbreviate: false, fallback: '-' })}x`}
```
Pearce542: we shouldn't render `-x` but `-` if the leverage is nullish

Pearce542 (dismissed):

OnlyJousting (approved):

apps/main/src/llamalend/features/borrow/components/LeverageInput.tsx
```diff
@@ -30,7 +30,11 @@ export const LeverageInput = ({
       endContent={
         <ActionInfo
           label={t`Leverage`}
-          value={`${formatNumber(leverage, { maximumFractionDigits: 2, abbreviate: false, fallback: '-' })}x`}
+          value={
+            leverage && leverage !== '0'
```
DanielSchiavini, line 34: how about '0.0'??
```suggestion
            leverage && +leverage
```

DanielSchiavini (approved):

apps/main/src/llamalend/features/borrow/components/LeverageInput.tsx
0xAlunara: Right, but im not giving up my approvals for that 😭

PR #2494, opened 2026-05-10
feat: add yield basis endpoints to prices api

0xAlunara:
Adds yield basis endpoints to prices api

I'm going to give the migration to Zod a higher prio, shoudl reduce the amount of code quit a bit once we migrate. We already have a zod dependency anyway.

### DIFF
--- a/packages/prices-api/src/yield-basis/models.ts
+++ b/packages/prices-api/src/yield-basis/models.ts
@@ -1,4 +1,4 @@
-import type { Address } from '@primitives/address.utils'
+import type { Address, Hex } from '@primitives/address.utils'
 import type { Chain } from '..'
 import type { Timestamp } from '../timestamp'
 
@@ -13,7 +13,7 @@ export type AdjacentPool = {
 }
 
 export type YieldBasisTransaction = {
-  txHash: string
+  txHash: Hex
   blockNumber: number
   timestamp: Timestamp
   volume: number
--- a/packages/prices-api/src/yield-basis/models.ts
+++ b/packages/prices-api/src/yield-basis/models.ts
@@ -1,0 +1,79 @@
+import type { Address } from '@primitives/address.utils'
+import type { Chain } from '..'
+import type { Timestamp } from '../timestamp'
+
+export type YieldBasisPool = {
+  name: string
+  address: Address
+}
+
+export type AdjacentPool = {
+  name: string
+  address: Address
+}
+
+export type YieldBasisTransaction = {
+  txHash: string
+  blockNumber: number
+  timestamp: Timestamp
+  volume: number
+  adjacentVolume: number
+  adjacentPools: AdjacentPool[]
+}
+
+export type YieldBasisPoolVolume = {
+  poolAddress: Address
+  poolName: string
+  transactions: YieldBasisTransaction[]
+}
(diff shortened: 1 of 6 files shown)


0xtutti (dismissed):

Pearce542 (dismissed):

packages/prices-api/src/yield-basis/models.ts
```diff
@@ -0,0 +1,79 @@
...
+  address: Address
+}
+
+export type AdjacentPool = {
+  name: string
+  address: Address
+}
+
+export type YieldBasisTransaction = {
+  txHash: string
```
DanielSchiavini:
```suggestion
  txHash: Hex
```

packages/prices-api/src/yield-basis/parsers.ts
```diff
@@ -0,0 +1,85 @@
+import { parseTimestamp } from '../timestamp'
+import type * as Models from './models'
+import type * as Responses from './responses'
+
+const parseNumberString = (value: string) => Number(value)
```
DanielSchiavini: Please never convert strings to numbers for no reason
```suggestion
const parseNumberString = (value: string) => value as `number`
```

DanielSchiavini (changes requested):

packages/prices-api/src/yield-basis/parsers.ts
0xAlunara: Oops, codex remainder. In every other place we use `parseFloat` though, so I've put it in this parser file as well,

I want to migrate to Zod, which will add support for proper validation when floats can't get parsed so we fail fast, but that's a diff PR. For now I just made it in line with the rest of the parsers in the prices api.

DanielSchiavini (approved):

packages/prices-api/src/yield-basis/parsers.ts
```diff
@@ -60,26 +58,26 @@ export const parseCrvUsdYieldBasisSupply = (
...
   cachedAt: response.cached_at == null ? undefined : parseTimestamp(response.cached_at),
-  ybFactoryBalance: parseNumberString(response.data.yb_factory_balance),
-  ybTotalAllocated: parseNumberString(response.data.yb_total_allocated),
-  ybTotalAmmBalance: parseNumberString(response.data.yb_total_amm_balance),
-  ybTotalAmmDebt: parseNumberString(response.data.yb_total_amm_debt),
-  ybMaxDebt: parseNumberString(response.data.yb_max_debt),
-  mintPegkeeperDebt: parseNumberString(response.data.mint_pegkeeper_debt),
-  mintMarketDebt: parseNumberString(response.data.mint_market_debt),
-  totalSupply: parseNumberString(response.data.total_supply),
+  ybFactoryBalance: parseFloat(response.data.yb_factory_balance),
```
DanielSchiavini, line 61: it seriously stick sucks that we are converting them to `number`, but at least it's consistent 🫠

0xAlunara: Back in the day we didn't have decimals yet. Could prob tackle that together with the Zod transformer

Pearce542 (approved):

PR #2495, opened 2026-05-10
feat: support disabling of sticky headers in datatable

0xAlunara:
I ran into the issue where the datatable's sticky headers are causing issues wrt the top & max height when used inside a card and not as a main part of the scrollable page. This PR adds a prop that allows us to optionally disable it.

Inside card with sticky headers by default:
image

Inside a card with `stickyHeader={false}`:
image

### DIFF
--- a/packages/curve-ui-kit/src/shared/ui/DataTable/DataRow.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/DataTable/DataRow.tsx
@@ -26,17 +26,19 @@ const onCellClick = (target: EventTarget, url: string, routerNavigate: (href: st
 export type DataRowProps<T extends TableItem> = {
   table: Table<T>
   row: Row<T>
-  isLast: boolean
   expandedPanel?: ExpandedPanel<T>
+  isLastRow: boolean
+  shouldStickLastRowToTop?: boolean
   shouldStickFirstColumn?: boolean
   verticalAlign?: 'top' | 'center' | 'bottom'
 }
 
 export const DataRow = <T extends TableItem>({
   table,
-  isLast,
   row,
   expandedPanel,
+  isLastRow,
+  shouldStickLastRowToTop,
   shouldStickFirstColumn,
   verticalAlign = 'center',
 }: DataRowProps<T>) => {
@@ -50,6 +52,7 @@ export const DataRow = <T extends TableItem>({
     [url, push, hasUrl],
   )
   const visibleCells = row.getVisibleCells()
+  const shouldApplyStickyLastRow = isLastRow && shouldStickLastRowToTop
   return (
     <>
       <InvertOnHover hoverColor={t => t.design.Table.Row.Hover} hoverRef={{ current: element }} disabled={isMobile}>
@@ -70,7 +73,7 @@ export const DataRow = <T extends TableItem>({
                   backgroundColor: t => t.design.Table.Row.Hover,
                 },
               },
-              ...(isLast && {
+              ...(shouldApplyStickyLastRow && {
                 // to avoid the sticky header showing without any rows, show the last row on top of it
                 position: 'sticky',
                 zIndex: t => t.zIndex.tableStickyLastRow,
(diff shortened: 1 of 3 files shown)


0xtutti (dismissed):

Pearce542 (dismissed):

packages/curve-ui-kit/src/shared/ui/DataTable/LegacyDataTable.tsx
```diff
@@ -78,6 +78,7 @@ export const LegacyDataTable = <T extends TableItem>({
   viewAllLabel,
   shouldStickFirstColumn = false,
   hideHeader = false,
+  stickyHeader = true,
```
DanielSchiavini: Boolean props work a bit better when we can make the default be false imo. First we don't have to worry too much about `undefined` cause it's falsey anyway. Second, when we want to override the default in JSX, the syntax sugar is nice because we don't need to pass any value to set a `prop?: boolean` to `true`. Finally, if we choose one or the other the code will be more consistent. It might be a question of taste tho.

packages/curve-ui-kit/src/shared/ui/DataTable/LegacyDataTable.tsx
```diff
@@ -78,6 +78,7 @@ export const LegacyDataTable = <T extends TableItem>({
   viewAllLabel,
   shouldStickFirstColumn = false,
```
DanielSchiavini, line 80: Do we want to connect the new prop to this one? I can imagine they will be use together.

packages/curve-ui-kit/src/shared/ui/DataTable/LegacyDataTable.tsx
```diff
@@ -153,7 +157,7 @@ export const LegacyDataTable = <T extends TableItem>({
               <DataRow<T>
                 key={row.id}
                 row={row}
-                isLast={index === visibleRows.length - 1}
+                isLast={stickyHeader && !hideHeader && index === visibleRows.length - 1}
```
DanielSchiavini: `isLast` should **never** depend on `stickyHeader` or `hideHeader`, that does not make any sense.
I assume this is currently only being used to define the stickyness of the row, but then we should rename the prop.

DanielSchiavini (changes requested):

packages/curve-ui-kit/src/shared/ui/DataTable/LegacyDataTable.tsx
0xAlunara: This prop seems to be more of a restriction on the sticky behavior if it's enabled (which it is by default)
Technically setting it to `true` would not make sense if sticky behavior will be turned off, but I don't think you can easily combine them into one 🤔

packages/curve-ui-kit/src/shared/ui/DataTable/LegacyDataTable.tsx
0xAlunara: Consistency is nice, since we also have `disableX` for other components like `SearchField`. And yeah no need for `={false}` either this way. Only 'downside' is that we need to negate the condition now: `...(!disableStickyHeader && {})`

packages/curve-ui-kit/src/shared/ui/DataTable/LegacyDataTable.tsx
0xAlunara: I was thinking maybe the sticky behavior and 'is last' should be split up into two different props, but it turns out the prop only defines sticky behavior. But then combining it into one prop in `DataRow` made it possible non-last rows could be sticky, which makes non sense either.

So in the end there is an `isLast` prop and `shouldStickLastRowToTop` as it's more readable.

packages/curve-ui-kit/src/shared/ui/DataTable/DataRow.tsx
```diff
@@ -26,17 +26,19 @@ const onCellClick = (target: EventTarget, url: string, routerNavigate: (href: st
 export type DataRowProps<T extends TableItem> = {
   table: Table<T>
   row: Row<T>
-  isLast: boolean
   expandedPanel?: ExpandedPanel<T>
+  isLastRow: boolean
```
DanielSchiavini, line 30: we probably could live with just `shouldStickToTop` but this works 👍

DanielSchiavini (approved):

0xtutti (approved):

PR #2502, opened 2026-05-11
feat: add metamask connector

0xAlunara:
Basically a bizdev decision; I know I know, it's going to introduce dependencies... But at least it's 'wagmi certified' as per https://wagmi.sh/react/api/connectors/metaMask

https://docs.metamask.io/metamask-connect

**Why use MetaMask Connect?**
If a user has the MetaMask browser extension installed, standards such as EIP-6963 and EIP-1193 support wallet discovery. When the extension isn't available (for example, on mobile, in a different browser, or on a new device), users need another way to connect. MetaMask Connect fills that gap and provides the following benefits:

* Cross-platform connections: Reach users on any device. When the MetaMask browser extension isn't available, MetaMask Connect connects users through the MetaMask mobile app with no additional setup.
* Multichain sessions: Request access to EVM, Solana, and future ecosystems in a single connection instead of connecting per chain.
* Persistent sessions: Keep sessions across page reloads and new tabs so users don't need to reconnect.
* Consistent API: Use the same interface whether users connect through the extension or the MetaMask mobile app.

### DIFF
--- a/packages/curve-ui-kit/package.json
+++ b/packages/curve-ui-kit/package.json
@@ -18,6 +18,7 @@
     "@emotion/react": "11.14.0",
     "@emotion/styled": "11.14.1",
     "@hookform/resolvers": "5.2.2",
+    "@metamask/connect-evm": "1.2.0",
     "@mui/icons-material": "7.3.10",
     "@mui/material": "7.3.10",
     "@safe-global/safe-apps-provider": "0.18.6",
--- a/packages/curve-ui-kit/src/features/connect-wallet/lib/wagmi/connectors.ts
+++ b/packages/curve-ui-kit/src/features/connect-wallet/lib/wagmi/connectors.ts
@@ -1,6 +1,6 @@
 import { assert } from '@primitives/objects.utils'
 import { t } from '@ui-kit/lib/i18n'
-import { coinbaseWallet, injected, safe, walletConnect } from '@wagmi/connectors'
+import { coinbaseWallet, injected, metaMask, safe, walletConnect } from '@wagmi/connectors'
 import type { CreateConnectorFn } from '@wagmi/core'
 
 // project managed at https://cloud.reown.com/ set up by Schiavini, Michael also has access.
@@ -11,6 +11,7 @@ export const INJECTED_CONNECTOR_ID = 'injected'
 // Order matters for the connect wallet modal, list grows from bottom to top, so first one is shown all the way down.
 export const connectors: CreateConnectorFn[] = [
   coinbaseWallet({ preference: { options: 'all', telemetry: false } }),
+  metaMask(),
   safe(),
   walletConnect({
     projectId: WALLET_CONNECT_PROJECT_ID,
--- a/packages/curve-ui-kit/src/features/connect-wallet/ui/WagmiConnectModal.tsx
+++ b/packages/curve-ui-kit/src/features/connect-wallet/ui/WagmiConnectModal.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useState } from 'react'
+import { useCallback, useMemo, useState } from 'react'
 import type { BaseError } from 'viem'
 import Alert from '@mui/material/Alert'
 import AlertTitle from '@mui/material/AlertTitle'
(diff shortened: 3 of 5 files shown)


packages/curve-ui-kit/src/features/connect-wallet/lib/wagmi/connectors.ts
```diff
@@ -11,6 +11,7 @@ export const INJECTED_CONNECTOR_ID = 'injected'
 // Order matters for the connect wallet modal, list grows from bottom to top, so first one is shown all the way down.
 export const connectors: CreateConnectorFn[] = [
   coinbaseWallet({ preference: { options: 'all', telemetry: false } }),
+  metaMask({ debug: true }),
```
DanielSchiavini:
```suggestion
  metaMask({ debug: isDevelopment }),
```

packages/curve-ui-kit/src/features/connect-wallet/ui/WagmiConnectModal.tsx
```diff
@@ -26,10 +27,16 @@ const WALLET_ICONS: Record<string, ReturnType<typeof createSvgIcon>> = {
...
   coinbaseWalletSDK: CoinbaseWalletIcon,
   safe: SafeWalletIcon,
+  metaMaskSDK: MetamaskWalletIcon,
 }
 
 const WALLET_ICON_SIZE = handleBreakpoints({ width: IconSize.xl, height: IconSize.xl })
 
+const getConnectorWalletIds = (connector: Connector) => [
+  connector.id,
+  ...(typeof connector.rdns === 'string' ? [connector.rdns] : (connector.rdns ?? [])),
```
DanielSchiavini: see [toArray](https://github.com/curvefi/curve-frontend/blob/81a8950b6142fca8c97343b26a9afc80fae4b3ef/packages/primitives/src/array.utils.ts#L29)
```suggestion
  ...toArray(connector.rdns),
```

packages/curve-ui-kit/src/features/connect-wallet/ui/WagmiConnectModal.tsx
```diff
@@ -124,6 +131,14 @@ export const WagmiConnectModal = () => {
           .toReversed()
           // Put browser injected wallet first as it's a good fallback that's supposed to work in most cases
           .toSorted((a, b) => (a.id === INJECTED_CONNECTOR_ID ? -1 : b.id === INJECTED_CONNECTOR_ID ? 1 : 0))
+          // Remove EIP-6963 dupes if there's already a connector manually added
+          // Keep the last duplicate so manual connectors win after EIP-6963 connectors are moved to the top.
+          .filter(
+            (connector, index, connectors) =>
+              connectors.findLastIndex(other =>
```
DanielSchiavini: filter + toReversed + toSorted + filter (with nested findLastIndex with nested some with nested includes) on every render

packages/curve-ui-kit/package.json
```diff
@@ -18,6 +18,7 @@
     "@emotion/react": "11.14.0",
     "@emotion/styled": "11.14.1",
     "@hookform/resolvers": "5.2.2",
+    "@metamask/connect-evm": "1.2.0",
```
DanielSchiavini, line 21: That has so many new [dependencies](https://npmgraph.js.org/?q=%40metamask%2Fconnect-evm) 😭 
We seriously need a way to only load that kind of crap when the user actually wants it.

Image

packages/curve-ui-kit/src/features/connect-wallet/lib/wagmi/connectors.ts
0xAlunara: Note to self: actually remove in final PR, I don't think we want this. It also didn't really give me any extra useful information

packages/curve-ui-kit/src/features/connect-wallet/ui/WagmiConnectModal.tsx
0xAlunara: According to multiple LLMs `useMemo` wasn't worth the additional overhead as the list very small anyway, with like 5 or 6 wallets 🤔

packages/curve-ui-kit/package.json
0xAlunara: I can't say for sure about metamask, but at least all the coinbase wallet dependencies are already lazily loaded until the connector is actually being used (by clicking on the coinbase wallet button in the connect wallet modal)

packages/curve-ui-kit/src/features/connect-wallet/ui/WagmiConnectModal.tsx
0xAlunara: I'm asking it again and it's really adamant about it 😂 

> So it’s not that `useMemo` would be wrong. It would be defensible. I just wouldn’t add it until there’s an actual reason beyond "we’re already extracting a const."

0xtutti (dismissed):

packages/curve-ui-kit/src/features/connect-wallet/ui/WagmiConnectModal.tsx
DanielSchiavini: For 6 connectors it's kind of fine (6^3=216) but this doesn't scale. With 10 connectors it will be already 1k iterations for every render with countless arrays being garbage collected. IMO it's definitely worth to extract a const here

DanielSchiavini: Actually - we are 4 layers deep here, but not only iterating on connectors but also their ids.
It's `connectorCount ^ 2 * idCount ^ 2`, with 10 connectors each of 3 ids, we get (up to) 900 iterations. 

But OK, in my machine I only get 29 lol

DanielSchiavini (dismissed):

packages/curve-ui-kit/src/features/connect-wallet/ui/WagmiConnectModal.tsx
0xAlunara: Meh I ended up wrapping it in useMemo anyway so this discussion doesn't come up in the future and we won't have a problem when there's more connectors 😅

packages/curve-ui-kit/src/features/connect-wallet/ui/WagmiConnectModal.tsx
```diff
@@ -75,6 +75,27 @@ export const WagmiConnectModal = () => {
   const { connectors, connect, showModal, closeModal } = useWallet()
   const [error, setError] = useState<unknown>(null)
   const [connectingToId, setConnectingToId] = useState<string | null>(null)
+  const isSafeApp = typeof window !== 'undefined' && window !== window.parent
```
DanielSchiavini, line 78: We don't have SSR anymore 🙌 
```suggestion
  const isSafeApp = window !== window.parent
```

DanielSchiavini (approved):

0xtutti (approved):

PR #2505, opened 2026-05-11
fix: slider labels positions

Pearce542:
In order to fix it properly I added a sliderProp to the SliderInput component, so the labels will always take the slider's width instead of using a Grid or a constant width 

- fix slider's labels position

<table>
  <tr>
    <th>Before</th>
    <td>
Screenshot 2026-05-11 at 20 47 53
</td>
  </tr>
  <tr>
    <th>After</th>
    <td>
Screenshot 2026-05-11 at 20 48 01
</td>
  </tr>
</table>

### DIFF
--- a/packages/curve-ui-kit/src/shared/ui/SliderInput.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/SliderInput.tsx
@@ -239,7 +239,7 @@ export const SliderInput = <T extends Decimal | DecimalRangeValue>({
   )
 
   const renderSlider = (
-    <Stack width="100%">
+    <Stack flexGrow={1}>
       {sliderLabel}
       <Slider
         size={size}
@@ -261,7 +261,12 @@ export const SliderInput = <T extends Decimal | DecimalRangeValue>({
   )
 
   return (
-    <Stack direction={layoutDirection} alignItems="end" columnGap={Spacing.sm} rowGap={Spacing.xs}>
+    <Stack
+      direction={layoutDirection}
+      alignItems={layoutDirection === 'row' ? 'end' : 'stretch'}
+      columnGap={Spacing.sm}
+      rowGap={Spacing.xs}
+    >
       {layoutDirection === 'row' ? (
         <>
           {isRange && renderInput(currentFirst, 0)}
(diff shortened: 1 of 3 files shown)


packages/curve-ui-kit/src/shared/ui/stories/SliderInput.stories.tsx
```diff
@@ -81,13 +81,23 @@ const meta: Meta<typeof SliderInputComponent> = {
       control: false,
       description: 'Callback triggered when the value changes.',
     },
+    sliderLabel: {
+      control: false,
```
DanielSchiavini: why false? We cannot test it in the storybook.

packages/curve-ui-kit/src/shared/ui/SliderInput.tsx
```diff
@@ -91,6 +93,7 @@ export const SliderInput = <T extends Decimal | DecimalRangeValue>({
   onChange,
   min = 0,
   max = Infinity,
+  sliderLabel,
```
DanielSchiavini, line 96: this is already in the slider, why add a prefix?
```suggestion
  label,
```

Pearce542: Because it could be the label of the `input`? Here it is explicitly bound to the width of the slider

packages/curve-ui-kit/src/shared/ui/stories/SliderInput.stories.tsx
Pearce542: mb

packages/curve-ui-kit/src/shared/ui/SliderInput.tsx
```diff
@@ -236,26 +239,29 @@ export const SliderInput = <T extends Decimal | DecimalRangeValue>({
...
-      }}
-      min={sliderMinValue}
-      max={sliderMaxValue}
-      step={sliderStepValue}
-      disabled={disabled}
-      scale={mapFromSliderValue}
-      data-testid={`slider-${name}`}
-      {...restSliderProps}
-    />
+    <Stack width="100%">
```
0xAlunara: `flexGrow={1}` is preferred over `width="100%"`, setting width can lead to some weird edge cases

0xAlunara (dismissed):

packages/curve-ui-kit/src/shared/ui/SliderInput.tsx
Pearce542: why is `width="100%"` not good? Changed to `flexGrow`

0xAlunara (approved):

DanielSchiavini (approved):

PR #2506, opened 2026-05-11
fix: isDirty boolean from react-hook-form

DanielSchiavini:
- isDirty is true even for automated form updates e.g. for max fields, so use touchedFields instead.


OnlyJousting (approved):
works great!

0xAlunara (approved):

PR #2510, opened 2026-05-12
refactor: use zod in prices api

0xAlunara:
Sorry for the big PR, can't make it any shorter I'm afraid.

* Refactors `@curvefi/prices-api` to use Zod schemas for runtime response validation. `models`, `responses` and `parsers` are replaced by Zod schema definitions inside `schema.ts` per endpoint. The `api` file has become the new `index.ts`, which also exports inferred schema types. Zod allows us to get rid of having to write custom types and parsers, and comes with built-in runtime validation.
* There's a new `test:endpoints` script for `@curvefi/prices-api` which tests every endpoint in a random fashion for liveness and to make sure the schemas are correct with actual live data. Even though the API uses OpenAPI and has an https://prices.curve.finance/feeds-docs/openapi.json, it does not 100% cover all cases and the production environment could deviate.
* Because of the heaviness and unreliability of the API itself, `test:endpoints` is not part of the CI/CD pipeline, but has to be either ran manually locally, or on Github Actions via a manual trigger.
* The `test:endpoints` has a specific test called the *catalog* test, which merely makes sure that every endpoint exposed by export has a test case. This specific test is described in the README.md and inspects typescript files; admittedly this has been completely vibe coded but it seems to work well!

Because of the sheer amount of code, even though not complex, I've used Codex with GPT 5.5 to aid in the migration. It has also aided in writing the testing framework, where code quality matters less than having an actual useful testing pipeline.

In addition to the automated tests I've also manually ran a smoke test through every page, comparing it with production and checking the logs for any Zod parsing errors.

In order to keep the changes local to the Prices API package none of the final schema property names have been untouched. This means there's still a bunch of `.transform()` calls inside the schemas which could be optimized, but it'd come with breaking changes for model consumers.


0xtutti (dismissed):

packages/prices-api/src/proposal/schema.ts
```diff
@@ -0,0 +1,128 @@
...
+  dt: timestampResponse,
+}
+
+const rawProposal = z.object(proposalShape).transform(camelizeKeys)
+type RawProposal = z.infer<typeof rawProposal>
+
+const transformProposal = (data: RawProposal) => ({
+  timestamp: parseTimestamp(data.dt),
+  id: data.voteId,
+  type: normalizeProposalType(data.voteType),
```
Pearce542: the `normalizeProposalType` is useless because zod already validated the proposal type, before normalizing the type

packages/prices-api/tests/seeds.ts
Pearce542, line 1: 🫠

Pearce542 (commented):
Nice this is much better!!

Pearce542:
I cannot trigger a Copilot review 😞 @DanielSchiavini

packages/prices-api/src/llamalend/schema.ts
```diff
@@ -0,0 +1,538 @@
...
+    collateral_token: token,
+    borrowed_token: token,
+    leverage: z.number(),
+    extra_reward_apr: z.array(extraRewardApr),
+    created_at: timestampResponse,
+    max_ltv: z.number(),
+  })
+  .transform(camelizeKeys)
+  .transform(data => {
+    const { version, lendApy, lendApr, lendAprCrv0Boost, lendAprCrvMaxBoost, createdAt, ...market } = data
```
DanielSchiavini:
```suggestion
  .transform(({ version, lendApy, lendApr, lendAprCrv0Boost, lendAprCrvMaxBoost, createdAt, ...market }) =>
```

packages/prices-api/src/llamalend/schema.ts
```diff
@@ -0,0 +1,538 @@
...
+
+    return {
+      ...market,
+      version: parseMarketVersion(version),
```
DanielSchiavini: shouldn't this kind of conversion be happening at the field level, e.g. by defining some custom number format? The distance between `version: z.number()` and `version: parseMarketVersion` is annoying

packages/prices-api/src/llamma/index.ts
```diff
@@ -1,4 +1,86 @@
...
+  const range = getTimeRange({ start, end })
+
+  const params = new URLSearchParams({
+    agg_number: interval.toString(),
+    agg_units: units,
+    start: range.start.toString(),
+    end: range.end.toString(),
+  })
+
+  const resp = await fetch(`${host}/v1/${endpointParam}/llamma_ohlc/${chain}/${llamma}?${params.toString()}`)
```
DanielSchiavini: The backquote already converts everything to string
```suggestion
  const resp = await fetch(`${host}/v1/${endpointParam}/llamma_ohlc/${chain}/${llamma}?${params}`)
```

packages/prices-api/src/llamma/index.ts
```diff
@@ -1,4 +1,86 @@
...
+
+  return Schema.getLlammaOHLCResponse.parse(resp)
```
DanielSchiavini: This is very repetitive. Every call has:
```
  const host = getHost(options)
  Schema.endpoint.parse(endpointParam)
  const params = new URLSearchParams(urlParams)
  const resp = await fetch(`${host}/v1/${endpointParam}/${path}?${params}`)
  return parser(resp)
```

.github/workflows/prices-api-endpoints.yaml
```diff
@@ -0,0 +1,24 @@
+name: Prices API endpoint validation
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  test-endpoints:
```
DanielSchiavini, line 10: I was hoping we'd have a job to generate all this code, not to just test it 🙃 
Is this all really hand/LLM written? Is that a good idea?

packages/prices-api/src/lending/schema.ts
```diff
@@ -0,0 +1,212 @@
...
+type CompleteOracleOHLC = RawOracleOHLC & {
+  close: number
+  high: number
+  low: number
+  open: number
+}
+
+const isCompleteOracleOHLC = (data: RawOracleOHLC): data is CompleteOracleOHLC =>
+  data.close !== null &&
+  data.close !== undefined &&
```
DanielSchiavini:
```suggestion
  data.close != null &&
```

packages/prices-api/tests/catalog.ts
```diff
@@ -0,0 +1,89 @@
...
+  status.missing.length === 0 && status.staleCases.length === 0 && status.staleExclusions.length === 0
+
+/** Formats catalog drift for the loud catalog test and live-test skip messages. */
+export const formatEndpointCatalogStatus = (status: EndpointCatalogStatus) =>
+  [
+    status.missing.length > 0 ? `missing: ${status.missing.join(', ')}` : undefined,
+    status.staleCases.length > 0 ? `stale cases: ${status.staleCases.join(', ')}` : undefined,
+    status.staleExclusions.length > 0 ? `stale exclusions: ${status.staleExclusions.join(', ')}` : undefined,
+  ]
+    .filter(Boolean)
```
DanielSchiavini:
```suggestion
  notFalsy(
    status.missing.length && ? `missing: ${status.missing.join(', ')}`,
    status.staleCases.length && ? `stale cases: ${status.staleCases.join(', ')}`,
    status.staleExclusions.length && `stale exclusions: ${status.staleExclusions.join(', ')}`,
  )
```

packages/prices-api/tests/catalog.ts
```diff
@@ -0,0 +1,89 @@
...
+    .filter(file => file.endsWith('.test.ts'))
+    .flatMap(file => getEndpointIdsFromTestFile(resolve(endpointTestsDir, file)))
+
+const getEndpointIdsFromTestFile = (file: string) => {
+  const source = readFileSync(file, 'utf8')
+  const moduleName = runEndpointCasesPattern.exec(source)?.[1]
+
+  return moduleName
+    ? matches(source, endpointCasePattern).map(functionName => endpointId(moduleName, functionName))
+    : []
```
DanielSchiavini:
```suggestion
  return notFalsyArray(moduleName && matches(source, endpointCasePattern).map(functionName => endpointId(moduleName, functionName)))
```

packages/prices-api/tests/endpoint-cases.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    throw error
+  }
+}
+
+const endpointLabel = (moduleName: EndpointModule, endpoint: EndpointCase) =>
+  endpoint.labelSuffix
+    ? `${moduleName}.${endpoint.functionName}(${endpoint.labelSuffix})`
+    : `${moduleName}.${endpoint.functionName}`
```
DanielSchiavini:
```suggestion
  `${moduleName}.${endpoint.functionName}${`endpoint.labelSuffix ? (${endpoint.labelSuffix})` : ''}`
```

packages/prices-api/tests/endpoint-cases.ts
```diff
@@ -0,0 +1,95 @@
...
+
+const getCatalogSkipReason = (() => {
+  let skipReason: string | undefined
+
+  return () => {
+    if (skipReason !== undefined) {
```
DanielSchiavini:
```suggestion
    if (skipReason) {
```

packages/prices-api/tests/seeds.ts
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    const fetchTracker = createFetchTracker()
+
+    try {
+      value = await fetchTracker.run(load)
+    } catch (seedError) {
+      error = formatSeedError(seedError, load.name || 'endpointSeed', fetchTracker.urls)
+    }
+  }, 180_000)
```
DanielSchiavini: T_T

packages/prices-api/tests/seeds.ts
```diff
@@ -0,0 +1,318 @@
...
+  return () => {
+    if (error) {
+      throw error
+    }
+
+    return requireSeed(value, load.name || 'endpointSeed')
+  }
+}
+
+const getCatalogSkipReason = (() => {
```
DanielSchiavini: duplicated from file above

packages/prices-api/tests/seeds.ts
```diff
@@ -0,0 +1,318 @@
...
+
+const formatSeedError = (error: unknown, source: string, urls: string[]) => {
+  const message = `Failed to load live seed from ${source}\n\n${error instanceof Error ? error.message : String(error)}\n\nPRICES_API_TEST_SEED=${endpointTestSeed}\n\nSeed URL:\n${formatTrackedFetchUrls(urls)}`
+
+  if (error instanceof Error) {
+    try {
+      Object.defineProperty(error, 'message', { configurable: true, value: message })
+      return error
+    } catch {
+      return new Error(message)
```
DanielSchiavini: maybe at least log it?

packages/prices-api/tests/seeds.ts
```diff
@@ -0,0 +1,318 @@
...
+const shuffled = <T>(items: readonly T[]) =>
+  items
+    .map((item, index) => ({ item, index, rank: random() }))
+    .sort((a, b) => a.rank - b.rank || a.index - b.index)
+    .map(({ item }) => item)
+
+const formatSeedError = (error: unknown, source: string, urls: string[]) => {
+  const message = `Failed to load live seed from ${source}\n\n${error instanceof Error ? error.message : String(error)}\n\nPRICES_API_TEST_SEED=${endpointTestSeed}\n\nSeed URL:\n${formatTrackedFetchUrls(urls)}`
+
+  if (error instanceof Error) {
```
DanielSchiavini: do we really need to change the message of **every** error?

packages/prices-api/tests/seeds.ts
```diff
@@ -0,0 +1,318 @@
...
+})
+
+export const getPoolSeed = once(async (): Promise<PoolSeed> => {
+  const supportedChains = await chains.getSupportedChains(requestOptions)
+
+  for (const chain of shuffled(supportedChains)) {
+    const response = await pools.getPools(chain, requestOptions)
+    const candidates = response.pools.filter(item => item.coins.length >= 2)
+
+    if (candidates.length === 0) {
```
DanielSchiavini:
```suggestion
    if (!candidates.length) {
```

packages/prices-api/tests/seeds.ts
```diff
@@ -0,0 +1,318 @@
...
+    'chains.getSupportedChains',
+  )
+})
+
+export const getPoolSeed = once(async (): Promise<PoolSeed> => {
+  const supportedChains = await chains.getSupportedChains(requestOptions)
+
+  for (const chain of shuffled(supportedChains)) {
+    const response = await pools.getPools(chain, requestOptions)
+    const candidates = response.pools.filter(item => item.coins.length >= 2)
```
DanielSchiavini: why do we need this filter?

OnlyJousting (dismissed):

packages/prices-api/README.md
```diff
@@ -22,17 +22,55 @@ const markets = await getMarkets('ethereum')
 - Chain data (TVL, volumes, fees)
 - crvUSD / LlamaLend markets and user positions
 - DAO data (votes, locks, gauges)
+- Runtime-validated responses using Zod schemas
```
DanielSchiavini, line 25: Not sure if this is a good idea tbh, do we wanna get errors when the prices API misbehaves or removes a field? I wanna know about it for sure, but maybe we can disable that in prod?

tests/cypress/e2e/all/error-boundary.cy.ts
```diff
@@ -3,31 +3,94 @@ import { API_LOAD_TIMEOUT, e2eBaseUrl, LOAD_TIMEOUT } from '@cy/support/ui'
...
+  })
   cy.intercept(`https://prices.curve.finance/v1/lending/markets`, {
     body: {
       chains: {
         ethereum: {
+          count: 1,
           data: [
-            // make an error occur on purpose by returning nonsense data (`address` should be string)
-            // note that the error only triggers the boundary if the query succeeds, but it fails during rendering
+            // Keep the API response valid, then trigger the render error from the icon path below.
```
DanielSchiavini: oh damn that gets long 😞

packages/prices-api/src/schemas.ts
```diff
@@ -0,0 +1,42 @@
...
+import type { TimestampResponse } from './timestamp'
+import { chains } from './index'
+
+type Primitive = bigint | boolean | null | number | string | symbol | undefined
+
+export type CamelCase<Value extends string> = Value extends `${infer Head}_${infer Tail}`
+  ? `${Head}${Capitalize<CamelCase<Tail>>}`
+  : Value
+
+export type Camelize<Value> = Value extends Primitive
```
DanielSchiavini, line 37: fancy stuff. Please add jsdoc

packages/prices-api/src/llamalend/index.ts
```diff
@@ -1,5 +1,135 @@
...
+
+export async function getAllMarkets(
+  params: {
+    page?: number
+    per_page?: number
+  } = {},
+  options?: Options,
+) {
+  const host = getHost(options)
+  const resp = await fetch(`${host}/v1/lending/markets${addQueryString(params)}`)
```
DanielSchiavini: Why abbreviate?
```suggestion
  const response = await fetch(`${host}/v1/lending/markets${addQueryString(params)}`)
```

packages/prices-api/src/proposal/schema.ts
0xAlunara: Ah good call, seems to be a legacy carry over. I've removed it.

.github/workflows/prices-api-endpoints.yaml
0xAlunara: I don't know what the guys are exactly using for the API server, but there's been a few cases where the `openapi.json` didn't match with nullability of the actual output, so I feel safer having an extensive testsuite anyway. For what it's worth, I've pointed it to hand-written models and the `openapi.json`, in order to convert to Zod schemas; it would have been too much to do by hand. The test suite in turn only found a few nullability issues, but other than that there's been no more errors on my end.

packages/prices-api/README.md
0xAlunara: The whole point is to fail early indeed, because otherwise errors might silently creep through the rest of the front-end. It doesn't appear Zod offers a way to ignore parsing errors; you either parse succesfully or you get an error. For now the errors only show up in the logs, but we could make it more visible with a notification so it's caught earlier?

Zod parsing errors should also show up in Sentry but I haven't been able to properly test that yet (there's none showing up yet in prod)

packages/prices-api/src/llamalend/index.ts
0xAlunara: Why not? It has always been this way too. Not sure what 'resp' could otherwise be 😄

DanielSchiavini: just in principle, it always takes a little bit longer to interpret abbreviated stuff :nerd_face:

packages/prices-api/src/llamalend/schema.ts
0xAlunara: For some reason we've decided to work with version enums on the llamalend app. But honestly that's very specific to just us, so I've migrated the actual parsing over from here to `llama-markets.ts` such that it remains a strict number here.

packages/prices-api/src/llamma/index.ts
0xAlunara: Yeah, but it's similar to how it currently already is. We could make it more generic but I'd like to save that for another PR. I've also tried to generalize it back in the day for CurveMonitor too, but regretted that decision and had to abandon that, as some calls sometimes deviated slightly from the norm.

tests/cypress/e2e/all/error-boundary.cy.ts
0xAlunara: Perhaps we need a .json fixture file like we do for `mint-markets.json`?

packages/prices-api/tests/seeds.ts
0xAlunara: Vitest will log any errors it gets thrown at

packages/prices-api/src/llamalend/index.ts
0xAlunara: Will change it and make sure it all looks consistent, but I can see us refactoring and generalizing this in the future so it'll probably go away anyway.

tests/cypress/e2e/all/error-boundary.cy.ts
0xAlunara: Ah, turns out I can just use `mockLendingVaults` instead!

packages/prices-api/tests/seeds.ts
0xAlunara: Just the seed errors, but I just inlined this function with a simpler formatting now

apps/main/src/llamalend/queries/market-list/llama-markets.ts
```diff
@@ -87,6 +87,17 @@ export type LlamaMarketsResult = {
...
+  switch (version) {
+    case 1:
+      return LlamaMarketVersion.v1
+    case 2:
+      return LlamaMarketVersion.v2
+    default:
+      throw new Error(`Unsupported LlamaLend market version: ${version}`)
+  }
+}
+
```
DanielSchiavini:
```suggestion
const toMarketVersion = (version: number): LlamaMarketVersion =>
  assert({1: LlamaMarketVersion.v1, 2: LlamaMarketVersion.v2}[version], `Unsupported LlamaLend market version: ${version}`)

```

tests/cypress/e2e/all/error-boundary.cy.ts
```diff
@@ -1,33 +1,38 @@
...
+    })),
+  }
+  mockLendingVaults(chains).as('error')
   const url = '/llamalend/ethereum/markets'
-  cy.visit(url, { timeout: API_LOAD_TIMEOUT.timeout })
+  cy.visit(url, {
+    timeout: API_LOAD_TIMEOUT.timeout,
+    onBeforeLoad: ({ String }) => {
+      const originalToLowerCase = String.prototype.toLowerCase
+      String.prototype.toLowerCase = function (this: string) {
```
DanielSchiavini, line 28: is cypress going to reload the window to recover this for the next test?

packages/prices-api/src/schemas.ts
```diff
@@ -0,0 +1,70 @@
...
+export const decimal = z.string().transform(value => value as Decimal)
+export const hex = z.string().transform(value => value as Hex)
+export const timestampResponse = z.union([z.number(), z.string()]).transform(value => value as TimestampResponse)
+
+type Primitive = bigint | boolean | null | number | string | symbol | undefined
+
+/**
+ * Converts a snake_case string literal into its camelCase equivalent.
+ *
+ * Used by `Camelize` to keep the inferred output type in sync with
```
DanielSchiavini: jsdoc is breaking lines at weird places

packages/prices-api/src/yield-basis/schema.ts
```diff
@@ -0,0 +1,133 @@
...
+  .object({
+    tx_hash: hex,
+    block_number: z.number(),
+    timestamp: timestampResponse,
+    volume: z.number(),
+    adjacent_volume: z.number(),
+    adjacent_pools: z.array(adjacentPool),
+  })
+  .transform(camelizeKeys)
+  .transform(({ timestamp, ...transaction }) => ({ ...transaction, timestamp: parseTimestamp(timestamp) }))
```
DanielSchiavini: why do we only parse it here and not in `timestampResponse`?

packages/prices-api/src/revenue/schema.ts
```diff
@@ -0,0 +1,209 @@
...
+    fee_amount: z.string(),
+    amount_received: z.number(),
+    router_received: z.number(),
+    epoch: z.number(),
+    tx_hash: address,
+    block_number: z.number(),
+    dt: timestampResponse,
+  })
+  .transform(camelizeKeys)
+  .transform(data => ({
```
DanielSchiavini: why are we not using the ...data syntax here?

packages/prices-api/src/revenue/schema.ts
```diff
@@ -0,0 +1,209 @@
...
+    timestamp: parseTimestamp(data.dt),
+    coin: data.coin,
+    amount: BigInt(data.amount),
+    amountFee: BigInt(data.feeAmount),
```
DanielSchiavini: shouldn't we define a `z.bigInt` field?

packages/prices-api/src/proposal/schema.ts
```diff
@@ -0,0 +1,125 @@
...
+  id: data.voteId,
+  type: data.voteType,
+  metadata: data.metadata?.startsWith('"') // Remove weird starting quote, if present.
+    ? data.metadata.substring(1)
+    : (data.metadata ?? ''),
+  proposer: data.creator,
+  block: data.snapshotBlock,
+  start: data.startDate,
+  end: data.startDate + 604800,
+  quorum: Number(BigInt(data.minAcceptQuorum)) / 10 ** 18, // Voting power in veCRV.
```
DanielSchiavini: The code and the comment are repeated (also in other places below), it seems worth a helper.
Shouldn't the calculation happen with more precise stuff than `number`?

packages/prices-api/src/proposal/index.ts
```diff
@@ -1,4 +1,79 @@
...
+  user: string,
+  proposalId: number,
+  proposalType: Schema.ProposalType,
+  txHash?: string,
+  options?: Options,
+) {
+  const pagination = 100
+  let page = 1
+
+  while (true) {
```
DanielSchiavini, line 56: can't we use the `paginate` helper? T_T

packages/prices-api/src/proposal/index.ts
```diff
@@ -1,4 +1,79 @@
...
+}
+
+export async function getUserProposalVotes(user: string, page = 1, pagination = 100, options?: Options) {
+  try {
+    const host = getHost(options)
+    const response = await fetch(`${host}/v1/dao/proposals/votes/user/${user}?pagination=${pagination}&page=${page}`)
+
+    return Schema.getUserProposalVotesResponse.parse(response)
+  } catch (err) {
+    if (err instanceof FetchError) {
```
DanielSchiavini: why would we want to catch any fetch error?

packages/prices-api/src/pools/schema.ts
```diff
@@ -0,0 +1,373 @@
...
+    fee: z.number(),
+    usd_fee: z.number().nullable(),
+    token_sold: tradeToken,
+    token_bought: tradeToken,
+    pool_state: z.unknown().nullable(),
+  })
+  .transform(camelizeKeys)
+  .transform(({ time, transactionHash, tokensSoldUsd, tokensBoughtUsd, usdFee, ...trade }) => ({
+    ...trade,
+    tokensSoldUsd: tokensSoldUsd ?? 0,
```
DanielSchiavini, line 151: null/undefined and 0 are definitely not the same 😞 
do we have a good reason to do that throughout this file?

packages/prices-api/src/ohlc/index.ts
```diff
@@ -1,4 +1,27 @@
...
+import * as Schema from './schema'
+
+export type * from './schema'
+
+export async function getOHLC(chain: Chain, poolAddr: string, tokenMain: string, tokenRef: string, options?: Options) {
+  const host = getHost(options)
+
+  const { start, end } = getTimeRange({ daysRange: 90 })
+
+  const url =
```
DanielSchiavini, line 13: what happened to URLSearchParams?

packages/prices-api/src/llamalend/schema.ts
```diff
@@ -0,0 +1,512 @@
...
+    current_assets_in_gauge: z.string(),
+    current_assets_in_convex: z.string(),
+    total_current_shares: z.string(),
+    total_current_assets: z.string(),
+    boost_multiplier: z.number().nullable(),
+  })
+  .transform(camelizeKeys)
+  .transform(data => ({
+    user: data.user,
+    earnings: parseFloat(data.earnings),
```
DanielSchiavini, line 235: that's a lot of repetition 😞 why aren't we using some other z.helper?

packages/prices-api/src/gauge/schema.ts
```diff
@@ -0,0 +1,186 @@
...
+    pool: data.pool
+      ? {
+          address: data.pool.address,
+          name: data.pool.name,
+          chain: data.pool.chain as Chain,
+          tvlUsd: data.pool.tvlUsd,
+          tradingVolume24h: data.pool.tradingVolume24h,
+        }
+      : undefined,
+    tokens: (data.tokens ?? []).map(token => ({
```
DanielSchiavini:
```suggestion
    tokens: data.tokens?.map(token => ({
```

packages/prices-api/src/gauge/schema.ts
```diff
@@ -0,0 +1,186 @@
...
+    last_vote_date: timestampResponse.nullable(),
+    last_vote_tx: address.nullable(),
+  })
+  .transform(camelizeKeys)
+  .transform(data => ({
+    address: data.address,
+    type: data.gaugeType,
+    name: data.name ?? undefined,
+    version: data.version ?? undefined,
+    lpToken: data.lpToken ? data.lpToken : undefined,
```
DanielSchiavini: this method has a lot of `x ? x : undefined` are we trying to filter out `0` or something? Anyway that's the same as `x || undefined`

packages/prices-api/src/dao/schema.ts
```diff
@@ -0,0 +1,145 @@
...
+    user: z.string().transform(value => value as Address | `Others(${string})`),
+    locked: z.string(),
+    weight: z.string(),
+    weight_ratio: z.string(),
+    unlock_time: timestampResponse.nullable(),
+  })
+  .transform(camelizeKeys)
+  .transform(data => ({
+    user: data.user,
+    locked: BigInt(Math.round(parseFloat(data.locked))),
```
DanielSchiavini: why is a float converted to bigint? 🤔  Also repeated elsewhere in this file.

tests/cypress/e2e/all/error-boundary.cy.ts
0xAlunara: Looks like it on my end yeah, it completely blanks out the frame it's showing and revisits the URL

packages/prices-api/src/schemas.ts
0xAlunara: I like it though 😭

packages/prices-api/src/yield-basis/schema.ts
0xAlunara: Initially the goal was to separate raw API validation and its transformation to a domain model. But I don't think there's any case where we want to use the raw timestamp response rather than the parsed one, so I'll go just ahead and parse it immediately.

packages/prices-api/src/gauge/schema.ts
0xAlunara: We're converting `null` to be `undefined` for consistency, we could revise that later but I don't dare to touch that behavior atm.

packages/prices-api/src/llamalend/schema.ts
0xAlunara: Because there's like two exceptions to the rule here. Don't want to spend time generalizing just yet

packages/prices-api/src/pools/schema.ts
0xAlunara: The only good reason is to maintain compatibility with the existing usages across curve front-end. I want to give these `0` / `null` / `undefined` default values another pass after the migration to Zod has gone through without issue.

packages/prices-api/src/proposal/schema.ts
0xAlunara: I'd rather keep the bigint but I don't want to also have to make changes on the DAO app to reparse it. Another time.

packages/prices-api/src/revenue/schema.ts
0xAlunara: z.bigInt() already exists, but the apps weren't made with that in mind so they're expecting numbers for now, will convert it later at some point I guess

packages/prices-api/src/dao/schema.ts
0xAlunara: For some reason the API can return something like "353785687565873647865345345355.23", some python rounding I guess

packages/prices-api/src/ohlc/index.ts
0xAlunara: Was never used here

packages/prices-api/src/proposal/index.ts
0xAlunara: Can't remember, must be a legacy fallback we don't want anymore I guess

packages/prices-api/src/proposal/index.ts
0xAlunara: The nuance here is that this isn't really paginating but rather it keeps on going indefinitely until a certain condition has been met. Could write a helper method for that but not now. I've added a comment though.

packages/prices-api/src/pools/schema.ts
```diff
@@ -0,0 +1,351 @@
...
+  tokensBought: data.tokensBought,
+  tokensBoughtUsd: data.tokensBoughtUsd ?? 0,
+  blockNumber: data.blockNumber,
+  time: data.time,
+  txHash: data.transactionHash,
+  buyer: data.buyer,
+  usdFee: data.usdFee ?? 0,
+})
+
+const requireTradeToken = (token: TradeTokenData | undefined, tokenId: number) => {
```
DanielSchiavini, line 118: this is a simple `assert`

packages/prices-api/src/proposal/schema.ts
```diff
@@ -42,15 +43,15 @@ const transformProposal = (data: RawProposal) => ({
   block: data.snapshotBlock,
   start: data.startDate,
   end: data.startDate + 604800,
-  quorum: Number(BigInt(data.minAcceptQuorum)) / 10 ** 18, // Voting power in veCRV.
-  support: Number(BigInt(data.supportRequired)) / 10 ** 18, // Voting power in veCRV.
+  quorum: Number(BigInt(data.minAcceptQuorum)) / 10 ** 18,
```
DanielSchiavini, line 46: lol the comments are removed but the code is still repeated 😆

DanielSchiavini (approved):

Pearce542 (approved):

OnlyJousting (approved):

PR #2518, opened 2026-05-13
fix: lend market loading

DanielSchiavini:
- fix lend hook to make sure we don't set isSuccess until we actually are done hydrating
- refactor lend market queries to match lend
- fix market rates query validation


apps/main/src/lend/hooks/useLendMarket.ts
```diff
@@ -26,11 +26,10 @@ const useLendMarketMapping = ({ chainId }: ChainParams<ChainId>) => {
         : undefined,
     [api, apiChainId, chainId, isHydrated, marketNames],
   )
-  return { data, isSuccess, error }
+  return { data, isSuccess: !!data, error, isLoading }
```
DanielSchiavini, line 29: isSuccess will give you data from the cache which we don't want in this case 😞

0xtutti (approved):

apps/main/src/lend/hooks/useLendMarket.ts
0xAlunara: Feels a bit odd this way. What if we invalidate the `useLendMarketNames` query (and possible the mint market one too) during hydration (or other bootstrap phase) so we keep the caching across the app's pages, but we're not keeping it across tabs and refreshes so it starts clean?

0xAlunara (approved):

apps/main/src/lend/hooks/useLendMarket.ts
DanielSchiavini: Sounds like a good idea, we can try!

OnlyJousting (approved):

apps/main/src/lend/hooks/useLendMarket.ts
DanielSchiavini: See https://github.com/curvefi/curve-frontend/pull/2521 but maybe that can get replaced by https://github.com/curvefi/curve-llamalend.js/pull/100

PR #2524, opened 2026-05-15
refactor: use some mui components on dex pool page

0xAlunara:
Don't judge this PR too harshly, I just couldn't help myself to get rid of some legacy boxes by using proper grids, stacks and card components. A variety of stuff.

Because I got rid of some boxes and nesting, the diff is best viewed with "Ignore whitespace" on

**This is not a proper redesign according to Figma, just an intermediate cleanup step**
**Two of the tabs will most likely be deleted in the future anyway**

image

image

image


apps/main/src/dex/components/OhlcAndActivityComp/index.tsx
```diff
@@ -48,46 +47,48 @@ export const OhlcAndActivityComp = ({
...
+          <ActivityTable
+            table={tradesTable.table}
+            isLoading={tradesTable.isLoading}
+            isError={tradesTable.isError}
+            emptyMessage={tradesTable.emptyMessage}
+            expandedPanel={PoolTradesExpandedPanel}
+          />
+        )}
+        {tab === 'chart' && (
+          <Stack sx={{ gap: Spacing.md, padding: Spacing.sm }}>
```
0xAlunara, line 72: Should prob be proper cards but I just copied the market page for now

apps/main/src/dex/components/PagePool/PoolDetails/PoolParameters/index.tsx
```diff
@@ -101,309 +107,288 @@ export const PoolParameters = ({ poolData, rChainId }: PoolParametersProps) => {
...
-                      <IndentDataTitle>{t`Oracle Address:`}</IndentDataTitle>
-                      {poolMetadata.oracles[idx].oracleAddress ? (
-                        <IndentDataAddressLink href={scanTokenPath(network, poolMetadata.oracles[idx].oracleAddress)}>
-                          {shortenAddress(poolMetadata.oracles[idx].oracleAddress)}
-                        </IndentDataAddressLink>
-                      ) : (
-                        <IndentData>-</IndentData>
+    <Card size="small">
+      <CardContent sx={{ padding: '0 !important' }} /** TODO: no data-inline support yet like in header */>
+        <Box display="grid" gap={Spacing.md} gridTemplateColumns="repeat(12, 1fr)">
```
0xAlunara: I think I'm going to ban using mui's `<Grid>`. The moment you add column spacing you end up with ugly wrapping, despite the sizes adding up to exactly 12. A box with display grid is way more versatile anyway. To be continued.

DanielSchiavini: That should definitely not happen, I've been using a lot of mui grid's for estamap.com and the spacing works fine 🤔 
Did you try to use `spacing` instead of `gap`?

apps/main/src/dex/components/PagePool/PoolDetails/PoolStats/index.tsx
```diff
@@ -44,89 +47,50 @@ export const PoolStats = ({ routerParams, poolAlert, poolData, poolDataCacheOrAp
...
-              {tokenAlert && tokenAlert.isInformationOnly && <AlertBox {...tokenAlert}>{tokenAlert.message}</AlertBox>}
-
-              {poolDataCacheOrApi.pool.referenceAsset === 'CRYPTO' && (
-                <AlertBox alertType="info" title={t`${poolDataCacheOrApi.pool.name} is a Cryptoswap pool`}>
-                  {t`Cryptoswap pools contain non pegged assets. Liquidity providers are exposed to all assets in the pools.`}{' '}
-                  <ExternalLink $noStyles href="https://docs.curve.finance/user/dex/overview">
-                    {t`Click here to learn more about Cryptoswap pools`}
-                  </ExternalLink>
+    <Card size="small">
+      <CardContent sx={{ padding: '0 !important' }} /** TODO: no data-inline support yet like in header */>
```
DanielSchiavini: why do we have nested CardContent? I think the whole point of this component is to add padding.

DanielSchiavini (commented):
The grid-like `Box` and the `CardContents` with overridden `padding: 0` could at least be shared code instead of repeated

apps/main/src/dex/components/OhlcAndActivityComp/index.tsx
Pearce542: we could move the gap and padding outside the `sx`

0xAlunara: I don't think so, because we only want it for the chart; the table tabs should be paddingless.

apps/main/src/dex/components/PagePool/PoolDetails/PoolParameters/index.tsx
0xAlunara: Aye, that did the trick x.x
Not sure why gap even exists if it's broken like that

apps/main/src/dex/components/PagePool/PoolDetails/PoolParameters/index.tsx
```diff
@@ -101,309 +108,292 @@ export const PoolParameters = ({ poolData, rChainId }: PoolParametersProps) => {
...
-                          {shortenAddress(poolMetadata.oracles[idx].oracleAddress)}
-                        </IndentDataAddressLink>
-                      ) : (
-                        <IndentData>-</IndentData>
+    <Card size="small">
+      <CardContent
+        component={Grid}
+        container
+        columnSpacing={Spacing.md}
+        sx={{ padding: '0 !important' /** The default padding interferes with background colors of the inner grids */ }}
```
0xAlunara, line 116: I'll make a ticket for a proper inline property for cases like these

Pearce542 (approved):

DanielSchiavini (approved):

apps/main/src/dex/components/OhlcAndActivityComp/index.tsx
DanielSchiavini: I think he meant this
```suggestion
          <Stack gap={Spacing.md} padding={Spacing.sm}>
```

OnlyJousting (approved):

PR #2526, opened 2026-05-15
feat: llama table active filters collapsible

Pearce542:
Added a collapsible bar for active filters as shown in the screenshot below
I wanted to add as little custom code for the filters as possible, because it’s easier to maintain.
Right now, we differentiate between:

- A list of values, such as collateral tokens: we render one `Chip` per value.
- A range of values, such as utilization or borrow APR: we display a custom chip with <, >, and -.
- Chains: we render `ChainFilterChips` with the chain icon.

Screenshot 2026-05-19 at 21 42 04


apps/main/src/llamalend/features/market-list/columns/column.definitions.tsx
```diff
@@ -33,9 +33,34 @@ import {
...
+/** Define a hidden column. */
+const hidden = (field: DeepKeys<LlamaMarket>, id: LlamaMarketColumnId, filterFn: FilterFnOption<LlamaMarket>) =>
+  columnHelper.accessor(field, {
+    id,
+    filterFn,
+    header: LLAMA_MARKET_TITLES[id],
+    meta: { hidden: true },
+  })
+
+export const LLAMA_MARKET_TITLES: Record<LlamaMarketColumnId, string> = {
```
Pearce542, line 55: Made the title required (for hidden column as well) since it's being used to display the active filters as well

apps/main/src/llamalend/features/market-list/columns/column.definitions.tsx
```diff
@@ -45,87 +70,75 @@ const headers = {
   [LlamaMarketColumnId.UserEarnings]: t`My Earnings`,
   [LlamaMarketColumnId.UserDeposited]: t`Supplied Amount`,
   [LlamaMarketColumnId.BorrowRate]: t`Borrow APR`,
-  [LlamaMarketColumnId.NetBorrowRate]: t`Net borrow APR`,
+  [LlamaMarketColumnId.NetBorrowRate]: t`Net Borrow APR`,
```
Pearce542, line 73: Changed all words to capitalized for consistency, but we should decide what to do, I saw a mix of both on Figma @0xtutti

apps/main/src/llamalend/features/market-list/filters/RangeFilter.tsx
```diff
@@ -79,6 +79,7 @@ export const RangeFilter = <TKey, TColumnId extends string>({
   }) => (
     <NumericTextField
       size="small"
+      variant="outlined"
```
Pearce542, line 82: the border of this input variant is a bit aggressive but that's the correct variant. the color will be update in the next color sync

0xtutti (approved):

0xtutti (dismissed):

apps/main/src/llamalend/features/market-list/filters/HiddenInlinedItems.tsx
```diff
@@ -0,0 +1,13 @@
+import { ReactNode } from 'react'
+
+/** Renders the overflow indicator for items hidden behind a max visibility limit.  */
+export const HiddenInlinedItems = ({
+  hiddenSelectedItemsLength,
+  renderItem,
+}: {
+  hiddenSelectedItemsLength: number
+  renderItem?: (value: string) => ReactNode
```
DanielSchiavini, line 9: using an object would make this more extendable (not important)
```suggestion
  renderItem?: (props: { value: string }) => ReactNode
```

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersCollapsible.tsx
```diff
@@ -0,0 +1,141 @@
...
+  return parseListFilter(serializedFilter)
+}
+
+const removeFilterValue = (
+  id: LlamaMarketColumnId,
+  value: string,
+  clickedValues: string | string[],
+  setColumnFilter: FilterProps<LlamaMarketColumnId>['setColumnFilter'],
+) => {
+  if (value.includes(RANGE_SEPARATOR)) {
```
DanielSchiavini: This is a bit of a weird function, I think we would want different 'remove' callbacks for the different types of filter.
For example, couldn't this RANGE_SEPARATOR exist in a chain name for example? Or break something if the separator is changed?
Seeing the usages of this function I don't see any reason to couple all fields into a single 'remove' function.

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersCollapsible.tsx
```diff
@@ -0,0 +1,141 @@
...
+export const LlamaTableFiltersCollapsible = <T extends TableItem>({
+  table,
+  resetFilters,
+  setColumnFilter,
+}: {
+  table: TanstackTable<T>
+  resetFilters: () => void
+} & FilterProps<LlamaMarketColumnId>) => {
+  const columnFiltersState = table.getState().columnFilters as { id: LlamaMarketColumnId; value: string }[]
+  const toggleFilterValue = useCallback(
```
DanielSchiavini: the function is called toggle but it only calls remove

packages/curve-ui-kit/src/shared/ui/DataTable/TableFilters.tsx
```diff
@@ -35,6 +37,8 @@ export const TableFilters = <ColumnIds extends string>({
   visibilityGroups: VisibilityGroup<ColumnIds>[]
   toggleVisibility?: (columns: string[]) => void
   popoverFilters?: ReactNode // filters shown in the popover menu
+  // collabsible bar that displays the active filters
+  collapsibleFilters?: { collapsible: ReactNode; hasActiveFilters?: boolean | undefined }
```
DanielSchiavini, line 40: it's not clear from the name or the comment that this is only for `!isMobile`

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersCollapsible.tsx
DanielSchiavini: also, it's memoized with useCallback but then the memoization is totally ignored by calling it inside arrow functions

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersCollapsible.tsx
Pearce542: I see, should we add a required `filterType` to the meta object of each column then? We need kind of a mapping between each column and the filter type in order to parse it correctly

DanielSchiavini: but why do we need to parse at all? I see the usages of this function, they could all be separate callbacks right?

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersCollapsible.tsx
```diff
@@ -0,0 +1,141 @@
...
+    (id: LlamaMarketColumnId, value: string, clickedValues: string | string[]) =>
+      removeFilterValue(id, value, clickedValues, setColumnFilter),
+    [setColumnFilter],
+  )
+
+  return (
+    <Stack
+      paddingBlock={Spacing.xs}
+      paddingInline="10px"
```
DanielSchiavini: let's use `Spacing`

apps/main/src/llamalend/features/market-list/columns/column.definitions.tsx
```diff
@@ -135,78 +148,81 @@ export const LLAMA_MARKET_COLUMNS = [
...
     cell: CompactUsdCell,
     meta: { type: 'numeric' },
     sortUndefined: 'last',
   }),
   columnHelper.accessor(LlamaMarketColumnId.Tvl, {
-    header: headers[LlamaMarketColumnId.Tvl],
+    header: LLAMA_MARKET_TITLES[LlamaMarketColumnId.Tvl],
     cell: TvlCell,
     meta: {
       type: 'numeric',
```
DanielSchiavini, line 224:
```typescript
export const LLAMA_RANGE_COLUMNS = LLAMA_MARKET_COLUMNS.filter(c => c.filterFn === rangeFilterFn).map(c => c.id)
export const LLAMA_LIST_COLUMNS = LLAMA_MARKET_COLUMNS.filter(c => c.filterFn === listFilterFn).map(c => c.id)
```

packages/curve-ui-kit/src/shared/ui/DataTable/DataTable.d.ts
```diff
@@ -10,6 +10,7 @@ import type { TooltipProps } from '../Tooltip'
 declare module '@tanstack/react-table' {
   interface ColumnMeta<TData extends RowData, TValue> {
     type?: 'numeric' // aligns cell content to the right
+    unit?: 'percentage' | 'dollar' // used when displaying the filter's serialized value
```
0xAlunara: Would it make sense to just give it the type of the parameters of `formatNumber`? 😄 
Otherwise we're going to endlessly expand it to suit our needs.

apps/main/src/llamalend/features/market-list/filters/utils.ts
```diff
@@ -0,0 +1,17 @@
+import { splitAt } from '@primitives/array.utils'
+import { notFalsyArray } from '@primitives/objects.utils'
+
+const MAX_VISIBLE_INLINED_ITEMS = 4
+
+/**  Split an array between visible and hidden items to maintain inline compaction.
```
0xAlunara: Weird doc formatting 🤔 
AI wouldn't make such mistake 😔

apps/main/src/llamalend/features/market-list/filters/RangeFilter.tsx
0xAlunara: I created a Notion ticket for this as I have the same issue with the refuel page's form.

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersCollapsible.tsx
```diff
@@ -0,0 +1,128 @@
...
+}: {
+  table: TanstackTable<T>
+  resetFilters: () => void
+} & FilterProps<LlamaMarketColumnId>) => {
+  const filtersState = table.getState().columnFilters as { id: LlamaMarketColumnId; value: string }[]
+  // Keep networks first and sort the remaining filters alphabetically to avoid making the filters jump to another position when filters are removed
+  const sortedFiltersState = useMemo(
+    () =>
+      filtersState.sort((a, b) => {
+        if (a.id === LlamaMarketColumnId.Chain) return -1
```
0xAlunara, line 64: I'm just here to say I prefer this over an arrow bodied ternary

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersCollapsible.tsx
```diff
@@ -0,0 +1,128 @@
...
+      paddingInline={Spacing.sm}
+      direction="row"
+      alignItems="end"
+      gap={Spacing.sm}
+      justifyContent="space-between"
+      sx={{ borderTop: t => `1px solid ${t.design.Layer[1].Outline}` }}
+    >
+      <Stack direction="row" gap={Spacing.sm} flexWrap="wrap">
+        {sortedFiltersState.map(({ id, value }) => {
+          const column = table.getColumn(id)
```
0xAlunara: I personally don't mind, but I'm sure somebody else will suggest extracting this into a subcomponent 😄

0xAlunara (commented):
Image

Just my opinion, but initially I didn't see the titles above the chips. Should we make them bolder? And perhaps increase the gap between categories by one t-shirt size?

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersCollapsible.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
+import { formatNumber } from '@ui-kit/utils'
+import { LlamaMarketColumnId } from '../columns/columns.enum'
+import { HiddenInlinedItems } from './HiddenInlinedItems'
+import { SelectedFilterChips } from './SelectedFilterChips'
+import { getInlinedItemsVisibility } from './utils'
+
+const { Spacing } = SizesAndSpaces
+
+const formatRangeValue = (value: number, unit?: 'percentage' | 'dollar') =>
```
DanielSchiavini:
```suggestion
const formatRangeValue = (value: number, unit?: ColumnMeta['unit']) =>
```

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersCollapsible.tsx
```diff
@@ -0,0 +1,128 @@
...
+  formatNumber(value, { abbreviate: true, ...(unit && { unit }) })
+
+// capitalize all labels except columns containing tokens symbols
+const formatLabel = (label: string, id: LlamaMarketColumnId) =>
+  [LlamaMarketColumnId.CollateralSymbol, LlamaMarketColumnId.BorrowedSymbol].includes(id)
+    ? label
+    : `${label.charAt(0).toUpperCase()}${label.slice(1)}`
```
DanielSchiavini: Did you consider using [`text-transform: capitalize`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/text-transform)?
capitalize
```suggestion
    : capitalize(label)
```

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersCollapsible.tsx
```diff
@@ -0,0 +1,128 @@
...
+export const LlamaTableFiltersCollapsible = <T extends TableItem>({
+  table,
+  resetFilters,
+  setColumnFilter,
+}: {
+  table: TanstackTable<T>
+  resetFilters: () => void
+} & FilterProps<LlamaMarketColumnId>) => {
+  const filtersState = table.getState().columnFilters as { id: LlamaMarketColumnId; value: string }[]
+  // Keep networks first and sort the remaining filters alphabetically to avoid making the filters jump to another position when filters are removed
```
DanielSchiavini: alphabetically by id not label, is that on purpose? Just to keep a constant order I guess. Maybe better to use the index in LLAMA_COLUMNS?

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersCollapsible.tsx
DanielSchiavini: Is no column with that ID a real use case? Otherwise:
```suggestion
          const column = assert(table.getColumn(id), `no column with id ${id}`)
```

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersCollapsible.tsx
```diff
@@ -0,0 +1,128 @@
...
+          const column = table.getColumn(id)
+          const isRangeFilterFn = column?.getFilterFn() === rangeFilterFn
+          const labels = isRangeFilterFn
+            ? notFalsy(getRangeLabel(value, column?.columnDef.meta?.unit))
+            : parseListFilter(value)
+          const [visibleLabels, hiddenLabels] = getInlinedItemsVisibility(labels)
+
+          const removeClickedValue = (clickedValue: string | string[]) => {
+            const activeValues = parseListFilter(value)
+            const remainingValues = activeValues?.filter(v => !toArray(clickedValue).includes(v))
```
DanielSchiavini: not a big deal but `toArray` will create a new array for every iteration
```suggestion
            const removedValues = new Set(toArray(clickedValue))
            const remainingValues = activeValues?.filter(v => !removedValues.has(v))
```

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersCollapsible.tsx
```diff
@@ -0,0 +1,128 @@
...
+            setColumnFilter(id, serializeListFilter(remainingValues))
+          }
+
+          return (
+            !!labels?.length && (
+              <SelectedFilterChips key={`selected-chip-${id}`} title={column?.columnDef.header as string}>
```
DanielSchiavini: `as string` isn't necessarily safe, do we really need this cast? I think `SelectedFilterChips` could allow `ReactNode` titles?

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersCollapsible.tsx
```diff
@@ -0,0 +1,128 @@
...
+                      <ActiveFilterChip
+                        key={`${label}-${id}`}
+                        label={formatLabel(label, id)}
+                        toggle={isRangeFilterFn ? () => setColumnFilter(id, null) : () => removeClickedValue(label)}
+                      />
+                    ))}
+                    <HiddenInlinedItems
+                      hiddenSelectedItemsLength={hiddenLabels.length}
+                      renderItem={label => (
+                        <ActiveFilterChip label={label} toggle={() => removeClickedValue(hiddenLabels)} />
```
DanielSchiavini, line 115: I expected that clicking the hidden items would show all of them, instead of removing all of them

packages/curve-ui-kit/src/shared/ui/DataTable/DataTable.d.ts
Pearce542: Yes good catch

apps/main/src/llamalend/features/market-list/filters/utils.ts
Pearce542: We should discuss how to format js doc consistently

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersCollapsible.tsx
Pearce542: This was initially meant for presentation only, we didn't want them to be clickable either, we should use the filter modal for more control

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersCollapsible.tsx
```diff
@@ -78,22 +80,26 @@ export const LlamaTableFiltersCollapsible = <T extends TableItem>({
...
+            const remainingValues = activeValues?.filter(v => !removedValues.has(v))
             setColumnFilter(id, serializeListFilter(remainingValues))
           }
 
           return (
             !!labels?.length && (
-              <SelectedFilterChips key={`selected-chip-${id}`} title={column?.columnDef.header as string}>
+              <SelectedFilterChips
+                key={`selected-chip-${id}`}
+                title={typeof column.columnDef.header === 'string' ? column.columnDef.header : id}
```
DanielSchiavini: Do we really wanna show the id? What's wrong with rendering the whole header when it's ReactNode?

apps/main/src/llamalend/features/market-list/filters/utils.ts
DanielSchiavini: Jsdoc **is** the spec on documentation formatting 😆

DanielSchiavini (dismissed):

0xtutti (dismissed):

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersCollapsible.tsx
Pearce542: I wanted to render the header from the column, but didn't manage to get all the header. Apparently using `table.getHeaderGroups()` or `table.getFlatHeaders()` only returns visible table headers and not all of the, (we can have filters for hidden column as well). So decided to just get the titles from `LLAMA_MARKET_TITLES`

0xAlunara (approved):

DanielSchiavini (approved):

PR #2529, opened 2026-05-18
feat: curve solver quotes

DanielSchiavini:
- add routes from the curve-solver to llamalend leveraging
- depends on #2597


0xPearce (approved):
Nice!

apps/router-api/src/curve-solver-router/curve-solver-router.ts
```diff
@@ -0,0 +1,107 @@
...
+): Promise<RouterRouteResponse[]> => {
+  const {
+    chainId,
+    tokenIn: [tokenIn],
+    tokenOut: [tokenOut],
+    blacklist = [],
+    amountIn: [amountIn] = [],
+    amountOut: [amountOut] = [],
+    userAddress,
+  } = query // todo: use slippage
```
0xPearce, line 53: we should add slippage though for the `min_out`

DanielSchiavini: that's done on the tx level, not sure we can do it here because we don't have a quote yet?

0xtutti (approved):

PR #2531, opened 2026-05-18
feat: share chart price formatting across charts

OnlyJousting:
Changes:
- Added a shared `formatChartAxisNumber` utility for compact chart label formatting

### DIFF
--- a/packages/curve-ui-kit/src/shared/ui/Chart/chart.utils.ts
+++ b/packages/curve-ui-kit/src/shared/ui/Chart/chart.utils.ts
@@ -1,8 +1,12 @@
 import { meanBy } from 'lodash'
 import { movingAverage } from '@primitives/array.utils'
 import { TIME_FRAMES } from '@ui-kit/lib/model/time'
+import { formatNumber, type NumberFormatOptions } from '@ui-kit/utils/number'
 
 export type ChartLineDashPattern = number[]
+
+export const DEFAULT_CHART_SIGNIFICANT_DIGITS = 5
+const DEFAULT_CHART_ABBREVIATE_FROM = 10000
 
 export const CHART_LINE_DASH_PATTERNS = {
   /** Matches lightweight-charts LineStyle.Dashed with the default 1px price line width. */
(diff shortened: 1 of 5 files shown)


michwill (approved):

packages/curve-ui-kit/src/shared/ui/Chart/chart.utils.ts
```diff
@@ -18,6 +22,46 @@ export const CHART_LINE_WIDTHS = {
...
+     */
+    significantDigits?: number
+    /**
+     * Axis labels only abbreviate above 10k by default so values like `4013` stay literal instead of becoming `4.013k`.
+     * Pass `false` for price ranges or tooltips that should never abbreviate.
+     */
+    abbreviateFrom?: number | false
+  } = {},
+) => {
+  const {
```
DanielSchiavini, line 50: this can be moved to the function declaration I think?

packages/curve-ui-kit/src/shared/ui/Chart/chart.utils.ts
```diff
@@ -18,6 +22,46 @@ export const CHART_LINE_WIDTHS = {
...
+ */
+export const formatChartAxisNumber = (
+  value: number,
+  options: Omit<
+    Partial<NumberFormatOptions>,
+    'abbreviate' | 'formatter' | 'maximumSignificantDigits' | 'minimumSignificantDigits'
+  > & {
+    /**
+     * Chart labels need compact, comparable values. Significant digits keep labels like `4013` readable while preserving
+     * small price moves better than a fixed decimal count.
```
DanielSchiavini, line 40: a fixed decimal count should only be used in very specific circumstances imo

packages/curve-ui-kit/src/shared/ui/Chart/chart.utils.ts
```diff
@@ -18,6 +22,46 @@ export const CHART_LINE_WIDTHS = {
...
+     */
+    significantDigits?: number
+    /**
+     * Axis labels only abbreviate above 10k by default so values like `4013` stay literal instead of becoming `4.013k`.
```
DanielSchiavini, line 44: this is not supposed to be specific to chart numbers, that is a bug in the existing code

DanielSchiavini (approved):
Approving because it fixes a bug, but the solution needs to be more generic imo (although it may be better handled by the work @0xAlunara is doing)

packages/curve-ui-kit/src/shared/ui/Chart/chart.utils.ts
OnlyJousting: we probably never want to have fixed decimals. the case would be when the the value's digits are more then the significant digits but in this case I think we're always showing all digits/the full number anyways

packages/curve-ui-kit/src/shared/ui/Chart/chart.utils.ts
OnlyJousting: yea maybe it's better but I will just merge like this if alu's work will make this redundant anyways

PR #2532, opened 2026-05-18
fix: duplicate debounce

DanielSchiavini:
- remove `useDebounce` from `TableFilters` since the `SearchField` is already debounced
- now each search flow has a single debounce boundary and a single source of truth
  - this avoids stale debounced values being echoed back into the input, which could drop characters under fast typing and make search tests flaky
  - example: `AssertionError: Timed out retrying after 5000ms: expected 'http://localhost:3000/llamalend/ethereum/markets/?search=wtETH+crvUSD' to include 'search=wstETH+crvUSD' `
- fix the search input so it always filters out leading/trailing spaces from URL


packages/curve-ui-kit/src/shared/ui/SearchField.tsx
```diff
@@ -2,52 +2,44 @@ import { RefObject, useCallback, useRef } from 'react'
...
 type SearchFieldProps = TextFieldProps & {
-  value?: string
+  value: string
   onSearch: (search: string) => void
   inputRef?: RefObject<HTMLInputElement | null>
   disableAutoFocus?: boolean
 }
 
-/** Compares two strings, ignoring leading and trailing whitespace. */
-const searchFieldEquals = (a: string, b: string) => a.trim() === b.trim()
```
Pearce542, line 17: Did you forget to implement this back or decided to remove it? I think it's useful

0xtutti (dismissed):

0xtutti (dismissed):

0xtutti (dismissed):

packages/curve-ui-kit/src/features/select-token/ui/modal/TokenList.tsx
```diff
@@ -52,6 +52,13 @@ export const TokenList = ({
   onSearch,
 }: TokenListProps) => {
   const [search, setSearch] = useState('')
+  const onSearchCallback = useCallback(
```
0xAlunara, line 55: Would've just called this `onSearch`.

packages/curve-ui-kit/src/hooks/useDebounce.ts
```diff
@@ -119,18 +119,21 @@ export function useFormDebounce<T>(givenValue: T, options?: DebouncedValueOption
  * @param callback - Function called when the debounced value changes
  * @param debounceMs - The debounce period in milliseconds (default: 166ms)
  * @param equals - Optional custom equality function to compare values
+ * @param clean - Optional custom cleaning function to transform values before comparison (default: identity)
```
0xAlunara: I think the usual term is `sanitize` 😄

0xAlunara: I'm btw not really sold on coupling this inside `useUniqueDebounce`. The only user of this parameter is `SearchField`, why not let that parent just do the sanitation? One less callback dependency and you don't need `idenitty` from lodash either.

0xAlunara (dismissed):
Comments are not severe, feel free to merge if you prefer it this way.

packages/curve-ui-kit/src/features/select-token/ui/modal/TokenList.tsx
DanielSchiavini: me too but that is the name of the prop

packages/curve-ui-kit/src/hooks/useDebounce.ts
DanielSchiavini: useUniqueDebounce is the place that handles the semantics of uniqueness. If we don't trim the string before comparing them, the uniqueness checks will not work. However, I still found issues when receiving callback echos. See https://github.com/curvefi/curve-frontend/pull/2532/commits/188913b6e3531a4f14da468ce85c548a09ade391

Pearce542 (approved):

0xAlunara (approved):

PR #2533, opened 2026-05-18
refactor: chart states, add empty state

OnlyJousting:
Changes:
- Added a reusable ChartEmpty state component for charts.
- Extended ChartStateWrapper to handle loading, error, empty, and content states.
- Updated the LlamaLend bands chart to show an empty state when no active bands exist, instead of treating empty data as loading.
- Preserved pending bands data as `undefined` through the chart flow so the empty state only appears after the market bands query resolves to an array.
- Removed the bands chart’s local empty-state component in favor of the shared chart empty state.
- Aligned error and no-position empty-state typography, spacing, and max-width styling with shared UI tokens.


packages/curve-ui-kit/src/shared/ui/Chart/ChartEmpty.tsx
```diff
@@ -0,0 +1,48 @@
...
+    sx={{
+      display: 'flex',
+      alignItems: 'center',
+      justifyContent: 'center',
+      width: '100%',
+      minHeight: height,
+    }}
+  >
+    <Stack flexDirection="column" alignItems="center" gap={Spacing.md} padding={Spacing.md}>
+      <LlamaIcon sx={{ width: IconSize.xxl, height: IconSize.xxl }} />
```
DanielSchiavini, line 28: Is this icon llamalend-specific?

apps/main/src/llamalend/features/bands-chart/BandsChart.tsx
```diff
@@ -73,7 +73,9 @@ const BandsChartComponent = ({
     >
       <ChartStateWrapper
         height={height}
-        isLoading={isLoading || !chartData?.length}
+        isLoading={isLoading}
+        isEmpty={chartData.length === 0}
```
DanielSchiavini: YES length==0 is not loading 🙏

DanielSchiavini (dismissed):

apps/main/src/llamalend/features/market-position-details/NoPosition.tsx
```diff
@@ -26,12 +24,12 @@ export const NoPosition = ({ type }: NoPositionProps) => (
...
     <LlamaIcon sx={{ width: IconSize.xxl, height: IconSize.xxl }} />
     <Stack alignItems="center">
-      <Typography variant="headingXsBold" sx={{ maxWidth: TYPOGRAPHY_MAX_WIDTH, textAlign: 'center' }}>
+      <Typography variant="headingXsBold" sx={{ maxWidth: MaxWidth.emptyStateCard, textAlign: 'center' }}>
         {title[type]}
       </Typography>
       <Typography
         variant="bodySRegular"
-        sx={{ maxWidth: TYPOGRAPHY_MAX_WIDTH, textAlign: 'center', color: t => t.design.Text.TextColors.Secondary }}
+        sx={{ maxWidth: MaxWidth.emptyStateCard, textAlign: 'center', color: t => t.design.Text.TextColors.Secondary }}
```
0xAlunara, line 32: Not that it matters much but I wonder if we could move the max width the the parent `<Stack>`

EDIT: similarly for `ErrorMessage.tsx`?
EDIT2: and `ChartEmpty.tsx`
EDIT3: actually they look very similar, perhaps it's also worthwhile to export a little subcomponent from `ErrorMessage.tsx` which can be re-used here?

packages/curve-ui-kit/src/shared/ui/Chart/ChartEmpty.tsx
0xAlunara: The intent is to indicate a 'user position', which means we'll also use it in the DEX in the future for LP positions and such.
I guess it also serves a double purpose and cute mascot though? 😄

0xAlunara (dismissed):
Technically it works, but sometimes when I open a market something happens in the following order:

1. Open a market like https://curve-dapp-git-refactor-chart-states-curvefi.vercel.app/lend/arbitrum/markets/0x83B85f3b08B5EE58dE9EF9604e7Eec087FCCf130
2. It shows the `ChartEmpty` component, telling me there's no active bands.
3. It then starts loading and the band chart shows.

I guess it's just weird that there's a brief second where it shows there's no bands before it actually starts loading.

packages/curve-ui-kit/src/shared/ui/Chart/ChartEmpty.tsx
OnlyJousting: it's an image of a llama!

apps/main/src/llamalend/features/market-position-details/NoPosition.tsx
OnlyJousting: it might work fine but I don't think it's correct. it's the width of the typography that is meant to be managed. the container is a separate issue and might present with different levels of whitespace

OnlyJousting:
> Technically it works, but sometimes when I open a market something happens in the following order:
> 
> 1. Open a market like https://curve-dapp-git-refactor-chart-states-curvefi.vercel.app/lend/arbitrum/markets/0x83B85f3b08B5EE58dE9EF9604e7Eec087FCCf130
> 2. It shows the `ChartEmpty` component, telling me there's no active bands.
> 3. It then starts loading and the band chart shows.
> 
> I guess it's just weird that there's a brief second where it shows there's no bands before it actually starts loading.

Dang, this is what data.length === 0 was managing. I tried to check if removing it was causing any issues but I didn't come across these premature empty states on my machine. need to add a check that the bands queries have completed

0xAlunara (dismissed):

apps/main/src/llamalend/features/bands-chart/hooks/useBandsData.ts
```diff
@@ -47,11 +47,15 @@ export const useBandsData = ({
     userBandsBalances,
   })
 
+  // Guards against prematurely showing a ChartStateWrapper empty state
```
DanielSchiavini: better fix the problem at the source:
https://github.com/curvefi/curve-frontend/blob/7fc1c3e62d2db46cbd2f80fa9daaa9bf92b5ce98/apps/main/src/llamalend/features/bands-chart/hooks/useProcessedBandsData.ts#L12
That is why your `chartData` is 0 length

0xtutti (dismissed):

0xAlunara (approved):

DanielSchiavini (approved):

PR #2538, opened 2026-05-19
refactor: move some metrics to chart cards

OnlyJousting:
Changes:
- Populates market page charts with relevant metrics
- Updates column configuration for user position metrics, advanced details metrics
- Removes irrelevant mint market utilization metric

Note: design has a "total backing" metric but it's still not available from an API endpoint so it can be added later


0xtutti (dismissed):

apps/main/src/llamalend/queries/crvusd-total-supply.query.ts
```diff
@@ -0,0 +1,16 @@
...
+import { decimal } from '@ui-kit/utils'
+
+/**
+ * Gets the full crvUSD supply from the Curve Finance API, including non-standard mint sources.
+ */
+export const { useQuery: useCrvUsdTotalSupply } = queryFactory({
+  queryKey: () => ['getCrvusdTotalSupplyNumber'] as const,
+  queryFn: async () => {
+    const resp = await fetch('https://api.curve.finance/api/getCrvusdTotalSupplyNumber')
+    return decimal(await resp.text()) ?? null
```
DanielSchiavini: Do we really expect invalid data?
```suggestion
    return await resp.text() as Decimal
```
```suggestion
    return decimal(await resp.text())!
```

apps/main/src/llamalend/features/market-advanced-information/AdvancedDetails.tsx
```diff
@@ -58,26 +59,30 @@ export const AdvancedDetails = ({ chainId, marketId, market, marketType }: Advan
       marketType,
     })
   const { utilization, utilizationBreakdown } = getUtilizationMetrics(availableLiquidity)
+  const hasRateCurveMetrics = useMarketInterestRatesAndUtilizationChart()
+  const showRateCurveMetrics = marketType === LlamaMarketType.Lend && hasRateCurveMetrics
```
DanielSchiavini: `marketType` checks should ideally not be part of components, if we have some other property we can check that'd be clearer imo

apps/main/src/llamalend/widgets/CrvUsdPriceChart.tsx
```diff
@@ -71,6 +81,22 @@ export const CrvUsdPriceChart = () => {
     )
   }, [priceHistory])
 
+  const oneWeekDeviation = useMemo(() => {
+    const now = Date.now()
+    const points = priceHistory.map(item => ({
+      timestamp: new Date(item.timestamp).getTime(),
+      price: Number(item.price),
+    }))
```
DanielSchiavini: `price` is already the result of `z.number()` and `timestamp` is the result of `getTime()`
```suggestion
    const points = priceHistory
```

apps/main/src/llamalend/widgets/CrvUsdPriceChart.tsx
```diff
@@ -71,6 +81,22 @@ export const CrvUsdPriceChart = () => {
...
+
+    return (
+      calculateAverageRates(
+        [...points, ...(currentPrice == null ? [] : [{ timestamp: now, price: currentPrice }])],
```
DanielSchiavini:
```suggestion
        notFalsyArray(points, currentPrice != null && [{ timestamp: now, price: currentPrice }])),
```

apps/main/src/llamalend/widgets/CrvUsdPriceChart.tsx
```diff
@@ -106,6 +132,33 @@ export const CrvUsdPriceChart = () => {
...
+            label={t`Current price`}
+            value={currentPrice}
+            loading={currentPrice == null && isCurrentPriceLoading}
+            valueOptions={{ unit: 'dollar', maximumSignificantDigits: 5 }}
+          />
+          <Metric
+            size="medium"
+            label={t`1W deviation`}
+            value={oneWeekDeviation}
+            loading={oneWeekDeviation == null && (showLoading || isCurrentPriceLoading)}
```
DanielSchiavini, line 148: `oneWeekDeviation == null` should be irrelevant for the loading state. `isCurrentPriceLoading` here doesn't make much sense since we are only adding an extra item to the array when the `currentPrice` is available. Otherwise, why do we have `oneWeekDeviation` without `currentPrice`?
```suggestion
            loading={showLoading}
```

apps/main/src/llamalend/widgets/MarketHistoricalRatesChart.tsx
```diff
@@ -101,6 +111,24 @@ export const MarketHistoricalRatesChart = ({
     )
   }, [snapshots, rateMode, marketRates])
 
+  const oneWeekAverageRate = useMemo(() => {
+    const now = Date.now()
+    const snapshotPoints = snapshots.map(snapshot => {
+      const timestamp = new Date(snapshot.timestamp).getTime()
```
DanielSchiavini:
```suggestion
      const { timestamp } = snapshot
```

apps/main/src/llamalend/widgets/MarketHistoricalRatesChart.tsx
```diff
@@ -101,6 +111,24 @@ export const MarketHistoricalRatesChart = ({
...
+      const rate = rateMode === 'borrow' ? snapshot.borrowApr : 'lendApy' in snapshot ? snapshot.lendApy * 100 : null
```
DanielSchiavini: why would we have `lendApy` missing in snapshot with `rateMode!='borrow'`?

apps/main/src/llamalend/widgets/MarketHistoricalRatesChart.tsx
```diff
@@ -131,14 +159,34 @@ export const MarketHistoricalRatesChart = ({
...
       />
       <CardContent component={Stack} gap={Spacing.md}>
+        <Stack
+          display="grid"
+          gap={Spacing.xl}
+          gridTemplateColumns={{ mobile: 'repeat(2, 1fr)', tablet: 'repeat(4, 1fr)' }}
+        >
+          <Metric
+            size="medium"
+            label={rateMode === 'borrow' ? t`Current APR` : t`Current APY`}
```
DanielSchiavini: this will break if we ever add more `rateMode` options (same below)

apps/main/src/llamalend/widgets/CrvUsdPriceChart.tsx
DanielSchiavini: the same is valid for most of the other loading states tbh

apps/main/src/llamalend/queries/crvusd-total-supply.query.ts
0xAlunara: Fwiw there's a ticket to remove this number from the `crvUSD` app's header, and the only other usage can be replaced with a prices api call. Which means this query will be deleted.

apps/main/src/llamalend/features/market-advanced-information/AdvancedDetails.tsx
```diff
@@ -58,26 +59,30 @@ export const AdvancedDetails = ({ chainId, marketId, market, marketType }: Advan
       marketType,
     })
   const { utilization, utilizationBreakdown } = getUtilizationMetrics(availableLiquidity)
+  const hasRateCurveMetrics = useMarketInterestRatesAndUtilizationChart()
```
0xAlunara: This feature flag has been in the stable channel for a while now, maybe it's time to remove it altogether?

apps/main/src/llamalend/features/market-advanced-information/AdvancedDetails.tsx
```diff
@@ -58,26 +59,30 @@ export const AdvancedDetails = ({ chainId, marketId, market, marketType }: Advan
...
-        loading={availableLiquidity?.loading}
-        valueOptions={{ unit: 'percentage' }}
-        notional={utilization == null ? undefined : utilizationBreakdown}
-        valueTooltip={{
-          title: t`Utilization ${MarketTypeSuffix[marketType]}`,
-          body: <UtilizationTooltip marketType={marketType} />,
-          ...TooltipOptions,
-        }}
-      />
+      {!showRateCurveMetrics && marketType === LlamaMarketType.Lend && (
```
0xAlunara: This `marketType ===` is already part of `showRateCurveMetrics`

apps/main/src/llamalend/features/market-advanced-information/AdvancedDetails.tsx
0xAlunara: I'm not sure why mint markets can't have their total collateral shown in the first place?

apps/main/src/llamalend/widgets/CrvUsdPriceChart.tsx
0xAlunara: I'm not sure what this code even does in the first place. `now` isn't even used anywhere, and `item.price` is already a number (in addition to what you said about the timestamp).

This whole `oneWeekDeviation` could be removed as it's essentially a noop?

apps/main/src/llamalend/widgets/MarketRateCurveChart.tsx
```diff
@@ -63,13 +68,52 @@ export const MarketRateCurveChart = ({
...
+      : `${formatNumber(totalBorrowed, { abbreviate: true })}/${formatNumber(capAndAvailable.totalAssets, {
+          abbreviate: true,
+        })} ${borrowToken?.symbol ?? ''}`
+
+  const collateralTotal = totalCollateral == null ? null : Number(totalCollateral.collateral)
+  const borrowedCollateralTotal = totalCollateral == null ? null : Number(totalCollateral.borrowed)
+  const collateralUsdValue =
+    collateralTotal == null || collateralUsdRate == null ? null : collateralTotal * collateralUsdRate
+  const borrowedCollateralUsdValue =
+    borrowedCollateralTotal == null || borrowedUsdRate == null ? null : borrowedCollateralTotal * borrowedUsdRate
```
0xAlunara, line 110: All those null checks are such an eye-sore 😵
Don't have a solution though, because `0` is a valid number 🤔

DanielSchiavini: We considered some map function but never got to implement it. I could do a refactor to include it. How would you name it? 🤔

0xAlunara:
```typescript
const maybe = <T, U>(x: T | null | undefined, f: (x: T) => U): U | undefined =>
  x == null ? undefined : f(x);
```
Because it kinda resembles mapping over an option? 

Other GPT suggestions are:
```
mapOptional
mapDefined
mapNonNull
mapMaybe
mapNil
nilMap
optMap
```

DanielSchiavini: mapMaybe and mapNotNull are nice. But maybe is shorter :nerd_face: 
I'm not sure if we should go for `null` or `undefined` as the default value

apps/main/src/llamalend/features/market-advanced-information/AdvancedDetails.tsx
OnlyJousting: what's the problem with checking marketType? it's the exact differentiator that determines what to show, seems like the most clear to me

OnlyJousting: > I'm not sure why mint markets can't have their total collateral shown in the first place?

Total collateral is shown in "advanced details" on mint markets. I guess it was moved to the IRM chart for lend markets because it's a part of the utilization calculation

apps/main/src/llamalend/features/market-advanced-information/AdvancedDetails.tsx
OnlyJousting: alright I will remove it!

apps/main/src/llamalend/queries/crvusd-total-supply.query.ts
OnlyJousting: I just moved it so it could be reused

apps/main/src/llamalend/widgets/CrvUsdPriceChart.tsx
OnlyJousting: now is used to pair with the "current" data point which is the latest we fetch on chain

apps/main/src/llamalend/widgets/CrvUsdPriceChart.tsx
OnlyJousting: there's been a problem with the metric component firing "N/A" for undefined values, hence all these extra == null checks. including isCurrentPriceLoading makes sense to me, otherwise we might have a value that is displaying an incorrect value for a moment before updating

apps/main/src/llamalend/widgets/MarketHistoricalRatesChart.tsx
OnlyJousting: ah yea, they are no longer `Date` 😊

apps/main/src/llamalend/widgets/MarketHistoricalRatesChart.tsx
OnlyJousting: checking rateMode wasn't enough so I added cast, there was also a fallback to 0 which was incorrect

0xAlunara (dismissed):

apps/main/src/llamalend/widgets/MarketRateCurveChart.tsx
DanielSchiavini: I created a task

apps/main/src/llamalend/widgets/CrvUsdPriceChart.tsx
DanielSchiavini: The correct approach would be to have `oneWeekDeviation` be a `Query` with it's own loading state, derived from `isCurrentPriceLoading || isLoading`.
If you don't want to show the `oneWeekDeviation` while the current price is loading, why does `oneWeekDeviation` have a value at all during that time?

apps/main/src/llamalend/features/market-advanced-information/AdvancedDetails.tsx
```diff
@@ -57,27 +31,14 @@ export const AdvancedDetails = ({ chainId, marketId, market, marketType }: Advan
       marketId,
       marketType,
     })
-  const { utilization, utilizationBreakdown } = getUtilizationMetrics(availableLiquidity)
+  const showRateCurveMetrics = marketType === LlamaMarketType.Lend
```
DanielSchiavini: can you add a comment that explains why lend markets?

apps/main/src/llamalend/features/market-advanced-information/AdvancedDetails.tsx
```diff
@@ -101,29 +62,31 @@ export const AdvancedDetails = ({ chainId, marketId, market, marketType }: Advan
...
-                { value: decimal(collateral?.totalBorrowed), symbol: collateral?.borrowedSymbol ?? undefined },
-              )
-        }
-        valueTooltip={{
-          title: t`Total Collateral`,
-          body: <TotalCollateralTooltip {...collateral} />,
-          ...TooltipOptions,
-        }}
-      />
+      {!showRateCurveMetrics && (
```
DanielSchiavini: Is this confusing or incorrect? We're showing something when `!show`

apps/main/src/llamalend/widgets/MarketHistoricalRatesChart.tsx
```diff
@@ -131,14 +159,34 @@ export const MarketHistoricalRatesChart = ({
...
+            value={currentLiveRate}
+            loading={currentLiveRate == null && (isMarketRatesLoading || !market)}
+            valueOptions={{ unit: 'percentage' }}
+          />
+          <Metric
+            size="medium"
+            label={rateMode === 'borrow' ? t`1W Avg APR` : t`1W Avg APY`}
```
DanielSchiavini:
```suggestion
            label={{ borrow: t`1W Avg APR`, supply: t`1W Avg APY`}[rateMode]}
```

apps/main/src/llamalend/widgets/MarketHistoricalRatesChart.tsx
OnlyJousting: this escalated into a better soltution from the comments about rateMode ternaries

apps/main/src/llamalend/features/market-advanced-information/AdvancedDetails.tsx
OnlyJousting: sure, mint markets doesn't have rate curves

apps/main/src/llamalend/features/market-advanced-information/AdvancedDetails.tsx
OnlyJousting: yea it's a bad name after removing the feature flag

apps/main/src/llamalend/widgets/MarketHistoricalRatesChart.tsx
OnlyJousting: outdated 😎

apps/main/src/llamalend/widgets/CrvUsdPriceChart.tsx
OnlyJousting: functionally it doesn't since it will be in a loading state, a query sounds reasonable but also a bit overkill since this is not getting reused anywhere?

0xAlunara (approved):

DanielSchiavini (approved):

PR #2542, opened 2026-05-19
feat: stop waiting for user data queries

DanielSchiavini:
- the user queries are sometimes taking much longer than the market queries
- the user gets stuck waiting for the data while most of it is already available
- ultimately we should improve the load time on the prices API, but for now let's at least show the rest of the data

### DIFF
--- a/packages/curve-ui-kit/src/hooks/useStateTimeout.ts
+++ b/packages/curve-ui-kit/src/hooks/useStateTimeout.ts
@@ -1,0 +1,30 @@
+import { useCallback, useRef, useState } from 'react'
+import { Duration } from '@ui-kit/themes/design/0_primitives'
+
+/**
+ * Timeout hook that returns a boolean indicating if the data has taken too long to load.
+ * @returns [isTimedOut, setIsReady]
+ * - isTimedOut is true if the data has been loaded for longer than Duration.LoadingTimeout
+ * - setIsReady is a function that takes a boolean indicating if the data is ready.
+ *   If the data is ready, the timeout is cleared and isTimedOut is set to false.
+ *   If the data is not ready and there is no timeout yet, a timer is started.
+ */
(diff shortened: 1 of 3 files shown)


0xtutti (approved):

packages/curve-ui-kit/src/hooks/useStateTimeout.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Timeout hook that returns a boolean indicating if the data has taken too long to load.
+ * @returns [isTimedOut, setIsReady]
+ * - isTimedOut is true if the data has been loaded for longer than Duration.LoadingTimeout
+ * - setIsReady is a function that takes a boolean indicating if the data is ready.
+ *   If the data is ready, the timeout is cleared and isTimedOut is set to false.
+ *   If the data is not ready and there is no timeout yet, a timer is started.
+ */
+export function useStateTimeout() {
```
0xAlunara, line 12: Not a fan of this custom timeout behavior in combination with queries, given TanStack has its own [retry functionality](https://tanstack.com/query/latest/docs/framework/react/guides/query-retries). In fact, you'll find that inside `factory.ts` we're already using that. Especially `setTimeout` with `useRef` gives me bad feelings.

The challenge here is to somehow signal to the consumer (`useQueries`) that the first of N tries for a specific query has failed.

Not blocking this PR as it's a valuable gain nonetheless, but my gut feeling tells me there's a better solution available. But perhaps later.

0xAlunara (approved):

packages/curve-ui-kit/src/hooks/useStateTimeout.ts
DanielSchiavini: I agree with you definitely something we can look into

PR #2548, opened 2026-05-19
feat: 0x router integration

DanielSchiavini:
- integrate 0x router into the router API
- disable price impact check for 0x, make it non-blocking for enso and odos
- depends on #2597


apps/router-api/src/zeroex-router/zeroex-router.ts
```diff
@@ -0,0 +1,81 @@
...
+  query: RoutesQuery,
+  log: FastifyBaseLogger,
+): Promise<RouterRouteResponse[]> => {
+  const {
+    chainId,
+    tokenIn: [sellToken],
+    tokenOut: [buyToken],
+    amountIn: [amountIn] = [],
+    userAddress: taker,
+  } = query
```
Pearce542, line 46: No slippage?

DanielSchiavini: I don't think they support that in the API we'll have to implement it on our side on another PR

0xAlunara (approved):

0xtutti (approved):

PR #2555, opened 2026-05-20
fix: use debounce reset to initial

DanielSchiavini:
- #2532 introduced a slight bug that broke the RPC tests.\
  - When the forms reset back to undefined, the check `initial!=default` will not trigger
  - The internal value in `useDebounce` would stay stale and the field would show the old value.
- This was impossible to solve without inlining the call to `useDebounced`, but it was trivial to fix once I did it

### DIFF
--- a/packages/curve-ui-kit/src/hooks/useDebounce.ts
+++ b/packages/curve-ui-kit/src/hooks/useDebounce.ts
@@ -135,7 +135,7 @@ export function useUniqueDebounce<T>({
   equals?: (a: T, b: T) => boolean
   sanitize?: (value: T) => T
 }) {
-  const [initialValue, setInitialValue] = useState<T>(defaultValue)
+  const [value, setValue] = useState<T>(defaultValue)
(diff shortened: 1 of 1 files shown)


packages/curve-ui-kit/src/hooks/useDebounce.ts
```diff
@@ -158,5 +158,5 @@ export function useUniqueDebounce<T>({
     [onChange, sanitize, equals],
   )
 
-  return useDebounce({ initialValue, debounceMs, callback })
+  return [value, ...useDebounced(callback, debounceMs, setValue)] as const
```
DanielSchiavini, line 161: same as the original call
https://github.com/curvefi/curve-frontend/blob/d6fc2e6ce4b946380b50a3ac86526170cb241246/packages/curve-ui-kit/src/hooks/useDebounce.ts#L84

OnlyJousting (approved):

0xAlunara (approved):

PR #2556, opened 2026-05-20
refactor: backend maintenance modal

Pearce542:
- generalized maintenance hook
- small review comments

### DIFF
--- a/packages/curve-ui-kit/src/hooks/useLocalStorage.ts
+++ b/packages/curve-ui-kit/src/hooks/useLocalStorage.ts
@@ -16,19 +16,6 @@ function getFromLocalStorage<T>(storageKey: string): T | null {
   return item && JSON.parse(item)
 }
 
-function getLocalStorageKeyStartingWith(prefix: string) {
-  if (typeof window === 'undefined') {
-    return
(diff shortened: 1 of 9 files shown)


packages/curve-ui-kit/src/hooks/useLocalStorage.ts
```diff
@@ -16,6 +16,19 @@ function getFromLocalStorage<T>(storageKey: string): T | null {
   return item && JSON.parse(item)
 }
 
+function getLocalStorageKeyStartingWith(prefix: string) {
```
DanielSchiavini: I'd just keep the old keys tbh

DanielSchiavini (dismissed):

packages/curve-ui-kit/src/hooks/useLocalStorage.ts
Pearce542: I kept the old keys and created a task to change the keys name when we do not have a maintenance planned, to  avoid conflicts and migrations

0xtutti (approved):

DanielSchiavini (approved):

PR #2478, opened 2026-05-06
feat: new small liquidation range chart

OnlyJousting:
Adds a compact liquidation range chart to LlamaLend loan action details so users can compare current and updated ranges alongside the oracle price directly in the action flow.

Changes:
- Adds the new SmallLiquidationRangeChart ECharts widget.
- Shows current and new liquidation ranges in loan action info lists.
- Adds oracle price markers, including a truncated rail for distant oracle prices.
- Adds Storybook examples for the new chart states.
- Updates notFalsyArray to support readonly arrays.

PR #2557, opened 2026-05-20
refactor: address remaining review comments

OnlyJousting:
This PR addresses review comments that weren't tackled before merging #2478


apps/main/src/llamalend/widgets/small-liquidation-range-chart/small-liquidation-range-chart.options.ts
```diff
@@ -51,7 +51,7 @@ const buildHiddenYAxis = (gridIndex?: number): YAXisComponentOption => ({
   min: 0,
   max: 1,
   show: false,
-  ...(gridIndex === undefined ? {} : { gridIndex }),
+  ...(gridIndex != null && { gridIndex }),
```
0xAlunara, line 54: `...(gridIndex && { gridIndex }),`

apps/main/src/llamalend/widgets/small-liquidation-range-chart/small-liquidation-range-chart.options.ts
```diff
@@ -352,27 +352,21 @@ export const buildSplitOption = ({
       }),
     ],
     yAxis: [buildHiddenYAxis(0), buildHiddenYAxis(1)],
-    series: [
+    series: notFalsyArray<SeriesOption>(
```
0xAlunara: wouldn't `notFalsy` be enough and then we don't need to wrap both arguments in an array of size 1?

Pearce542: all 3 `notFalsyArray` of this file could be rewritten using `notFalsy` and without wrapping with an array

DanielSchiavini (dismissed):

apps/main/src/llamalend/widgets/small-liquidation-range-chart/small-liquidation-range-chart.options.ts
DanielSchiavini: Good one, I missed that!

0xtutti (dismissed):

apps/main/src/llamalend/widgets/small-liquidation-range-chart/small-liquidation-range-chart.options.ts
0xAlunara: actually I think `...gridIndex` would work even if it's undefined?

OnlyJousting: gridIndex 0 is valid so the first suggestion won't work as it would omit 0, the second one apparently doesn't work because spreading a number becomes {}

DanielSchiavini (approved):

0xtutti (approved):

PR #2530, opened 2026-05-18
feat: new market list filters for mobile

Pearce542:
Next: #2558 

- added loading state to chain chips component
- enabled new market list filter for mobile
- refactored the market list filters into a ~~modal~~ drawer for mobile instead of popover
- refactored the llamalend table filters logic into a single component instead of `FilterChip` and `LlamaTableFiltersOverlay`

Tests are failing becasue they have been implemented in anther PR https://github.com/curvefi/curve-frontend/pull/2558

Screenshot 2026-05-26 at 10 16 58

PR #2558, opened 2026-05-21
feat: new market list filters tests

Pearce542:
- depends on https://github.com/curvefi/curve-frontend/pull/2530
- added back [previously deleted](https://github.com/curvefi/curve-frontend/pull/2407#discussion_r3160248481) llamalend market list tests 
- added tests for new market list filters
- refactored llamalend markets test helpers into separeate file
- reordered and grouped llamalend market tests


packages/primitives/src/array.utils.ts
```diff
@@ -46,3 +46,13 @@ export function movingAverage(values: number[], timestamps: number[], windowMs:
     return windowSum / (i - windowStart + 1)
   })
 }
+
+/**
+ * Returns the median value from an array of numbers or `undefined` when the array is empty.
+ * For even-length arrays, this returns the lower-middle value.
```
DanielSchiavini, line 52:
```suggestion
 * For even-length arrays, this returns the lower-middle value. For empty arrays, returns undefined.
```

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
```diff
@@ -144,82 +198,14 @@ testCases.forEach(([width, height, breakpoint]) => {
...
-    // todo: filter by chain
 
-    it(`should allow filtering by token`, () => {
-      openFilters(breakpoint)
-      checkCoinSelection('collateral')
-      checkCoinSelection('borrowed')
-    })
-
-    it('should allow filtering favorites', { scrollBehavior: false }, () => {
+    itNotMobile('should allow filtering favorites', { scrollBehavior: false }, () => {
```
DanielSchiavini: Is this feature not available on mobile? A comment would be nice.
We should add it though, it's weird to mark markets as favorites and then have no way of finding them

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
```diff
@@ -32,34 +43,43 @@ const testCases = [oneViewport()] as const
...
-        return cy.get(`[data-testid^="data-table-row"]`).should('have.length.above', 1)
-      })
+    it(`should copy the market address`, RETRY_IN_CI, () => {
+      expandFirstRowOnMobile(breakpoint)
+      // unfortunately we need to click twice on Chromium, the first one doesn't work (maybe due to the tooltip)
+      range(2).forEach(() =>
+        breakpoint === 'desktop'
+          ? // on desktop, the copy button is not visible until hovered - but cypress doesn't support that so use force
+            cy.get(`[data-testid^="copy-market-address"]`).first().click({ force: true })
+          : cy.get(`[data-testid^="copy-market-address"]:visible`).first().click(),
```
DanielSchiavini:
```suggestion
        // on desktop, the copy button is not visible until hovered - but cypress doesn't support that so use force
        cy.get(`[data-testid^="copy-market-address"]`).first().click({ force: breakpoint === 'desktop' })
```

DanielSchiavini: the `:visible` is implied in cypress

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
```diff
@@ -32,34 +43,43 @@ const testCases = [oneViewport()] as const
...
+
+    it(`should navigate to the market details`, () => {
+      const [type, urlRegex] = oneOf(
+        [LlamaMarketType.Mint, /\/crvusd\/\w+\/markets\/[\w-]+\/?$/],
+        [LlamaMarketType.Lend, /\/lend\/\w+\/markets\/.+\/?$/],
+      )
+      filterByMarketType([width, height], type)
+      cy.get(`[data-testid^="market-link-"]`).first().click()
+      if (breakpoint === 'mobile') {
+        cy.get(`[data-testid^="llama-market-go-to-market"]:visible`).click()
```
DanielSchiavini: `:visible` is not needed
```suggestion
        cy.get(`[data-testid^="llama-market-go-to-market"]`).click()
```

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
```diff
@@ -71,16 +91,50 @@ testCases.forEach(([width, height, breakpoint]) => {
       cy.get('[data-testid^="data-table-row"]').eq(10).invoke('outerHeight').should('equal', 65)
     })
 
+    it('should hide columns', () => {
+      if (breakpoint == 'mobile') {
+        // mobile viewports do not have this feature
+        const [width, height] = oneDesktopViewport()
+        cy.viewport(width, height)
+      }
+      const { toggle, element } = oneOf(
```
DanielSchiavini, line 100: I think this is just moved? I don't think these are the only columns that can be hidden, these days.

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
```diff
@@ -71,16 +91,50 @@ testCases.forEach(([width, height, breakpoint]) => {
...
+      cy.get(`[data-testid="${element}"]`).should('not.exist')
+    })
+
+    it('should display Borrow APR by default', () => {
+      const borrowColumnId = LlamaMarketColumnId.BorrowRate
+
+      if (breakpoint === 'mobile') {
+        // On mobile, expand the first row and check the metric is visible in the expanded panel
+        expandFirstRowOnMobile(breakpoint)
+        cy.get('[data-testid="data-table-expansion-row"]').contains('Borrow APR').should('be.visible')
```
DanielSchiavini, line 118: no test ID?

DanielSchiavini: tbh it doesn't look like this test needs a branch on breakpoint, we can call `expandFirstRowOnMobile` on desktop and it'll be ignored

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
```diff
@@ -233,133 +219,114 @@ testCases.forEach(([width, height, breakpoint]) => {
...
-        [LlamaMarketType.Mint, /\/crvusd\/\w+\/markets\/[\w-]+\/?$/],
-        [LlamaMarketType.Lend, /\/lend\/\w+\/markets\/.+\/?$/],
-      )
-      filterByMarketType([width, height], type)
-      cy.get(`[data-testid^="market-link-"]`).first().click()
-      if (breakpoint === 'mobile') {
-        cy.get(`[data-testid^="llama-market-go-to-market"]:visible`).click()
-      }
-      cy.url(LOAD_TIMEOUT).should('match', urlRegex)
+    it(`should allow filtering by token`, () => {
```
DanielSchiavini, line 235: the test body doesn't seem to be filtering anything

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
```diff
@@ -233,133 +219,114 @@ testCases.forEach(([width, height, breakpoint]) => {
...
-        breakpoint === 'desktop'
-          ? // on desktop, the copy button is not visible until hovered - but cypress doesn't support that so use force
-            cy.get(`[data-testid^="copy-market-address"]`).first().click({ force: true })
-          : cy.get(`[data-testid^="copy-market-address"]:visible`).first().click(),
-      )
-      cy.get(`[data-testid="copy-confirmation"]`).should('be.visible')
+      // filter by multiple chains at the same time
+      const otherChain = oneOf(...chains.filter(c => c !== chain))
+      withFiltersPopover(() => cy.get(`[data-testid="chip-chain-${otherChain}"]`).click())
+      checkChainSelection(chain, otherChain)
```
DanielSchiavini, line 232: should we repeat `checkChainSelection(chain)` to test for multi-select?

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
```diff
@@ -233,133 +219,114 @@ testCases.forEach(([width, height, breakpoint]) => {
...
+        withFiltersPopover(() =>
+          cy
+            .get(`[data-testid="slider-${columnId}"]`)
+            .as('slider')
+            .then(
+              (
+                $el, // With log slider a click from the left is not enough to filter
+              ) =>
+                // Click 20px from the right edge and vertically centered
+                [($el.width() ?? 80) - 20, ($el.height() ?? 24) / 2],
```
DanielSchiavini: where is the 80 coming from? I think we better throw an error than use the wrong width (that'd be hard to debug). `assert` could be useful here

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
```diff
@@ -233,133 +219,114 @@ testCases.forEach(([width, height, breakpoint]) => {
...
-      cy.get('body').click(0, 0) // close popover
-
-      cy.get(`[data-testid="data-table-cell-assets"] [data-testid^="token-icon-${symbol}"]`).should('exist') // token might be hidden behind other tokens
-      cy.url().should('include', `assets_${type}_symbol=${encodeURIComponent(symbol)}`)
-
-      cy.get(`[data-testid="multi-select-filter-${columnId}"]`).click() // open the menu
-      cy.get(`[data-testid="multi-select-clear"]`).click() // deselect previously selected tokens
-      cy.url().should('not.include', `assets_${type}_symbol`)
-    }
+    itNotMobile('should show active filters in the collapsible', () => {
```
DanielSchiavini: argh `itNotMobile` calls work technically but smell like UX issues

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
```diff
@@ -233,133 +219,114 @@ testCases.forEach(([width, height, breakpoint]) => {
...
+      withFiltersPopover(() => {
+        cy.get(`[data-testid="table-filter-btn-market-version-${LlamaMarketVersion.v1}"]`).click()
+        cy.get(`[data-testid="table-filter-btn-market-type-${LlamaMarketType.Lend}"]`).click()
+        return cy.get(`[data-testid="chip-chain-ethereum"]`).click()
+      })
+
+      cy.get(`[data-testid="table-filters-collapsible"]`).should('be.visible').as('collapsible')
+      cy.get('@collapsible').children().first().first().children().should('have.length', 3)
```
DanielSchiavini: `be.visible` is not needed, again implied by `click`
`children().first().first().children()` is tricky to read and debug, besides being fragile to DOM changes.
Couldn't we have a testId(prefix) to filter by?
```suggestion
      cy.get(`[data-testid=^"table-filter-"]`).should('have.length', 3)
```

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
```diff
@@ -32,34 +43,43 @@ const testCases = [oneViewport()] as const
 testCases.forEach(([width, height, breakpoint]) => {
   describe(`LlamaLend Markets`, () => {
     let vaultData: Record<Chain, GetMarketsResponse>
-    const itMobileOnly = breakpoint === 'mobile' ? it : it.skip
+    const itNotMobile = breakpoint !== 'mobile' ? it : it.skip
```
DanielSchiavini:
```suggestion
    const itSkipOnMobile = breakpoint === 'mobile' ? it.skip : it
```

packages/primitives/src/array.utils.ts
Pearce542: what's wrong with the previous sentence? `Returns the median value from an array of numbers or `undefined` when the array is empty.`

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
Pearce542: Not available currently

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
Pearce542: that's already the case, `checkChainSelection ` can check for multiple chains

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
Pearce542: I just moved this when I reordered the tests

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
Pearce542: Yes it was moved

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
Pearce542: also moved

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
Pearce542: good catch!

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
Pearce542: used `assert` thanks

Pearce542:
@DanielSchiavini a lot of your comments were about old llamalend markets tests that I just reordered, do you think it is worth it to make a different task to update/refactor those tests?

packages/primitives/src/array.utils.ts
DanielSchiavini: oh sorry I was not paying enough attention

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
DanielSchiavini: Do we have a task for tutti to figure out a solution?

DanielSchiavini:
> @DanielSchiavini a lot of your comments were about old llamalend markets tests that I just reordered, do you think it is worth it to make a different task to update/refactor those tests?

That's the disadvantage of mixing refactoring and new code in the same PR :nerd_face: 
If it's simple we can fix directly, otherwise add a TODO.

Since these tests are random it would be nice to run them once with 20 jobs just to find flakyness issues (we can create a commit outside this branch with 20 parallel jobs for example)

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
Pearce542: Yes

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
```diff
@@ -32,34 +35,51 @@ const testCases = [oneViewport()] as const
...
-      expandFirstRowOnMobile(mobile)
-      cy.get(`[data-testid="rewards-icons"]`).should('be.visible')
-      withFilterChips(mobile, () => {
-        cy.get(`[data-testid="chip-rewards"]`).click()
-        return cy.get(`[data-testid^="data-table-row"]`).should('have.length.above', 1)
+    it(`should copy the market address`, () => {
+      expandFirstRowOnMobile(breakpoint)
+      cy.window().then(win => {
+        // with cypress some browsers deny clipboard permissions so we use a stub
+        cy.stub(win.navigator.clipboard, 'writeText').as('writeClipboard').resolves()
```
Pearce542, line 49: tests were silently passing but sometimes failing on mobile because we were only checking if the `copy-confirmation` notification was visible.

example: this test was passing even though the browser was denying clipboard write permissions
Image

Pearce542:
@DanielSchiavini ran [~20 successfull jobs](https://github.com/curvefi/curve-frontend/actions/runs/26405630574).  Fixed the `copy market address` tests that were a bit flaky, here are the [failed runs](https://github.com/curvefi/curve-frontend/actions/runs/26398889831)

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
DanielSchiavini: I think that was known at the moment we wrote the test, if the browser blocks the clipboard this is expected behavior tbh

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
```diff
@@ -49,16 +42,23 @@ testCases.forEach(([width, height, breakpoint]) => {
...
       expandFirstRowOnMobile(breakpoint)
-      // unfortunately we need to click twice on Chromium, the first one doesn't work (maybe due to the tooltip)
-      range(2).forEach(() =>
-        cy
-          .get(`[data-testid^="copy-market-address"]`)
-          .first()
-          // on desktop, the copy button is not visible until hovered - but cypress doesn't support that so use force
-          .click({ force: breakpoint === 'desktop' }),
+      cy.window().then(win => {
+        // with cypress some browsers deny clipboard permissions so we use a stub
```
DanielSchiavini, line 48: I don't think we need this tbh, browser denying permissions is kind of expected app behavior

DanielSchiavini (approved):

OnlyJousting (approved):

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
Pearce542: mmh I still think that's the cause of tests sometimes failing on mobile. will make a test run without it

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
Pearce542: you were right, [seems](https://github.com/curvefi/curve-frontend/actions/runs/26446903227) we don't need it

PR #2559, opened 2026-05-21
fix: incorrect TAC decimals

0xAlunara:
Not sure why but apparantly TAC no longer uses 8 but 18 decimals? All works fien again if we use the TAC config shipped with wagmi.


OnlyJousting (approved):
very strange but seems to work ok

mimaklas (approved):

mimaklas:
Looks to work now
image

PR #2563, opened 2026-05-21
feat: add backend maintenance link

Pearce542:
- add the backend maintenance [X announcement](https://x.com/CurveFinance/status/2057435934526702083?s=20) to the banner and modal

### DIFF
--- a/apps/main/src/maintenances.ts
+++ b/apps/main/src/maintenances.ts
@@ -6,5 +6,5 @@ export const BACKEND_MAINTENANCE: MaintenanceConfig = {
   dateISO: '2026-05-25T07:00:00.000Z', // 25 May 2026, 09h00 CEST
   warnBefore: 'week',
   expectedDurationLabel: t`20 minutes to 1 hour`,
-  learnMoreLink: 'https://x.com/CurveFinance/status/2057435934526702083?s=20',
+  learnMoreLink: 'https://x.com/CurveFinance/status/2057435934526702083',
 }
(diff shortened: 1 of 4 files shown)


apps/main/src/maintenances.ts
```diff
@@ -6,4 +6,5 @@ export const BACKEND_MAINTENANCE: MaintenanceConfig = {
   dateISO: '2026-05-25T07:00:00.000Z', // 25 May 2026, 09h00 CEST
   warnBefore: 'week',
   expectedDurationLabel: t`20 minutes to 1 hour`,
+  learnMoreLink: 'https://x.com/CurveFinance/status/2057435934526702083?s=20',
```
DanielSchiavini: Is x our official channel? ☠️

DanielSchiavini (dismissed):

0xtutti (dismissed):

apps/main/src/maintenances.ts
DanielSchiavini: that s=20 doesn't make any sense btw

Pearce542: yeah don't know where it came from, I removed it

DanielSchiavini (approved):

packages/curve-ui-kit/src/features/maintenance/BackendMaintenanceModal.tsx
```diff
@@ -49,6 +46,11 @@ export const BackendMaintenanceModal = ({
         {t`Underlying smart contracts will continue to operate normally. This maintenance only affects parts of the app experience.`}
       </Typography>
       <Typography>{t`Thank you for your patience.`}</Typography>
+      {learnMoreLink && (
+        <Stack direction="row" justifyContent="end">
```
0xAlunara, line 50: I don't think this stack is necessary but it's not worth blocking

0xAlunara (approved):

PR #2568, opened 2026-05-22
refactor: market list filters drawer for mobile

Pearce542:
Depends on https://github.com/curvefi/curve-frontend/pull/2530, Next: #2558 

> Created another PR so it's easier to see the changes. Le @0xtutti what do you think of having a modal instead of a popover for mobile? Suggested by @OnlyJousting 

Answer: after talking with Tutti, a drawer is more appropriate for mobile, since we are already using it for sorting. Sorry for the changes!
  
- refactored the market list filters into a ~~modal~~ drawer for mobile instead of popover
- refactored the llamalend table filters logic into a single component instead of `FilterChip` and `LlamaTableFiltersOverlay`

Screenshot 2026-05-25 at 18 32 47


apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersOverlay.tsx
```diff
@@ -0,0 +1,86 @@
...
+      onClose={onClose}
+      anchorEl={anchorEl}
+      anchorOrigin={{ vertical: 'bottom', horizontal: 'left' }}
+      slotProps={{
+        paper: {
+          sx: { backgroundColor: t => t.design.Layer[3].Fill, width: Width.modal.md },
+        },
+      }}
+    >
+      <Stack sx={{ '& > * + *': { borderTop: t => `1px solid ${t.design.Layer[1].Outline}` } }}>
```
0xAlunara: nice selector lmao

0xAlunara (approved):
Right side doesn't seem to extend the fullwidth, but perhaps that's because of the Vercel thing only visible on PRs.

Image

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersOverlay.tsx
Pearce542: Dani 😉

Pearce542:
> Right side doesn't seem to extend the fullwidth, but perhaps that's because of the Vercel thing only visible on PRs.

Do you mean the scrollbar width? It doens't appear on mobile devices usually

0xtutti (approved):

PR #2570, opened 2026-05-22
refactor: maybe helper and eslint rule

DanielSchiavini:
- create maybe helper
- create eslint rule
- update existing code


package.json
```diff
@@ -8,6 +8,7 @@
   "workspaces": [
     "apps/*",
     "packages/*",
+    "scripts",
```
0xAlunara: Leftover from codemod?

packages/primitives/src/objects.utils.ts
```diff
@@ -68,3 +68,18 @@ export const isEmpty = (obj: object) => Object.keys(obj).length === 0
...
+  value: [T1 | null | undefined, T2 | null | undefined] | null | undefined,
+  mapper: (value: [T1, T2]) => R,
+): R | undefined
+export function maybe<T1, T2, T3, R>(
+  value: [T1 | null | undefined, T2 | null | undefined, T3 | null | undefined] | null | undefined,
+  mapper: (value: [T1, T2, T3]) => R,
+): R | undefined
+export function maybe<T, R>(value: T | null | undefined, mapper: (value: T) => R): R | undefined
+export function maybe<T, R>(value: T | null | undefined, mapper: (value: T) => R): R | undefined {
+  if (value === null || value === undefined) return undefined
```
0xAlunara:
```suggestion
  if (value == null) return undefined
```

packages/primitives/src/objects.utils.ts
```diff
@@ -68,3 +68,18 @@ export const isEmpty = (obj: object) => Object.keys(obj).length === 0
 
 export const pick = <T, K extends keyof T>(obj: T, ...keys: K[]) =>
   Object.fromEntries(keys.map(key => [key, obj[key]])) as { [P in K]: T[P] }
+
+export function maybe<T1, T2, R>(
+  value: [T1 | null | undefined, T2 | null | undefined] | null | undefined,
+  mapper: (value: [T1, T2]) => R,
+): R | undefined
+export function maybe<T1, T2, T3, R>(
```
0xAlunara: Do we really need that many generic overloads for the tuples? I figured simply having `value: T | T[] | null | undefined` would suffice?

EDIT: I guess if you want to `maybe` over `[foo, bar]` where `foo` is `number | undefined` and `bar` is `string | undefined` or any other non overlapping type combination you have to use this.

EDIT2: Codex was able to generalize it at least to tuples of size `N`, up to you if it's worth it or not:
```typescript
type NonNullishTuple<T extends readonly unknown[]> = {
  [K in keyof T]: NonNullable<T[K]>
}

export function maybe<const T extends readonly unknown[], R>(
  value: readonly [...T] | null | undefined,
  mapper: (value: NonNullishTuple<T>) => R,
): R | undefined
export function maybe<T, R>(value: T | null | undefined, mapper: (value: NonNullable<T>) => R): R | undefined
export function maybe<T, R>(value: T | null | undefined, mapper: (value: T) => R): R | undefined {
  if (value == null) return undefined
  if (Array.isArray(value) && value.some(x => x == null)) return undefined
  return mapper(value)
}

```

packages/curve-ui-kit/src/types/util.ts
```diff
@@ -71,7 +72,7 @@ export const mapQuery = <TSource, TResult>(
 ) =>
   q({
     isLoading,
-    data: data == null ? undefined : (selector(data) ?? undefined),
+    data: maybe(data, data => selector(data) ?? undefined),
```
0xAlunara, line 75: Was thinking to myself what the point of `?? undefined` was, but I guess it's in the case `selector(data)` return `null` which doesn't sit well with TanStack?

apps/main/src/llamalend/widgets/MarketRateCurveChart.tsx
```diff
@@ -94,24 +94,24 @@ export const MarketRateCurveChart = ({
...
+    ) ?? null
+  const utilizationBreakdown = maybe(
+    [totalBorrowed, capAndAvailable?.totalAssets],
+    ([borrow, available]) =>
+      `${formatNumber(borrow, { abbreviate: true })}/${formatNumber(available, {
+        abbreviate: true,
+      })} ${borrowToken?.symbol ?? ''}`,
+  )
+
+  const collateralTotal = maybe(totalCollateral, totalCollateral => Number(totalCollateral.collateral)) ?? null
```
0xAlunara, line 109: Not related nor in scope but I wonder if those `?? null` is really what we want here

EDIT: maybe we could return `null` if the value itself is `null` as well rather than `undefined`. Although that's a bit smelly as well.

apps/main/src/llamalend/features/market-list/cells/LineGraphCell.tsx
```diff
@@ -49,8 +50,8 @@ export const LineGraphCell = ({ market, type, graphSize = defaultGraphSize }: Ra
     isIntersecting,
   )
   const { design } = useTheme()
-  if (rate == null) return null // supply yield disabled for mint markets
-  return (
+  // supply yield disabled for mint markets
+  return maybe(rate, () => (
```
0xAlunara, line 54: You're not exactly returning `null` anymore but I don't think that matters for TSX components?

0xAlunara (dismissed):
All comments are just me being philosophical

packages/curve-ui-kit/src/types/util.ts
DanielSchiavini: There is some distinction in the code rn between `null` (query has no data) and `undefined` query did not run. I just kept it for now.

apps/main/src/llamalend/widgets/MarketRateCurveChart.tsx
DanielSchiavini: Maybe not, but at least for now I'm trying not to break anything

apps/main/src/llamalend/features/market-list/cells/LineGraphCell.tsx
DanielSchiavini: indeed, it does not matter

packages/primitives/src/objects.utils.ts
DanielSchiavini: thanks!

0xAlunara (dismissed):

0xAlunara:
Uh oh the lint check caught a few missing cases 👀

0xAlunara (approved):

OnlyJousting (approved):

packages/curve-ui-kit/src/types/util.ts
Pearce542, line 75: if the data is an array, `maybe` will return `undefined` if at least one element is nullish. Is it intended? Wasn't like this before

DanielSchiavini: - https://github.com/curvefi/curve-frontend/pull/2604

PR #2571, opened 2026-05-23
feat: refuel

0xAlunara:
Final version after having a sync call with tutti and michael k. To save time and effort I've re-used the chart logic from the analytics page, seems like a perfect fit. The page should be fast and visible without having to connect a wallet.

Had to add refuel endpoints to prices api. Apparantly it's all nullable so that is a bit ugly, but it is what it is.

[This pool is useful for testing](https://curve-dapp-git-feat-refuel-curvefi.vercel.app/dex/ethereum/pools/0x6e5492f8ea2370844ee098a56dd88e1717e4a9c2/refuel): 
[This pool should show a low reserve warning](https://curve-dapp-git-feat-refuel-curvefi.vercel.app/dex/xdai/pools/0x0b9036F639f4E7498aFA34751a2710159373fC7B/refuel)

image


0xtutti (dismissed):

apps/main/src/analytics/features/charts/components/EChartsCard.tsx
```diff
@@ -24,12 +24,15 @@ type ChartCardProps = {
   loading: boolean
   /** EChart options */
   option: EChartsOption
-  /** Whether the card should be shown as a fullscreen modal or */
+  /** Content for above the chart, like metrics */
+  topContent?: ReactNode
```
DanielSchiavini: 'like' metrics or always just metrics? If it's only used for metrics we might as well make it opinionated imo

```suggestion
  /** Element that renders the chart metrics */
  metrics?: ReactNode
```

apps/main/src/dex/components/PagePool/index.tsx
```diff
@@ -91,6 +93,7 @@ export const Transfer = (pageTransferProps: PageTransferProps) => {
   const snapshotData = snapshots?.[0]
 
   const pricesApiPoolData = poolData && pricesApiPoolsMapper?.[poolData.pool.address]
+  const hasRefuel = pricesApiPoolData?.poolMethods.includes('donation_shares')
```
DanielSchiavini: This magic string in a component isn't great
```suggestion
  const hasRefuel = !!pricesApiPoolData && hasRefuelMethod(pricesApiPoolData)
```

DanielSchiavini: Is the prices API data always available?
Could we have the same feature with curve-js data?

apps/main/src/dex/components/PageRefuel/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <DetailPageLayout
+        header={<Header blockchainId={blockchainId} poolAddress={poolAddress} />}
+        formTabs={<RefuelFormTabs chainId={chainId} blockchainId={blockchainId} poolAddress={poolAddress} />}
+        testId="refuel-page"
+      >
+        <Grid container columnSpacing={Spacing.md}>
+          {reserves?.lowReserves && (
+            <Grid size={12}>
+              <Banner
+                subtitle={t`Current reserve ratio: ${formatNumber(reserves.reserveRatio, { unit: 'percentage', decimals: 6, abbreviate: false })}. Refuel the pool to maintain balance`}
```
DanielSchiavini:
```suggestion
                subtitle={t`Current reserve ratio` + `: ${formatNumber(reserves.reserveRatio, { unit: 'percentage', decimals: 6, abbreviate: false })}. ` + t`Refuel the pool to maintain balance`}
```

apps/main/src/dex/features/refuel/components/DailyRefuelsChart.tsx
```diff
@@ -0,0 +1,156 @@
...
+            data: chartData.map(point => point.totalUsd),
+            type: 'line',
+            z: 2,
+          },
+          {
+            name: REFUELS_COUNT_LABEL,
+            data: chartData.map(point => point.count),
+            type: 'bar',
+            yAxisIndex: 1,
+            barWidth: '55%',
```
DanielSchiavini: is this safe for (very) small screens?

apps/main/src/dex/features/refuel/components/Header.tsx
```diff
@@ -0,0 +1,43 @@
...
+        <ArrowLeft />
+      </IconButton>
+
+      <WithSkeleton loading={isLoading} variant="rectangular" width={35} height={35}>
+        {pool?.coins && (
+          <TokenPair chain={blockchainId} assets={{ primary: pool.coins[0], secondary: pool.coins[1] }} hideChainIcon />
+        )}
+      </WithSkeleton>
+
+      <WithSkeleton loading={isLoading} width={140} height={24}>
```
DanielSchiavini, line 38: are the hardcoded sizes really needed? Skeleton should generally take the size of its content!

apps/main/src/dex/features/refuel/components/BorrowInformationContainer.tsx
```diff
@@ -0,0 +1,75 @@
...
+  poolAddress: Address
+}) => {
+  const { data: pool, isLoading: isPoolLoading, error: poolError } = useRefuelPool({ blockchainId, poolAddress })
+
+  return (
+    <Card size="small" data-testid="refuel-pool-information">
+      <CardHeader title={t`Pool Information`} />
+      <CardContent>
+        <Grid container columnSpacing={Spacing.md}>
+          <Grid size={2}>
```
DanielSchiavini, line 29: Size sums up to 8 shouldn't it be 12? And isn't 2 very small for mobiles?
Anyway if all columns are the same and non-responsive, we might not need a `Grid` at all.

apps/main/src/dex/features/refuel/components/RefuelForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+          name="tokenAAmount"
+          balance={values.tokenAAmount}
+          onBalance={tokenAAmount => form.update({ tokenAAmount })}
+          isError={!!tokenA.amountError}
+          walletBalance={{
+            balance: tokenA.balance,
+            symbol: tokenA.symbol,
+            usdRate: tokenA.usdRate,
+            loading: tokenA.isLoading,
+          }}
```
DanielSchiavini:
```suggestion
          walletBalance={tokenA}
```

apps/main/src/dex/features/refuel/components/RefuelForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+          name="tokenBAmount"
+          balance={values.tokenBAmount}
+          onBalance={tokenBAmount => form.update({ tokenBAmount })}
+          isError={!!tokenB.amountError}
+          walletBalance={{
+            balance: tokenB.balance,
+            symbol: tokenB.symbol,
+            usdRate: tokenB.usdRate,
+            loading: tokenB.isLoading,
+          }}
```
DanielSchiavini:
```suggestion
          walletBalance={tokenB}
```

apps/main/src/dex/features/refuel/components/RefuelForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+          inputBalanceUsd={decimal(values.tokenBAmount && tokenB.usdRate && tokenB.usdRate * +values.tokenBAmount)}
+          tokenSelector={
+            <TokenLabel
+              blockchainId={blockchainId}
+              address={tokenB.address}
+              label={tokenB.symbol ?? '?'}
```
DanielSchiavini, line 80: hardcoded and repeated placeholder should probably be handled in `TokenLabel` for consistency

apps/main/src/dex/features/refuel/components/RefuelFormList.tsx
```diff
@@ -0,0 +1,92 @@
...
+export const RefuelFormList = ({ values, tokenA, tokenB, poolTvl: pool }: RefuelFormListProps) => {
+  const refuelTvlPercentage = useMemo(
+    () =>
+      tokenA.usdRate == null ||
+      tokenB.usdRate == null ||
+      pool.usd == null ||
+      pool.usd === 0 ||
+      (values.tokenAAmount == null && values.tokenBAmount == null)
+        ? undefined
+        : ((+(values.tokenAAmount ?? 0) * tokenA.usdRate + +(values.tokenBAmount ?? 0) * tokenB.usdRate) / pool.usd) *
```
DanielSchiavini, line 48: shouldn't we be using the `decimal` helpers such as `decimalMultiply` and `decimalDiv`?

apps/main/src/dex/features/refuel/components/RefuelPricesChart.tsx
```diff
@@ -0,0 +1,197 @@
...
+import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
+import { formatNumber } from '@ui-kit/utils'
+import { REFUEL_TIMESERIES_PAGE_SIZE, useRefuelTimeseries } from '../queries/timeseries.query'
+
+const { Spacing } = SizesAndSpaces
+
+const LAST_PRICE_LABEL = t`Last price`
+const ORACLE_PRICE_LABEL = t`Oracle price`
+const PRICE_SCALE_LABEL = t`Price scale`
+const PERIODS = ['7d', '1m', '3m', '6m', '1y'] as const satisfies Period[]
```
DanielSchiavini, line 35: https://github.com/curvefi/curve-frontend/blob/02f82be196d4336081cc0e0820216e79c74681c4/apps/main/src/analytics/features/charts/types.ts#L2

apps/main/src/dex/features/refuel/components/RefuelPricesChart.tsx
```diff
@@ -0,0 +1,197 @@
...
+  formatNumber(value, { abbreviate: false, decimals: 6, fallback: '-' })
+
+// Helper functions and filters that don't type check nicely when inlined
+const toExportPoint = (time: number, value: number | null) => (value == null ? [] : [{ time, value }])
+const isFinitePrice = (value: number | null): value is number => value != null && Number.isFinite(value)
+const getLatestMetric = (values: (number | null | undefined)[]) =>
+  values.findLast((value): value is number => value != null && Number.isFinite(value))
+
+/** Used to calculate padded minimum for the price axis with some extra padding */
+const getPaddedMin = (values: number[]) => {
```
DanielSchiavini, line 55: suspiciously similar to this
https://github.com/curvefi/curve-frontend/blob/bb742d8a084db50b124df458fb7f30486edc7074/apps/main/src/llamalend/widgets/small-liquidation-range-chart/small-liquidation-range-chart.utils.ts#L51

apps/main/src/dex/features/refuel/components/RefuelPricesChart.tsx
```diff
@@ -0,0 +1,197 @@
...
+          })),
+        },
+        palette,
+      }),
+    [chartData, legendSets, palette, yAxisMin],
+  )
+
+  return (
+    <EChartsCard
+      title={`${data?.tokens[1]?.symbol ?? '?'} prices in ${data?.tokens[0]?.symbol ?? '?'}`}
```
DanielSchiavini, line 148: again the same placeholder constants

apps/main/src/dex/features/refuel/components/RefuelPricesChart.tsx
```diff
@@ -0,0 +1,197 @@
...
+              valueOptions={{ abbreviate: true, unit: 'dollar' }}
+              loading={loading}
+              testId="refuel-lp-token-value"
+            />
+          </Grid>
+
+          <Grid size={{ mobile: 6, tablet: 4 }}>
+            <Metric
+              label={t`Virtual price`}
+              value={virtualPrice && virtualPrice / 10 ** 18}
```
DanielSchiavini: shouldn't we be using `fromWei`? And shouldn't the `18` be in a constant?
https://github.com/curvefi/curve-frontend/blob/e26deb916f029a105606c013b4a7c8192e8c6ed6/packages/curve-ui-kit/src/utils/decimal.ts#L63

apps/main/src/dex/features/refuel/components/RefuelSharesChart.tsx
```diff
@@ -0,0 +1,218 @@
...
+import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
+import { formatNumber } from '@ui-kit/utils'
+import { refuelPoolAbi } from '../abi'
+import { REFUEL_TIMESERIES_PAGE_SIZE, useRefuelTimeseries } from '../queries/timeseries.query'
+
+const { Spacing } = SizesAndSpaces
+
+const REFUEL_SHARES_LABEL = t`Refuel shares`
+const UNLOCKED_SHARES_LABEL = t`Unlocked refuel shares`
+const PERIODS = ['7d', '1m', '3m', '6m', '1y'] as const satisfies Period[]
```
DanielSchiavini, line 38: again 🫠

apps/main/src/dex/features/refuel/components/RefuelSharesChart.tsx
```diff
@@ -0,0 +1,218 @@
...
+      })),
+    [palette.colors, visibility],
+  )
+
+  const chartData = useMemo(
+    () =>
+      llama(data?.data)
+        .map(point => ({
+          time: new Date(point.timestamp).getTime(),
+          donationShares: point.donationShares != null ? point.donationShares / 10 ** 18 : 0,
```
DanielSchiavini: - double-negative
- 0 is probably a bad placeholder
- hardcoded 18
- `fromWei` usage

apps/main/src/dex/features/refuel/components/ReservesCompositionChart.tsx
```diff
@@ -0,0 +1,156 @@
...
+  prices: (number | null | undefined)[] | null | undefined
+  decimals: (number | undefined)[]
+}) => {
+  const reserveUsdValues =
+    reserves?.map((reserve, index) => {
+      const amount = reserve == null || decimals[index] == null ? 0 : reserve / 10 ** decimals[index]
+      const relativeUsdPrice = index === 0 ? 1 : (prices?.[index - 1] ?? 1)
+      return amount * relativeUsdPrice
+    }) ?? []
+  const totalUsd = reserveUsdValues.reduce((sum, value) => sum + value, 0)
```
DanielSchiavini, line 49:
```suggestion
  const totalUsd = sum(reserveUsdValues)
```

apps/main/src/dex/features/refuel/components/recent-refuels/cells/AmountCell.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { formatNumber } from '@ui-kit/utils'
+
+const { Spacing } = SizesAndSpaces
+
+type AmountCellProps = {
+  amount?: number | null
+  usdAmount?: number | null
+}
+
+export const AmountCell = ({ amount, usdAmount }: AmountCellProps) => {
```
DanielSchiavini, line 13: isn't this duplicated from llamalend?

apps/main/src/dex/features/refuel/components/recent-refuels/cells/AmountCell.tsx
```diff
@@ -0,0 +1,26 @@
+import Stack from '@mui/material/Stack'
+import Typography from '@mui/material/Typography'
+import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
+import { formatNumber } from '@ui-kit/utils'
+
+const { Spacing } = SizesAndSpaces
+
+type AmountCellProps = {
+  amount?: number | null
+  usdAmount?: number | null
```
DanielSchiavini: This shouldn't be optional
```suggestion
  amount: number | null | undefined
  usdAmount: number | null | undefined
```

apps/main/src/dex/features/refuel/components/recent-refuels/columns/columns.definitions.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export type RecentRefuelsToken = {
+  symbol: string
+  address: Address
+  decimals: number
+}
+
+export type RecentRefuelRow = TableItem & {
+  timestamp: number
+  donor?: Address | null
```
DanielSchiavini, line 17: why is everything optional?

apps/main/src/dex/features/refuel/hooks/useLowReserves.ts
```diff
@@ -0,0 +1,46 @@
...
+    address: poolAddress,
+    abi: refuelPoolAbi,
+    functionName: 'donation_shares',
+    chainId,
+  })
+
+  const reserveRatio =
+    refuelShares.data == null || pool.data == null || pool.data.lpTokenSupply === 0
+      ? undefined
+      : Number(refuelShares.data) / 10 ** 18 / pool.data.lpTokenSupply
```
DanielSchiavini: This doesn't belong in a component, plus see earlier comments

apps/main/src/dex/features/refuel/hooks/useRefuelForm.ts
```diff
@@ -0,0 +1,101 @@
...
+  const tokenBBalance = useTokenBalance({ chainId, userAddress, tokenAddress: coins?.[1]?.address })
+
+  const tokenAUsdRate = useTokenUsdRate({ chainId, tokenAddress: coins?.[0]?.address })
+  const tokenBUsdRate = useTokenUsdRate({ chainId, tokenAddress: coins?.[1]?.address })
+
+  const tokens = useMemo(
+    (): Tokens | undefined =>
+      coins != null && tokenADecimals.data != null && tokenBDecimals.data != null
+        ? {
+            tokenA: { address: coins[0].address, symbol: coins[0].symbol, decimals: tokenADecimals.data },
```
DanielSchiavini, line 46: Why do we only use A/B while pools can have 8 tokens?
Why do we A/B instead of a tuple or array of tokens?

apps/main/src/dex/features/refuel/mutations/refuel.mutation.ts
```diff
@@ -0,0 +1,93 @@
...
+  onReset: () => void
+}
+
+export const useRefuelMutation = ({ chainId, poolAddress, tokens, userAddress, onReset }: RefuelMutationOptions) => {
+  const config = useConfig()
+  const { mutate, error, isPending } = useTransactionMutation<RefuelFormValues>({
+    mutationKey: [...rootKeys.chain({ chainId }), 'refuel'] as const,
+    mutationFn: async (form: RefuelFormValues) => {
+      // Not sure what to do here, this shouldn't be happening but i cant make the type not undefined
+      // TODO: handle this with validation or something?
```
DanielSchiavini: The validation suite only applies to `RefuelFormValues` which don't include the `RefuelMutationOptions`.
So this is kind of expected. For llamalend we used `buildContext` combined with `assert`:
https://github.com/curvefi/curve-frontend/blob/2e4d4f743004bfaf69478ebb696ffec8de20c512/apps/main/src/llamalend/mutations/useLlammaMutation.ts#L54-L59

apps/main/src/dex/features/refuel/mutations/refuel.mutation.ts
```diff
@@ -0,0 +1,93 @@
...
+    },
+    validationSuite: refuelFormValidationSuite,
+    validationParams: {},
+    pendingMessage: () => t`Refueling pool...`,
+    successMessage: () => t`Pool refueled!`,
+    onReset: () => {
+      void invalidateTokenBalances(config, {
+        chainId,
+        userAddress: userAddress!,
+        tokenAddresses: [tokens!.tokenA.address, tokens!.tokenB.address],
```
DanielSchiavini, line 82: we can easily pass `context` to `onReset` too

apps/main/src/dex/features/refuel/queries/pools.query.ts
```diff
@@ -0,0 +1,21 @@
...
+  queryFn: async ({ blockchainId }: ChainNameQuery) => getRefuelPools(blockchainId),
+  validationSuite: pricesApiChainValidationSuite,
+  category: 'dex.pools',
+})
+
+type RefuelPoolParams = ChainNameParams & { poolAddress?: Address | null }
+
+export const useRefuelPool = ({ blockchainId, poolAddress }: RefuelPoolParams) =>
+  mapQuery(
+    useRefuelPools({ blockchainId }, !!poolAddress),
```
DanielSchiavini: why is the `!!poolAddress` needed?

apps/main/src/dex/features/refuel/queries/pools.query.ts
```diff
@@ -0,0 +1,21 @@
...
+import { mapQuery } from '@ui-kit/types/util'
+
+export const { useQuery: useRefuelPools, fetchQuery: fetchRefuelPools } = queryFactory({
+  queryKey: ({ blockchainId }: ChainNameParams) => [...rootKeys.chainName({ blockchainId }), 'getRefuelPools'] as const,
+  queryFn: async ({ blockchainId }: ChainNameQuery) => getRefuelPools(blockchainId),
+  validationSuite: pricesApiChainValidationSuite,
+  category: 'dex.pools',
+})
+
+type RefuelPoolParams = ChainNameParams & { poolAddress?: Address | null }
```
DanielSchiavini: The hook won't work without `poolAddress` so it shouldn't be optional
```suggestion
type RefuelPoolParams = ChainNameParams & { poolAddress: Address | null | undefined }
```

packages/curve-ui-kit/src/hooks/useTokenDecimals.ts
```diff
@@ -0,0 +1,18 @@
+import { type Address, erc20Abi } from 'viem'
+import { useReadContract } from 'wagmi'
+import type { FieldsOf } from '@ui-kit/lib'
+import type { ChainQuery } from '@ui-kit/lib/model'
+
+type TokenDecimalsQuery = ChainQuery & { tokenAddress: Address }
+
+/** Fetch ERC-20 token decimals for a token on a specific chain. */
```
DanielSchiavini, line 8: We already have that data from curve-js why would we add yet another onchain call?

packages/curve-ui-kit/src/lib/model/entities/allowance.ts
```diff
@@ -0,0 +1,53 @@
...
+      abi: erc20Abi,
+      functionName: 'allowance',
+      args: [userAddress, spenderAddress],
+    }),
+    staleTime: 0,
+  })
+
+/** Helper function for @see {@link fetchAllowance} that checks if the allowance is sufficient for a given amount. */
+export const fetchHasEnoughAllowance = async (config: Config, params: AllowanceQuery & { amount: bigint }) =>
+  params.amount <= 0n ? true : (await fetchAllowance(config, params)) >= params.amount
```
DanielSchiavini:
```suggestion
  params.amount <= 0n || (await fetchAllowance(config, params)) >= params.amount
```

packages/curve-ui-kit/src/utils/errors.ts
```diff
@@ -13,9 +13,9 @@ export function getErrorMessage(error: CustomError | null, defaultErrorMessage?:
   let errorMessage = defaultErrorMessage ?? ''
 
   if (error?.message) {
-    if (error.message.includes('user rejected transaction') || error?.code === 'ACTION_REJECTED') {
+    if (error.message.toLocaleLowerCase().includes('user rejected') || error?.code === 'ACTION_REJECTED') {
       errorMessage = t`User rejected transaction`
-    } else if (error.message.includes('Bad swap type')) {
+    } else if (error.message.toLocaleLowerCase().includes('Bad swap type')) {
```
DanielSchiavini: that can't possibly work

tests/cypress/e2e/dex/refuel.cy.ts
```diff
@@ -0,0 +1,77 @@
...
+  const visitRefuelPage = () => {
+    cy.visitWithoutTestConnector(REFUEL_ROUTE)
+    cy.location('pathname', LOAD_TIMEOUT).should('equal', `/${REFUEL_ROUTE}`)
+    getTestId('page', LOAD_TIMEOUT).should('be.visible')
+  }
+
+  it('renders the crvUSD/WETH refuel dashboard and disconnected form state', () => {
+    visitRefuelPage()
+
+    cy.contains('Refuel setup').should('be.visible')
```
DanielSchiavini: can we use test-ids instead of hardcoding strings in tests?

tests/cypress/e2e/dex/refuel.cy.ts
```diff
@@ -0,0 +1,77 @@
+import type { AppRoute } from '@cy/support/routes'
+import { API_LOAD_TIMEOUT, LOAD_TIMEOUT } from '@cy/support/ui'
+
+const REFUEL_POOL_ADDRESS = '0x6e5492f8ea2370844ee098a56dd88e1717e4a9c2'
+const REFUEL_ROUTE = `dex/ethereum/pools/${REFUEL_POOL_ADDRESS}/refuel` satisfies AppRoute
+
+/** Smol helper to reduce amount of repetition */
+const getTestId = (testId: string, options?: Partial<Cypress.Timeoutable>) =>
```
DanielSchiavini:
```suggestion
const getTestById = (testId: string, options?: Partial<Cypress.Timeoutable>) =>
```

tests/cypress/e2e/dex/refuel.cy.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  it('renders the crvUSD/WETH refuel dashboard and disconnected form state', () => {
+    visitRefuelPage()
+
+    cy.contains('Refuel setup').should('be.visible')
+
+    cy.get('input[name="tokenAAmount"]').should('be.visible').and('have.attr', 'placeholder', '0.00')
+    cy.get('input[name="tokenBAmount"]').should('be.visible').and('have.attr', 'placeholder', '0.00')
+
+    cy.contains('crvUSD', API_LOAD_TIMEOUT).should('be.visible')
```
DanielSchiavini: It'd be nice to leave the token names in constants
Also, we use `API_LOAD_TIMEOUT` 20 times, shouldn't this info be loaded all at once?

apps/main/src/dex/components/PagePool/index.tsx
0xAlunara: Only the prices api returns a list of available pool methods. Hopefully this'll be consolidated once we get rid of the legacy API and rely purely on prices api. Not sure if there's a reliable on-chain fallback possible though.

apps/main/src/dex/components/PageRefuel/index.tsx
0xAlunara: Figured you'd like the single interpolated one 😔

apps/main/src/dex/features/refuel/components/DailyRefuelsChart.tsx
0xAlunara: A bit smoshed, but it turns out it's almost identical to the ECharts default, so I've just removed it as it didn't offer any benefit.

image

apps/main/src/dex/features/refuel/components/Header.tsx
0xAlunara: It's copied from `page-header.tsx` of llamalend. For the token icons it's needed because if there's no tokens there's also no icons. For the text it's needed as by default it'll span the entire header width which is a bit too long. And for some reason the height isn't correctly inferred by mui for the typography. Could refactor it to shared constants but this is a temp component anyway as the jsdoc above says, so I'll leave it for now.

apps/main/src/dex/features/refuel/components/BorrowInformationContainer.tsx
0xAlunara: It indeed sums up to 8, but that's because I want similar size and spacing as the other card with 3 metrics, even though this one only has 2 metrics.

apps/main/src/dex/features/refuel/components/RefuelForm.tsx
0xAlunara: Hah, I've refactored this so many times I didn't see the obvious simplification at the end.

apps/main/src/dex/features/refuel/components/RefuelForm.tsx
0xAlunara: Will make a ticket for that, probably something we need to adjust in other places too

apps/main/src/dex/features/refuel/components/RefuelPricesChart.tsx
0xAlunara: I'll create a ticket for this; looks very obvious but there's a few exception wrt some prices api endpoints.

apps/main/src/dex/features/refuel/components/RefuelFormList.tsx
0xAlunara: Everything but the amounts are `number` rather than `Decimal`, so I found this a bit more readable. It's purely presentational anyway, not contract input, so any loss of precision if it occurs isn't too important.

apps/main/src/dex/features/refuel/components/RefuelPricesChart.tsx
0xAlunara: Will ask JJ if it's solving the exact same issue and see if it can be standardized as a helper function.

apps/main/src/dex/features/refuel/components/RefuelPricesChart.tsx
0xAlunara: Noted this additional case for the TokenLabel ticket

apps/main/src/dex/features/refuel/components/RefuelPricesChart.tsx
0xAlunara: All those helper functions are for converting between strings and bigints. This time however the price is returned as a normal number by prices api; already parsed, so a normal division suffices. I did add `DEFAULT_DECIMALS` though

apps/main/src/dex/features/refuel/components/RefuelSharesChart.tsx
0xAlunara: Idk I can't let the placeholder be `null` or `undefined` as it's per datapoint, and using these instead of 0 for charting libraries leads to all kinds of breaking results. Removed the double negatives.

apps/main/src/dex/features/refuel/components/recent-refuels/cells/AmountCell.tsx
0xAlunara: We should rather look if we can consolidate all token amount cells with the one from the activity table (`TokenAmountCell`). Will make a ticket.

apps/main/src/dex/features/refuel/components/recent-refuels/columns/columns.definitions.tsx
0xAlunara: Zod typing because for some rason prices api says it can be null. If I make this `T | null | undefined` we need to painstakingly modify the Zod transform. I'd rather keep it like this.

apps/main/src/dex/features/refuel/hooks/useRefuelForm.ts
0xAlunara: At the moment refueling only supports two tokens. This is easier than trying to generalize to N all over the place when in fact it's always two for now. Would rather refactor when we get there.

apps/main/src/dex/features/refuel/hooks/useLowReserves.ts
0xAlunara: Figured it'd be nice as a separate file 😭

apps/main/src/dex/features/refuel/mutations/refuel.mutation.ts
0xAlunara: Refactored this to use `assert` as well, but then I lose the null type checking and need to null assert with `!`, which is ugly 😦

packages/curve-ui-kit/src/hooks/useTokenDecimals.ts
0xAlunara: Same reason we have `useTokenBalance`, so this page works with on-chain data without relying on a hydrated curve-js.

packages/curve-ui-kit/src/utils/errors.ts
0xAlunara: I was a bit to eager there 🙈

tests/cypress/e2e/dex/refuel.cy.ts
0xAlunara: I could remove it, but some of the data relies on prices api so if that's slow it might timeout

tests/cypress/e2e/dex/refuel.cy.ts
0xAlunara: Removed it for now, don't want to bother in this PR to deep select those LTI labels...

0xtutti (approved):

OnlyJousting (approved):

apps/main/src/dex/components/PageRefuel/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+        formTabs={<RefuelFormTabs chainId={chainId} blockchainId={blockchainId} poolAddress={poolAddress} />}
+        testId="refuel-page"
+      >
+        <Grid container columnSpacing={Spacing.md}>
+          {reserveRatio != null && reserveRatio < LOW_RESERVES_RATIO && (
+            <Grid size={12}>
+              <Banner
+                subtitle={
+                  t`Current reserve ratio: ` +
+                  formatNumber(reserveRatio, { unit: 'percentage', decimals: 6, abbreviate: false }) +
```
Pearce542, line 61: shouldn't `reserveRatio` be converted to a percentage before formatting it?
```suggestion
                  formatNumber(reserveRatio * 100, { unit: 'percentage', decimals: 6, abbreviate: false }) +
```

apps/main/src/dex/components/PageRefuel/index.tsx
DanielSchiavini: that'd mess up translations (formatNumber shoudn't be in the `t`)

apps/main/src/dex/features/refuel/mutations/refuel.mutation.ts
DanielSchiavini: that should not be needed! `assert` filters nullables

PR #2572, opened 2026-05-23
refactor: add inline size to cards

0xAlunara:
Small refactor that gets rid of the padding exceptions and introduces `inline` as a proper size, like `small`. The dividing line between `size` and a new prop that also changes sizes (for the header) is blurry, so I've decided to just make it a proper `size`.

Inline looks a bit stupid but it's ment for cards within cards.

image

### DIFF
--- a/packages/curve-ui-kit/src/themes/components/card/mui-card.ts
+++ b/packages/curve-ui-kit/src/themes/components/card/mui-card.ts
@@ -1,9 +1,10 @@
 /// <reference types="./mui-card.d.ts" />
 import type { Components, TypographyVariantsOptions } from '@mui/material/styles'
-import { cardContentSmallStyles } from '../card-content'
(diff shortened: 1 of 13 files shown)


0xtutti (approved):

packages/curve-ui-kit/src/themes/components/card/mui-card.ts
```diff
@@ -18,5 +19,12 @@ export const defineMuiCard = (typography: TypographyVariantsOptions): Components
         '& .MuiCardContent-root': cardContentSmallStyles,
       },
     },
+    {
```
DanielSchiavini, line 22: I wish we could use `variant` instead of `size` but this is a huge improvement

DanielSchiavini (approved):

Pearce542 (approved):

OnlyJousting (approved):

PR #2573, opened 2026-05-23
feat: remove crv supply and price from llamalend header

0xAlunara:
We have the analytics page for these two metrics now

image


0xtutti (dismissed):

apps/main/src/llamalend/queries/crv-usd-supply-total.query.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/** Gets the total latest crvUSD supply, no matter the source of the mint */
+export const { useQuery: useCrvUsdSupplyTotal } = queryFactory({
+  queryKey: () => ['crvusd-supply-total'] as const,
+  queryFn: async () => {
+    const supply = await getCrvUsdSupply('ethereum', SUPPLY_LOOKBACK_DAYS)
+    const latestTimestamp = supply.reduce<number | null>(
+      (latest, item) => (latest == null || item.timestamp > latest ? item.timestamp : latest),
+      null,
+    )
```
DanielSchiavini:
```suggestion
    const latestTimestamp = max(supply, i => i.timestamp)
```

apps/main/src/llamalend/queries/crv-usd-supply-total.query.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    return latestTimestamp == null
+      ? null
+      : supply.reduce((total, item) => (item.timestamp === latestTimestamp ? total + item.supply : total), 0)
```
DanielSchiavini:
```suggestion
    return latestTimestamp && sum(supply.filter(i => i.timestamp === latestTimestamp), i => i.supply)
```

DanielSchiavini (commented):
We could use lodash or our own helpers, but reduce in business logic is quite hard to read imo

apps/main/src/llamalend/queries/crv-usd-supply-total.query.ts
0xAlunara: Have to keep the ternary as queries can't return `undefined` but should return `null`.

DanielSchiavini (approved):

Pearce542 (approved):

PR #2580, opened 2026-05-25
feat: use echarts for historical gauge weights chart

OnlyJousting:
Changes:
- Replaced DAO gauge weight history charts with the shared EChartsLineChart.
- Added a TanStack query for gauge weight history via @curvefi/prices-api.
- Removed legacy gauge history fetching/storage from the gauges Zustand slice.
- Moved shared chart tooltip components into curve-ui-kit.
- Updated Prices API gauge weight history parsing for normalized weights, relative weight %, emissions, and timestamps.
- Rename `userAddressValidation` -> `evmAddressValidation` and update callsites


apps/main/src/dao/components/Charts/GaugeWeightHistoryChart.tsx
```diff
@@ -0,0 +1,104 @@
...
+interface GaugeWeightHistoryChartProps {
+  gaugeAddress: string
+  height?: number
+}
+
+export const GaugeWeightHistoryChart = ({ gaugeAddress, height = Height.chart }: GaugeWeightHistoryChartProps) => {
+  const {
+    design: { Chart },
+  } = useTheme()
+  const lineColor = Chart.Lines[1]
```
Pearce542: We can save couple of lines 

```suggestion
  const lineColor = useTheme().design.Chart.Lines[1]
```

apps/main/src/dao/components/Charts/GaugeWeightHistoryChart.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const {
+    data = EMPTY_GAUGE_WEIGHT_HISTORY_DATA,
```
Pearce542: Do we need a constant just for = []?
```suggestion
    data = [],
```

Pearce542 (dismissed):

apps/main/src/dao/components/Charts/GaugeWeightHistoryChart.tsx
```diff
@@ -0,0 +1,104 @@
...
+    isLoading,
+    isSuccess,
+    error,
+    refetch,
+  } = useGaugeWeightHistoryQuery({
+    gaugeAddress,
+  })
+
+  const series = useMemo<LineSeriesConfig<GaugeWeightSeriesKey>[]>(
```
DanielSchiavini: I think the useMemo calls could be inlined, they aren't used more than once and not inside any loop or conditional

apps/main/src/dao/entities/gauge-weight-history.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export type { WeightHistory as GaugeWeightHistoryData } from '@curvefi/prices-api/gauge'
+
+export const { useQuery: useGaugeWeightHistoryQuery } = queryFactory({
+  queryKey: (params: { gaugeAddress: string }) =>
+    ['gauge-weight-history', { gaugeAddress: params.gaugeAddress.toLowerCase() }] as const,
+  queryFn: async ({ gaugeAddress }: { gaugeAddress: string }) =>
+    sortBy(await getWeightHistory(gaugeAddress), 'timestamp'),
+  category: 'dao.gauges',
+  validationSuite: EmptyValidationSuite,
```
DanielSchiavini: I think we should be reusing `ContractQuery`/`ContractParams`/`rootKeys.contract`/`contractValidationGroup`
I think it's also nice to reuse the called url/function in the query key so queries are easier to debug

0xtutti (dismissed):

apps/main/src/dao/components/Charts/GaugeWeightHistoryChart.tsx
OnlyJousting: 100%, there's no toggling/conditional series in this chart

packages/curve-ui-kit/src/shared/ui/Chart/ChartTooltipComponents.tsx
```diff
@@ -51,3 +51,10 @@ export const ChartTooltipSeriesRow = ({
     <Typography variant="bodySBold">{value}</Typography>
```
0xAlunara, line 51: Need to use `sx` props here with mui v9 and in some other places too. Prob best to sync with `main` and run `typecheck`.

apps/main/src/llamalend/queries/market-list/mint-markets.ts
```diff
@@ -28,6 +28,10 @@ export const { getQueryOptions: getMintMarketOptions, invalidate: invalidateMint
   validationSuite: EmptyValidationSuite,
 })
 
+const userEvmAddressValidationSuite = createValidationSuite(({ userAddress }: UserParams) => {
```
0xAlunara: Is this a dupe of the suite inside `@ui-kit/lib/model/query/evm-address-validation`?

apps/main/src/llamalend/queries/market-list/lending-vaults.ts
```diff
@@ -31,6 +31,10 @@ export const { getQueryOptions: getLendingVaultsOptions, invalidate: invalidateL
   validationSuite: EmptyValidationSuite,
 })
 
+const userEvmAddressValidationSuite = createValidationSuite(({ userAddress }: UserParams) => {
```
0xAlunara: Same dupe here

packages/curve-ui-kit/src/lib/model/query/evm-address-validation.ts
```diff
@@ -0,0 +1,29 @@
+import { enforce, group, skipWhen, test } from 'vest'
+import { isAddress } from 'viem'
+import { createValidationSuite } from '@ui-kit/lib/validation'
+
+export const evmAddressValidationGroup = <T extends string>({
```
0xAlunara: So I'm not sure if we still need `contract-validation.ts`, could we remove that too now?

0xtutti (dismissed):

packages/curve-ui-kit/src/lib/model/query/evm-address-validation.ts
OnlyJousting: contract validation takes a blockchainId too so it's a little different

packages/curve-ui-kit/src/shared/ui/Chart/ChartTooltipComponents.tsx
OnlyJousting: I'm not getting any errors in this file 🤔

apps/main/src/dao/components/PageGauge/index.tsx
```diff
@@ -27,7 +31,9 @@ export const Gauge = () => {
       <Stack sx={{ gap: PAGE_SPACING, backgroundColor: t => t.design.Layer[1].Fill }}>
         <GaugeHeader gaugeData={gaugeData} dataLoading={gaugesIsLoading} />
         <GaugeMetrics gaugeData={gaugeData} dataLoading={gaugesIsLoading} />
-        <Stack>{gaugeData && <GaugeWeightHistoryChart gaugeAddress={gaugeData.address} minHeight={25} />}</Stack>
+        <Stack sx={{ padding: Spacing.md }}>
+          {gaugeData && <GaugeWeightHistoryChart gaugeAddress={gaugeData.address as Address} />}
```
DanielSchiavini, line 35: Do we need skeletons?

DanielSchiavini (dismissed):

packages/curve-ui-kit/src/shared/ui/Chart/ChartTooltipComponents.tsx
0xAlunara: Probably not a thing anymore now it's synched with main?

0xAlunara (dismissed):

DanielSchiavini (approved):

0xtutti (approved):

apps/main/src/dao/components/PageGauge/index.tsx
OnlyJousting: I think it's ok as it is now but it could be improved with skeletons in the future, feels a little out of scope for this PR

PR #2584, opened 2026-05-27
fix: handle invalid and small prices

OnlyJousting:
Changes:
- Filter invalid and negative candle prices before calculating chart ranges
- Keep chart padding from pushing non-negative price ranges below zero
- Improve small price tick formatting for sub-unit assets
- Simplify chart price formatting by using locale rounding directly
- Add focused tests for range calculation and price formatting edge cases


0xtutti (approved):

packages/curve-ui-kit/src/features/candle-chart/utils.ts
```diff
@@ -106,24 +109,16 @@ export const priceFormatter = (x: number, delta: number) => {
...
+  // Let a tight visible range drive precision, while allowing extra decimals for sub-unit assets.
   const magnitude = deltaMagnitude < valueMagnitude - 1 ? deltaMagnitude : valueMagnitude
-  const decimals = Math.min(4, Math.max(0, 3 - magnitude))
+  const maxDecimals = x < 1 ? MAX_SUB_UNIT_PRICE_DECIMALS : MAX_PRICE_DECIMALS
+  const decimals = Math.min(maxDecimals, Math.max(0, 3 - magnitude))
 
-  const formatted = formatNumber(x, {
-    decimals,
-    abbreviate: false,
+  return x.toLocaleString('en-US', {
```
0xAlunara, line 118: rip to using `formatNumber`

0xAlunara (dismissed):
Presentation code is always hairy, so if it works it works despite code quality 🫠

0xtutti (dismissed):

packages/curve-ui-kit/src/features/candle-chart/utils.ts
```diff
@@ -1,8 +1,13 @@
-import { TIME_OPTION_MS } from '@ui-kit/lib/model/time'
-import { formatNumber } from '@ui-kit/utils'
+import { TIME_OPTION_MS } from '../../lib/model/time'
+import { formatNumber } from '../../utils/number'
 import type { TimeOption } from './types'
 
 const toSeconds = (timeOption: TimeOption) => TIME_OPTION_MS[timeOption] / 1000
+const isValidPrice = (value: number) => Number.isFinite(value) && value >= 0
```
DanielSchiavini: why do we expect invalid values? A comment would be nice

packages/curve-ui-kit/src/features/candle-chart/utils.ts
OnlyJousting: the reason is to allow rounding to 0 in the chart instead of "<0.00001" etc, 0 is more desirable for chart ticks

packages/curve-ui-kit/src/features/candle-chart/utils.ts
OnlyJousting: hm yea this might be covering up issues in the data which I guess is counter to what we would want, we want to see those issues. I'll remove it (I don't expect invalid values)

0xAlunara (approved):

0xtutti (approved):

DanielSchiavini (approved):

PR #2587, opened 2026-05-27
feat: vsdCRV token alert

Pearce542:
- add vsdCRV token alert


Pearce542:
pool alerts and markets alerts should be enough

PR #2591, opened 2026-05-27
feat: new llamalend user position table

Pearce542:
- refactored the user position tbale into two vertical tables
- removed the user position tab logic

Screenshot 2026-05-27 at 19 00 40


0xAlunara (approved):
More of a personal opinion, but if one of the two tables is empty I'd rather see nothing. It's taking up  a whole chunk of vertical space just to say there's no position lol

Image

OnlyJousting (approved):
this looks so much better ❤️

apps/main/src/llamalend/features/market-list/UserPositionsEmptyState.tsx
```diff
@@ -17,15 +17,15 @@ type EmptyStateConfig = {
 type UserPositionsEmptyStateProps = {
   table: Table<LlamaMarket>
   state: PositionsEmptyState
-  tab: MarketRateType
+  marketRateType: MarketRateType
```
DanielSchiavini, line 20: rateType would be clear enough

apps/main/src/llamalend/features/market-list/UserPositionsEmptyState.tsx
```diff
@@ -17,15 +17,15 @@ type EmptyStateConfig = {
...
   onReload: () => void
-  resetFilters: () => void
+  resetFilters?: () => void
```
DanielSchiavini, line 22: I see this becoming optional, but I don't see it being removed somewhere - why is it optional now?

DanielSchiavini (approved):

apps/main/src/llamalend/features/market-list/UserPositionsEmptyState.tsx
Pearce542: I did that initially, but then I thought that it could lead to confusion with APR / APY rates

Pearce542: Also is `MarketRateType` in general the best name for `Borrow` | `Supply` 🤔

apps/main/src/llamalend/features/market-list/UserPositionsEmptyState.tsx
Pearce542: In the new user position table we don't have filters anymore, so It's not used there

PR #2595, opened 2026-05-27
styling: fix numeric text field border and chip colors to match figma

0xAlunara:
image
image

image
image

### DIFF
--- a/packages/curve-ui-kit/src/themes/design/2_theme.ts
+++ b/packages/curve-ui-kit/src/themes/design/2_theme.ts
@@ -275,11 +275,44 @@ export const createLightDesign = (
     Transition: 'Transition',
   } as const
 
+  const Inputs = {
+    Base: {
+      Default: {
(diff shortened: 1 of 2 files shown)


packages/curve-ui-kit/src/themes/design/2_theme.ts
```diff
@@ -275,11 +275,44 @@ export const createLightDesign = (
     Transition: 'Transition',
   } as const
 
+  const Inputs = {
```
0xAlunara, line 278: Just moved these `Inputs` above `Chip` so I could use `Stroke: Inputs.Base.Default.Border.Default,` there so it matches more closely with Figma.

DanielSchiavini (approved):

Pearce542:
image
wind-in-hair-fabulous-wind

Pearce542 (approved):

PR #2597, opened 2026-05-27
feat: new slippage modal

DanielSchiavini:
- implement new slippage modal
- remove slippage per pool
- add slippage for leverage
- remove support for slippage in query string for router


0xtutti (dismissed):

packages/curve-ui-kit/src/widgets/SlippageSettings/useSlipageSettingsForm.ts
```diff
@@ -0,0 +1,43 @@
...
+function isSlippage(nr: Decimal) {
+  enforce(nr)
+    .message(t`Invalid percentage number`)
+    .isDecimal()
+  enforce(nr)
+    .message(t`Slippage cannot be smaller than ${MIN_SLIPPAGE}%`)
+    .gte(MIN_SLIPPAGE)
+  enforce(nr)
+    .message(t`Slippage cannot be larger than ${MAX_SLIPPAGE}%`)
+    .lt(MAX_SLIPPAGE)
```
0xAlunara: `lte` would make more sense I think but not important

apps/main/src/dex/components/PagePool/utils.ts
```diff
@@ -46,3 +49,9 @@ export const DEFAULT_SLIPPAGE: Slippage = {
 }
 
 export const DEFAULT_ESTIMATED_GAS: EstimatedGas = { loading: false, estimatedGas: null, error: null }
+
+export function getSlippageType(poolData: PoolData): SlippageType
+export function getSlippageType(poolData: PoolData | undefined): SlippageType | undefined
+export function getSlippageType(poolData: PoolData | undefined): SlippageType | undefined {
+  return maybe(poolData, ({ pool }) => (pool.isCrypto ? 'crypto' : 'stable'))
```
0xAlunara: This may sound stupid, but is there merit in adding an overload for `maybe` where if the value given is of type `T` the return type will be `R` rather than `R | undefined`? Saves us from many overloads like `getSlippageType` when the value is not nullable to begin with.

Not in scope of PR, naturally

apps/main/src/dex/components/PageRouterSwap/Page.tsx
```diff
@@ -40,7 +39,7 @@ export const PageRouterSwap = () => {
   const nativeToken = curveApi?.getNetworkConstants()?.NATIVE_TOKEN
   const paramsFromAddress = searchParams?.get('from')?.toLowerCase() || nativeToken?.address || ''
   const paramsToAddress = searchParams?.get('to')?.toLowerCase() || nativeToken?.wrappedAddress || ''
-  const paramsMaxSlippage = searchParams?.get('slippage')
+  const paramsMaxSlippage = decimal(searchParams?.get('slippage'))
```
0xAlunara: This isn't used anymore and can be removed. The only usage is fake as it's in the dependency array of the `useEffect` but then not used. Linting can't catch this because it's a legacy form with the exhaustive deps rule turned off.

packages/curve-ui-kit/src/features/user-profile/store.ts
```diff
@@ -1,85 +1,35 @@
...
-  typeof window !== 'undefined'
-    ? window.matchMedia('(prefers-color-scheme: dark)').matches
-      ? 'dark'
-      : 'light'
-    : 'light'
-
 const INITIAL_STATE: UserProfileState = {
-  theme: INITIAL_THEME,
-  maxSlippage: { crypto: '0.1', stable: '0.03' },
+  theme: window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light',
```
0xAlunara, line 24: We're no longer checking if window is `undefined` but might not be needed after all

packages/curve-ui-kit/src/features/user-profile/store.ts
```diff
@@ -1,85 +1,35 @@
...
-          state.maxSlippage[key] = maxSlippage
-        } else {
-          Object.keys(state.maxSlippage).forEach(k => (state.maxSlippage[k] = maxSlippage))
-        }
-      }),
-    )
-
-    return true
-  },
+  setMaxSlippage: maxSlippage => set(state => ({ ...state, maxSlippage: { ...maxSlippage } })),
```
0xAlunara: Why wouldn't `({ ...state, maxSlippage }),` suffice; you need a shallow clone?

packages/curve-ui-kit/src/shared/ui/ModalDialog.tsx
```diff
@@ -64,10 +65,22 @@ type ModalDialogProps = {
   /** Predefined width for the modal dialog */
   width?: keyof typeof modalWidth
 
+  /**
+   * When a form is used inside the modal, we need to wrap the contents of the modal with a form element,
+   * so that we can keep the form submission in the footer.
+   */
+  formProps?: Omit<FormHTMLAttributes<HTMLFormElement>, 'children'>
```
0xAlunara, line 72: I suppose we should update `CreateVoteModal.tsx` too, as there we still use the `FormProvider` inside the `ModalDialog` instead of the other way around. Not really in scope of this PR, feel free to modify if you need to work some more on this PR anyway

packages/curve-ui-kit/src/themes/components/index.ts
```diff
@@ -76,6 +76,8 @@ export const createComponents = (
   },
   MuiFormControlLabel: {
     styleOverrides: {
+      // MUI offsets radio padding with margins; inside tight grids those margins overflow.
```
0xAlunara: Comment talks about radio padding but doesn't this selector hold for literally every form control?

packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.tsx
```diff
@@ -13,13 +13,14 @@ import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
 import { type QueryProp } from '@ui-kit/types/util'
 import { formatPercent, getErrorMessage } from '@ui-kit/utils'
 import { getPriceImpactSeverity } from '@ui-kit/widgets/DetailPageLayout/price-impact.util'
+import type { SlippageType } from '@ui-kit/widgets/SlippageSettings'
 
 type FormErrors<Field extends string> = readonly (readonly [Field, string])[]
 
 type FormAlertProps<Field extends string> = {
-  error: Error | null
+  error?: Error | null
```
0xAlunara, line 22: We sure we want to make this property optional rather than `Error | null | undefined`?

0xAlunara (commented):
I don't know if it's me locally, but on the PR preview build if I try to swap ETH to anything I get hit with an error.

Image

Image

Image

apps/main/src/dex/components/PageRouterSwap/Page.tsx
DanielSchiavini: Good find, I had removed it in some commit that I discarded later on

packages/curve-ui-kit/src/features/user-profile/store.ts
DanielSchiavini: Indeed, not needed since we don't use SSR

apps/main/src/dex/components/PagePool/utils.ts
DanielSchiavini: It sounds stupid indeed, but it's actually a good idea :laughing:

packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.tsx
DanielSchiavini: yes in the slippage modal we just set the store, there is no exception handling or promise that can fail. I'll explain in  jsdoc

packages/curve-ui-kit/src/features/user-profile/store.ts
DanielSchiavini: I got some issues during development, but probably this was not the cause as I can't reproduce. I'll remove the cloning.

packages/curve-ui-kit/src/shared/ui/ModalDialog.tsx
DanielSchiavini: The change is super small, I'll just do it

DanielSchiavini:
> I don't know if it's me locally, but on the PR preview build if I try to swap ETH to anything I get hit with an error.

I have no idea why, but it worked with my own address. However, I could reproducing by using yours! Fixed in https://github.com/curvefi/curve-frontend/pull/2597/commits/562564739f8e77682c13c695ebf671310852c40c

apps/main/src/dex/components/PagePool/utils.ts
```diff
@@ -46,3 +49,6 @@ export const DEFAULT_SLIPPAGE: Slippage = {
 }
 
 export const DEFAULT_ESTIMATED_GAS: EstimatedGas = { loading: false, estimatedGas: null, error: null }
+
+export const getSlippageType = <T extends PoolData | undefined>(poolData: T) =>
```
0xAlunara, line 53: Is the generic needed? I don't know if there's any subtype of `PoolData` 🤔

packages/curve-ui-kit/src/features/user-profile/store.ts
```diff
@@ -1,85 +1,35 @@
...
-    ? window.matchMedia('(prefers-color-scheme: dark)').matches
-      ? 'dark'
-      : 'light'
-    : 'light'
-
 const INITIAL_STATE: UserProfileState = {
-  theme: INITIAL_THEME,
-  maxSlippage: { crypto: '0.1', stable: '0.03' },
+  theme: window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light',
+  maxSlippage: { crypto: '0.1', stable: '0.03', leverage: '0.5' },
```
0xAlunara: Shouldn't we use `SLIPPAGE` from `slippage.utils.ts` here?

DanielSchiavini: Kind of, that leads to a circular reference, let me see if I can fix that.

apps/main/src/dex/components/PagePool/utils.ts
DanielSchiavini: yes, if we type this as `(poolData: PoolData | undefined)` the return type will be `SlippageType | undefined`.
Currently, it is `const getSlippageType: <T extends PoolData | undefined>(poolData: T) => null extends T ? "stable" | "crypto" | "leverage" | undefined : undefined extends T ? "stable" | "crypto" | "leverage" | (T & undefined) : "stable" | "crypto" | "leverage"`

packages/curve-ui-kit/src/features/user-profile/store.ts
DanielSchiavini: it was fine, `type SlippageSettings` was already imported directly from `slippage.utils`, avoiding the barrel file.

0xtutti (dismissed):

0xAlunara (dismissed):

DanielSchiavini:
Sorry @0xAlunara @0xtutti, the test seemed to be flaky!

0xAlunara (approved):

0xtutti (approved):

PR #2598, opened 2026-05-27
feat: fetch ohlc data with react query

OnlyJousting:
Refactors existing OHLC chart data loading to use React Query instead of legacy store-managed state, while preserving the existing chart behavior.

Changes:
- Remove legacy OHLC Zustand slices and store actions.
- Move OHLC fetching, pagination state, and refetch/fetch-more handling into query-based hooks.
- Centralize candle-chart pagination, query, timestamp, and formatting helpers.
- Update chart components to consume query loading/error/refetch state directly.
- Use lightweight-charts autosize support and simplify chart resize handling.
- Keep existing historical range loading behavior while simplifying the data flow.
- Prevent transient candle-chart queries from being persisted.
- Extend existing prices-api clients to accept the range/interval parameters needed by the query hooks and bumps version to 3.1.1


packages/prices-api/src/ohlc/index.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+  low: 0.5,
+}
+
+const response = (data = [completeRow, incompleteRow]) =>
+  new Response(JSON.stringify({ chain: 'ethereum', address: '0x0000000000000000000000000000000000000001', data }), {
+    status: 200,
+    headers: { 'Content-Type': 'application/json' },
+  })
+
+const mockFetch = () => {
```
DanielSchiavini: It looks like something we should have in a shared file not directly in a test

packages/curve-ui-kit/src/features/candle-chart/hooks/useChartTimeSettings.ts
```diff
@@ -65,5 +46,5 @@ export const useChartTimeSettings = (
     return 'day'
```
DanielSchiavini: I'd expect a mapping { timeOption: timeUnit }
```suggestion
    return timeUnits[timeOption]
```

packages/curve-ui-kit/src/features/candle-chart/query-utils.ts
```diff
@@ -0,0 +1,259 @@
+import type { UTCTimestamp } from 'lightweight-charts'
+import { useCallback, useMemo, useRef } from 'react'
+import { toLocalTimestampSeconds } from '@primitives/timestamp.utils'
+import { type InfiniteData, type QueryKey, useInfiniteQuery } from '@tanstack/react-query'
+import { getOhlcPageStart, getPreviousOhlcPageEnd, hasFullOhlcPage } from './time-utils'
+import type { LpPriceOhlcDataFormatted, OraclePriceData, TimeOption } from './types'
+
+export const CANDLE_CHART_QUERY_GC_TIME = 10 * 60 * 1000
```
DanielSchiavini: We should be using the existing query categories

packages/curve-ui-kit/src/features/candle-chart/query-utils.ts
```diff
@@ -0,0 +1,259 @@
...
+export const CANDLE_CHART_QUERY_OPTIONS = {
+  staleTime: Infinity,
+  gcTime: CANDLE_CHART_QUERY_GC_TIME,
+  refetchOnMount: false,
```
DanielSchiavini: why are those configured manually?

packages/curve-ui-kit/src/features/candle-chart/query-utils.ts
```diff
@@ -0,0 +1,259 @@
...
+
+export const getOhlcPaginationState = (query: OhlcPaginationQuery) => ({
+  canFetchMore: canFetchMoreOhlc(query),
+  isFetchingMore: query.isFetchingNextPage,
+})
+
+export const fetchMoreOhlcQueries = (queries: readonly OhlcFetchMoreQuery[]) => {
+  const requests = queries.filter(canFetchMoreOhlc).map(query => query.fetchNextPage(FETCH_NEXT_PAGE_OPTIONS))
+
+  return requests.length > 0 ? Promise.all(requests) : undefined
```
DanielSchiavini: whats wrong with `promise.all([])`?

packages/curve-ui-kit/src/features/candle-chart/query-utils.ts
```diff
@@ -0,0 +1,259 @@
...
+
+type UseOhlcPagesAdapterParams<TPage, TData> = {
+  query: OhlcQuery<TPage>
+  selectData: (pages: TPage[] | undefined) => TData
+}
+
+const canFetchMoreOhlc = ({ hasNextPage, isFetchingNextPage, isSuccess }: OhlcPaginationQuery) =>
+  isSuccess && hasNextPage && !isFetchingNextPage
+
+// Historical scrolling can emit repeated fetch attempts before React Query's
```
DanielSchiavini: This should be jsdoc + I don't understand what it means 😆

packages/curve-ui-kit/src/features/candle-chart/query-utils.ts
```diff
@@ -0,0 +1,259 @@
...
+export const formatCandleOhlcData = (data: OhlcPoint[]): LpPriceOhlcDataFormatted[] =>
+  data.filter(hasCompleteOhlcValues).map(({ time, open, close, high, low }) => ({
+    time: toLocalTimestampSeconds(time) as UTCTimestamp,
+    open,
+    close,
+    high,
+    low,
+  }))
+
+export const formatOraclePriceData = (data: NullableOraclePoint[]): OraclePriceData[] =>
```
DanielSchiavini: some of the helpers in this file aren't really query-related. Since the file is getting large it would be nice to split it.

packages/curve-ui-kit/src/features/candle-chart/query-utils.ts
```diff
@@ -0,0 +1,259 @@
...
+  data.flatMap(({ time, oraclePrice }) =>
+    oraclePrice == null
```
DanielSchiavini: `notFalsy(oraclePrice && { ...})`

packages/curve-ui-kit/src/features/candle-chart/query-utils.ts
```diff
@@ -0,0 +1,259 @@
...
+
+export const flattenOhlcPages = <TPage, TItem>(pages: TPage[] | undefined, selectItems: (page: TPage) => TItem[]) =>
+  [...(pages ?? [])].reverse().flatMap(selectItems)
+
+export const applyLatestOraclePrice = (data: OraclePriceData[], oraclePrice: number | string | undefined) => {
+  if (oraclePrice == null || data.length === 0) return data
+
+  const latestValue = Number(oraclePrice)
+  const lastItem = data[data.length - 1]
+  if (!Number.isFinite(latestValue) || lastItem.value === latestValue) return data
```
DanielSchiavini: do we expect it to be infinite?

packages/curve-ui-kit/src/features/candle-chart/query-utils.ts
```diff
@@ -0,0 +1,259 @@
...
+    isFetchingNextPage,
+    isLoading,
+    isSuccess,
+    refetch: refetchQuery,
+  } = query
+  const pages = queryData?.pages
+  const data = useMemo(() => selectData(pages), [pages, selectData])
+  const paginationState = getOhlcPaginationState({ hasNextPage, isFetchingNextPage, isSuccess })
+
+  const refetch = useCallback(() => refetchQuery(), [refetchQuery])
```
DanielSchiavini: refetch is the same as refetchQuery

packages/curve-ui-kit/src/features/candle-chart/query-utils.ts
```diff
@@ -0,0 +1,259 @@
...
+
+  const fetchMore = useCallback(() => {
+    if (!paginationState.canFetchMore) return undefined
```
DanielSchiavini: shouldn't react-query get rid of it? I'd prefer to disable the button instead of ignoring clicks.

packages/curve-ui-kit/src/shared/ui/Chart/ChartError.tsx
```diff
@@ -6,12 +6,14 @@ import { ErrorMessage } from '@ui-kit/shared/ui/ErrorMessage'
...
   height,
+  error,
   errorMessage,
   refetchFunction,
 }: {
   height: number
+  error: Error
   errorMessage: string
-  refetchFunction?: () => void
+  refetchFunction?: () => Promise<unknown> | void
```
DanielSchiavini: We shouldn't need `Function` as a suffix 🙈

packages/curve-ui-kit/src/shared/ui/ErrorMessage.tsx
```diff
@@ -23,7 +23,7 @@ export const ErrorMessage = ({
   subtitle?: string
   error?: Error | string
   errorMessage: string
-  refreshData?: () => void
+  refreshData?: () => Promise<unknown> | void
```
DanielSchiavini, line 26: refreshData, refreshFunction, refreshQuery, refresh.. How many names do we need? 😁

packages/prices-api/src/ohlc/index.spec.ts
DanielSchiavini: NB I just found a duplicate in `packages/prices-api/src/llamma/index.spec.ts`

packages/curve-ui-kit/src/features/candle-chart/query-utils.spec.ts
```diff
@@ -0,0 +1,221 @@
...
+
+    expect(page).toEqual({
+      start: 1_674_166_400,
+      end: 1_700_000_000,
+    })
+  })
+
+  it('derives the next historical page from the returned oldest timestamp', () => {
+    vi.useFakeTimers()
+    vi.setSystemTime(new Date('2030-01-01T00:00:00.000Z'))
```
DanielSchiavini: is this going to break after 2030?

packages/curve-ui-kit/src/features/candle-chart/query-utils.spec.ts
```diff
@@ -0,0 +1,221 @@
...
+  formatCandleOhlcData,
+  formatOraclePriceData,
+  getOldestOhlcTimestampSeconds,
+  getNextOhlcPageParam,
+  getOhlcPaginationState,
+  refetchOhlcQueries,
+} from './query-utils'
+import type { OraclePriceData } from './types'
+
+describe('candle chart query utils', () => {
```
DanielSchiavini: these tests do not contribute to checking the app actually works tbh, it just makes it harder to refactor the functions. e2e tests would give us much better verification

packages/curve-ui-kit/src/features/candle-chart/index.ts
```diff
@@ -1,6 +1,24 @@
...
+  formatCandleOhlcData,
+  formatOraclePriceData,
+  getOldestOhlcTimestampSeconds,
+  getOhlcPaginationState,
+  refetchOhlcQueries,
+  useOhlcInfiniteQuery,
+  useOhlcPagesAdapter,
+  useOhlcQueryAdapter,
+  useStableOhlcAnchorEnd,
+} from './query-utils'
```
DanielSchiavini: Why do we need to export query-utils?

packages/curve-ui-kit/src/features/candle-chart/CandleChart.tsx
```diff
@@ -569,14 +517,20 @@ export const CandleChart = ({
...
   // Update oracle price data when it changes
   useEffect(() => {
     if (!oraclePriceSeriesRef.current || !oraclePriceData) return
 
     oraclePriceSeriesRef.current.setData(oraclePriceData)
-  }, [oraclePriceData])
+    // Oracle data drives historical pagination only for fallback-only charts.
+    // In candle charts, candle updates restore the viewport; restoring here can
+    // consume the pending range before the candle page arrives.
+    if (ohlcData.length === 0) {
```
DanielSchiavini: data shouldn't have length 0 during loading

packages/curve-ui-kit/src/features/candle-chart/CandleChart.tsx
```diff
@@ -75,6 +75,8 @@ type LiquidationRangeSeriesApi = ISeriesApi<
   LiquidationRangeSeriesOptions
 >
 
+const getAdjustedChartWidth = (width: number) => (width > 0 ? Math.max(1, width - 1) : undefined)
```
DanielSchiavini: why do we need to adjust the width at all?

packages/curve-ui-kit/src/features/candle-chart/hooks/useHistoricalChartPagination.ts
```diff
@@ -0,0 +1,235 @@
...
+
+export const useHistoricalChartPagination = ({
+  candlestickSeriesRef,
+  chartRef,
+  fetchMoreChartData,
+  ohlcData,
+  oraclePriceData,
+  oraclePriceSeriesRef,
+}: UseHistoricalChartPaginationParams) => {
+  const ohlcDataRef = useLatestValueRef(ohlcData)
```
DanielSchiavini, line 33: why do we need all these refs? They contain data from the previous render 😞 and appear to try to cheat the linter

packages/curve-ui-kit/src/features/candle-chart/hooks/useHistoricalChartPagination.ts
```diff
@@ -0,0 +1,235 @@
...
+      return
+    }
+
+    if (!chartRef.current || !update.visibleRangeToRestore) return
+
+    // Older history is prepended, which shifts logical bar indices. Restore
+    // the user's viewport after setData applies the new points; doing it
+    // before setData lets the prepend move the x-axis back in time.
+    chartRef.current.timeScale().setVisibleRange(update.visibleRangeToRestore)
+  }, [chartRef, ohlcDataRef, oraclePriceDataRef])
```
DanielSchiavini, line 72: this is basically never going to update

packages/curve-ui-kit/src/features/candle-chart/hooks/useHistoricalChartPagination.ts
```diff
@@ -0,0 +1,235 @@
...
+export const resolveHistoricalDataUpdate = ({
+  nextDataLength,
+  pendingVisibleRange,
+  previousDataLength,
+}: {
+  nextDataLength: number
+  pendingVisibleRange: VisibleTimeRange | null
+  previousDataLength: number
+}): HistoricalDataUpdate => {
+  if (nextDataLength < previousDataLength) {
```
DanielSchiavini: the returns in this function are almost the same, I think the difference would be much more clear by unifying into a single return.

packages/curve-ui-kit/src/features/candle-chart/hooks/useHistoricalChartPagination.ts
```diff
@@ -0,0 +1,235 @@
...
+
+const shouldFetchHistoricalPage = (barsInfo: { barsBefore: number } | null) =>
+  !!barsInfo && barsInfo.barsBefore < HISTORICAL_PAGE_THRESHOLD_BARS
+
+export const getHistoricalPaginationDataLength = (
+  ohlcData: LpPriceOhlcDataFormatted[],
+  oraclePriceData?: OraclePriceData[],
+) => (ohlcData.length > 0 ? ohlcData.length : (oraclePriceData?.length ?? 0))
+
+export const startHistoricalPageFetch = ({
```
DanielSchiavini: Why do we need this?

packages/curve-ui-kit/src/features/candle-chart/hooks/useHistoricalChartPagination.spec.ts
```diff
@@ -0,0 +1,132 @@
+import type { Time } from 'lightweight-charts'
+import { describe, expect, it, vi } from 'vitest'
+import { resolveHistoricalDataUpdate, startHistoricalPageFetch } from './useHistoricalChartPagination'
+
+const visibleRange = { from: 1 as Time, to: 2 as Time }
+
+describe('historical chart pagination helpers', () => {
```
DanielSchiavini: this file is the only usage of startHistoricalPageFetch, the whole thing is just useless

apps/main/src/llamalend/queries/ohlc-chart.query.ts
```diff
@@ -0,0 +1,197 @@
...
+  borrowedToken: TokenInfo
+  collateralToken: TokenInfo
+}
+
+export type LlammaOhlcPage = OhlcPageResult & {
+  page: OhlcPageParam
+  oraclePriceData: OraclePriceData[]
+}
+
+const EMPTY_TOKEN: TokenInfo = {
```
DanielSchiavini: please never use empty strings for representing missing data

apps/main/src/llamalend/hooks/useLlammaOhlcChartStateModel.ts
```diff
@@ -0,0 +1,163 @@
...
+  )
+
+  const newLiqPrices = previewPrices ?? legacyPreviewPrices
+
+  const { oraclePriceVisible, liqRangeCurrentVisible, liqRangeNewVisible, legendSets } = useChartLegendToggles({
+    hasNewLiquidationRange: !!newLiqPrices,
+    hasLiquidationRange: !!userPrices,
+    llammaEndpoint: shouldUseFallbackChart,
+  })
+  const shouldFetchFallbackOracleLine =
```
DanielSchiavini: why do we need a fallback here?

apps/main/src/llamalend/hooks/useLlammaOhlcChartStateModel.ts
```diff
@@ -0,0 +1,163 @@
...
+    shouldUseFallbackChart || (oraclePriceVisible && oraclePoolOraclePriceData.length === 0)
+  const neededHistoricalSelection = useMemo(
+    () => ({
+      oraclePool: !shouldUseFallbackChart,
+      llamma: shouldFetchFallbackOracleLine,
+    }),
+    [shouldFetchFallbackOracleLine, shouldUseFallbackChart],
+  )
+  const refetchPricesData = useCallback(() => refetch(neededHistoricalSelection), [neededHistoricalSelection, refetch])
+  const fetchMoreChartData = useCallback(
```
DanielSchiavini: useless useCallback

apps/main/src/llamalend/hooks/useLlammaOhlcChartStateModel.ts
```diff
@@ -0,0 +1,163 @@
...
+  oraclePrice,
+  previewPrices,
+  userPrices,
+}: LlammaOhlcChartStateModelParams) => {
+  const { timeOption, setTimeOption, chartInterval, timeUnit } = useChartTimeSettings()
+  const anchorKey = `${chainKey}:${marketId}:${controllerAddress}:${llammaAddress}:${timeOption}`
+  const anchorEnd = useStableOhlcAnchorEnd(anchorKey)
+
+  const {
+    oraclePool: {
```
DanielSchiavini: we should use Query<> type

apps/main/src/llamalend/hooks/useLlammaOhlcChartStateModel.ts
```diff
@@ -0,0 +1,163 @@
...
+  legacyPreviewPrices,
+  llammaAddress,
+  marketId,
+  network,
+  oraclePrice,
+  previewPrices,
+  userPrices,
+}: LlammaOhlcChartStateModelParams) => {
+  const { timeOption, setTimeOption, chartInterval, timeUnit } = useChartTimeSettings()
+  const anchorKey = `${chainKey}:${marketId}:${controllerAddress}:${llammaAddress}:${timeOption}`
```
DanielSchiavini: why do we need anchor keys?

apps/main/src/llamalend/hooks/useLlammaOhlcChartStateModel.ts
```diff
@@ -0,0 +1,163 @@
...
+    oraclePool: {
+      error: oraclePoolError,
+      isLoading: oraclePoolIsLoading,
+      ohlcData: oraclePoolData,
+      oraclePriceData: oraclePoolOraclePriceData,
+    },
+    llammaFallback: {
+      error: llammaFallbackError,
+      isLoading: llammaFallbackIsLoading,
+      oraclePriceData: llammaFallbackOraclePriceData,
```
DanielSchiavini: We should be using the Query<> type
```suggestion
      data: llammaFallbackOraclePriceData,
```

apps/main/src/llamalend/hooks/useLlammaOhlcChartData.ts
```diff
@@ -0,0 +1,147 @@
...
+  )
+
+  return {
+    fetchMore,
+    llammaFallback: {
+      error: rawLlammaOraclePrice.error,
+      isLoading: rawLlammaOraclePrice.isLoading,
+      oraclePriceData: llammaOraclePriceData,
+    },
+    oraclePool: {
```
DanielSchiavini: we should be using Query<> type

apps/main/src/llamalend/hooks/useLlammaOhlcChartData.ts
```diff
@@ -0,0 +1,147 @@
...
+    (selection: HistoricalSelection) => fetchMoreOhlcQueries(selectHistoricalQueries(selection)),
+    [selectHistoricalQueries],
+  )
+
+  return {
+    fetchMore,
+    llammaFallback: {
+      error: rawLlammaOraclePrice.error,
+      isLoading: rawLlammaOraclePrice.isLoading,
+      oraclePriceData: llammaOraclePriceData,
```
DanielSchiavini:
```suggestion
      data: llammaOraclePriceData,
```

apps/main/src/llamalend/hooks/useLlammaOhlcChartStateModel.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  // Oracle-pool is primary because it has the richest candle data. LLAMMA is
+  // only needed as an oracle-line fallback when oracle-pool does not supply
+  // oracle price data.
+  const hasOraclePoolData = oraclePoolData.length > 0 || oraclePoolOraclePriceData.length > 0
+  const hasFallbackOracleLine = llammaFallbackOraclePriceData.length > 0
+  const shouldUseFallbackChart = !hasOraclePoolData && hasFallbackOracleLine
+  const hasChartData = hasOraclePoolData || hasFallbackOracleLine
+
+  const isLoading = oraclePoolIsLoading || (!hasOraclePoolData && llammaFallbackIsLoading)
```
DanielSchiavini: We should not put `isLoading` to `true` when data has 0 length

apps/main/src/llamalend/hooks/useLlammaOhlcChartStateModel.ts
```diff
@@ -0,0 +1,163 @@
...
+  const selectedChartKey = isLoading ? undefined : shouldUseFallbackChart ? 'llamma' : 'oracle'
+  const currentError = hasChartData ? null : (llammaFallbackError ?? oraclePoolError)
+  const noDataAvailable = enabled && !isLoading && !currentError && !hasChartData
+
+  const ohlcData = shouldUseFallbackChart ? [] : oraclePoolData
+  const selectedOraclePriceData =
+    oraclePoolOraclePriceData.length > 0 ? oraclePoolOraclePriceData : llammaFallbackOraclePriceData
+  const oraclePriceData = selectedOraclePriceData.length > 0 ? selectedOraclePriceData : undefined
+  const currentOraclePriceData = shouldUseFallbackChart ? llammaFallbackOraclePriceData : oraclePoolOraclePriceData
```
DanielSchiavini: These lines are very hard to review, there seem to be some fallback logic for some reason, lots of data selection based on other checks, quite hard to read and understand why we need the complexity

apps/main/src/llamalend/hooks/useLlammaOhlcChartData.ts
```diff
@@ -0,0 +1,147 @@
...
+    [oracleTokenPage],
+  )
+  const selectHistoricalQueries = useCallback(
+    (selection: HistoricalSelection) =>
+      [
+        { selected: selection.oraclePool, query: oraclePoolQuery },
+        { selected: selection.llamma && shouldFetchLlammaQuery, query: llammaQuery },
+      ]
+        .filter(({ selected }) => selected)
+        .map(({ query }) => query),
```
DanielSchiavini:
```suggestion
      notFalsy(
        oraclePool && oraclePoolQuery,
        llamma && shouldFetchLlammaQuery && llammaQuery,
      )
```

apps/main/src/llamalend/hooks/useLlammaOhlcChartData.ts
```diff
@@ -0,0 +1,147 @@
...
+    anchorEnd,
+    enabled,
+  })
+  const oraclePool = useOhlcPagesAdapter({
+    query: oraclePoolQuery,
+    selectData: selectOraclePoolChartData,
+  })
+  const oraclePoolIsSettled = oraclePoolQuery.isSuccess || oraclePoolQuery.isError
+  const oraclePoolHasOraclePriceData = oraclePool.data.oraclePriceData.length > 0
+  // Oracle-pool is primary because it has the richest candle data. LLAMMA is
```
DanielSchiavini: duplicated comment that I don't understand 😅

apps/main/src/dex/queries/ohlc-chart.query.ts
```diff
@@ -0,0 +1,139 @@
...
+  units: 'minute' | 'hour' | 'day'
+}
+
+type DexOhlcPage = OhlcPageResult & {
+  page: OhlcPageParam
+  ohlcData: LpPriceOhlcDataFormatted[]
+}
+
+export const getDexChartSelectionKey = (chartSelection: ChartSelection) => {
+  if (chartSelection.type !== 'pair') return chartSelection.type
```
DanielSchiavini: double negative

apps/main/src/dex/queries/ohlc-chart.query.ts
```diff
@@ -0,0 +1,139 @@
...
+export const useDexOhlcQuery = ({
+  anchorEnd,
+  chain,
+  chartSelection,
+  interval,
+  poolAddress,
+  timeOption,
+  units,
+}: DexOhlcQueryParams) => {
+  const chartSelectionKey = useMemo(() => getDexChartSelectionKey(chartSelection), [chartSelection])
```
DanielSchiavini: unnecessary useMemo, getDexChartSelectionKey is too simple to deserve it

apps/main/src/llamalend/hooks/useLlammaOhlcChartData.ts
OnlyJousting: I'll deduplicate and see if I can explain better

apps/main/src/llamalend/hooks/useLlammaOhlcChartStateModel.ts
OnlyJousting: roger that, will try to make it easier to follow

apps/main/src/llamalend/hooks/useLlammaOhlcChartStateModel.ts
OnlyJousting: sometimes the main data source (endpoint serving data from "oracle pools") doesn't have oracle price data, then we use the fallback which is data derived from the actual market/llamma pool. for markets like that we're most likely only showing the oracle price line

apps/main/src/llamalend/hooks/useLlammaOhlcChartStateModel.ts
OnlyJousting: to make sure anchorEnd is stable, I updated it to a useMemo so it's a little easier to understand

packages/curve-ui-kit/src/features/candle-chart/CandleChart.tsx
OnlyJousting: it's not a loading check but I'll make it a little more clear

packages/curve-ui-kit/src/features/candle-chart/query-utils.ts
OnlyJousting: there's not a button, fetching is triggered on scroll events inside the chart

packages/curve-ui-kit/src/features/candle-chart/query-utils.ts
OnlyJousting: it's fine tbh, it allows some simplification too so will allow an empty array to be passed instead of undefined

apps/main/src/llamalend/hooks/useLlammaOhlcChartStateModel.ts
OnlyJousting: it's a special scenario where the primary data source (oracle pool data) is empty so we enable and wait for the fallback data source to load. we don't want to disable the loading state just to pick it up again. I will create a variable that makes this a little easier to understand (I hope)

packages/curve-ui-kit/src/features/candle-chart/query-utils.spec.ts
OnlyJousting: yea they are/were not intended to be left when the PR is ready to merge

packages/curve-ui-kit/src/features/candle-chart/hooks/useHistoricalChartPagination.ts
OnlyJousting: we do not

packages/curve-ui-kit/src/shared/ui/ErrorMessage.tsx
OnlyJousting: 😅 I'll settle on refreshData then

packages/curve-ui-kit/src/features/candle-chart/hooks/useHistoricalChartPagination.ts
OnlyJousting: lightweight-charts is an imperative canvas API so all the instances, callbacks live outside reacts render model. we need refs all over the candle-chart to keep the data and our callbacks stable

packages/curve-ui-kit/src/features/candle-chart/hooks/useHistoricalChartPagination.ts
OnlyJousting: it updates on data changes inside the refs, flow:

1. User scrolls near the left edge of the chart.
2. handleVisibleLogicalRangeChange stores the current visible range in pendingVisibleRangeRef and calls fetchMoreChartData.
3. React Query loads/prepends older data.
4. ohlcData or oraclePriceData changes in CandleChart.
5. The setData(...) effect in CandleChart runs and calls restoreVisibleRangeAfterDataUpdate.
6. That callback reads the latest data via ohlcDataRef.current / oraclePriceDataRef.current, compares the new length to the previous length, and restores the saved range.

packages/curve-ui-kit/src/features/candle-chart/CandleChart.tsx
OnlyJousting: we don't anymore 🥳

packages/curve-ui-kit/vitest.config.ts
```diff
@@ -0,0 +1,20 @@
+import { dirname, resolve } from 'node:path'
+import { fileURLToPath } from 'node:url'
+import { defineConfig } from 'vitest/config'
+
+const rootDir = dirname(fileURLToPath(import.meta.url))
+
+export default defineConfig({
```
0xAlunara, line 7: What's this file for? I can't find any newly added spec files here

packages/curve-ui-kit/src/lib/api/provider.tsx
```diff
@@ -8,9 +8,14 @@ type QueryProviderWrapperProps = {
   queryClient: QueryClient
 }
 
+const shouldDehydrateQuery = (query: Query) => query.state.status === 'success' && query.meta?.persist !== false
```
0xAlunara, line 11: Just curious what this is actually for. Is it do manually disabling caching of certain queries?

EDIT: ah yeah

packages/curve-ui-kit/src/features/candle-chart/utils.ts
```diff
@@ -1,24 +1,70 @@
...
+
+type OhlcPoint = {
+  time: number
+  open?: number | null
+  close?: number | null
+  high?: number | null
+  low?: number | null
+}
+
+type CompleteOhlcPoint = {
```
0xAlunara:
```suggestion
`type CompleteOhlcPoint = Required<OhlcPoint>`
```

packages/curve-ui-kit/src/features/candle-chart/utils.ts
```diff
@@ -1,24 +1,70 @@
...
+type NullableOraclePoint = {
+  time: number
+  oraclePrice?: number | null
+}
+
+const hasCompleteOhlcValues = (data: OhlcPoint): data is CompleteOhlcPoint =>
+  data.close != null && data.high != null && data.low != null && data.open != null
+
+export const formatCandleOhlcData = (data: OhlcPoint[]): LpPriceOhlcDataFormatted[] =>
+  data.filter(hasCompleteOhlcValues).map(({ time, open, close, high, low }) => ({
```
0xAlunara:
```suggestion
  data.filter(hasCompleteOhlcValues).map(({ time, ...ohlc }) => ({ time: toLocalTimestampSeconds(time) as UTCTimestamp, ...ohlc })
```

packages/curve-ui-kit/src/features/candle-chart/utils.ts
```diff
@@ -1,24 +1,70 @@
...
+export const applyLatestOraclePrice = (data: OraclePriceData[], oraclePrice: number | undefined) => {
+  if (oraclePrice == null || data.length === 0) return data
+
+  const lastItem = data[data.length - 1]
+  if (lastItem.value === oraclePrice) return data
+
+  return [...data.slice(0, -1), { ...lastItem, value: oraclePrice }]
+}
+
+export const flattenOhlcPages = <TPage, TItem>(pages: TPage[] | undefined, selectItems: (page: TPage) => TItem[]) =>
```
0xAlunara: It says flatten but it also reverses. Given that we're only using it in one place we might as well inline it so we don't have to bother with both naming and typing?

EDIT: ah shit typescript lied to me it's used in 2 places 😭

packages/curve-ui-kit/src/features/candle-chart/utils.ts
```diff
@@ -1,24 +1,70 @@
...
+  data.flatMap(({ time, oraclePrice }) =>
+    notFalsy(
+      oraclePrice != null && {
+        time: toLocalTimestampSeconds(time) as UTCTimestamp,
+        value: oraclePrice,
+      },
+    ),
+  )
+
+export const applyLatestOraclePrice = (data: OraclePriceData[], oraclePrice: number | undefined) => {
```
0xAlunara, line 42:
```suggestion
export const applyLatestOraclePrice = (data: OraclePriceData[], oraclePrice: number | undefined) =>
  oraclePrice == null || !data.length || data.at(-1).value === oraclePrice
    ? data
    : [...data.slice(0, -1), { ...data.at(-1), value: oraclePrice }]
```

packages/curve-ui-kit/src/features/candle-chart/CandleChart.tsx
```diff
@@ -33,30 +34,34 @@ const normalizeLiquidationRangePoints = (range?: LlammaLiquididationRange | null
...
 
-  if (!orderedEntries.length) {
+  if (!orderedPoints.length) {
     return []
   }
 
-  const fallbackStart = orderedEntries[0][0] as Time
-  const fallbackEnd = orderedEntries[orderedEntries.length - 1][0] as Time
+  const fallbackStart = orderedPoints[0].time as Time
+  const fallbackEnd = orderedPoints[orderedPoints.length - 1].time as Time
```
0xAlunara: `.at(-1)`

packages/curve-ui-kit/src/features/candle-chart/query-utils.ts
```diff
@@ -0,0 +1,222 @@
...
+ *    request.
+ *
+ * TanStack Query defaults `cancelRefetch` to true, which would let those
+ * duplicate calls cancel/restart the active historical page request. Setting it
+ * to false makes duplicate calls during an active page fetch reuse the existing
+ * request instead.
+ */
+const FETCH_NEXT_PAGE_OPTIONS = { cancelRefetch: false } as const
+
+export const useOhlcInfiniteQuery = <TPage extends OhlcPageResult, TQueryKey extends QueryKey>({
```
0xAlunara: shouldn't this be in a separate hook file inside the hooks folder?

OnlyJousting: yea that makes sense!

packages/curve-ui-kit/src/features/candle-chart/utils.ts
OnlyJousting: we can't accept null values here but I can just use the lightweight charts type directly instead

packages/curve-ui-kit/src/features/candle-chart/utils.ts
OnlyJousting: will update the name anyways

packages/curve-ui-kit/vitest.config.ts
OnlyJousting: ah shit, forgot to clean it up, I let codex have it's test spec files while working but got rid of them but forgot this´

Pearce542: still here 😄

packages/curve-ui-kit/src/features/candle-chart/hooks/useOhlcQueries.ts
```diff
@@ -0,0 +1,148 @@
...
+  const pages = query.data?.pages
+  const data = useMemo(() => selectData(pages), [pages, selectData])
+
+  return useOhlcAdapter({ query, data })
+}
+
+export const useOhlcQueryAdapter = <TPage, TItem>({ query, selectItems }: UseOhlcQueryAdapterParams<TPage, TItem>) =>
+  useOhlcPagesAdapter({ query, selectData: pages => flattenOhlcPagesChronologically(pages, selectItems) })
+
+const createOhlcAnchorEnd = (_resetKey: string) => Math.floor(Date.now() / 1000)
```
DanielSchiavini: This function is not needed, also we shouldn't use Date.now in a hook. See [useCurrentDate](https://github.com/curvefi/curve-frontend/blob/c68943fc5283e7902d45c5eec7673391e41c7ac6/packages/curve-ui-kit/src/hooks/useCurrentDate.ts#L5)

apps/main/src/dex/components/OhlcAndActivityComp/hooks/useOhlcChartState.ts
```diff
@@ -15,86 +17,54 @@ type UseOhlcChartStateArgs = {
...
-        chartInterval,
-        timeUnit,
-        +startTime,
-        endTime,
-      )
-    },
-    [chartInterval, fetchMorePricesApiCharts, pricesApiPoolData.address, rChainId, selectedChart, timeOption, timeUnit],
+  const chartSelectionKey = getDexChartSelectionKey(selectedChart)
+  const anchorEnd = useStableOhlcAnchorEnd(
+    `${rChainId}:${pricesApiPoolData.address}:${chartSelectionKey}:${timeOption}`,
```
DanielSchiavini: I think the hook should be responsible for the string concatenation.
Do we really want to only reset this on load? It would be nice to update every x minutes

apps/main/src/dex/queries/ohlc-chart.query.ts
```diff
@@ -0,0 +1,140 @@
...
+      { poolAddress },
+      { chartSelectionKey },
+      { interval },
+      { units },
+      { timeOption },
+      { anchorEnd },
+    ),
+    anchorEnd,
+    timeOption,
+    enabled: !!chain && !!poolAddress,
```
DanielSchiavini: Shall we create a task to integrate this properly with the existing queryFactory and validation?

apps/main/src/dex/queries/ohlc-chart.query.ts
```diff
@@ -0,0 +1,140 @@
...
+      },
+      { signal },
+    )
+  }
+
+  return getLpOHLC(
+    {
+      chain,
+      poolAddress,
+      priceUnits: chartSelection.type === 'lp-usd' ? 'usd' : 'token0',
```
DanielSchiavini: I'm not sure what this is doing. Should we use a mapping and/or constants?

apps/main/src/dex/queries/ohlc-chart.query.ts
```diff
@@ -0,0 +1,140 @@
...
+          end: page.end,
+        },
+        signal,
+      )
+      const ohlcData = formatCandleOhlcData(responseData)
+
+      assertInitialOhlcPageHasData({
+        anchorEnd,
+        dataLength: ohlcData.length,
+        message: 'No OHLC data found. Data may be unavailable for this pool.',
```
DanielSchiavini: It's not clear why we are doing assertions on data length, shouldn't this simply be an empty state instead of an error?

apps/main/src/llamalend/hooks/useLlammaOhlcChartData.ts
```diff
@@ -0,0 +1,134 @@
...
+    timeOption,
+    units,
+    anchorEnd,
+    enabled,
+  })
+  const oraclePoolsChartAdapter = useOhlcPagesAdapter({
+    query: oraclePoolQuery,
+    selectData: selectOraclePoolChartData,
+  })
+  const oraclePoolIsSettled = oraclePoolQuery.isSuccess || oraclePoolQuery.isError
```
DanielSchiavini, line 74: I'd be nice to avoid repeating this elsewhere
```suggestion
  const oraclePoolIsSettled = isSettled(oraclePoolQuery)
```

apps/main/src/llamalend/hooks/useLlammaOhlcChartData.ts
```diff
@@ -0,0 +1,134 @@
...
+  )
+  const oracleTokens = useMemo(() => {
+    const { collateralToken, borrowedToken } = oracleTokenPage ?? {}
+    return collateralToken && borrowedToken
+      ? {
+          collateralSymbol: collateralToken.symbol,
+          borrowedSymbol: borrowedToken.symbol,
+        }
+      : null
+  }, [oracleTokenPage])
```
DanielSchiavini:
```suggestion
  const oracleTokens = useMemo(() => maybe(oracleTokenPage, ({ collateralToken, borrowedToken }) =>
    ({ collateralSymbol: collateralToken.symbol, borrowedSymbol: borrowedToken.symbol })
  ), [oracleTokenPage])
```

apps/main/src/llamalend/hooks/useLlammaOhlcChartData.ts
```diff
@@ -0,0 +1,134 @@
...
+  const selectHistoricalQueries = useCallback(
+    (selection: HistoricalSelection) =>
+      notFalsy(selection.oraclePool && oraclePoolQuery, selection.llamma && shouldFetchLlammaQuery && llammaQuery),
+    [llammaQuery, oraclePoolQuery, shouldFetchLlammaQuery],
```
DanielSchiavini: depending on whole queries is quite useless usually, as they can change every render

apps/main/src/llamalend/hooks/useLlammaOhlcChartStateModel.ts
```diff
@@ -0,0 +1,164 @@
...
+  marketId,
+  network,
+  oraclePrice,
+  previewPrices,
+  userPrices,
+}: LlammaOhlcChartStateModelParams) => {
+  const { timeOption, setTimeOption, chartInterval, timeUnit } = useChartTimeSettings()
+  const anchorKey = `${chainKey}:${marketId}:${timeOption}`
+  const anchorEnd = useStableOhlcAnchorEnd(anchorKey)
+  const latestOraclePrice = oraclePrice ? Number(oraclePrice) : undefined
```
DanielSchiavini: Can we only convert to Number on the latest needed moment possible? And keep it Decimal as long as we can

apps/main/src/llamalend/queries/ohlc-chart.query.ts
```diff
@@ -0,0 +1,198 @@
...
+    ...(borrowedPool
+      ? {
+          borrowedToken: {
+            address: borrowedPool.borrowedAddress,
+            symbol: borrowedPool.borrowedSymbol,
+          },
+        }
+      : {}),
+  }
+}
```
DanielSchiavini, line 67: `{...undefined}` is totally fine!
```suggestion
  return {
    ...maybe(collateralPool, ({ collateralAddress, collateralSymbol }) =>
      ({ collateralToken: { address: collateralAddress, symbol: collateralSymbol }
    }),
    ...maybe(borrowedPool, ({borrowedAddress, borrowedSymbol}) => ({
        borrowedToken: { address: borrowedAddress, symbol: borrowedSymbol }
      }))
  }
}
```

apps/main/src/llamalend/queries/ohlc-chart.query.ts
```diff
@@ -0,0 +1,198 @@
...
+      { units },
+      { timeOption },
+      { anchorEnd },
+    ),
+    anchorEnd,
+    timeOption,
+    enabled: enabled && !!chain && !!controller,
+    fetchPage: async ({ pageParam, signal }): Promise<OraclePoolOhlcPage> => {
+      if (!chain) {
+        throw new Error('Cannot fetch oracle-pool OHLC data without a chain.')
```
DanielSchiavini:
```typescript
import { assert } from '@primitives/objects.utils'
```

apps/main/src/llamalend/queries/ohlc-chart.query.ts
```diff
@@ -0,0 +1,198 @@
...
+          start: page.start,
+          end: page.end,
+        },
+        { signal },
+      )
+
+      assertInitialOhlcPageHasData({
+        anchorEnd,
+        dataLength: data.length,
+        message: 'No oracle OHLC data found. Data may be unavailable for this pool.',
```
DanielSchiavini: Same question as above

apps/main/src/llamalend/queries/ohlc-chart.query.ts
```diff
@@ -0,0 +1,198 @@
...
+    ),
+    anchorEnd,
+    timeOption,
+    enabled: enabled && !!chain && !!controller,
+    fetchPage: async ({ pageParam, signal }): Promise<OraclePoolOhlcPage> => {
+      if (!chain) {
+        throw new Error('Cannot fetch oracle-pool OHLC data without a chain.')
+      }
+
+      const page = pageParam
```
DanielSchiavini: why do we have page and pageParam if they are the same?

packages/curve-ui-kit/src/features/candle-chart/hooks/useHistoricalChartPagination.ts
```diff
@@ -0,0 +1,150 @@
...
+const HISTORICAL_PAGE_THRESHOLD_BARS = 50
+const HISTORICAL_FETCH_DEBOUNCE_MS = 500
+
+const shouldFetchHistoricalPage = (barsInfo: { barsBefore: number } | null) =>
+  !!barsInfo && barsInfo.barsBefore < HISTORICAL_PAGE_THRESHOLD_BARS
+
+const getHistoricalPaginationDataLength = (
+  ohlcData: LpPriceOhlcDataFormatted[],
+  oraclePriceData?: OraclePriceData[],
+) => (ohlcData.length > 0 ? ohlcData.length : (oraclePriceData?.length ?? 0))
```
DanielSchiavini:
```suggestion
) => (ohlcData.length || (oraclePriceData?.length ?? 0))
```

packages/curve-ui-kit/src/features/candle-chart/hooks/useHistoricalChartPagination.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  const fetchHistoricalPage = useCallback(() => {
+    if (historicalFetchInFlightRef.current) return
+
+    historicalFetchInFlightRef.current = true
+
+    let request: Promise<unknown>
+    try {
+      request = fetchMoreChartDataRef.current()
+    } catch {
```
DanielSchiavini: 💀 please never use empty catch
why do we even expect this to throw?

packages/curve-ui-kit/src/features/candle-chart/hooks/useHistoricalChartPagination.ts
```diff
@@ -0,0 +1,150 @@
...
+    void request
+      .catch(() => {
+        pendingVisibleRangeRef.current = null
+      })
+      .finally(() => {
+        historicalFetchInFlightRef.current = false
+      })
+  }, [fetchMoreChartDataRef])
+
+  const debouncedFetchHistoricalPage = useMemo(
```
DanielSchiavini: [`'@ui-kit/hooks/useDebounce'`](https://github.com/curvefi/curve-frontend/blob/424563c5f0da8b24e0fce345c55573b3baea4797/packages/curve-ui-kit/src/hooks/useDebounce.ts) pls

packages/curve-ui-kit/src/features/candle-chart/hooks/useHistoricalChartPagination.ts
```diff
@@ -0,0 +1,150 @@
...
+  fetchMoreChartData,
+  ohlcData,
+  oraclePriceData,
+  oraclePriceSeriesRef,
+}: UseHistoricalChartPaginationParams) => {
+  const ohlcDataRef = useLatestValueRef(ohlcData)
+  const oraclePriceDataRef = useLatestValueRef(oraclePriceData)
+  const fetchMoreChartDataRef = useLatestValueRef(fetchMoreChartData)
+  // lightweight-charts emits many imperative range events during a drag. Refs
+  // keep that hot path out of React render state while still guarding the
```
DanielSchiavini, line 37: I'm not sure I believe this statement, are the refs really necessary @OnlyJousting ??
This file is extremely hard to review

packages/curve-ui-kit/src/features/candle-chart/hooks/useOhlcQueries.ts
```diff
@@ -0,0 +1,148 @@
...
+export const refetchOhlcQueries = (queries: readonly { refetch: () => Promise<unknown> }[]) =>
+  Promise.all(queries.map(query => query.refetch()))
+
+const useOhlcAdapter = <TPage, TData>({ query, data }: UseOhlcAdapterParams<TPage, TData>) => {
+  const { error, fetchNextPage, hasNextPage, isError, isFetchingNextPage, isLoading, isSuccess, refetch } = query
+  const canFetchMore = canFetchMoreOhlc({ hasNextPage, isFetchingNextPage, isSuccess })
+  const isFetchingMore = isFetchingNextPage
+
+  const fetchMore = useCallback(
+    () => Promise.all(canFetchMore ? [fetchNextPage(FETCH_NEXT_PAGE_OPTIONS)] : []),
```
DanielSchiavini:
```suggestion
    () => Promise.all(notFalsy(canFetchMore && fetchNextPage(FETCH_NEXT_PAGE_OPTIONS))),
```

packages/curve-ui-kit/src/features/candle-chart/hooks/useOhlcQueries.ts
```diff
@@ -0,0 +1,148 @@
...
+    [canFetchMore, fetchNextPage],
+  )
+
+  return {
+    canFetchMore,
+    data,
+    error: isError ? error : null,
```
DanielSchiavini:
```suggestion
    error,
```

packages/curve-ui-kit/src/features/candle-chart/CandleChart.tsx
```diff
@@ -33,30 +34,34 @@ const normalizeLiquidationRangePoints = (range?: LlammaLiquididationRange | null
...
-    .sort((a, b) => a[0] - b[0])
+  const orderedPoints = sortBy(
+    Array.from(pointMap, ([time, { upper, lower }]) =>
+      typeof upper === 'number' && typeof lower === 'number'
+        ? {
+            time,
+            upper: Math.max(upper, lower),
+            lower: Math.min(upper, lower),
+          }
+        : null,
```
DanielSchiavini:
```suggestion
      maybes(([upper, lower]), points => ({ time, upper: Math.max(...points), lower: Math.min(...points) })) ?? null,
```

packages/curve-ui-kit/src/features/candle-chart/CandleChart.tsx
```diff
@@ -550,35 +481,51 @@ export const CandleChart = ({
     candlestickSeriesRef.current.setData(ohlcData)
     candlestickSeriesRef.current.applyOptions({ priceFormat })
     chartRef.current.applyOptions({ localization: { priceFormatter: priceFormat.formatter } })
-  }, [ohlcData])
+    restoreVisibleRangeAfterDataUpdate()
+  }, [ohlcData, restoreVisibleRangeAfterDataUpdate])
 
-  // Apply initial right-side spacing (10% of chart width) only on first data load between chart data and y-axis price scale
-  // This effect runs when wrapperWidth changes to ensure chart is rendered with proper dimensions
+  // Apply initial right-side spacing once after the chart has data and a non-zero rendered width.
```
DanielSchiavini: this doesn't say why the code is necessary at all. Given the comment is needed, the code probably belongs in a separate hook/function

packages/curve-ui-kit/src/features/candle-chart/CandleChart.tsx
```diff
@@ -550,35 +481,51 @@ export const CandleChart = ({
...
-  }, [oraclePriceData])
+    // Oracle data drives historical pagination only for fallback-only charts.
+    // In candle charts, candle updates restore the viewport; restoring here can
+    // consume the pending range before the candle page arrives.
+    const isOracleOnlyChart = ohlcData.length === 0
+
+    if (isOracleOnlyChart) {
+      restoreVisibleRangeAfterDataUpdate()
+    }
+  }, [ohlcData.length, oraclePriceData, restoreVisibleRangeAfterDataUpdate])
```
DanielSchiavini, line 503: is depending on .length safe here? What if the new data has the same length?

packages/curve-ui-kit/src/features/candle-chart/CandleChart.tsx
```diff
@@ -550,35 +481,51 @@ export const CandleChart = ({
...
 
   // Update oracle price data when it changes
   useEffect(() => {
     if (!oraclePriceSeriesRef.current || !oraclePriceData) return
 
     oraclePriceSeriesRef.current.setData(oraclePriceData)
-  }, [oraclePriceData])
+    // Oracle data drives historical pagination only for fallback-only charts.
+    // In candle charts, candle updates restore the viewport; restoring here can
+    // consume the pending range before the candle page arrives.
```
DanielSchiavini, line 497: what do you mean with consuming the pending range? Is length==0 the right check here? If the data is loading I expect `ohlcData==undefined` instead of `[]`

packages/curve-ui-kit/src/features/candle-chart/query-utils.ts
```diff
@@ -0,0 +1,73 @@
...
+  page,
+}: {
+  anchorEnd: number
+  dataLength: number
+  message: string
+  page: OhlcPageParam
+}) => {
+  if (dataLength === 0 && page.end === anchorEnd) {
+    throw new Error(message)
+  }
```
DanielSchiavini:
```suggestion
  assert(!dataLength && page.end === anchorEnd, message)
```

apps/main/src/dex/components/OhlcAndActivityComp/hooks/useOhlcChartState.ts
OnlyJousting: it becomes really messy if a user is a few pages in since each subsequent page is depending on the previous, it would require some extra logic to keep the users scroll position on the x-axis and each page would need to be updated to keep the data intact. it would be nice to have reload so the first page (latest data) get rolling updates but I didn't want to start digging into that complexity for this refactor

apps/main/src/llamalend/hooks/useLlammaOhlcChartData.ts
OnlyJousting: oooh, I forgot about `maybe`!

apps/main/src/dex/queries/ohlc-chart.query.ts
OnlyJousting: it sets what the chart is denominated in (usd or token0 in in the array of the pools tokens). I can add a constant

packages/curve-ui-kit/src/features/candle-chart/hooks/useOhlcQueries.ts
OnlyJousting: hm.. the function isn't needed but it's important that Date.now() only gets calculated once or when the anchor key is changed (new query) so we don't get refetches that will cause the chart to re-render a new data set. is there a bigger reason to not set Date.now in a hook that I'm missing?

packages/curve-ui-kit/src/features/candle-chart/query-utils.ts
OnlyJousting: I'll use assert but have to modify it a little bit to maintain the behaviour

packages/curve-ui-kit/src/features/candle-chart/hooks/useHistoricalChartPagination.ts
OnlyJousting: yea unfortunately, here are the [docs](https://tradingview.github.io/lightweight-charts/tutorials/react/advanced) if you want to take a look

apps/main/src/llamalend/hooks/useLlammaOhlcChartData.ts
OnlyJousting: are we repeating this anywhere else? this only happens here because we have to use a fallback query on occasion and don't want to needlesly fetch the fallback when it's not needed

apps/main/src/llamalend/queries/ohlc-chart.query.ts
OnlyJousting: good point, we don't even need to return page(pageParam) anymore either!

packages/curve-ui-kit/src/features/candle-chart/CandleChart.tsx
OnlyJousting: there is only one source of ohlcData, the fallback only provides oracle price data (used for price line).

Three scenarios:
the ohlc (oracle pool) endpoint provides both ohlc and oracle price data
the ohlc (oracle pool) endpoint provides ohlc data but has no oracle price data, llamma fallback endpoint will provide oracle price data
the ohlc (oracle pool) endpoint has no data/doesn't exist, the chart will only show the oracle price data from the fallback llamma endpoint

restore visible range is recorded before fetching more data and restore after updating the chart data. when you update the chart data in lightweight-charts the chart resets the users scroll position. this makes it important to determine WHEN we restore based on the three scenarios above

apps/main/src/dex/queries/ohlc-chart.query.ts
OnlyJousting: yea you're right about that actually!

apps/main/src/llamalend/hooks/useLlammaOhlcChartData.ts
OnlyJousting: ouff, will tighten it up

packages/curve-ui-kit/src/features/candle-chart/hooks/useHistoricalChartPagination.ts
OnlyJousting: would need to add support for leading debounce 🤔

OnlyJousting: but maybe `cancelRefetch` is enough protection to not even need a debounce anymore

packages/curve-ui-kit/src/features/candle-chart/CandleChart.tsx
DanielSchiavini: OK good job! In the future it would be nice to have that kind of fallback logic in a separate PR to keep things rolling

packages/curve-ui-kit/src/features/candle-chart/hooks/useOhlcQueries.ts
```diff
@@ -0,0 +1,150 @@
...
+    enabled,
+    queryFn: ({ pageParam, signal }) => fetchPageRef.current({ pageParam, signal }),
+    getNextPageParam: (lastPage, _allPages, lastPageParam) => getNextOhlcPageParam(timeOption, lastPage, lastPageParam),
+  })
+}
+
+export const fetchMoreOhlcQueries = (queries: readonly OhlcFetchMoreQuery[]) =>
+  Promise.all(queries.filter(canFetchMoreOhlc).map(query => query.fetchNextPage(FETCH_NEXT_PAGE_OPTIONS)))
+
+export const refetchOhlcQueries = (queries: readonly { refetch: () => Promise<unknown> }[]) =>
```
DanielSchiavini: The refetch member seems to be unnecessary
```suggestion
export const refetchOhlcQueries = (queries: readonly () => Promise<unknown>[]) =>
```

packages/curve-ui-kit/src/features/candle-chart/hooks/useOhlcQueries.ts
DanielSchiavini: https://react.dev/reference/rules/components-and-hooks-must-be-pure#why-does-purity-matter

apps/main/src/lend/hooks/useOhlcChartState.ts
```diff
@@ -60,159 +54,31 @@ export const useOhlcChartState = ({ rChainId, marketId, previewPrices }: UseOhlc
...
-  const llammaRefetchingCapped = useStore(state => state.ohlcCharts.chartLlammaOhlc.refetchingCapped)
-  const llammaLastFetchEndTime = useStore(state => state.ohlcCharts.chartLlammaOhlc.lastFetchEndTime)
-  const fetchLlammaOhlcData = useStore(state => state.ohlcCharts.fetchLlammaOhlcData)
-  const fetchOraclePoolOhlcData = useStore(state => state.ohlcCharts.fetchOraclePoolOhlcData)
-  const fetchMoreData = useStore(state => state.ohlcCharts.fetchMoreData)
-  const resetOhlcState = useStore(state => state.ohlcCharts.resetState)
+  const market = useLendMarket(rChainId, marketId).data
   const priceInfo = useStore(state => state.markets.pricesMapper[rChainId]?.[marketId]?.prices ?? null)
-
   const { oraclePrice } = priceInfo ?? {}
```
DanielSchiavini: The `oraclePrice` is still coming from the store!

OnlyJousting: yep, updating in a separate PR

packages/curve-ui-kit/src/features/candle-chart/CandleChart.tsx
OnlyJousting: it became necesarry in this PR because the two data sources used to be always fetched together even if the fallback wasn't being used. then it was easy to determine when to update the user scroll position. I didn't want to keep doing excessive fetching of data that is not being used when doing this refactor though.

apps/main/src/llamalend/hooks/useLlammaOhlcChartStateModel.ts
```diff
@@ -0,0 +1,168 @@
...
+    chartHeight: Height.chart,
+    isLoading,
+    isEmpty: noDataAvailable,
+    emptyMessage,
+    error: currentError,
+    ohlcData,
+    oraclePriceData,
+    liquidationRange: selectedLiqRange,
+    timeOption,
+    selectedChartKey: selectedChartKey ?? '',
```
DanielSchiavini: It would be great to remove these dummy strings
```suggestion
    selectedChartKey,
```

packages/curve-ui-kit/src/features/candle-chart/hooks/useHistoricalChartPagination.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    return null
+  }, [candlestickSeriesRef, ohlcDataRef, oraclePriceDataRef, oraclePriceSeriesRef])
+
+  const handleVisibleLogicalRangeChange = useCallback(() => {
+    if (!chartRef.current) {
+      return
+    }
+
+    const paginationSeries = getPaginationSeries()
```
DanielSchiavini: Inlining some of these calls and/or using ternaries could make it easier to read imo

DanielSchiavini (dismissed):

apps/main/src/llamalend/hooks/useLlammaOhlcChartStateModel.ts
```diff
@@ -0,0 +1,168 @@
...
+  const oraclePoolOracleLine = oraclePoolsChartQuery.data?.oraclePriceData ?? []
+  const llammaOracleLine = oraclePriceFallbackQuery.data ?? []
+  const { hasAnySeries, isOracleLineOnly, liquidationFallbackData, ohlcData, oraclePriceData } =
+    resolveLlammaChartSeries({
+      oraclePoolCandles,
+      oraclePoolOracleLine,
+      llammaOracleLine,
+    })
+
+  const isLoading = !isAnchorEndReady || oraclePoolsChartQuery.isLoading || isWaitingForFallbackChartData
```
0xPearce: Nice!! there is a very short time range where I believe the `market` object hasn't resolved yet and we show an empty state instead of loading. This is a very small detail the rest looks nice

```suggestion
  const isLoading = !enabled || !isAnchorEndReady || oraclePoolsChartQuery.isLoading || isWaitingForFallbackChartData
```

0xPearce (dismissed):

packages/curve-ui-kit/src/features/candle-chart/hooks/useHistoricalChartPagination.ts
```diff
@@ -72,27 +71,23 @@ export const useHistoricalChartPagination = ({
...
+    const oraclePriceSeries = oraclePriceSeriesRef.current
+    const hasCandlestickData = ohlcDataRef.current.length > 0
+    const hasOraclePriceData = Boolean(oraclePriceDataRef.current?.length)
+    const paginationSeries =
+      hasCandlestickData && candlestickSeries
+        ? candlestickSeries
+        : hasOraclePriceData && oraclePriceSeries
+          ? oraclePriceSeries
+          : null
     if (!paginationSeries) return
```
DanielSchiavini, line 88:
```suggestion
    if (!(ohlcDataRef.current.length && candlestickSeriesRef.current) && !(oraclePriceDataRef.current?.length && oraclePriceSeriesRef.current)) return
```

packages/curve-ui-kit/src/features/candle-chart/hooks/useHistoricalChartPagination.ts
```diff
@@ -72,27 +71,23 @@ export const useHistoricalChartPagination = ({
...
-  }, [candlestickSeriesRef, ohlcDataRef, oraclePriceDataRef, oraclePriceSeriesRef])
-
   const handleVisibleLogicalRangeChange = useCallback(() => {
-    if (!chartRef.current) {
-      return
-    }
-
-    const paginationSeries = getPaginationSeries()
+    const chart = chartRef.current
+    if (!chart) return
```
DanielSchiavini, line 76:
```suggestion
    if (!chartRef.current) return
```

packages/curve-ui-kit/src/features/candle-chart/ChartWrapper.tsx
```diff
@@ -56,7 +56,10 @@ export const ChartWrapper = ({
 }: OhlcChartProps) => {
   const colors = useChartPalette({ backgroundOverride: betaBackgroundColor })
 
-  const errorMessage = t`Unable to fetch "${selectChartList?.find(c => c.key === selectedChartKey)?.label ?? ''}" data.`
+  const selectedChartLabel = selectChartList.find(c => c.key === selectedChartKey)?.label
+  const errorMessage = selectedChartLabel
+    ? t`Unable to fetch "${selectedChartLabel}" data.`
+    : t`Unable to fetch chart data.`
```
DanielSchiavini, line 62:
```suggestion
  const errorMessage = t`Unable to fetch ${selectedChartLabel ? `"${selectedChartLabel}"` : 'chart'} data.`
```

packages/curve-ui-kit/src/features/candle-chart/hooks/useDexChartList.ts
```diff
@@ -75,16 +75,19 @@ export const useDexChartList = ({ coins, nCoins }: UseDexChartListArgs) => {
         ? 'lp-token'
         : `pair-${selectedPairIndex}`
 
-  const selectChartList: ChartSelections[] = useMemo(
-    () => [
+  const selectChartList: ChartSelections[] = useMemo(() => {
+    const lpTokenSymbol = coins[0]?.symbol
+    const lpTokenLabel = lpTokenSymbol ? t`LP Token (${lpTokenSymbol})` : t`LP Token`
```
DanielSchiavini, line 80:
```suggestion
    const lpTokenLabel = notFalsy(t`LP Token`, coins[0] && `(${coins[0]?.symbol})`).join(' ')
```

DanielSchiavini (approved):
Very minor suggestions don't bother too much about them

0xAlunara (approved):

PR #2599, opened 2026-05-28
Update input and select look and feel

0xtutti:
## Summary
- add semantic input/select and LargeTokenInput sizing/spacing tokens
- update input/select hover, text roles, and LargeTokenInput layout to match Figma
- fix numeric adornment spacing so mobile filter fields use shared input padding semantics


packages/curve-ui-kit/src/themes/design/1_sizes_spaces.ts
```diff
@@ -57,6 +57,81 @@ const MappedButtonSize = {
...
+} as const
+
+const MappedInputSpacing = {
+  PaddingX: {
+    tiny: fixedResponsive(Spacing[200]),
+    small: fixedResponsive(Spacing[300]),
+    medium: fixedResponsive(Spacing[300]),
+    large: fixedResponsive(Spacing[300]),
+  },
+  ContentPaddingRight: fixedResponsive(Spacing[300]),
```
0xAlunara: I don't get why `fixedResponsive` is a thing when we could just use `Spacing[300]` directly. I think we can get rid of that util function?

packages/curve-ui-kit/src/themes/components/mui-menu-item.ts
```diff
@@ -28,7 +31,10 @@ export const defineMuiMenuItem = (design: DesignSystem): Components['MuiMenuItem
       },
     },
     dense: {
-      height: ButtonSize.sm,
+      ...handleBreakpoints({
```
0xAlunara: I don't think we need to use `handleBreakpoints`, given none of the properties actually use `mobile`, `tablet` or `desktop` breakpoints.

packages/curve-ui-kit/src/themes/components/mui-input-base.ts
```diff
@@ -5,15 +5,18 @@ import type { DesignSystem } from '@ui-kit/themes/design'
...
 export const defineMuiInputBase = (
-  { Inputs: { Base } }: DesignSystem,
+  { Inputs, InputSelect }: DesignSystem,
   typography: TypographyVariantsOptions,
 ): Components['MuiInputBase'] => ({
   styleOverrides: {
     root: {
-      backgroundColor: Base.Default.Fill.Default,
+      height: InputSize.medium,
+      boxSizing: 'border-box',
```
0xAlunara: Are we sure about this? We usually never have to set border-box and can go by the default

packages/curve-ui-kit/src/shared/ui/LargeTokenInput/BalanceTextField.tsx
```diff
@@ -18,18 +17,24 @@ export const BalanceTextField = ({ balance, name, isError, onChange, disabled }:
...
-          maxHeight: Sizing.lg,
-          backgroundColor: t => t.design.Inputs.Large.Default.Fill,
-          fontFamily: t => t.typography.highlightXl.fontFamily,
-          fontSize: FontSize.xl,
-          fontWeight: FontWeight.Bold,
-          color: t => (isError ? t.design.Layer.Feedback.Error : t.design.Text.TextColors.Primary),
-        },
+        sx: theme => ({
+          height: SizesAndSpaces.LargeTokenInput.AmountHeight,
+          maxHeight: SizesAndSpaces.LargeTokenInput.AmountHeight,
```
0xAlunara: Having both `maxHeight` and `height` with the same value makes no sense. Also not sure why the Balance (sub)compoment relies on LTI constants

packages/curve-ui-kit/src/shared/ui/LargeTokenInput/BalanceTextField.tsx
```diff
@@ -18,18 +17,24 @@ export const BalanceTextField = ({ balance, name, isError, onChange, disabled }:
...
+          backgroundColor: Transparent,
+          color: isError ? theme.design.Inputs.Text.Error : theme.design.Inputs.Text.Value,
+          ...theme.typography.headingSBold,
+          '&& input.MuiInputBase-input': {
+            height: SizesAndSpaces.LargeTokenInput.AmountHeight,
+            paddingInlineStart: 0,
+            paddingLeft: 0,
```
0xAlunara: `paddingLeft` and `paddingInlineStart` are the same. If it's required to reset a MUI default we need to add a comment indicating so

packages/curve-ui-kit/src/shared/ui/LargeTokenInput/LargeTokenInput.tsx
```diff
@@ -266,12 +266,21 @@ export const LargeTokenInput = ({
       id={componentId}
       data-testid={testId}
       sx={{
-        backgroundColor: t => t.design.Inputs.Large.Default.Fill,
+        backgroundColor: t => t.design.InputSelect.Large.Default.Fill,
```
0xAlunara: Why are we going from Inputs to InputSelect? This isn't selecting anything

packages/curve-ui-kit/src/shared/ui/LargeTokenInput/LargeTokenInput.tsx
```diff
@@ -266,12 +266,21 @@ export const LargeTokenInput = ({
...
         outline: t =>
-          `1px solid ${isError ? t.design.Layer.Feedback.Error : t.design.Inputs.Base.Default.Border.Default}`,
+          `1px solid ${
+            isError ? t.design.InputSelect.Base.Default.Border.Error : t.design.InputSelect.Base.Default.Border.Default
+          }`,
+        ...(!disabled &&
```
0xAlunara: It's odd to add in default styling by checking if there's no special case. It'd be better if these two hover styles were always present and either `disabled` or `isError` would add modifications.

packages/curve-ui-kit/src/shared/ui/Select.tsx
```diff
@@ -1,14 +1,28 @@
...
+const ghostSx = (theme: Theme) => ({
+  '&.MuiInputBase-root': { backgroundColor: Transparent },
+  '& .MuiOutlinedInput-notchedOutline': { borderColor: Transparent, backgroundColor: Transparent },
+  '&:hover:not(.Mui-focused):not(.Mui-error):not(.Mui-disabled)': {
+    backgroundColor: Transparent,
+    '& .MuiOutlinedInput-notchedOutline': {
+      borderColor: theme.design.InputSelect.Base.Default.Border.Hover,
+    },
+  },
+  '&.Mui-focused .MuiOutlinedInput-notchedOutline': {
```
0xAlunara: Are you double sure this should only be a thing for the `ghost` variant? Sounds like these should be general `sx` props for any variant to me.

packages/curve-ui-kit/src/features/select-token/ui/TokenSelectButton.tsx
```diff
@@ -1,22 +1,79 @@
...
+const TokenSelectButtonLabel = ({
+  token,
+  disabled,
+  size,
+}: {
+  token: TokenOption
+  disabled: boolean
+  size: SelectProps['size']
+}) =>
+  size === 'small' ? (
```
0xAlunara: Tell Codex to try harder: why isn't this a new size or modification for TokenLabel? It's almost a recreation of that component for no good reason.

packages/curve-ui-kit/src/features/select-token/ui/TokenSelectButton.tsx
```diff
@@ -1,22 +1,79 @@
...
+  desktop: `calc(${IconSize.lg.desktop} + ${SelectSpacing.IconPaddingRight.small.desktop})`,
+}
+
+const smallTokenSelectorSx = {
+  // Token selector triggers use the visual Select pattern, but token symbols do not need the generic value/icon gap
+  // because the trigger label is an intrinsic token chip rather than free-form select text.
+  '&& .MuiSelect-select.MuiSelect-select': {
+    ...handleBreakpoints({
+      paddingInlineEnd: smallTokenSelectorIconSpace,
+      paddingRight: smallTokenSelectorIconSpace,
```
0xAlunara: Why are we setting both `paddingRight` and `paddingInlineEnd` which are the same?

packages/curve-ui-kit/src/features/select-token/ui/TokenSelectButton.tsx
```diff
@@ -1,22 +1,79 @@
...
+import { TokenIcon } from '@ui-kit/shared/ui/TokenIcon'
 import { TokenLabel } from '@ui-kit/shared/ui/TokenLabel'
+import { handleBreakpoints } from '@ui-kit/themes/basic-theme'
+import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
 import type { TokenOption } from '../types'
 
+const { IconSize, SelectSpacing, Spacing } = SizesAndSpaces
+
+const smallTokenSelectorIconSpace = {
+  mobile: `calc(${IconSize.lg.desktop} + ${SelectSpacing.IconPaddingRight.small.mobile})`,
```
0xAlunara: It probably works out fine in the calculation but using desktop icon sizes in both the `mobile` and `tablet` breakpoints looks a bit odd.

0xAlunara (changes requested):

apps/main/src/dex/components/PagePool/Swap/index.tsx
```diff
@@ -390,6 +390,7 @@ export const Swap = ({
               onClose={closeModalFromToken}
               isOpen={!!isOpenFromToken}
               onOpen={openModalFromToken}
+              size="small"
```
DanielSchiavini, line 393: We should make this the default at some point I think

packages/curve-ui-kit/src/shared/ui/LargeTokenInput/Balance.tsx
```diff
@@ -83,17 +87,18 @@ export const Balance = <T extends Amount>({
         </BalanceAmount>{' '}
         <Typography
           variant="highlightXs"
-          color={disabled ? 'textDisabled' : 'textPrimary'}
-          sx={{ ...VERTICAL_CENTER_TEXT }}
+          sx={{
+            ...VERTICAL_CENTER_TEXT,
+            color: t => (disabled ? t.design.Inputs.Text.Disabled : t.design.Inputs.Text.Unit),
```
DanielSchiavini:
```suggestion
            color: t => t.design.Inputs.Text[disabled ? 'Disabled' : 'Unit'],
```

packages/curve-ui-kit/src/shared/ui/LargeTokenInput/BalanceAmount.tsx
```diff
@@ -25,10 +25,18 @@ export const BalanceAmount = <T extends Amount>({
...
       data-value={children ?? ''}
-      sx={{ ...VERTICAL_CENTER_TEXT, ...sx }}
+      sx={{
+        ...VERTICAL_CENTER_TEXT,
+        color: t =>
+          disabled
+            ? t.design.Inputs.Text.Disabled
+            : children == null
+              ? t.design.Inputs.Text.MetaSubtle
+              : t.design.Inputs.Text.Value,
```
DanielSchiavini:
```suggestion
          t.design.Inputs.Text[disabled ? 'Disabled' : children == null ? 'MetaSubtle' : 'Value'],
```

packages/curve-ui-kit/src/shared/ui/LargeTokenInput/BalanceTextField.tsx
```diff
@@ -18,18 +19,22 @@ export const BalanceTextField = ({ balance, name, isError, onChange, disabled }:
...
-          backgroundColor: t => t.design.Inputs.Large.Default.Fill,
-          fontFamily: t => t.typography.highlightXl.fontFamily,
-          fontSize: FontSize.xl,
-          fontWeight: FontWeight.Bold,
-          color: t => (isError ? t.design.Layer.Feedback.Error : t.design.Text.TextColors.Primary),
-        },
+        sx: theme => ({
+          height: AmountHeight,
+          backgroundColor: Transparent,
+          color: isError ? theme.design.Inputs.Text.Error : theme.design.Inputs.Text.Value,
```
DanielSchiavini:
```suggestion
          color: theme.design.Inputs.Text[isError ? 'Error' : 'Value'],
```

packages/curve-ui-kit/src/shared/ui/LargeTokenInput/LargeTokenInput.tsx
```diff
@@ -268,10 +268,27 @@ export const LargeTokenInput = ({
       sx={{
         backgroundColor: t => t.design.Inputs.Large.Default.Fill,
         outline: t =>
-          `1px solid ${isError ? t.design.Layer.Feedback.Error : t.design.Inputs.Base.Default.Border.Default}`,
+          `1px solid ${isError ? t.design.Inputs.Base.Default.Border.Error : t.design.Inputs.Base.Default.Border.Default}`,
```
DanielSchiavini: same comment as above should be applied to this file, too

packages/curve-ui-kit/src/shared/ui/NumericTextField.tsx
```diff
@@ -93,10 +94,20 @@ const getDisplayValue = (val?: Decimal) => (val == null ? '' : String(val))
...
-  <Typography variant={size === 'tiny' ? 'bodySBold' : 'bodyMBold'} color="textSecondary">
-    {children}
-  </Typography>
+const AdornmentTypography = ({
+  children,
+  position,
+  size,
+}: {
+  children: ReactNode
+  position: 'start' | 'end'
```
DanielSchiavini: all instances are passing `position="end"` I think? If that's the case we don't need this at all.
```suggestion
```

packages/curve-ui-kit/src/themes/components/mui-input-base.d.ts
```diff
@@ -11,6 +11,7 @@ import type { TextFieldClasses } from '@mui/material'
  */
 interface InputSizeOverrides {
   tiny: true // called xxs in the design system
+  large: true // called xl in the design system
```
DanielSchiavini: XL should be called `extraLarge` 🙏 
mapping XL to `large` is confusing and blocks a future addition of the L size

packages/curve-ui-kit/src/themes/components/mui-select.ts
```diff
@@ -1,108 +1,166 @@
...
 
-const { Spacing, Sizing, IconSize } = SizesAndSpaces
+const { IconSize, SelectSize, SelectSpacing } = SizesAndSpaces
+
+type ResponsiveOrStatic = Responsive | string
+
+const getResponsiveValue = (value: ResponsiveOrStatic, breakpoint: keyof Responsive) =>
+  typeof value === 'string' ? value : value[breakpoint]
+
+const getSelectIconSpace = (iconSize: Responsive, iconPaddingRight: ResponsiveOrStatic): Responsive<string> => ({
```
DanielSchiavini: this should be using `mapBreakpoints` and/or `handleBreakpoints`

packages/curve-ui-kit/src/themes/components/mui-select.ts
```diff
@@ -1,108 +1,166 @@
...
 import { DesignSystem } from '@ui-kit/themes/design'
 import { handleBreakpoints, type Responsive } from '../basic-theme'
 import { SizesAndSpaces } from '../design/1_sizes_spaces'
 
-const { Spacing, Sizing, IconSize } = SizesAndSpaces
+const { IconSize, SelectSize, SelectSpacing } = SizesAndSpaces
+
+type ResponsiveOrStatic = Responsive | string
+
+const getResponsiveValue = (value: ResponsiveOrStatic, breakpoint: keyof Responsive) =>
```
DanielSchiavini: this doesn't belong in this file and can be probably removed altogether

packages/curve-ui-kit/src/themes/components/mui-select.ts
```diff
@@ -1,108 +1,166 @@
...
+  iconPaddingRight: ResponsiveOrStatic
+  paddingBlock: ResponsiveOrStatic
+  paddingInlineStart: ResponsiveOrStatic
+  typography: 'bodySBold' | 'bodyMBold' | 'headingSBold'
 }
 
 type SelectSizes = NonNullable<SelectProps['size']>
 
 const selectSizes: Record<SelectSizes, SelectSizeDefinition> = {
   tiny: {
```
DanielSchiavini: most of the blocks for each size is duplicated, we probably can share some logic. See for example `sizeBreakpoint` for (toggle)buttons

apps/main/src/dex/components/PagePool/Swap/index.tsx
0xtutti: Agreed. I kept this explicit for this PR so we do not change every TokenSelector usage as part of the input/select parity work; making `small` the default can be a follow-up.

packages/curve-ui-kit/src/themes/components/mui-select.ts
```diff
@@ -1,108 +1,140 @@
...
 type SelectSizeDefinition = {
-  height: Responsive
+  height: string
   iconSize: Responsive
-  paddingBlock: Responsive
-  paddingInlineStart: Responsive
+  iconPaddingRight: string
+  paddingBlock: string
+  paddingInlineStart: string
+  typography: 'bodySBold' | 'bodyMBold' | 'headingSBold'
```
DanielSchiavini, line 18:
```suggestion
  typography: TypographyVariantKey
```

DanielSchiavini (approved):

0xAlunara (approved):
Guess it's time to FAFO, if something's broken design wise I suppose we could easily fix it in a follow up PR.

PR #2592, opened 2026-05-27
refactor: llamalend tables improvements

Pearce542:
Depends on https://github.com/curvefi/curve-frontend/pull/2591, next https://github.com/curvefi/curve-frontend/pull/2607

Small improvements:
- In order to improve the vertical space of the llama markets, the table title is no longer sticky, only the filters
- added mui border style and selector helpers 
- added card header for titles for consistency with the swap and market page
- hide a specific user position table when empty

PR #2607, opened 2026-05-28
feat: user positions table empty state

Pearce542:
Depends on https://github.com/curvefi/curve-frontend/pull/2592
Improve user positions table empty states

Not connected: 
Screenshot 2026-05-28 at 17 39 56

Query loading:
Screenshot 2026-05-28 at 17 42 10

No positions:
Screenshot 2026-05-28 at 17 40 20


0xtutti (dismissed):

apps/main/src/llamalend/features/market-list/UserPositionsTables.tsx
```diff
@@ -28,12 +35,14 @@ export const UserPositionsTables = ({
   tableQuery,
   tableQuery: { data: queryData, isLoading },
 }: UserPositionsTableProps) => {
+  const { address, isConnecting } = useConnection()
+  const { connect } = useWallet()
   // Tracks whether the user has any positions for each market rate type.
   const hasUserPositions = maybe(queryData?.userHasPositions, userHasPositions =>
```
DanielSchiavini, line 41: Is this really necessary? See
https://github.com/curvefi/curve-frontend/blob/95ff0cc59fdc4424329ad4b355cac4ba9e3b1395/apps/main/src/llamalend/queries/market-list/llama-markets.ts#L201-L207

apps/main/src/llamalend/features/market-list/UserPositionsTables.tsx
```diff
@@ -64,6 +73,29 @@ export const UserPositionsTables = ({
             onReload={onReload}
           />
         )}
+        {[!hasUserPositions, !address].some(Boolean) && (
```
DanielSchiavini:
```suggestion
        {(!hasUserPositions || !address) && (
```

DanielSchiavini (dismissed):

apps/main/src/llamalend/features/market-list/UserPositionsTables.tsx
Pearce542: the advantage with the current approach is that we avoid filtering all the markets if there are no borrow position. Same for supply.

DanielSchiavini: We have that calculated in the result afaik

DanielSchiavini: Isn't this the same as
https://github.com/curvefi/curve-frontend/blob/95ff0cc59fdc4424329ad4b355cac4ba9e3b1395/apps/main/src/llamalend/queries/market-list/llama-markets.ts#L393

Pearce542: Exactly we have a summary of the `userHasPositions`, for each market type and each market rate type in the `llama-markets` query's result, with the following type: 
`Record<LlamaMarketType, Record<MarketRateType, boolean>>`. 

I use that result and convert it into something like this: `Record<MarketRateType, boolean>` where the llama market type is implicit

DanielSchiavini: ah cool!

DanielSchiavini (dismissed):

apps/main/src/llamalend/features/market-list/UserPositionsTables.tsx
DanielSchiavini: We could consider adding this to the query result as well. `userHasPositions` or something.

0xAlunara (dismissed):

Pearce542:
The new empty state card is making some tests fail because previously we didn't show any empty state when there was no positions. Let me fix that 

Screenshot 2026-05-29 at 15 56 57

0xAlunara (dismissed):

0xtutti (dismissed):

DanielSchiavini (approved):

0xAlunara (approved):

PR #2612, opened 2026-05-29
chore: fix refuel timeouts as timeseries endpoint can be slower than pools endpoint for pool-information

0xAlunara:
The flaky `undefined` timeout error comes most likely from the price metrics, not the pool metrics. The test waits for `pool-tvl-value`, which only proves the pool info query loaded, while `lp-token-value-value` and `virtual-price-value` depend on a separate refuel timeseries query (which can be way slower).


DanielSchiavini (approved):
Image

OnlyJousting (approved):

PR #2616, opened 2026-05-29
feat: add address cell explorer urls for pool and market pages

0xAlunara:
Adds copyable addresses and explorer urls to refuel table, pool tables and market tables.

Decided not to touch the similar looking functionality from ActionInfo, because values in that component don't necessarily have to be addresses. The additional complexity wouldn't be worth the DRY gain.

image


0xtutti (approved):

Pearce542 (approved):

packages/curve-ui-kit/src/features/activity-table/types.ts
```diff
@@ -4,8 +4,9 @@ import type { AllPoolTrade, PoolLiquidityEvent } from '@curvefi/prices-api/pools
 import type { Token } from '@primitives/address.utils'
 
 // LLAMMA Types (for lending/crvusd markets)
-export type LlammaTradeRow = LlammaTrade & { txUrl?: string; url?: never; network: Chain }
+export type LlammaTradeRow = LlammaTrade & { buyerUrl?: string; txUrl?: string; url?: never; network: Chain }
 export type LlammaEventRow = LlammaEvent & {
+  providerUrl?: string
```
DanielSchiavini, line 9: providerUrl, buyerUrl and donorUrl are passed to explorerUrl, is there a way to make them a bit more consistent?

DanielSchiavini (approved):

packages/curve-ui-kit/src/features/activity-table/types.ts
0xAlunara: Well, they're consistent with the name of the property their URL is dervied from.

PR #2618, opened 2026-05-29
feat: add TokenInfo info widget

0xAlunara:
I saw the same logic for displaying a token icon with 2 values duplicated four times over the new designs, so I've deduplicated them into a more general `TokenInfo` widget, which

1. Will show a token icon on either the left or right side.
2. Has a primary value, which can be anything from token amount to token symbol.
3. Has an optional secondary value, usually some kind of notional value but can also be the token symbol or anything else really.
4. In the new pool advanced details they'll be used to denote assets with their price.

EDIT: This might completely replace `TokenLabel`, as you can see from the bottom pic, although `TokenLabel` also has a size prop which this one doesn't. Might look at that in a follow up PR.

image

image

image

image


apps/main/src/llamalend/features/supply/components/columns/column.definitions.tsx
```diff
@@ -20,11 +22,25 @@ const headers = {
...
+const formatBalance = (balance: ClaimableToken['amount'] | undefined) =>
+  balance == null ? '-' : formatNumber(balance, { abbreviate: false })
+
 export const CLAIM_TAB_COLUMNS = [
   columnHelper.accessor('amount', {
     id: ClaimTabColumnId.Token,
     header: headers[ClaimTabColumnId.Token],
-    cell: TokenBalanceCell<ClaimableToken>,
+    cell: ({ getValue, row }) => (
+      <InlineTableCell>
```
0xAlunara, line 30: Found it easier to just inline these cells with an `InlineTabelCell` and `TokenInfo` than to define a whole file with prop types for them

packages/curve-ui-kit/src/features/activity-table/cells/LlammaEventChangeCell.tsx
```diff
@@ -1,51 +1,37 @@
...
-  amount: number
-  amountUsd?: number | null
-  token: Token | undefined
-}
-
-const AmountRow = ({ amount, amountUsd, token, chain }: AmountRowProps) => (
-  <Stack direction="row" sx={{ alignItems: 'center', justifyContent: 'right', gap: Spacing.xs }}>
-    <Stack sx={{ gap: Spacing.sm }}>
-      <Typography variant="tableCellMBold">{formatNumber(amount, { abbreviate: false })}</Typography>
-      {amountUsd != null && (
```
0xAlunara, line 31: `amountUsd` was never passed lol

0xtutti (dismissed):

Pearce542 (dismissed):

packages/curve-ui-kit/src/features/activity-table/columns/pool-liquidity-columns.tsx
```diff
@@ -45,13 +48,16 @@ export const createPoolLiquidityColumns = ({ poolTokens }: CreatePoolLiquidityCo
...
         return (
-          <TokenAmountCell
-            amount={displayAmount}
-            symbol={token.symbol}
-            tokenAddress={token.address}
-            chainId={network}
-            align="right"
-          />
+          <InlineTableCell sx={{ alignItems: 'end' }}>
+            {amount !== 0 && (
```
DanielSchiavini: I don't see this logic in the old TokenAmountCell - why is this added now? It's not really obvious to hide elements when the value is 0.

packages/curve-ui-kit/src/shared/ui/TokenInfo.tsx
```diff
@@ -0,0 +1,45 @@
...
+  primary: string
+  secondary?: string
+  showChainIcon?: boolean
+}) => {
+  const isIconRight = iconPosition === 'right'
+  const tokenIcon = <TokenIcon blockchainId={blockchainId} address={address} size="lg" showChainIcon={showChainIcon} />
+
+  return (
+    <Stack direction="row" sx={{ gap: Spacing.xs, alignItems: 'center' }}>
+      {!isIconRight && tokenIcon}
```
DanielSchiavini: `!isIconRight` is not really correct imo
```suggestion
      {iconPosition === 'left' && tokenIcon}
```

packages/curve-ui-kit/src/shared/ui/TokenInfo.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+      <Stack sx={{ gap: Spacing.xxs, alignItems: isIconRight ? 'end' : 'start' }}>
```
DanielSchiavini:
```suggestion
      <Stack sx={{ gap: Spacing.xxs, alignItems: {left: 'start', right: 'end'}[iconPosition] }>
```

packages/curve-ui-kit/src/shared/ui/TokenInfo.tsx
0xAlunara: yeah it reads a bit funky with the double negative, i guess this is a bit better

packages/curve-ui-kit/src/shared/ui/TokenInfo.tsx
0xAlunara: kept it a ternary for now; reads a bit better

packages/curve-ui-kit/src/features/activity-table/columns/pool-liquidity-columns.tsx
0xAlunara: I think it looks better but you're right, it's inconsistent with other tables

packages/curve-ui-kit/src/features/activity-table/columns/pool-liquidity-columns.tsx
```diff
@@ -45,13 +48,18 @@ export const createPoolLiquidityColumns = ({ poolTokens }: CreatePoolLiquidityCo
...
         return (
-          <TokenAmountCell
-            amount={displayAmount}
-            symbol={token.symbol}
-            tokenAddress={token.address}
-            chainId={network}
-            align="right"
-          />
+          <InlineTableCell sx={{ alignItems: 'end' }}>
+            {amount !== 0 ? (
```
0xAlunara: Oops, double negative

packages/curve-ui-kit/src/shared/ui/TokenInfo.tsx
DanielSchiavini: It wasn't about readability but correctness when we introduce a new icon position

DanielSchiavini (approved):

OnlyJousting (approved):

PR #2622, opened 2026-06-01
lint: enable consistent-type-definitions

DanielSchiavini:
- enable eslint [rule](https://typescript-eslint.io/rules/consistent-type-definitions/) and run lint:fix
- [why the default is interface](https://typescript-eslint.io/rules/consistent-type-definitions/#why-is-the-default-interface)


0xAlunara:
I'd really prefer `type` over `interface` actually. Not only have we already been primarily using it, I like it better as it works better when constructing more complex types, including `infer`, non-shape types, unions and intersections, and all the useful [utility types](https://www.typescriptlang.org/docs/handbook/utility-types.html). I'm actually surprised they're recommending interfaces over types, I figured types would be considered a more modern practice these days. Feels like a leftover from the `class` and object oriented days.

DanielSchiavini:
> I'd really prefer `type` over `interface` actually. 

That has also been my personal preference tbh, even just because it's shorter.
However, interface seems to be the better option after reading the referenced links

Pearce542 (approved):
Nice!

0xtutti (approved):

PR #2624, opened 2026-06-01
refactor: data table scrollable and sticky

Pearce542:
- Made the table headers sticky when the table's width <= parent's width. Otherwise, horizontal scrolling is enabled and sticky headers are disabled, as I haven't found a straightforward way to support both. Credits to @0xAlunara for the idea.
- Refactored useResizeObserver to always return an array, since this is how we've been using it everywhere:
```typescript
const [selectWidth] = useResizeObserver(selectRef) ?? []
```


packages/curve-ui-kit/src/hooks/useResizeObserver.ts
```diff
@@ -66,5 +66,5 @@ export function useResizeObserver(
     }
   }, [elementRef, threshold])
 
-  return dimensions
+  return dimensions ?? []
```
DanielSchiavini: Can we put the [] in the global scope? I kept the `?? []` in the components to make sure we use destructuring in order to avoid memoization churn.
```suggestion
  return dimensions ?? EMPTY
```

packages/curve-ui-kit/src/shared/ui/DataTable/DataTable.tsx
```diff
@@ -108,83 +113,96 @@ export const DataTable = <T extends TableItem>({
...
-        {showFooter && (
-          <TableFooter>
-            {footerRow && <TableRow>{footerRow}</TableRow>}
-            {showViewAllButton && (
-              <TableRow>
-                <TableViewAllCell colSpan={columnCount} onClick={onShowAll} isLoading={isLoadingViewAll}>
-                  {viewAllLabel || t`View all`}
-                </TableViewAllCell>
-              </TableRow>
+          <TableBody>
```
DanielSchiavini, line 171: Can't we put the horizontal scrollbar in the TableBody instead?

Pearce542: We can't because we want the table's header (which contains the column's headers) to scroll with the body as well

0xAlunara (dismissed):

DanielSchiavini (dismissed):
Looks good but tests failing

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
```diff
@@ -69,20 +69,31 @@ testCases.forEach(([width, height, breakpoint]) => {
       cy.url(LOAD_TIMEOUT).should('match', urlRegex)
     })
 
-    it('should have sticky headers', () => {
-      cy.get('[data-testid^="data-table-row"]').last().then(assertNotInViewport)
-      cy.get('[data-testid^="data-table-row"]').eq(10).scrollIntoView()
-      cy.get('[data-testid="data-table-head"] th').eq(1).then(assertInViewport)
+    it('should have sticky headers or horizontally scrollable table', () => {
```
DanielSchiavini, line 70: the test failure sucks, it provides no useful information
```
     AssertionError: [object Object] should be in the viewport: expected false to be true
```

OnlyJousting (dismissed):

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
Pearce542: fixed the test. I will make a 20+ parallel llamalend tests run after finishing this one https://github.com/curvefi/curve-frontend/pull/2629 due to all the small changes I made to the market list

0xAlunara (dismissed):

DanielSchiavini (dismissed):

Pearce542:
@0xAlunara @DanielSchiavini sorry about dismissing your reviews and the longer PR, I thought I was pushing the last commits on another PR on top of it 🙈

packages/curve-ui-kit/src/shared/ui/DataTable/DataRow.tsx
```diff
@@ -52,7 +53,9 @@ export const DataRow = <T extends TableItem>({
     [url, push, hasUrl],
   )
   const visibleCells = row.getVisibleCells()
-  const shouldApplyStickyLastRow = isLastRow && shouldStickLastRowToTop
+  // to do: remove shouldApplyStickyLastRow and sx code once LegacyDataTable is deleted
```
DanielSchiavini, line 56: why? How did you solve the problem that was fixed by this?

packages/curve-ui-kit/src/shared/ui/DataTable/DataRow.tsx
```diff
@@ -27,7 +28,7 @@ export type DataRowProps<T extends TableItem> = {
   table: Table<T>
   row: Row<T>
   expandedPanel?: ExpandedPanel<T>
-  isLastRow: boolean
+  isLastRow?: boolean
```
DanielSchiavini, line 31: Why is this made optional?

packages/curve-ui-kit/src/shared/ui/DataTable/DataTable.tsx
```diff
@@ -123,8 +122,8 @@ export const DataTable = <T extends TableItem>({
       position: 'sticky',
       top: maxHeight ? 0 : top,
       zIndex: t.zIndex.tableHeader,
+      marginBlockEnd: Height.row, // last row should not be hidden by the sticky header
```
DanielSchiavini, line 125: how is hiding the sticky header related to this margin? 🤔

packages/curve-ui-kit/src/shared/ui/DataTable/DataRow.tsx
Pearce542: By adding a margin bottom to the sticky header itself

packages/curve-ui-kit/src/shared/ui/DataTable/DataRow.tsx
Pearce542: because no longer needed for the `DataTable` but still needed for the `LegacyDataTable`

packages/curve-ui-kit/src/shared/ui/DataTable/DataRow.tsx
Pearce542: also this was not working, the last row haad 1. a higher z-index then the table header, 2. and was not pushing away the header as you can see in this images: 
Screenshot 2026-06-03 at 13 12 04
Screenshot 2026-06-03 at 13 12 16

packages/curve-ui-kit/src/shared/ui/DataTable/DataTable.tsx
Pearce542: without it the sticky header would scroll on top of the last row. By adding a margin bottom of the same height than one row, we avoid that

DanielSchiavini: doesn't that introduce empty space? :thinking:

packages/curve-ui-kit/src/shared/ui/DataTable/DataRow.tsx
DanielSchiavini: a TODO would be nice

DanielSchiavini: but maybe not needed since you already have a todo and then this will become unused

packages/curve-ui-kit/src/shared/ui/DataTable/DataTable.tsx
Pearce542: Not sure I understand what you mean, but no, the header stops being sticky at the last row like this: 

Screenshot 2026-06-03 at 14 56 49

DanielSchiavini: oh nice solution :+1:

DanielSchiavini: image

DanielSchiavini (approved):

0xtutti (approved):

PR #2625, opened 2026-06-01
Sync Chad typography with Figma

0xtutti:
Brings back Mono font to chad mode to align with the classic UI.

## Summary
- maps Chad typography font families to Ioskeley Mono across design tokens and legacy CSS variables
- adds an Ioskeley Mono webfont asset for Chad-only loading
- syncs typography variants and tab-specific text variants to the live Figma values


0xtutti (commented):
Daniel + Alunara review pass:

Daniel lens — token/runtime correctness:
- No blockers found in the token mapping. Chad now resolves Heading/Body/Mono/Button to Ioskeley Mono in the design theme, the MUI font stack knows that family, and the legacy CSS vars cover the old packages/ui surfaces.
- Good catch keeping this Chad-only: there is no global preload, so Light/Dark should not eagerly download the new font.
- I found one coverage gap during review: curve-ui-kit Storybook had its own font-face setup and would not load Ioskeley for Chad stories. Fixed in follow-up commit 20ab1ec71.

Alunara lens — UI/component feel:
- Tab variants now use dedicated ButtonLabel/Tabs styles instead of borrowing generic button/heading variants, which better matches the Figma type intent.
- The Chad bodyM override is a reasonable compatibility shim for dense legacy surfaces while the rest of the typography follows token aliases.
- Remaining caveat: the checked-in asset is a static regular WOFF2, not a real variable font. Non-400 weights will be synthesized until we replace packages/curve-ui-kit/public/fonts/Ioskeley-Mono.woff2 with the licensed variable webfont and change the @font-face range to 300 700.

Validation observed:
- git diff --check
- yarn workspace curve-ui-kit typecheck

apps/main/src/index.css
```diff
@@ -2,12 +2,13 @@
 
 body {
   --font: 'Mona Sans', 'Helvetica Neue', Helvetica, sans-serif;
-  --font-mono: Minecraft, 'SF Mono Regular 11', 'Ubuntu Mono', monospace;
+  --font-mono: 'Mona Sans', 'Helvetica Neue', Helvetica, sans-serif;
```
DanielSchiavini: It seems `--font-mono` should be deleted if it's not needed anymore. It doesn't make sense now it's not even `monospace`

packages/curve-ui-kit/src/shared/ui/LargeTokenInput/BalanceTextField.tsx
```diff
@@ -29,6 +30,12 @@ export const BalanceTextField = ({ balance, name, isError, onChange, disabled }:
           backgroundColor: Transparent,
           color: theme.design.Inputs.Text[isError ? 'Error' : 'Value'],
           ...theme.typography.headingSBold,
+          '&&:hover, &&.Mui-focused, &&.Mui-error, &&.Mui-disabled': {
+            backgroundColor: Transparent,
+          },
+          '&&:hover:not(.Mui-focused):not(.Mui-error):not(.Mui-disabled)': {
+            backgroundColor: Transparent,
+          },
```
DanielSchiavini: `backgroundColor: Transparent,` is repeated 3 times here and it's not even overridden in other places. It's also adding all the same selectors with a `:not` 😆 .
It seems that it's trying to override some MUI defaults - which would deserve some comments.
However, if we just want it to be always transparent, just add extra ampersands like
```suggestion
          '&&&': { backgroundColor: Transparent },
          color: theme.design.Inputs.Text[isError ? 'Error' : 'Value'],
          ...theme.typography.headingSBold,
```

packages/curve-ui-kit/src/themes/components/button/mui-button.ts
```diff
@@ -30,13 +30,34 @@ const sizeBreakpoint = (
...
+    LineHeight[lineHeight],
+    'tablet',
+  ),
+  ...sizeBreakpoint(
+    ButtonSize[height],
+    FontSize[fontSize],
+    fontWeightTokens[fontWeight],
+    LineHeight[lineHeight],
+    'desktop',
+  ),
```
DanielSchiavini: The same code is called for all breakpoints, we could just iterate over `basicMuiTheme.breakpoints`

packages/curve-ui-kit/src/themes/design/2_theme.ts
```diff
@@ -1179,11 +1211,12 @@ export const createChadDesign = (Chad: typeof SurfacesAndText.plain.Chad | typeo
       },
     },
     FontFamily: {
```
DanielSchiavini: all fonts are the same for all themes, we could just make this a simple string instead of separating per type.

packages/curve-ui-kit/src/themes/typography.ts
```diff
@@ -73,51 +81,59 @@ const variant = ({
   ...responsiveValues(fontSize, lineHeight, 'desktop'),
 })
 
+const chadTypographyVariantOverrides = {
```
DanielSchiavini: Argh this file shouldn't know anything about the themes, if chad has a different font size and line height it should be part of 2_theme.ts

packages/curve-ui-kit/src/themes/typography.ts
```diff
@@ -126,7 +142,11 @@ export const createTypography = ({ Text }: DesignSystem) =>
 
     ...Object.fromEntries(
       Object.entries(TYPOGRAPHY_VARIANTS).map(([key, def]) => {
-        const { fontFamily, ...value } = variant(def)
+        const definition =
+          Text.FontFamily.Body === 'Ioskeley Mono'
```
DanielSchiavini: Exactly to avoid this kind of slop

packages/curve-ui-kit/src/themes/components/button/mui-button.ts
DanielSchiavini: tip: use helpers from `objects.utils.ts` such as `recordValues` and `fromEntries`

apps/main/src/index.css
0xtutti: Actually the new Mona Sans is variable and has a mono option now.

0xtutti: Addressed in 07667461d/ad6143c09: `--font-mono` now points to the real `Mona Sans Mono` variable WOFF2, and the font stack helper keeps CSS generic fallbacks like `monospace` unquoted.

packages/curve-ui-kit/src/shared/ui/LargeTokenInput/BalanceTextField.tsx
0xtutti: Addressed in 07667461d: replaced the repeated transparent background selectors with the single `&&&` specificity override while keeping the input text color on design tokens.

packages/curve-ui-kit/src/themes/components/button/mui-button.ts
0xtutti: Addressed in 07667461d: button sizing now uses `handleBreakpoints` instead of repeating the same breakpoint block manually. I also cleaned the adjacent icon spacing and toggle button sizing pattern the same way.

packages/curve-ui-kit/src/themes/design/2_theme.ts
0xtutti: Addressed in 07667461d: `Text.FontFamily` is now a simple theme-level string instead of separate Heading/Body/Mono/Button entries.

packages/curve-ui-kit/src/themes/typography.ts
0xtutti: Addressed in 07667461d/ad6143c09: Chad-specific body size and line-height are now supplied through `Text.TypographyVariantOverrides` in `2_theme.ts`; `typography.ts` only consumes the token.

packages/curve-ui-kit/src/themes/typography.ts
0xtutti: Addressed in 07667461d: removed the font-family/theme sniffing from `typography.ts`; typography variants now merge theme-provided overrides without knowing about Chad or Ioskeley.

Pearce542 (dismissed):
Looking very nice 🔥

packages/curve-ui-kit/src/themes/components/button/mui-toggle-button.ts
```diff
@@ -47,46 +25,42 @@ type ButtonSize = BaseButtonSize & {
...
+    minWidth: ButtonSize[size],
+    height: ButtonSize[size],
+    width: ButtonSize[size],
+    paddingBlock: Spacing[padding],
+    paddingInline: Spacing[padding],
+    fontSize: FontSize[fontSize],
+    fontWeight: fontWeightTokens[fontWeight],
+    lineHeight: LineHeight[lineHeight],
+  }),
+  overflow: 'hidden',
```
DanielSchiavini: I think this can be inside handleBreakpoints so we can get rid of the surrounding `{...}`

packages/curve-ui-kit/src/themes/design/2_theme.ts
```diff
@@ -1178,12 +1214,9 @@ export const createChadDesign = (Chad: typeof SurfacesAndText.plain.Chad | typeo
         Inverted: Grays[50],
       },
     },
-    FontFamily: {
-      Heading: 'Minecraft',
-      Body: 'Hubot Sans',
-      Mono: 'Hubot Sans',
-      Button: 'Minecraft',
-    },
+    FontFamily: 'Ioskeley Mono',
```
DanielSchiavini, line 1219: so now the font family is simpler, but you still kept the mono font. Either we use it here for the buttons or we should delete that `--font-mono`

packages/curve-ui-kit/src/themes/components/button/mui-toggle-button.ts
0xtutti: Addressed in c88cf6b0e: moved `overflow: hidden` into the `handleBreakpoints` call so `buttonSizeSquare` no longer needs the surrounding spread/object wrapper.

packages/curve-ui-kit/src/themes/design/2_theme.ts
0xtutti: Addressed in c88cf6b0e: `Button.FontFamily` now uses `Mona Sans Mono` for Light/Dark and `Ioskeley Mono` for Chad, and legacy `--button--font` now points at `--font-mono`, so the mono font is actively used for buttons.

DanielSchiavini (approved):

0xAlunara (approved):

PR #2609, opened 2026-05-28
refactor: improve market list table (part 1)

Pearce542:
Next https://github.com/curvefi/curve-frontend/pull/2629

- do not show all columns while the table is loading
- refactor the maintenance local storage key to something more general

PR #2629, opened 2026-06-01
refactor: improve market list table (part 2)

Pearce542:
- Depends on https://github.com/curvefi/curve-frontend/pull/2609
- remove `unknown` llama column visibility variant
- center the empty user position card for all browser
- active chain chip wrapper was not aligned with the other chips (height of 25.27px instead of 24px 🤓 )
- consistent borders for the table: removed `TableCell` border bottom, it is now controlled by the parent. Each row has a border bottom eccept last row, removed duplicate borders between end of data rows and table footer

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>Screenshot 2026-06-01 at 19 14 46</td>
    <td>Screenshot 2026-06-01 at 19 14 51
</td>
  </tr>
  <tr>
    <td>
Screenshot 2026-06-03 at 16 24 52
</td>
    <td>
Screenshot 2026-06-03 at 16 25 09
</td>
  </tr>
</table>


packages/curve-ui-kit/src/shared/ui/DataTable/data-table.utils.ts
```diff
@@ -120,7 +120,6 @@ export function useCellSx<T extends TableItem>({
         zIndex: (t: Theme) => t.zIndex.tableStickyColumn,
         backgroundColor: (t: Theme) => t.design.Table.Row.Default,
       }),
-      borderBlockEnd: borderStyle,
```
Pearce542, line 123: moved logic to the parent `TableBody`

0xAlunara (approved):

apps/main/src/llamalend/features/market-list/UserPositionsMarketRateTable.tsx
```diff
@@ -72,6 +72,8 @@ export const UserPositionsMarketRateTable = ({
         max: DEFAULT_VISIBLE_ROWS,
         buttonLabel: t`View all ${data.length} ${marketRateType.toLowerCase()} positions`,
       }}
+      // with two user positions tables let's avoid having too many skeletons rows when loading
+      increasingLengthOptions={{ initialLength: 1, maxLength: 3 }}
```
DanielSchiavini, line 76: I'd hope to avoid these magic numbers in components, they should either be in constants or as `variants` of the component

DanielSchiavini (approved):

apps/main/src/llamalend/features/market-list/UserPositionsMarketRateTable.tsx
Pearce542: will change it

PR #2637, opened 2026-06-02
fix: legacy deleverage action card overflowing

Pearce542:
The legacy `DetailInfoTradeRoutes` was making the action card overflow for the legacy deleverage tab. This is a minimal fix and a workaround since these component and tab are legacy

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
Screenshot 2026-06-02 at 18 13 46

</td>
    <td>
Screenshot 2026-06-02 at 18 07 46
</td>
  </tr>
</table>


0xAlunara (approved):
Who found this btw? If it's not a dev but a user I'm more interested on the reason why they're using legacy in the first place 🤔

0xtutti (approved):

PR #2639, opened 2026-06-03
fix: use on-chain rates

DanielSchiavini:
- use onchain rates for the future loan rates
- this should be more up-to-date and works also for LLv2
- update llamalend to make the useApi flag available https://github.com/curvefi/curve-llamalend.js/pull/106


Pearce542 (approved):
Nice

0xtutti (approved):

PR #2641, opened 2026-06-03
feat: new pool advanced details

0xAlunara:
This PR is merely a coat of paint and does not attempt to refactor the dependency on `poolDataCacheOrApi` and the many pool stores. That's for a later PR. What this PR does offer is:

1. A new advanced details card for the pool page, behind the beta feature flag.
2. Currency reserves fall back to the prices API is available (which it is not for lite networks)

A few examples:

1. Spark.fi PYUSD Reserve - `0xa632d59b9b804a956bfaa9b48af3a1b74808fc1f` - a basic pool
image

2. weETH/WETH-ng - `0xdb74dfdd3bb46be8ce6c33dc9d82777bcfc3ded5` - Pool with points campaign
image

3. TRICRV - `0x4ebdf703948ddcea3b11f675b4d1fba9d2414a14` - Pool with CRV rewards, 3 coins for market composition and a lot of parameters
image


apps/main/src/dex/features/advanced-details/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+            <Metrics
+              chainId={chainId}
+              poolDataCacheOrApi={poolDataCacheOrApi}
+              poolId={resolvedPoolId}
+              pricesApiPoolData={pricesApiPoolData}
+            />
+            <YieldBreakdown chainId={chainId} poolDataCacheOrApi={poolDataCacheOrApi} poolId={resolvedPoolId} />
+            <MarketComposition
+              chainId={chainId}
+              poolDataCacheOrApi={poolDataCacheOrApi}
```
0xAlunara, line 81: I started with the intention of not having to pass both `poolDataCacheOrApi` or `pricesApiPoolData` as props, but it got too hairy too quickly. This'll have to do for now so we can refactor it with proper hooks and queries later (so when we get rid of that pool mapper zustand store)

apps/main/src/dex/features/advanced-details/components/market-composition/columns/columns.definitions.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <Tooltip
+          title={maybe(
+            getValue(),
+            value =>
+              !isNaN(value) &&
+              `${value} ${row.original.symbol} ${maybe(row.original.tokenAmountUsd, value => ` / ${value}`) ?? ''}`,
+          )}
+          placement="top"
+        >
+          <Box>
```
0xAlunara: Forgot to add a comment like for the other table, but `Box` is necessary somehow to get `Tooltip` to work properly

apps/main/src/dex/components/PagePool/index.tsx
```diff
@@ -83,6 +85,10 @@ export const Transfer = (pageTransferProps: PageTransferProps) => {
   const snapshotData = snapshots?.[0]
 
   const pricesApiPoolData = poolData && pricesApiPoolsMapper?.[poolData.pool.address]
+  const managePoolPath = hasRefuelMethod(pricesApiPoolData?.poolMethods)
+    ? getPath(params, `${ROUTE.PAGE_POOLS}/${poolAddress}/refuel`)
```
DanielSchiavini: shouldn't we have a proper getRefuelPath helper?

apps/main/src/dex/components/PagePool/index.tsx
```diff
@@ -83,6 +85,10 @@ export const Transfer = (pageTransferProps: PageTransferProps) => {
   const snapshotData = snapshots?.[0]
 
   const pricesApiPoolData = poolData && pricesApiPoolsMapper?.[poolData.pool.address]
+  const managePoolPath = hasRefuelMethod(pricesApiPoolData?.poolMethods)
```
DanielSchiavini: the path is to the refuel page, not a manage pool page

apps/main/src/dex/components/PagePool/index.tsx
```diff
@@ -220,47 +226,59 @@ export const Transfer = (pageTransferProps: PageTransferProps) => {
...
-              </Button>
+        {isPoolAdvancedDetailsEnabled ? (
+          <AdvancedDetails
+            curve={curve}
+            routerParams={routerParams}
+            poolData={poolData}
+            poolDataCacheOrApi={poolDataCacheOrApi}
+            poolAlert={poolAlert}
+            pricesApiPoolData={pricesApiPoolData}
+            managePoolPath={managePoolPath}
```
DanielSchiavini: `AdvancedDetails` should be able to calculate the url by itself, why are we passing it as a prop?

apps/main/src/dex/features/advanced-details/components/market-composition/FooterRow.tsx
```diff
@@ -0,0 +1,31 @@
...
+  <>
+    <TableCell sx={{ paddingInline: Spacing.md, paddingBlock: Spacing.sm }}>
+      <Typography variant="tableCellMBold">{t`USD Total`}</Typography>
+    </TableCell>
+
+    {/** Empty address column */}
+    <TableCell />
+
+    <TableCell sx={{ paddingInline: Spacing.sm, paddingBlock: Spacing.sm, textAlign: 'right' }}>
+      <Typography variant="tableCellMBold">100%</Typography>
```
DanielSchiavini, line 22: It's hard to know what this means without any column info

apps/main/src/dex/features/advanced-details/components/market-composition/FooterRow.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { t } from '@ui-kit/lib/i18n'
+import { WithSkeleton } from '@ui-kit/shared/ui/WithSkeleton'
+import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
+import { amount, formatNumber } from '@ui-kit/utils'
+
+const { Spacing } = SizesAndSpaces
+
+export const FooterRow = ({ isLoading, totalUsd }: { isLoading: boolean; totalUsd: string | undefined }) => (
+  <>
+    <TableCell sx={{ paddingInline: Spacing.md, paddingBlock: Spacing.sm }}>
```
DanielSchiavini:
```suggestion
    <TableCell sx={{ padding: Spacing.md }}>
```

apps/main/src/dex/features/advanced-details/components/market-composition/FooterRow.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <Typography variant="tableCellMBold">{t`USD Total`}</Typography>
+    </TableCell>
+
+    {/** Empty address column */}
+    <TableCell />
+
+    <TableCell sx={{ paddingInline: Spacing.sm, paddingBlock: Spacing.sm, textAlign: 'right' }}>
```
DanielSchiavini, line 21:
```suggestion
    <TableCell sx={{ padding: Spacing.sm, textAlign: 'right' }}>
```

apps/main/src/dex/features/advanced-details/components/market-composition/FooterRow.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <Typography variant="tableCellMBold">100%</Typography>
+    </TableCell>
+
+    <TableCell sx={{ paddingInline: Spacing.md, paddingBlock: Spacing.sm, textAlign: 'right' }}>
```
DanielSchiavini, line 26:
```suggestion
    <TableCell sx={{ padding: Spacing.sm, textAlign: 'right' }}>
```

apps/main/src/dex/features/advanced-details/components/market-composition/FooterRow.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <WithSkeleton loading={isLoading} sx={{ justifySelf: 'end' }}>
+        <Typography variant="tableCellMBold">
+          {formatNumber(amount(totalUsd), { unit: 'dollar', abbreviate: true, fallback: '-' })}
```
DanielSchiavini: isn't `-` the default fallback?

apps/main/src/dex/features/advanced-details/components/market-composition/columns/columns.definitions.tsx
```diff
@@ -0,0 +1,100 @@
...
+  columnHelper.accessor('tokenAmount', {
+    id: MarketCompositionColumnId.TokenAmount,
+    header: headers[MarketCompositionColumnId.TokenAmount],
+    cell: ({ getValue, row }) => (
+      <InlineTableCell sx={{ alignItems: 'end' }}>
+        <Tooltip
+          title={maybe(
+            getValue(),
+            value =>
+              !isNaN(value) &&
```
DanielSchiavini: why do we expect NaN values?

apps/main/src/dex/features/advanced-details/components/market-composition/columns/columns.definitions.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { InlineTableCell } from '@ui-kit/shared/ui/DataTable/inline-cells/InlineTableCell'
+import { TokenInfo } from '@ui-kit/shared/ui/TokenInfo'
+import { Tooltip } from '@ui-kit/shared/ui/Tooltip'
+import { amount, formatNumber } from '@ui-kit/utils'
+import { MarketCompositionColumnId } from './columns.enum'
+
+export type MarketCompositionRow = TableItem & {
+  symbol: string
+  tokenAddress: string
+  tokenAddressUrl?: string
```
DanielSchiavini: why are there so many optional props?

apps/main/src/dex/features/advanced-details/components/yield-breakdown/columns/columns.definitions.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { InlineTableCell } from '@ui-kit/shared/ui/DataTable/inline-cells/InlineTableCell'
+import { TokenInfo, type TokenInfoProps } from '@ui-kit/shared/ui/TokenInfo'
+import { Tooltip } from '@ui-kit/shared/ui/Tooltip'
+import { formatPercent } from '@ui-kit/utils'
+import { YieldBreakdownColumnId } from './columns.enum'
+
+export type YieldBreakdownRow = TableItem & {
+  source: TokenInfoProps
+  address?: string
+  addressUrl?: string
```
DanielSchiavini: Why so many optional props?

apps/main/src/dex/features/advanced-details/components/yield-breakdown/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const columns = useMemo(() => createYieldBreakdownColumns({ showPointsMultiplier }), [showPointsMultiplier])
+  const table = useTable({
+    data: rows,
+    columns,
+    ...getTableOptions(rows),
+  })
+
+  if (rows.length === 0) return null
+
+  return (
```
DanielSchiavini:
```suggestion
  return rows.length > 0 && (
```

apps/main/src/dex/features/advanced-details/components/Alerts.tsx
```diff
@@ -0,0 +1,53 @@
...
+  poolAlert: PoolAlert | null
+  isCryptoPool: boolean
+  poolName: string
+  tokenAlert: PoolAlert | null
+}
+
+const PoolMessageAlert = ({ alert: { alertType, message } }: { alert: PoolAlert }) => (
+  <Alert
+    variant="filled"
+    severity={alertType === 'danger' || alertType === 'error' ? 'error' : alertType === 'warning' ? 'warning' : 'info'}
```
DanielSchiavini: This is similar to:
https://github.com/curvefi/curve-frontend/blob/d0a3b10116d29d2e72d9f87b6e6c8f596c1c5218/apps/main/src/llamalend/widgets/banners/MarketAlertBanner.tsx#L11
https://github.com/curvefi/curve-frontend/blob/615b69209a316909b26f485d8f2b204d6b17111a/apps/main/src/dex/components/PoolAlertBanner.tsx#L6
https://github.com/curvefi/curve-frontend/blob/d5ee15f0d9f01a9a04409e3d6425316bfc3f57c8/apps/main/src/dex/utils/alerts.tsx#L16

apps/main/src/dex/features/advanced-details/components/Metrics.tsx
```diff
@@ -0,0 +1,66 @@
...
+}) => {
+  const { tvl, volume, gaugeTotalSupply, totalStakedPercent } = useMetrics({
+    chainId,
+    poolDataCacheOrApi,
+    poolId,
+    pricesApiPoolData,
+  })
+
+  return (
+    <Grid container columnSpacing={Spacing.md} rowSpacing={Spacing.md}>
```
DanielSchiavini:
```suggestion
    <Grid container spacing={Spacing.md}>
```

apps/main/src/dex/features/advanced-details/components/Parameters.tsx
```diff
@@ -0,0 +1,197 @@
...
+    metadata,
+    network,
+    poolAddress,
+    poolType,
+    snapshotData,
+    virtualPrice,
+  } = useParameters({ chainId, poolDataCacheOrApi, poolId })
+
+  const rampADetails =
+    initial_A && future_A_time && future_A
```
DanielSchiavini: This should trigger a lint warning for `maybes` usage 😞

apps/main/src/dex/features/advanced-details/components/Parameters.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+          {(A != null || snapshotData?.a != null) && (
+            <ActionInfo
+              label={t`A factor`}
+              value={formatADisplay(A ?? snapshotData?.a ?? undefined)}
+              valueTooltip={
+                <Stack sx={{ gap: Spacing.sm }}>
+                  {t`Amplification coefficient chosen from fluctuation of prices around 1.`}
+                  {rampADetails?.isFutureATimePassedToday &&
+                    initial_A_time != null &&
```
DanielSchiavini: maybes again

apps/main/src/dex/features/advanced-details/hooks/useMarketComposition.ts
```diff
@@ -0,0 +1,73 @@
...
+}: {
+  chainId: ChainId
+  poolDataCacheOrApi: PoolDataCacheOrApi
+  poolId: string
+  pricesApiPoolData?: PricesApiPool
+}) => {
+  const { data: network } = useNetworkByChain({ chainId })
+  const currencyReserves = useStore(state => state.pools.currencyReserves[getChainPoolIdActiveKey(chainId, poolId)])
+
+  const usePricesApiReserves = isNaN(Number(currencyReserves?.total)) && !network.isLite
```
DanielSchiavini, line 26: Do we use prices api while loading? Do we really expect NaNs?

apps/main/src/dex/features/advanced-details/hooks/useMarketComposition.ts
```diff
@@ -0,0 +1,73 @@
...
+  const pricesApiTotalUsd = pricesApiPoolData?.balancesUsd.reduce((total, balanceUsd) => total + balanceUsd, 0)
```
DanielSchiavini: sum pls 😭

apps/main/src/dex/features/advanced-details/hooks/useMarketComposition.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  const rows = useMemo<MarketCompositionRow[]>(
+    () =>
+      poolDataCacheOrApi.tokens
+        .map((symbol, index) => {
+          const tokenAddress = poolDataCacheOrApi.tokenAddresses[index]
+          const reserve = currencyReserves?.tokens.find(
+            token => token.tokenAddress.toLowerCase() === tokenAddress.toLowerCase(),
+          )
+          const balance = usePricesApiReserves ? pricesApiPoolData?.balances[index] : reserve?.balance
```
DanielSchiavini: do we need to check usePricesApiReserves 4 times for every item!? I'd expect to see something similar to `poolDataCacheOrApi`, i.e. one list that comes from either source

apps/main/src/dex/features/advanced-details/hooks/useMetrics.ts
```diff
@@ -0,0 +1,29 @@
+import { usePoolTotalStaked } from '@/dex/hooks/usePoolTotalStaked'
+import { usePoolTvl } from '@/dex/queries/pool-tvl.query'
+import { usePoolVolume } from '@/dex/queries/pool-volume.query'
+import type { ChainId, PoolDataCacheOrApi } from '@/dex/types/main.types'
+import type { Pool as PricesApiPool } from '@curvefi/prices-api/pools'
+import { maybe } from '@primitives/objects.utils'
+
+export const useMetrics = ({
```
DanielSchiavini, line 8: What kind of metrics?

apps/main/src/dex/features/advanced-details/hooks/useYieldBreakdown.tsx
```diff
@@ -0,0 +1,124 @@
...
+  })
+
+  // Construct all yield rows imperatively rather than functional to improve readability
+  const rows = useMemo(() => {
+    const rows: YieldBreakdownRow[] = []
+    const crvBase = rewardsApy?.crv?.[0]
+    const crvMax = rewardsApy?.crv?.[1] ?? crvBase
+    const haveCrv = Number(crvBase) > 0 || Number(crvMax) > 0
+
+    if (haveCrv) {
```
DanielSchiavini: We might not even need the `some`
```suggestion
    if (rewardsApy?.crv?.some(Boolean)) {
      const crvBase = rewardsApy.crv[0]
      const crvMax = rewardsApy.crv[1] ?? crvBase
```

apps/main/src/dex/features/advanced-details/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+            </Button>
+          )
+        }
+      />
+      <CardContent
+        size="inline"
+        component={Grid}
+        container
+        columnSpacing={Spacing.md}
+        rowSpacing={Spacing.md}
```
DanielSchiavini:
```suggestion
        spacing={Spacing.md}
```

apps/main/src/dex/components/PagePool/index.tsx
0xAlunara: Forward thinking; it's refuel only for now but tutti wants it to become a fully fledged pool management page, hence the link description.

DanielSchiavini: Then we should change the url

apps/main/src/dex/features/advanced-details/components/yield-breakdown/columns/columns.definitions.tsx
0xAlunara: All thanks to prices api types being optional and/or nullable

apps/main/src/dex/features/advanced-details/components/market-composition/FooterRow.tsx
0xAlunara: Not the same 🤓

apps/main/src/dex/features/advanced-details/components/market-composition/FooterRow.tsx
0xAlunara: It is if we use the new formatter category

apps/main/src/dex/features/advanced-details/components/market-composition/columns/columns.definitions.tsx
0xAlunara: I didn't at first but it showed up during testing. It's a value curve-js returns when there's no wallet connected.

apps/main/src/dex/features/advanced-details/components/market-composition/columns/columns.definitions.tsx
0xAlunara: All thanks to prices api types being optional and/or nullable

apps/main/src/dex/features/advanced-details/hooks/useMarketComposition.ts
0xAlunara: We use prices API as a fallback for non-lite networks, and `.total` is `NaN` when no wallet is connected. Added commetnf or clarification.

apps/main/src/dex/features/advanced-details/hooks/useMetrics.ts
0xAlunara: Just a bunch of generic metrics, couldn't find a better name. Prone to change anyway

apps/main/src/dex/features/advanced-details/hooks/useYieldBreakdown.tsx
0xAlunara: I don't trust it being an array, so I'll stick to just `if (crvBase || crvMax)`

0xAlunara:
Processed the review comments and did a resync with tutti. A couple of big changes:

* Points campaigns are now separate from the yield breakdown.
* Asset / Source cells now show a copyable address as secondary value, and show a "View on explorer" tooltip on hover of entire cell.
* Token prices have been moved to their own column and the address column is gone.
* Yield breakdown has token prices fallback as for some reason curve-js does not always include token price. This might lead to slight differences in token prices compared to prices in the market compensation table. There's no easy way to fix this without a refactor in curve-js. Might fix in a later PR if it's really bothersome, but for now it's an edge case that only pops up when a pool is rewarding yield with one of the pool's underlying tokens. A fix would be to disregard all curve-js prices and use our own token rates query in both places.
* APRs are now shown as APYs based on weekly compounding.

The column width deviate slightly so all 3 tables aren't neatly aligned if you have OCD, but that's out of scope as it requires some DataTable refactoring.

Mobile can be a bit scuffed, but want to fix that in a follow up PR.

The "15+x" in the screenshot below is because somebody did not correctly enter the multiplier data!

image

apps/main/src/dex/features/advanced-details/components/market-composition/columns/columns.definitions.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+      return (
+        <InlineTableCell sx={{ alignItems: 'end' }}>
+          <Tooltip
+            title={maybe(
+              getValue(),
+              value =>
+                !isNaN(value) &&
+                typeof symbol === 'string' &&
+                `${value} ${symbol} ${maybe(row.original.amountUsd, value => ` / ${value}`) ?? ''}`,
```
DanielSchiavini: Why would we expect `symbol` to be non-string?
If the NaN is really expected please add a comment

apps/main/src/dex/features/advanced-details/components/market-composition/columns/columns.definitions.tsx
```diff
@@ -0,0 +1,95 @@
...
+            )}
+            placement="top"
+          >
+            {/** Needed for tooltip to work for whatever reason */}
+            <Box>
+              <TokenInfo
+                icon={null}
+                iconPosition="right"
+                primary={formatNumber(getValue(), 'token.compact')}
+                secondary={maybe(row.original.amountUsd, x => formatNumber(x, 'usd.notional'))}
```
DanielSchiavini: no loading state or fallback for this?

apps/main/src/dex/features/advanced-details/components/market-composition/columns/columns.definitions.tsx
```diff
@@ -0,0 +1,95 @@
...
+    header: headers[MarketCompositionColumnId.Asset],
+    cell: ({ getValue, row }) => <TokenCell source={getValue()} explorerUrl={row.original.explorerUrl} />,
+    enableSorting: false,
+  }),
+  columnHelper.accessor('price', {
+    id: MarketCompositionColumnId.Price,
+    header: headers[MarketCompositionColumnId.Price],
+    cell: ({ getValue }) => (
+      <InlineTableCell>
+        <Typography variant="tableCellMRegular">{formatNumber(getValue(), 'usd.amount')}</Typography>
```
DanielSchiavini: This should just be a standard UsdAmountColumn, why do we need to create a new column just for this table? Same about the PercentRateColumn below.

apps/main/src/dex/features/advanced-details/hooks/useYieldBreakdown.tsx
DanielSchiavini: Why not? I don't think we should handle unexpected stuff

apps/main/src/dex/features/advanced-details/hooks/useMarketComposition.ts
DanielSchiavini: Can we fix this at the prices api or handle the bad values in the prices API interface?

apps/main/src/dex/features/advanced-details/components/points-campaigns/columns/columns.definitions.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <TokenInfo {...getValue()} boldPrimary />
+      </InlineTableCell>
+    ),
+    enableSorting: false,
+  }),
+  columnHelper.accessor('points', {
+    id: PointsCampaignsColumnId.Points,
+    header: headers[PointsCampaignsColumnId.Points],
+    cell: ({ getValue }) => (
+      <InlineTableCell sx={{ alignItems: 'end' }}>
```
DanielSchiavini: `type: numeric` was meant to align stuff to the right, already
This should just be a generic column imo, there's nothing market-specific to it

apps/main/src/dex/features/advanced-details/components/points-campaigns/columns/columns.definitions.tsx
```diff
@@ -0,0 +1,57 @@
...
+  points: string
+  campaignUrl: string
+}
+
+const columnHelper = createColumnHelper<PointsCampaignsRow>()
+
+const headers = {
+  [PointsCampaignsColumnId.Source]: t`Source`,
+  [PointsCampaignsColumnId.Points]: t`Points`,
+  [PointsCampaignsColumnId.CampaignUrl]: t`Link`,
```
DanielSchiavini: It's a bit of a weird pattern to have a 'link' column 🤔

apps/main/src/dex/features/advanced-details/components/Metrics.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+      <Grid size={METRIC_GRID_SIZE}>
+        <Metric size="medium" label={t`TVL`} value={tvl} valueOptions={{ unit: 'dollar', abbreviate: true }} />
+      </Grid>
+
+      <Grid size={METRIC_GRID_SIZE}>
+        <Metric
+          size="medium"
+          label={t`Liquidity utilization`}
+          value={maybes([tvl, volume], ([tvl, volume]) => (tvl === 0 ? 0 : (volume / tvl) * 100))}
```
DanielSchiavini:
```suggestion
          value={maybes([tvl, volume], ([tvl, volume]) => (tvl && (volume / tvl) * 100))}
```

apps/main/src/dex/features/advanced-details/components/Parameters.tsx
```diff
@@ -0,0 +1,207 @@
...
+  return (
+    <Card size="inline">
+      <CardHeader title={t`Parameters`} />
+      <CardContent component={Stack}>
+        <Section>
+          <ActionInfo
+            label={t`AMM fee`}
+            loading={isLoadingParameters}
+            value={formatNumber(amount(fee ?? snapshotData?.fee), {
+              maximumFractionDigits: 4,
```
DanielSchiavini, line 69: please don't hardcode magic numbers, we need to have consistent formatting 😭

apps/main/src/dex/features/advanced-details/components/Parameters.tsx
```diff
@@ -0,0 +1,207 @@
...
+        <Section>
+          {gamma && (
+            <ActionInfo
+              label={t`Gamma`}
+              value={formatNumber(amount(gamma), { useGrouping: false, abbreviate: false, fallback: '-' })}
+            />
+          )}
+
+          {maybe(snapshotData?.feeGamma, x => (
+            <ActionInfo label={t`Fee Gamma`} value={formatParam(x / 10 ** 18)} />
```
DanielSchiavini, line 160: `/10**18` does not belong in components imo and shouldn't be repeated

apps/main/src/dex/features/advanced-details/components/Parameters.tsx
```diff
@@ -0,0 +1,207 @@
...
+          ))}
+
+          {maybe(snapshotData?.allowedExtraProfit, x => (
```
DanielSchiavini, line 163: should all the parameters be hidden while loading? I think it's nice to have a skeleton?

apps/main/src/dex/features/advanced-details/hooks/useYieldBreakdown.tsx
```diff
@@ -0,0 +1,132 @@
...
+          blockchainId: 'ethereum',
+          iconPosition: 'left',
+          primary: 'CRV',
+        },
+        address: MAINNET_CRV_ADDRESS,
+        explorerUrl: scanTokenPath(defaultNetworks[Chain.Ethereum], MAINNET_CRV_ADDRESS),
+        price: crvPrice,
+        apy: gaugeIsKilled ? undefined : crvMax,
+        apySecondary: gaugeIsKilled ? undefined : crvBase,
+        apyTooltip: gaugeIsKilled ? undefined : t`Max CRV APY can be reached with max boost for this pool.`,
```
DanielSchiavini:
```suggestion
        ...!gaugeIsKilled && {
          apy: crvMax,
          apySecondary: crvBase,
          apyTooltip: t`Max CRV APY can be reached with max boost for this pool.`,
        }
```

apps/main/src/dex/features/advanced-details/components/market-composition/columns/columns.definitions.tsx
0xAlunara: No, this is a trap, we already havr like 3 different usd cells that started out with the intention of being generic cells when they suddenly become niche cells due to tooltips and the surrounding context and such. At this point I'd rather just have a normal cell and call the correct `formatNumber` and be done with it.

apps/main/src/dex/features/advanced-details/components/market-composition/columns/columns.definitions.tsx
0xAlunara: I can remove the `maybe` so the default fallback is `-` from `formatNumber(x, 'usd.notional')`

apps/main/src/dex/features/advanced-details/components/market-composition/columns/columns.definitions.tsx
0xAlunara: Removed the NaN check, maybe that was a thing in earlier development. Added docs explaining symbol is a ReactNode as that's how TokenInfo works in orde to support customized stuff like underline on hover and such.

apps/main/src/dex/features/advanced-details/hooks/useMarketComposition.ts
0xAlunara: This isn't about the prices API, this is about curve-js returning `NaN` if you try to fetch the `currencyReserves` without a connected wallet. I prefer this check over `isConnected` from `useWallet`, because at some point I hope curve-js is refactored into not having that wallet requirement and then it'll still work. It's a refactor for later, out of scope for now.

apps/main/src/dex/features/advanced-details/components/points-campaigns/columns/columns.definitions.tsx
0xAlunara: I've renamed it to 'Details' now, sounds a bit less technical

apps/main/src/dex/features/advanced-details/components/points-campaigns/columns/columns.definitions.tsx
0xAlunara: Was able to remove it, but not in all cases, for example if there's an inner stack like in `yield-breakdown`.

apps/main/src/dex/features/advanced-details/components/Parameters.tsx
0xAlunara: Legacy shit, I want to give this another pass when this and the number categorization are both merged

apps/main/src/dex/features/advanced-details/components/Parameters.tsx
0xAlunara: I argue it should as it's very specific to these parameters. See how there's a mix of `10**8`, `10**10` and `10**18`. We shouldn't generalize contract specific domain logic.

apps/main/src/dex/features/advanced-details/components/Parameters.tsx
0xAlunara: We can't know beforehand which parameters will pop up with a value or not aside from a few, sadly. A big skeleton block wouldn't look better in my view.

apps/main/src/dex/features/advanced-details/components/market-composition/columns/columns.definitions.tsx
DanielSchiavini: Ok I understand - do realize this will inevitably lead to UI inconsistencies

apps/main/src/dex/features/advanced-details/hooks/useYieldBreakdown.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const rewardsApy = useStore(state => state.pools.rewardsApyMapper[chainId]?.[poolId])
+
+  const { data: campaigns } = useCampaignsByAddress({
+    blockchainId: network?.networkId as BlockchainId | undefined,
+    address: poolAddress,
+  })
+
+  // For some reason it might be curve-js returns no token price, so we'll try to fall back to our own token rates query.
+  const missingTokenRates = useMemo(
+    () =>
```
DanielSchiavini, line 45: notFalsyArray

apps/main/src/dex/features/advanced-details/hooks/useYieldBreakdown.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const { data: fallbackTokenRates } = useTokenUsdRates(
+    { chainId, tokenAddresses: missingTokenRates },
+    missingTokenRates.length > 0,
+  )
+
+  const { data: crvPrice } = useTokenUsdRate({ chainId: Chain.Ethereum, tokenAddress: MAINNET_CRV_ADDRESS })
+
+  // Construct all yield rows imperatively rather than functional to improve readability
+  const rows: YieldBreakdownRow[] = useMemo(() => {
+    const rows: YieldBreakdownRow[] = []
```
DanielSchiavini, line 58: This would become simpler with a single `return notFalsy`

DanielSchiavini (approved):

apps/main/src/dex/features/advanced-details/hooks/useYieldBreakdown.tsx
DanielSchiavini: I see the comment two lines above, I do not agree 😂 but np

OnlyJousting (approved):

PR #2642, opened 2026-06-03
feat: add semantic number format categories

0xAlunara:
Add `NUMBER_FORMAT_CATEGORIES` and a category overload for `formatNumber` so common number formatting presets can be reused by semantic intent.

* Migrate existing `formatUsd` and `formatPercent` usages to category-based `formatNumber` calls, then remove the legacy wrappers.
* Add category coverage for USD, percentage, token amount, compact token amount, and token balance formats, including fallback behavior and token balance significant-digit handling.

Categories added:
image

`token.balance` is complicated because "just use 5 significant digets" is not as easy as it sounds. For example when you have 123,456,789 max sig digits of 5 as it'll be rounded to 123,450,000. But you do want 0.00001234567 to round to 0.000012346. Check the test cases in `number.spec.ts` to see what's covered.
image

### DIFF
--- a/apps/main/src/dex/features/refuel/components/DailyRefuelsChart.tsx
+++ b/apps/main/src/dex/features/refuel/components/DailyRefuelsChart.tsx
@@ -20,7 +20,7 @@ import { useSwitch } from '@ui-kit/hooks/useSwitch'
 import { t } from '@ui-kit/lib/i18n'
 import type { LegendItem } from '@ui-kit/shared/ui/Chart/LegendSet'
 import { SelectTimeOption } from '@ui-kit/shared/ui/Chart/SelectTimeOption'
-import { formatNumber, formatUsd } from '@ui-kit/utils'
+import { formatNumber } from '@ui-kit/utils'
 import { useRefuelDailyRefuels } from '../queries/daily-refuels.query'
 
 const REFUELS_LABEL = t`Daily refuels`
@@ -30,7 +30,9 @@ const REFUELS_VALUE_KEY = 'totalUsd'
 const REFUELS_COUNT_KEY = 'count'
 
 const formatDonationTooltip = (value: number, seriesName: string) =>
-  seriesName === REFUELS_COUNT_LABEL ? formatNumber(value, { abbreviate: false, decimals: 0 }) : formatUsd(value)
+  seriesName === REFUELS_COUNT_LABEL
+    ? formatNumber(value, { abbreviate: false, decimals: 0 })
+    : formatNumber(value, 'usd.notional')
 
 export const DailyRefuelsChart = ({ blockchainId, poolAddress }: { blockchainId: Chain; poolAddress: Address }) => {
   const [period, setPeriod] = useState<(typeof PERIODS)[number]>('6m')
@@ -96,7 +98,7 @@ export const DailyRefuelsChart = ({ blockchainId, poolAddress }: { blockchainId:
             ...yAxisBase,
             position: 'right',
             splitLine: { lineStyle: { color: palette.gridLinesColor } },
-            axisLabel: { ...yAxisBase.axisLabel, formatter: (value: number) => formatUsd(value) },
+            axisLabel: { ...yAxisBase.axisLabel, formatter: (value: number) => formatNumber(value, 'usd.notional') },
           },
           {
             type: 'value',
--- a/packages/curve-ui-kit/src/utils/number.spec.ts
+++ b/packages/curve-ui-kit/src/utils/number.spec.ts
@@ -500,7 +500,9 @@ describe('formatNumber', () => {
     })
 
     it('formats percent categories', () => {
+      expect(formatNumber(1200.345, 'percent.value')).toBe('1,200.35%')
       expect(formatNumber(12.345, 'percent.value')).toBe('12.35%')
+      expect(formatNumber(1200.345, 'percent.rate')).toBe('1.20k%')
       expect(formatNumber(12.3, 'percent.rate')).toBe('12.30%')
       expect(formatNumber(0.1234, 'percent.rate')).toBe('0.12%')
     })
--- a/packages/curve-ui-kit/src/utils/number.spec.ts
+++ b/packages/curve-ui-kit/src/utils/number.spec.ts
@@ -9,7 +9,6 @@ import {
   scaleSuffix,
   log10Exp,
   getFractionDigitsOptions,
-  NUMBER_FORMAT_CATEGORIES,
 } from './number'
 
 describe('getFractionDigitsOptions', () => {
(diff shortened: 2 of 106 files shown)


0xtutti (approved):

packages/curve-ui-kit/src/utils/number.spec.ts
```diff
@@ -474,6 +482,52 @@ describe('formatNumber', () => {
...
+      expect(formatNumber(-123456789.1234, 'token.balance')).toBe('-123,456,789.12')
+    })
+
+    it('formats USD categories', () => {
+      expect(formatNumber(1234567, 'usd.notional')).toBe('$1.23m')
+      expect(formatNumber(1234.56, 'usd.amount')).toBe('$1,234.56')
+    })
+
+    it('formats percent categories', () => {
+      expect(formatNumber(1200.345, 'percent.value')).toBe('1,200.35%')
```
Pearce542, line 516: I would have expected the `percent.value` category to format a percentage value like this `0.1 => 10%` so both `x 100` and adding a `%`, instead of `0.1 => 0.1%`. But I guess it has already been like this

packages/curve-ui-kit/src/utils/number.ts
```diff
@@ -176,6 +180,46 @@ export type NumberFormatOptions = {
...
 
+const TOKEN_BALANCE_SIGNIFICANT_DIGITS = 5
+
+const NUMBER_FORMAT_CATEGORIES = {
+  'token.amount': { abbreviate: false, fallback: '-' },
+  'token.compact': { abbreviate: true, fallback: '-' },
+  'token.balance': {
+    abbreviate: false,
+    formatter: (value: Amount) => {
+      const absValue = Math.abs(Number(value))
```
Pearce542, line 191: shouldn't we use bigNumber?

packages/curve-ui-kit/src/utils/number.ts
```diff
@@ -176,6 +180,46 @@ export type NumberFormatOptions = {
...
+      const options: Partial<NumberFormatOptions> | undefined =
+        absValue > 0 && absValue < 1
+          ? { maximumSignificantDigits: TOKEN_BALANCE_SIGNIFICANT_DIGITS, trailingZeroDisplay: 'auto' }
+          : absValue >= 1 && Number.isFinite(absValue)
+            ? {
+                maximumFractionDigits: Math.max(
+                  2,
```
Pearce542, line 198: magic number ?

packages/curve-ui-kit/src/utils/number.ts
```diff
@@ -176,6 +180,46 @@ export type NumberFormatOptions = {
...
+                  TOKEN_BALANCE_SIGNIFICANT_DIGITS - Math.floor(Math.log10(absValue)) - 1,
```
Pearce542, line 199: maybe a comment here cna be helpful

Pearce542 (approved):
Nice! You mentioned that not all use cases were covered. Which one is left?

apps/main/src/dex/features/refuel/components/DailyRefuelsChart.tsx
```diff
@@ -30,7 +30,9 @@ const REFUELS_VALUE_KEY = 'totalUsd'
 const REFUELS_COUNT_KEY = 'count'
 
 const formatDonationTooltip = (value: number, seriesName: string) =>
-  seriesName === REFUELS_COUNT_LABEL ? formatNumber(value, { abbreviate: false, decimals: 0 }) : formatUsd(value)
+  seriesName === REFUELS_COUNT_LABEL
+    ? formatNumber(value, { abbreviate: false, decimals: 0 })
```
DanielSchiavini, line 34: Can we get rid of the custom formatting options soon?

packages/curve-ui-kit/src/utils/number.ts
```diff
@@ -176,6 +180,46 @@ export type NumberFormatOptions = {
...
+  'token.amount': { abbreviate: false, fallback: '-' },
+  'token.compact': { abbreviate: true, fallback: '-' },
+  'token.balance': {
+    abbreviate: false,
+    formatter: (value: Amount) => {
+      const absValue = Math.abs(Number(value))
+      const options: Partial<NumberFormatOptions> | undefined =
+        absValue > 0 && absValue < 1
+          ? { maximumSignificantDigits: TOKEN_BALANCE_SIGNIFICANT_DIGITS, trailingZeroDisplay: 'auto' }
+          : absValue >= 1 && Number.isFinite(absValue)
```
DanielSchiavini, line 195: Do we really expect infinite? In that case I'd prefer to see the mistake than hide it
```suggestion
          : absValue >= 1
```

packages/curve-ui-kit/src/utils/number.ts
```diff
@@ -176,6 +180,46 @@ export type NumberFormatOptions = {
   fallback?: string
 } & Omit<Intl.NumberFormatOptions, 'unit' | 'style' | 'compact' | 'notation' | 'currency'>
 
+const TOKEN_BALANCE_SIGNIFICANT_DIGITS = 5
+
+const NUMBER_FORMAT_CATEGORIES = {
```
DanielSchiavini, line 185: IMO these should be the number format types akin to the [query types](https://github.com/curvefi/curve-frontend/blob/0985a94293f2060f28e3be51cf978676560413d5/packages/curve-ui-kit/src/lib/model/query/query-types.ts#L29) that we have.
The categories should be a lot more specific, maybe even more specific than the [query categories](https://github.com/curvefi/curve-frontend/blob/0985a94293f2060f28e3be51cf978676560413d5/packages/curve-ui-kit/src/lib/model/query/query-categories.ts#L11)

DanielSchiavini (approved):
This is a huge improvement, I hope we can get rid of all custom formatters and only use these

packages/curve-ui-kit/src/utils/number.ts
Pearce542: agree with this as well

OnlyJousting (approved):
😍

0xAlunara:
> Nice! You mentioned that not all use cases were covered. Which one is left?

Basically any usage of `formatNumber` that's left with their own custom options. I'll tackle these issues in follow up PRs to minimize conflicts.

packages/curve-ui-kit/src/utils/number.ts
0xAlunara: Note to self: I should prob extract a `const DEFAULT_DECIMALS = 2` out of `defaultNumberFormatter`

apps/main/src/dex/features/refuel/components/DailyRefuelsChart.tsx
0xAlunara: Most likely, after we kill off the rest of the stragglers

packages/curve-ui-kit/src/utils/number.ts
0xAlunara: The formatting function itself doesn't really work with bigints anyway. I don't think we format actual raw bigints anywhere and probably shouldn't anyway 🤔

PR #2644, opened 2026-06-04
feat: pool-list v2 using prices-api data

OnlyJousting:
This PR adds the new prices-api v2 pool list as data source for the dex pool-list page.

Changes:
- Use prices-api v2 pool-list endpoint via react-query
- Use API based filtering, searching and pagination
- Use `DataTable` in new pool list
- Add v2 pools client/schema coverage in `@curvefi/prices-api`
- Add new cell components
- Preserve the old pool list behind `Legacy*` modules for unsupported chains and in `stable` release channel
- Update e2e tests and add pool-list mock data

Notes:
- ~~There is a token blacklist maintained in the front-end, I just added that filtering on top of the 50 paginated results returned from the API. I'm not so sure if this is how we want this to work as it screws a little bit with pool counts, pool counts when filtering etc.~~ _all blacklisting will be done on the API_
- It's clunky having to use pool ids in order to map to user positions, when curve-js gets migrated to use the same pools endpoint we can get rid of this clunkiness. 
- There is currently no "My positions" filter like the old UI has, it's not possible until we get an API based version, I think passing an array of pools the front-end identifies as pools the user has positions in would work
- The API only returns yields in APR while the old list/the table declares in APY, APY fields will be added to the API


apps/main/src/dex/features/pool-list/apiPoolList.utils.ts
```diff
@@ -0,0 +1,113 @@
...
+import { fromEntries, notFalsy, recordValues } from '@primitives/objects.utils'
+import { DEX_ROUTES } from '@ui-kit/shared/routes'
+import { decimal } from '@ui-kit/utils'
+import type { PoolListItem } from './types'
+
+const VYPER_EXPLOIT_VERSIONS = new Set(['0.2.15', '0.2.16', '0.3.0'])
+
+export type PoolIdByAddressSource = Record<string, { pool: Pick<PoolData['pool'], 'address' | 'id'> }>
+export type PoolListItemOptions = {
+  hasPosition?: boolean
```
DanielSchiavini: why would these be optional?

apps/main/src/dex/features/pool-list/PoolListApiTable.tsx
```diff
@@ -0,0 +1,220 @@
...
+      emptyState={
+        <EmptyStateRow table={table}>
+          <PoolListEmptyState columnFiltersById={emptyStateFilters} resetFilters={resetFilters} isError={isError} />
+        </EmptyStateRow>
+      }
+      expandedPanel={PoolMobileExpandedPanel}
+      shouldStickFirstColumn={isTablet}
+      loading={isLoading}
+    >
+      <LegacyTableFilters<PoolColumnId>
```
DanielSchiavini: Why are we using the legacy table?

apps/main/src/dex/features/pool-list/PoolListApiTable.tsx
```diff
@@ -0,0 +1,220 @@
...
+      searchNavigate(
+        { sort: apiSorting.map(({ id, desc }) => `${desc ? '-' : ''}${id}`), page: null },
+        { replace: true },
+      )
+    },
+    [defaultSort, searchNavigate, sorting],
+  )
+  const onPaginationChange: OnChangeFn<PaginationState> = useCallback(
+    updater => {
+      const nextPagination = typeof updater === 'function' ? updater(pagination) : updater
```
DanielSchiavini: this does not belong in a component, it is already in usePageFromQueryString

apps/main/src/dex/features/pool-list/PoolListApiTable.tsx
```diff
@@ -0,0 +1,220 @@
...
+    (type: PoolTypeFilter | null) => searchNavigate({ pool_type: type, page: null }, { replace: true }),
+    [searchNavigate],
+  )
+  const resetFilters = useCallback(
+    () => searchNavigate({ search: null, pool_type: null, page: null }, { replace: true }),
+    [searchNavigate],
+  )
+  const onSortingChange: OnChangeFn<SortingState> = useCallback(
+    updater => {
+      const nextSorting = typeof updater === 'function' ? updater(sorting) : updater
```
DanielSchiavini: This does not belong in a component, we have useSortFromQueryString

apps/main/src/dex/features/pool-list/PoolListApiTable.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const setSearch = useCallback(
+    (value: string) => searchNavigate({ search: value || null, page: null }, { replace: true }),
+    [searchNavigate],
+  )
+  const setPoolType = useCallback(
+    (type: PoolTypeFilter | null) => searchNavigate({ pool_type: type, page: null }, { replace: true }),
+    [searchNavigate],
+  )
+  const resetFilters = useCallback(
+    () => searchNavigate({ search: null, pool_type: null, page: null }, { replace: true }),
```
DanielSchiavini: this will break when we add new filters

apps/main/src/dex/features/pool-list/PoolListApiTable.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const searchText = searchParams.get('search') ?? ''
+  const poolTypeParam = searchParams.get('pool_type')
+  const poolType = isPoolType(poolTypeParam) ? poolTypeParam : undefined
+
+  const setSearch = useCallback(
+    (value: string) => searchNavigate({ search: value || null, page: null }, { replace: true }),
+    [searchNavigate],
+  )
+  const setPoolType = useCallback(
+    (type: PoolTypeFilter | null) => searchNavigate({ pool_type: type, page: null }, { replace: true }),
```
DanielSchiavini: we shouldn't be resetting the page in every single function that's a lot of duplication

apps/main/src/dex/features/pool-list/PoolListApiTable.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const sortDirection = sorting[0].desc ? 'desc' : 'asc'
+  const [urlPagination, setUrlPagination] = usePageFromQueryString(PER_PAGE)
+  const pagination = useMemo(
+    () => ({
+      pageIndex: Number.isFinite(urlPagination.pageIndex) ? Math.max(urlPagination.pageIndex, 0) : 0,
+      pageSize: urlPagination.pageSize,
+    }),
+    [urlPagination.pageIndex, urlPagination.pageSize],
+  )
+  const searchText = searchParams.get('search') ?? ''
```
DanielSchiavini: please don't use `?? ''`

apps/main/src/dex/features/pool-list/PoolListApiTable.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const [urlSorting] = useSortFromQueryString(defaultSort)
+  const sorting = useMemo(() => getApiSorting(urlSorting, defaultSort), [defaultSort, urlSorting])
+  const sortColumn = sorting[0].id as keyof typeof API_SORT_FIELDS
+  const sortField = sortColumn as PoolColumnId
+  const sortBy = API_SORT_FIELDS[sortColumn]
+  const sortDirection = sorting[0].desc ? 'desc' : 'asc'
+  const [urlPagination, setUrlPagination] = usePageFromQueryString(PER_PAGE)
+  const pagination = useMemo(
+    () => ({
+      pageIndex: Number.isFinite(urlPagination.pageIndex) ? Math.max(urlPagination.pageIndex, 0) : 0,
```
DanielSchiavini: Do we expect infinite??

apps/main/src/dex/features/pool-list/PoolListApiTable.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const getUserPositionOptions = usePoolListApiUserPositionOptions(network.chainId)
+  const searchParams = useSearchParams()
+  const searchNavigate = useSearchNavigate(searchParams)
+  const defaultSort = useMemo<SortingState>(() => getDefaultSort(network.isLite), [network.isLite])
+  const [urlSorting] = useSortFromQueryString(defaultSort)
+  const sorting = useMemo(() => getApiSorting(urlSorting, defaultSort), [defaultSort, urlSorting])
+  const sortColumn = sorting[0].id as keyof typeof API_SORT_FIELDS
+  const sortField = sortColumn as PoolColumnId
+  const sortBy = API_SORT_FIELDS[sortColumn]
+  const sortDirection = sorting[0].desc ? 'desc' : 'asc'
```
DanielSchiavini: mapping pls

apps/main/src/dex/features/pool-list/Page.tsx
```diff
@@ -1,8 +1,36 @@
...
+  return isBetaPoolListEnabled ? <BetaPoolListPage /> : <StablePoolListPage />
+}
+
+const StablePoolListPage = () => {
   const network = useNetworkFromUrl()
+
   return <ListPageWrapper>{network && <PoolListTable network={network} />}</ListPageWrapper>
 }
+
+const BetaPoolListPage = () => {
```
DanielSchiavini: We should rename the old one to legacy, and change the non-legacy files so we can see the diff in the new list page

apps/main/src/dex/features/pool-list/Page.tsx
```diff
@@ -1,8 +1,36 @@
...
+  /**
+   * Prices API v2 does not cover every full DEX network that's supplied by the curve-api.
+   * Unsupported: Moonbeam, Kava, Avalanche, Celo, Aurora, X-Layer, zkSync, Mantle.
+   */
+  const isSupported = supportedPoolChains?.some(({ chainId }) => chainId === network?.chainId)
+
+  return (
+    <ListPageWrapper>
+      {network &&
+        !isLoading &&
```
DanielSchiavini: Skeletons pls

apps/main/src/dex/features/pool-list/apiPoolList.utils.ts
```diff
@@ -0,0 +1,118 @@
...
+) => (userPoolIds ? notFalsy(poolId, poolAddress).some(id => userPoolIds.has(normalizeAddress(id))) : undefined)
+
+const getPoolRewards = (pool: V2Pool, poolId: string): RewardsApy => {
+  const poolAddress = normalizeAddress(pool.address)
+  const gaugeAddress = pool.gauge?.address ? normalizeAddress(pool.gauge.address) : ''
+
+  return {
+    poolId,
+    base: {
+      day: `${pool.baseDailyApr ?? 0}`,
```
DanielSchiavini: Please never use ??0 that is just wrong

apps/main/src/dex/features/pool-list/components/PoolListEmptyState.tsx
```diff
@@ -24,9 +24,10 @@ enum ERROR {
 type Props = {
   columnFiltersById: PartialRecord<PoolColumnId, string>
   resetFilters: () => void
+  isError?: boolean
```
DanielSchiavini: why optional?

apps/main/src/dex/features/pool-list/chips/PoolListChips.tsx
```diff
@@ -27,14 +28,15 @@ export const PoolListChips = ({
   sortField: PoolColumnId
   searchText: string
   onSearch: (value: string) => void
+  sortOptions?: readonly { id: PoolColumnId; label: string }[]
```
DanielSchiavini: why optional?

apps/main/src/dex/features/pool-list/apiPoolList.utils.ts
```diff
@@ -0,0 +1,118 @@
...
+export const getPoolIdByAddressEntries = (poolMapper: PoolIdByAddressSource | undefined) =>
+  recordValues(poolMapper ?? {}).map(({ pool }) => [normalizeAddress(pool.address), pool.id] as const)
+
+export const getPoolIdByAddress = (...poolMappers: (PoolIdByAddressSource | undefined)[]) =>
+  fromEntries(poolMappers.flatMap(getPoolIdByAddressEntries))
+
+export const getCurvePoolIdByAddressEntries = (curve: CurveApi | undefined) =>
+  curve?.getPoolList().map(poolId => [normalizeAddress(curve.getPool(poolId).address), poolId] as const) ?? []
+
+export const getPoolIdByAddressFromEntries = (entries: (readonly [string, string])[]) => fromEntries(entries)
```
DanielSchiavini: completely unnecessary function

apps/main/src/dex/features/pool-list/apiPoolList.utils.ts
```diff
@@ -0,0 +1,118 @@
...
+
+const VYPER_EXPLOIT_VERSIONS = new Set(['0.2.15', '0.2.16', '0.3.0'])
+
+export type PoolIdByAddressSource = Record<string, { pool: Pick<PoolData['pool'], 'address' | 'id'> }>
+export type PoolListItemOptions = {
+  hasPosition?: boolean
+  poolId?: string
+}
+
+export const normalizeAddress = (address: string) => address.toLowerCase()
```
DanielSchiavini: We shouldn't need to normalize addresses

apps/main/src/dex/features/pool-list/apiPoolList.utils.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export const getPoolIdByAddressEntries = (poolMapper: PoolIdByAddressSource | undefined) =>
+  recordValues(poolMapper ?? {}).map(({ pool }) => [normalizeAddress(pool.address), pool.id] as const)
+
+export const getPoolIdByAddress = (...poolMappers: (PoolIdByAddressSource | undefined)[]) =>
+  fromEntries(poolMappers.flatMap(getPoolIdByAddressEntries))
+
+export const getCurvePoolIdByAddressEntries = (curve: CurveApi | undefined) =>
```
DanielSchiavini: this function shouldn't be necessary and why default to []?

apps/main/src/dex/features/pool-list/PoolListApiTable.tsx
OnlyJousting: this PR's scope is to just replace the data source and move filtering/searching to the API so that means just reusing the current UI components

apps/main/src/dex/features/pool-list/apiPoolList.utils.ts
OnlyJousting: it's needed to stay within scope of this PR and fit into legacy components

michwill (dismissed):

apps/main/src/dex/features/pool-list/PoolListTable.tsx
```diff
@@ -1,94 +1,253 @@
...
-  const { globalFilter, setGlobalFilter, columnFilters, columnFiltersById, setColumnFilter, resetFilters } = useFilters(
-    {
-      columns: PoolColumnId,
+const getApiSorting = (sorting: SortingState, defaultSort: SortingState): ApiSorting => {
+  const sort = sorting.find(isApiColumnSort) ?? defaultSort.find(isApiColumnSort)
+
+  return sort ? [{ id: sort.id, desc: sort.desc }] : [{ id: PoolColumnId.Tvl, desc: true }]
+}
+
+const usePoolListApiUrlState = ({ isLite }: Pick<NetworkConfig, 'isLite'>) => {
```
0xPearce: maybe worth to move it into a different file

apps/main/src/dex/features/pool-list/PoolListTable.tsx
```diff
@@ -1,94 +1,253 @@
...
 
-export const PoolListTable = ({ network }: { network: NetworkConfig }) => {
-  const { isLite, poolFilters } = network
+// The API bundles crypto, factory_crypto, and twocrypto_ng pools under the "crypto" filter.
+// Omitted "main" and "factory" from available filters.
+const POOL_TYPE_FILTERS = [
+  { key: 'stableswapng', label: t`Stable NG` },
+  { key: 'crvusd', label: t`crvUSD` },
+  { key: 'crypto', label: t`Twocrypto` },
+  { key: 'factory_tricrypto', label: t`Tricrypto` },
```
0xPearce: these keys don't match the legacy ones e.g.: `stableswapng` !== `stableng`, old links with filters will not work.

0xPearce: Also, we should decide whether we use camelcase, snakecase or lowercase for the pools and llamalend, I have seen all 3 (outside the scope of this PR)

apps/main/src/dex/queries/user-pools.query.ts
```diff
@@ -10,7 +10,7 @@ type UserPoolsParams = FieldsOf<UserPoolsQuery>
 
 export const { useQuery: useUserPools, fetchQuery: fetchUserPools } = queryFactory({
   queryKey: ({ chainId, userAddress }: UserPoolsParams) =>
-    [...rootKeys.chain({ chainId }), ...rootKeys.user({ userAddress }), 'pools'] as const,
+    [...rootKeys.chain({ chainId }), ...rootKeys.user({ userAddress }), 'user-pools'] as const,
```
0xAlunara: Dani would say use `getUserPoolList` as key 🤓

apps/main/src/dex/features/pool-list/PoolListTable.tsx
OnlyJousting: the keys are just derived from the API filters but ofc we could create a mapping to bridge that if need be

apps/main/src/dex/features/pool-list/Page.tsx
```diff
@@ -1,8 +1,36 @@
...
+        (isSupported ? <PoolListTable network={network} /> : <LegacyPoolListTable network={network} />)}
+    </ListPageWrapper>
+  )
 }
```
0xAlunara, line 31: It's a bit verbose, could just keep it a single component I guess

```tsx
const Page = () => {
  const network = useNetworkFromUrl()
  const { data: supportedPoolChains, isLoading } = usePoolChains({})
  /**
   * Prices API v2 does not cover every full DEX network that's supplied by the curve-api.
   * Unsupported: Moonbeam, Kava, Avalanche, Celo, Aurora, X-Layer, zkSync, Mantle.
   */
  const isSupported = supportedPoolChains?.some(({ chainId }) => chainId === network?.chainId)

  return (
    <ListPageWrapper>
      {network &&
        !isLoading &&
        (isSupported ? <PoolListTable network={network} /> : <LegacyPoolListTable network={network} />)}
    </ListPageWrapper>
  )
}
```

apps/main/src/dex/features/pool-list/PoolListTable.tsx
```diff
@@ -1,94 +1,253 @@
...
 import { PoolListEmptyState } from './components/PoolListEmptyState'
 import { PoolMobileExpandedPanel } from './components/PoolMobileExpandedPanel'
-import { usePoolListData } from './hooks/usePoolListData'
+import { usePoolListApiUserPositionOptions } from './hooks/usePoolListApiUserPositionOptions'
 import { usePoolListVisibilitySettings } from './hooks/usePoolListVisibilitySettings'
-import { usePoolsGlobalFilterFn } from './poolsGlobalFilter'
 
 const LOCAL_STORAGE_KEY = 'dex-pool-list'
-
 const PER_PAGE = 50
```
0xAlunara: I believe this is already the default though in prices api package?

apps/main/src/dex/features/pool-list/PoolListTable.tsx
```diff
@@ -1,94 +1,253 @@
...
+    manualPagination: true,
+    manualSorting: true,
+    manualFiltering: true,
+    pageCount: poolList ? Math.ceil(poolList.count / PER_PAGE) : -1,
+    autoResetPageIndex: false,
   })
 
-  const resultCount = table.getFilteredRowModel().rows.length
   return (
     <LegacyDataTable
```
0xAlunara: Might as well use the new DataTable?

apps/main/src/dex/features/pool-list/PoolListTable.tsx
```diff
@@ -1,94 +1,253 @@
...
         <EmptyStateRow table={table}>
-          <PoolListEmptyState columnFiltersById={columnFiltersById} resetFilters={resetFilters} />
+          <PoolListEmptyState columnFiltersById={emptyStateFilters} resetFilters={resetFilters} isError={isError} />
         </EmptyStateRow>
       }
       expandedPanel={PoolMobileExpandedPanel}
-      shouldStickFirstColumn={Boolean(useIsTablet() && userHasPositions)}
-      loading={isLoading}
+      shouldStickFirstColumn={isTablet}
+      loading={loading}
```
0xAlunara: When changing sorting or any other kind of I briefly notice it's showing those 3 skeleton rows, very annoying. If we use the new DataTable we could keep have the same amount of skeleton rows as we currently are presenting, would reduce flickering.

0xAlunara (dismissed):
Approved as it's beta only anyway, you might prefer processing feedback in follow up PRs

apps/main/src/dex/features/pool-list/hooks/usePoolListUserHasPosition.ts
```diff
@@ -0,0 +1,45 @@
...
+  type PoolListItemOptions,
+} from '../apiPoolList.utils'
+
+export const usePoolListApiUserPositionOptions = (chainId: ChainId) => {
+  const { curveApi, isHydrated } = useCurve()
+  const { address: userAddress } = useConnection()
+  const poolDataMapper = useStore(state => state.pools.poolsMapper[chainId])
+  const poolDataCacheMapper = useStore(state => state.storeCache.poolsMapper[chainId])
+  const poolIdSource = poolDataMapper ?? poolDataCacheMapper
+  const curveApiForChain = isHydrated && curveApi?.chainId === chainId ? curveApi : undefined
```
DanielSchiavini, line 16: Do we really need this check? I'm trying to get rid of this kind of stuff.

apps/main/src/dex/features/pool-list/apiPoolList.utils.ts
DanielSchiavini: Could we keep this only when using legacy stores?

apps/main/src/dex/features/pool-list/apiPoolList.utils.ts
```diff
@@ -0,0 +1,114 @@
...
+  recordValues(poolMapper ?? {}).map(({ pool }) => [normalizeAddress(pool.address), pool.id] as const)
+
+export const getCurvePoolIdByAddressEntries = (curve: CurveApi) =>
+  curve.getPoolList().map(poolId => [normalizeAddress(curve.getPool(poolId).address), poolId] as const)
+
+export const getHasPosition = (
+  userPoolIds: Set<string> | undefined,
+  poolAddress: string,
+  poolId: string | undefined,
+) => (userPoolIds ? notFalsy(poolId, poolAddress).some(id => userPoolIds.has(normalizeAddress(id))) : undefined)
```
DanielSchiavini:
```suggestion
) => maybe(userPoolIds, userPoolIds => notFalsy(poolId, poolAddress).some(id => userPoolIds.has(normalizeAddress(id))))
```

apps/main/src/dex/features/pool-list/apiPoolList.utils.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const getPoolRewards = (pool: V2Pool, poolId: string): RewardsApy => {
+  const poolAddress = normalizeAddress(pool.address)
+  const gaugeAddress = pool.gauge?.address ? normalizeAddress(pool.gauge.address) : ''
+
+  return {
+    poolId,
+    // Match curve-js RewardsApy: missing base APRs become 0 and no CRV rewards are represented as [0, 0].
```
DanielSchiavini: matching curve-js will just make it harder on ourselves later on 😞

apps/main/src/dex/features/pool-list/LegacyPoolListTable.tsx
```diff
@@ -0,0 +1,97 @@
...
+        chips={
+          <PoolListChips
+            poolFilters={poolFilters}
+            hiddenCount={getHiddenCount(table)}
+            resetFilters={resetFilters}
+            onSortingChange={onSortingChange}
+            sortField={sortField}
+            searchText={globalFilter}
+            onSearch={setGlobalFilter}
+            resultCount={data ? resultCount : undefined}
```
DanielSchiavini, line 90:
```suggestion
            resultCount={data && resultCount}
```

apps/main/src/dex/features/pool-list/PoolListTable.tsx
```diff
@@ -1,94 +1,253 @@
...
+  )
+  const setPoolType = useCallback(
+    (type: PoolTypeFilter | null) => updateQueryAndResetPage({ [PoolColumnId.PoolTags]: type }),
+    [updateQueryAndResetPage],
+  )
+  const resetFilters = useCallback(
+    () => updateQueryAndResetPage({ search: null, [PoolColumnId.PoolTags]: null }),
+    [updateQueryAndResetPage],
+  )
+  const onSortingChange: OnChangeFn<SortingState> = useCallback(
```
DanielSchiavini: what's wrong with useSortFromQueryString?

apps/main/src/dex/features/pool-list/PoolListTable.tsx
```diff
@@ -1,94 +1,253 @@
...
+  const sorting = useMemo(() => getApiSorting(urlSorting, defaultSort), [defaultSort, urlSorting])
+  const [{ id: sortField, desc }] = sorting
+  const { sortBy } = API_SORT_COLUMNS[sortField]
+  const sortDirection: 'desc' | 'asc' = desc ? 'desc' : 'asc'
+  const [pagination, onPaginationChange] = usePageFromQueryString(PER_PAGE)
+  const searchText = searchParams.get('search') ?? ''
+  const poolTypeParam = searchParams.get(PoolColumnId.PoolTags)
+  const poolType = getPoolType(poolTypeParam)
+
+  const updateQueryAndResetPage = useCallback(
```
DanielSchiavini: I would prefer not repeating this for every table.

apps/main/src/dex/features/pool-list/PoolListTable.tsx
```diff
@@ -1,94 +1,253 @@
...
+    (update: Record<string, string | string[] | null>) => searchNavigate({ ...update, page: null }, { replace: true }),
+    [searchNavigate],
+  )
+  const setSearch = useCallback(
```
DanielSchiavini: what's wrong with useFilters?

apps/main/src/dex/features/pool-list/PoolListTable.tsx
```diff
@@ -1,94 +1,253 @@
...
+        onSearch={setSearch}
         hasSearchBar
         chips={
           <PoolListChips
-            poolFilters={poolFilters}
-            hiddenCount={getHiddenCount(table)}
-            resetFilters={resetFilters}
+            columnFiltersById={emptyStateFilters}
+            filterGroups={POOL_TYPE_FILTER_GROUPS}
+            hiddenCount={poolType ? 1 : 0}
```
DanielSchiavini: What does this mean, how's hidden count related to pool type at all?

apps/main/src/dex/features/pool-list/PoolListTable.tsx
```diff
@@ -1,94 +1,253 @@
...
             onSortingChange={onSortingChange}
+            searchText={searchText}
+            onSearch={setSearch}
+            poolFilters={POOL_TYPES}
+            resetFilters={resetFilters}
+            resultCount={apiResultCount}
+            setColumnFilter={(_, value) => setPoolType(isPoolType(value) ? value : null)}
+            showHiddenCountReset={false} // "hidden" makes no sense unless returned from the API
```
DanielSchiavini: why?

apps/main/src/dex/queries/pool-list.query.ts
```diff
@@ -0,0 +1,36 @@
...
+      { searchString },
+      { poolType },
+      { sortBy },
+      { sortDirection },
+    ] as const,
+  queryFn: async ({ chainId, page, pageSize, searchString, poolType, sortBy, sortDirection }: PoolListQuery) =>
+    listPools({ chainId, page, pagination: pageSize, searchString, poolType, sortBy, sortDirection }),
+  validationSuite: createValidationSuite(chainValidationGroup),
+  category: 'dex.pools',
+  keepPreviousData: true,
```
DanielSchiavini, line 40: is this necessary?

apps/main/src/dex/queries/pools-blacklist.query.ts
```diff
@@ -225,6 +225,9 @@ const getBlacklist = (blacklistPricesApi: QueryData<typeof usePricesApiBlacklist
     ...blacklistPricesApi.filter(({ chain }) => chain === blockchainId).map(({ address }) => address),
   ].map(address => getAddress(address)) // just to be sure there's no missing checksums from the prices api
 
+export const getLocalPoolsBlacklist = (blockchainId: string | undefined) =>
+  blockchainId ? (blacklist[blockchainId as ChainBlacklist] ?? []).map(address => getAddress(address)) : []
```
DanielSchiavini: notFalsyArray

apps/main/src/dex/queries/pools-blacklist.query.ts
```diff
@@ -225,6 +225,9 @@ const getBlacklist = (blacklistPricesApi: QueryData<typeof usePricesApiBlacklist
     ...blacklistPricesApi.filter(({ chain }) => chain === blockchainId).map(({ address }) => address),
   ].map(address => getAddress(address)) // just to be sure there's no missing checksums from the prices api
 
+export const getLocalPoolsBlacklist = (blockchainId: string | undefined) =>
```
DanielSchiavini: Why is blockchainId a string and not `Chain`?

apps/main/src/dex/features/pool-list/apiPoolList.utils.ts
DanielSchiavini: I was hoping we would have a new table with new data model so we can easily get rid of the old one

apps/main/src/dex/features/pool-list/PoolListTable.tsx
OnlyJousting: because filtering happens on the backend. in order to see how many pools are hidden we would need to look at the `count` from a query without filters and subtract the `count` from the query that applies the filters. I suppose this is not too complicated too do but it feels clunky. I think it would be nicer then if the API returned a `hidden` value next to `count`

apps/main/src/dex/features/pool-list/PoolListTable.tsx
OnlyJousting: yea I set it to 50 there too, API default is 20. I'll just let it use the default as I don't see anyone changing that

OnlyJousting: never mind, usePageFromQueryString requires `pageSize` so I'm gonna leave it

apps/main/src/dex/features/pool-list/PoolListTable.tsx
DanielSchiavini: the hidden count is hard, but the reset button should be available right?

apps/main/src/dex/features/pool-list/PoolListTable.tsx
OnlyJousting: Yea I'm gonna do some refactoring and implement the new DataTable

apps/main/src/dex/features/pool-list/PoolListTable.tsx
OnlyJousting: yea it's still available

DanielSchiavini: The hidden count is just total - found right?

apps/main/src/dex/queries/pools-blacklist.query.ts
OnlyJousting: getLocalPoolsBlacklist was removed from the PR as we won't do any blacklisting in the new pool list (handled on the API)

apps/main/src/dex/features/pool-list/hooks/usePoolListUserHasPosition.ts
OnlyJousting: it's there to help serve marks for pools users have positions in as soon as possible either via curve-js data or cache, I'd say needed for now

apps/main/src/dex/queries/pool-list.query.ts
OnlyJousting: if you mean `keepPreviousData` we need it so we don't end up with an empty table while waiting for results from searching/filtering/paginating

0xtutti (dismissed):

apps/main/src/dex/features/pool-list/PoolListTable.tsx
```diff
@@ -1,97 +1,138 @@
...
-            resetFilters={resetFilters}
-            onSortingChange={onSortingChange}
-            sortField={sortField}
-            searchText={globalFilter}
-            onSearch={setGlobalFilter}
-            resultCount={data ? resultCount : undefined}
-            {...filterProps}
-          />
+    <Stack>
+      <CardHeader title={t`Pools`} subheader={t`Find your next opportunity`} />
```
0xPearce: We can add the `onReload` as well like for the llamalend table (can be done in another PR)

tests/cypress/e2e/dex/dex-markets.cy.ts
```diff
@@ -141,10 +140,10 @@ describe('DEX Pools', () => {
     if (breakpoint === 'mobile') {
       cy.get('[data-testid="btn-expand-search-dex-pool-list"]').click()
```
0xPearce: this will break btw, search bar is always expanded for mobile with the new data table

0xPearce (commented):
That's a lot better and very fast! Just searched for weETH and found a possible regression in the rewards columns for the second market:

<table>
  <tr>
    <th>Before</th>
    <td>Image</td>
  </tr>
  <tr>
    <th>After</th>
    <td>Image</td>
  </tr>
</table>

apps/main/src/dex/features/pool-list/cells/PoolListBaseApyCell.tsx
```diff
@@ -0,0 +1,40 @@
...
+import type { CellContext } from '@tanstack/react-table'
+import { t } from '@ui-kit/lib/i18n'
+import { Tooltip } from '@ui-kit/shared/ui/Tooltip'
+import { formatNumber } from '@ui-kit/utils'
+import type { PoolListItem } from '../poolList.types'
+import { Placeholder } from './Placeholder'
+
+const BaseApyTooltip = ({ day, week }: { day: number; week: number | null | undefined }) => (
+  <Stack component="span">
+    <strong>{t`Pool APY`}</strong>
```
DanielSchiavini: We should be using typography with design variants

apps/main/src/dex/features/pool-list/cells/PoolListBaseApyCell.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export const PoolListBaseApyCell = ({
+  row: { original: pool },
+  getValue,
+}: CellContext<PoolListItem, number | null>) => {
+  const day = getValue()
+
+  return day == null ? (
+    <Placeholder />
+  ) : day > LARGE_APY ? (
```
DanielSchiavini: This variable name is confusing, it looks like we are comparing day counts with APYs

apps/main/src/dex/features/pool-list/cells/PoolListRewards.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+type CrvRewards = Pick<PoolListItem, 'crvApr' | 'crvAprBoosted'>
+type ExtraRewards = Pick<PoolListItem, 'address' | 'extraRewardsApr'>
+
+const getCrvRewardsLabel = ({ crvApr, crvAprBoosted }: CrvRewards): ReactNode => {
+  const baseApr = crvApr ?? 0
+  const boostedApr = crvAprBoosted ?? 0
+
+  if (!baseApr && !boostedApr) return null
+  if (baseApr < 0.01) return '< 0.01% CRV'
```
DanielSchiavini: no magic numbers pls

apps/main/src/dex/features/pool-list/cells/PoolListRewards.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  return (
+    <>
+      {formatNumber(baseApr, 'percent.value')}
+      {boostedApr ? ' \u2192 ' : null}
+      {boostedApr ? <span style={{ whiteSpace: 'nowrap' }}>{formatNumber(boostedApr, 'percent.value')}</span> : null}
```
DanielSchiavini: - We can use unicode directly in the jsx 
- We should use typography
- We don't need `: null` in JSX
```suggestion
      {!!boostedApr && (
        <>
          {' → '}
          <Typography style={{ whiteSpace: 'nowrap' }}>{formatNumber(boostedApr, 'percent.value')}</span>
        </>
      )}
```

apps/main/src/dex/features/pool-list/cells/PoolListTitleCell.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { PoolAlertIcons } from './PoolTitleCell/PoolAlertIcons'
+import { PoolTokens } from './PoolTitleCell/PoolTokens'
+
+const { Spacing, Height } = SizesAndSpaces
+
+export const PoolListTitleCell = ({
+  row: { original: pool },
+  column: { getFilterValue },
+}: CellContext<PoolListItem, string>) => {
+  const tokenList = useMemo(() => pool.coins.map(({ symbol, address }) => ({ symbol, address })), [pool.coins])
```
DanielSchiavini: What's wrong with
```suggestion
  const tokenList = pool.coins
```

apps/main/src/dex/features/pool-list/cells/PoolListRewards.tsx
```diff
@@ -0,0 +1,75 @@
...
+          tooltipProps={{ placement: 'bottom-end' }}
+        >
+          {crvRewardsLabel}
+        </Chip>
+      )}
+      {extraRewards.map(({ apr, id, symbol }) => (
+        <Chip key={id} size="md">
+          {formatNumber(apr, 'percent.value')} {symbol}
+        </Chip>
+      ))}
```
DanielSchiavini: getExtraRewards could be inlined here, we end up calling `map` twice which causes an unnecessary layer of indirection. `apr`, `id` and `symbol` are only used once! We don't need to add `poolAddress` to `id`, using `index` will give a lint warning but it's defensible in this case.
```suggestion
      {extraRewardsApr.map(({ apr, address, symbol }, index) => r.apr > 0 && (
        <Chip key={address ?? index} size="md">
          {formatNumber(apr, 'percent.value')} {symbol ?? ''}
        </Chip>
      ))}
```

apps/main/src/dex/features/pool-list/cells/PoolListUsdCell.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { Tooltip } from '@ui-kit/shared/ui/Tooltip'
+import { formatNumber } from '@ui-kit/utils'
+import type { PoolListItem } from '../poolList.types'
+
+export const PoolListUsdCell = ({ getValue }: CellContext<PoolListItem, number | null | undefined>) => {
+  const value = getValue()
+
+  return (
+    <Tooltip title={maybe(value, value => formatNumber(value, 'usd.amount'))}>
+      <Stack>{value == null ? '-' : formatNumber(value, 'usd.notional')}</Stack>
```
DanielSchiavini: `formatNumber` should be handling fallbacks now
```suggestion
      <Stack>{formatNumber(value, 'usd.notional')}</Stack>
```

apps/main/src/dex/features/pool-list/columns/legacy-column.definitions.ts
```diff
@@ -0,0 +1,109 @@
...
+  [LegacyPoolColumnId.Tvl]: t`TVL`,
+}
+
+type LegacyPoolColumn = ColumnDefinition<LegacyPoolListItem>
+
+/** Sorts pool rows by a numeric reward value, placing rows with no value last. */
+const sortByReward =
+  (getValue: (row: LegacyPoolListItem) => number | undefined) =>
+  (x: Row<LegacyPoolListItem>, y: Row<LegacyPoolListItem>) =>
+    (getValue(x.original) ?? -Infinity) - (getValue(y.original) ?? -Infinity)
```
DanielSchiavini, line 36: `-Infinity` is always going to result in `NaN`

apps/main/src/dex/features/pool-list/columns/legacy-column.definitions.ts
```diff
@@ -0,0 +1,109 @@
...
+  }),
+  columnHelper.accessor(row => (row.volume ? +row.volume : null), {
+    id: LegacyPoolColumnId.Volume,
+    header: headers[LegacyPoolColumnId.Volume],
+    cell: LegacyUsdCell,
+    meta: { type: 'numeric' },
+    sortUndefined: 'last',
+    filterFn: rangeFilterFn,
+  }),
+  columnHelper.accessor(row => (row.tvl ? +row.tvl : null), {
```
DanielSchiavini, line 85: No reason to convert this to number
```suggestion
  columnHelper.accessor('tvl', {
```

apps/main/src/dex/features/pool-list/columns/legacy-column.definitions.ts
```diff
@@ -0,0 +1,109 @@
...
+  }),
+  columnHelper.accessor('rewards', {
+    id: LegacyPoolColumnId.RewardsOther,
+    header: headers[LegacyPoolColumnId.RewardsOther],
+    cell: LegacyRewardsOtherCell,
+    enableSorting: false, // that's done in the separate columns RewardsCrv and RewardsIncentives
+    enableMultiSort: false, // that's done in the separate columns RewardsCrv and RewardsIncentives
+    meta: { type: 'numeric', tooltip: { title: t`Token APR based on current prices of tokens and reward rates` } },
+  }),
+  columnHelper.accessor(row => (row.volume ? +row.volume : null), {
```
DanielSchiavini, line 77:
```suggestion
  columnHelper.accessor('volume', {
```

apps/main/src/dex/features/pool-list/components/PoolListEmptyState.tsx
```diff
@@ -22,54 +12,31 @@ enum ERROR {
...
-      errorKey === ERROR.search
-        ? searchText
-        : errorKey === ERROR.filter
-          ? columnFiltersById[PoolColumnId.UserHasPositions] && userAddress
-            ? shortenAccount(userAddress)
-            : (columnFiltersById[PoolColumnId.PoolTags] as PoolTag | undefined)
-          : undefined,
-    [errorKey, searchText, columnFiltersById, userAddress],
-  )
+export const PoolListEmptyState = ({ poolType, poolTypeFilters, resetFilters, searchText, isError }: Props) => {
```
DanielSchiavini, line 23: Conflict with #2737 fyi

apps/main/src/dex/features/pool-list/components/PoolListEmptyState.tsx
```diff
@@ -22,54 +12,31 @@ enum ERROR {
...
-          <br />{' '}
-          <Button variant="text" onClick={resetFilters}>
-            View all
-          </Button>
-        </Trans>
-      )}
-    </>
+    <Trans>
+      No pool found {errorSearchedValue && `for "${errorSearchedValue}"`}.
+      <br />{' '}
```
DanielSchiavini:
```suggestion
      <p>No pool found {errorSearchedValue && `for "${errorSearchedValue}"`}.</p>
```

apps/main/src/dex/features/pool-list/drawers/PoolListSortDrawer.tsx
```diff
@@ -0,0 +1,75 @@
...
+  onSortingChange: OnChangeFn<SortingState>
+  options: readonly { id: PoolListSortableColumn; label: string }[]
+  sortField: PoolListSortableColumn
+}
+
+export const PoolListSortDrawer = ({ onSortingChange, options, sortField }: Props) => {
+  const [open, , closeDrawer, toggleDrawer, setOpen] = useSwitch(false)
+  const menuRef = useRef<HTMLLIElement | null>(null)
+
+  const handleSort = useCallback(
```
DanielSchiavini: this useCallback is useless because it's only called in an arrow function

apps/main/src/dex/features/pool-list/hooks/usePoolListUrlState.ts
```diff
@@ -0,0 +1,146 @@
...
+const SORT_OPTIONS = recordEntries(SORT_COLUMNS).map(([id, { label }]) => ({ id, label }))
+
+const isPoolType = (value: string | null): value is PoolListPoolType => value != null && POOL_TYPE_SET.has(value)
+
+const getPoolType = (value: string | null): PoolListPoolType | undefined => {
+  if (value == null) return undefined
+  if (CRYPTO_POOL_TYPE_ALIASES.has(value)) return 'crypto'
+
+  return isPoolType(value) ? value : undefined
+}
```
DanielSchiavini:
```suggestion
const getPoolType = (value: string | null): PoolListPoolType | undefined =>
  maybe(value, value =>
    CRYPTO_POOL_TYPE_ALIASES.has(value) ? 'crypto' : isPoolType(value) ? value : undefined
  )
```

apps/main/src/dex/features/pool-list/hooks/usePoolListUrlState.ts
```diff
@@ -0,0 +1,146 @@
...
+
+const isPoolListSortColumn = (value: string | undefined): value is PoolListSortableColumn =>
+  value != null && value in SORT_COLUMNS
+
+const isColumnSort = (sort: SortingState[number]): sort is ColumnSort => isPoolListSortColumn(sort.id)
+
+const getPoolListSorting = (sorting: SortingState, defaultSort: SortingState): PoolListSorting => {
+  const sort = sorting.find(isColumnSort) ?? defaultSort.find(isColumnSort)
+
+  return sort ? [{ id: sort.id, desc: sort.desc }] : [{ id: PoolListColumnId.Tvl, desc: true }]
```
DanielSchiavini: why are we using `desc: boolean` instead of the standard `sort: 'asc' | 'desc'`?

apps/main/src/dex/features/pool-list/hooks/usePoolListUrlState.ts
```diff
@@ -0,0 +1,146 @@
...
+  const setSearch = useCallback(
+    (value: string) => updateQueryAndResetPage({ [SEARCH_QUERY_FIELD]: value || null }),
+    [updateQueryAndResetPage],
+  )
+  const setPoolType = useCallback(
+    (value: PoolListPoolType | null) => updateQueryAndResetPage({ [POOL_TYPE_QUERY_FIELD]: value }),
+    [updateQueryAndResetPage],
+  )
+  const resetFilters = useCallback(
+    () => updateQueryAndResetPage(mapRecord(activeFilters, () => null)),
```
DanielSchiavini: resetFilters shouldn't set the active filters to null, it should reset the filters to the default like the other tables

apps/main/src/dex/features/pool-list/hooks/usePoolListUrlState.ts
```diff
@@ -0,0 +1,146 @@
...
+
+const isColumnSort = (sort: SortingState[number]): sort is ColumnSort => isPoolListSortColumn(sort.id)
+
+const getPoolListSorting = (sorting: SortingState, defaultSort: SortingState): PoolListSorting => {
+  const sort = sorting.find(isColumnSort) ?? defaultSort.find(isColumnSort)
+
+  return sort ? [{ id: sort.id, desc: sort.desc }] : [{ id: PoolListColumnId.Tvl, desc: true }]
+}
+
+export const usePoolListUrlState = ({ isLite }: Pick<NetworkConfig, 'isLite'>) => {
```
DanielSchiavini: This hook does WAAAY too much and seems mostly unnecessary, we should separate sorting, redirects, filters, default filters, etc all in separate hooks, just like done in the other tables.
We already have all those helpers such as `useSortFromQueryString`, `useFilters`, etc

apps/main/src/dex/features/pool-list/poolList.utils.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+export const normalizeAddress = (address: string) => address.toLowerCase()
+
+export const getPoolIdByAddressEntries = (poolMapper: PoolIdByAddressSource | undefined) =>
+  recordValues(poolMapper ?? {}).map(({ pool }) => [normalizeAddress(pool.address), pool.id] as const)
+
+export const getCurvePoolIdByAddressEntries = (curve: CurveApi) =>
+  curve.getPoolList().map(poolId => [normalizeAddress(curve.getPool(poolId).address), poolId] as const)
+
+export const getHasPosition = (userPoolIds: Set<string> | undefined, poolAddress: string, poolId: string | undefined) =>
```
DanielSchiavini: This is just used once, I think it's simpler to inline it.

apps/main/src/dex/features/pool-list/PoolListTable.tsx
```diff
@@ -1,97 +1,138 @@
...
+    sortOptions,
+  } = usePoolListUrlState(network)
 
-  const { globalFilter, setGlobalFilter, columnFilters, columnFiltersById, setColumnFilter, resetFilters } = useFilters(
-    {
-      columns: PoolColumnId,
-    },
+  const {
+    data: poolList,
+    isPlaceholderData,
```
DanielSchiavini, line 42: Is this going to work on filter changes? Given the query key changes

apps/main/src/dex/features/pool-list/PoolListTable.tsx
```diff
@@ -1,97 +1,138 @@
...
+  } = usePoolList({
+    chainId: network.chainId,
+    page: pagination.pageIndex + 1,
+    pageSize: POOL_LIST_PAGE_SIZE,
+    searchString: searchText || undefined,
+    poolType,
+    sortBy,
+    sortDirection,
+  })
+  const loading = isPending
```
DanielSchiavini: We use isLoading everywhere else, let's keep it consistent pls

apps/main/src/dex/features/pool-list/PoolListTable.tsx
```diff
@@ -1,97 +1,138 @@
...
-    globalFilterFn,
-    ...getTableOptions(data),
-    getPaginationRowModel: getPaginationRowModel(),
+    onSortingChange,
+    getCoreRowModel: getCoreRowModel(),
+    getExpandedRowModel: getExpandedRowModel(),
+    manualPagination: true,
+    manualSorting: true,
+    manualFiltering: true,
+    pageCount: poolList ? Math.ceil(poolList.count / POOL_LIST_PAGE_SIZE) : -1,
```
DanielSchiavini: We should get page count from the backend or calculate it in the query

packages/curve-ui-kit/src/shared/ui/DataTable/hooks/useTableScroll.ts
```diff
@@ -43,7 +43,7 @@ export function useScrollToTopOnPageChange<T extends TableItem>({
   const { pageIndex } = table.getState().pagination
   useEffectOnValueChange(pageIndex, () => {
     // scroll after the user changes pages
-    if (containerRef) containerRef.current?.scrollTo({ top: 0, behavior: 'smooth' })
+    if (containerRef?.current) containerRef.current.scrollTo({ top: 0, behavior: 'smooth' })
```
DanielSchiavini, line 46: It's not clear why the containerRef is optional at all
```suggestion
   containerRef?.current?.scrollTo({ top: 0, behavior: 'smooth' })
```

apps/main/src/dex/features/pool-list/Page.tsx
```diff
@@ -1,8 +1,31 @@
...
+import { LegacyPoolListTable } from './LegacyPoolListTable'
+import { PoolListTable } from './PoolListTable'
 
 export const Page = () => {
   const network = useNetworkFromUrl()
-  return <ListPageWrapper>{network && <PoolListTable network={network} />}</ListPageWrapper>
+  const isBetaPoolListEnabled = useDexPoolListV2()
+  const { data: supportedPoolChains = [], isLoading } = usePoolChains({}, isBetaPoolListEnabled)
+  /**
+   * Prices API v2 does not cover every full DEX network that's supplied by the curve-api.
```
DanielSchiavini, line 13: we should be adapting the data from the lite API into the same format so the rest of the table stays the same

apps/main/src/dex/features/pool-list/hooks/usePoolListUserPositionOptions.ts
```diff
@@ -0,0 +1,44 @@
...
+import { useCurve } from '@ui-kit/features/connect-wallet'
+import {
+  getCurvePoolIdByAddressEntries,
+  getHasPosition,
+  getPoolIdByAddressEntries,
+  normalizeAddress,
+  type PoolListItemOptions,
+} from '../poolList.utils'
+
+export const usePoolListUserPositionOptions = (chainId: ChainId) => {
```
DanielSchiavini: useOptions returns a callback - the name doesn't make much sense

OnlyJousting:
> That's a lot better and very fast! Just searched for weETH and found a possible regression in the rewards columns for the second market:
> 
> Before	Image
> After	Image

there's a lot of discrepancies in this screenshot, vAPR and CRV emissions rates 🤔

EDIT: resolved on the API

apps/main/src/dex/features/pool-list/Page.tsx
OnlyJousting: we can do that in a future PR, this one is already blown up in scope 😓

apps/main/src/dex/features/pool-list/PoolListTable.tsx
OnlyJousting: don't even need the variable anymore

apps/main/src/dex/features/pool-list/PoolListTable.tsx
OnlyJousting: I will request to get it added to the API

OnlyJousting: or maybe just calculate in the query then

apps/main/src/dex/features/pool-list/columns/legacy-column.definitions.ts
OnlyJousting: this file just has an updated file name and updated imports so I'm just gonna leave it

apps/main/src/dex/features/pool-list/hooks/usePoolListUrlState.ts
OnlyJousting: reset clears the url and the UI falls back to default set here: https://github.com/curvefi/curve-frontend/blob/071bc5bad26882ac922153e1008bf368f056f597/apps/main/src/dex/features/pool-list/columns/column.options.ts#L50-L52, maybe this is confusing though 🤔

apps/main/src/dex/features/pool-list/hooks/usePoolListUserPositionOptions.ts
OnlyJousting: very true, updated the hook and name

apps/main/src/dex/features/pool-list/PoolListTable.tsx
OnlyJousting: you mean for the filter count? it's there to set it to undefined instead of quickly showing the previous queries count on the newly selected filter

packages/curve-ui-kit/src/shared/ui/DataTable/hooks/useTableScroll.ts
OnlyJousting: so scrolling wasn't working without setting `maxHeight` on the table, it doesn't really make sense to set a `maxHeight` on this table. without `maxHeight` there is no internal scroll container so containerRef.current is null which means `scrollTo` doesn't do anything. The change enabled falling back on the else if

packages/curve-ui-kit/src/utils/pagination.ts
```diff
@@ -0,0 +1,3 @@
+/** Compute page count from total item count and page size. */
+export const getPageCount = (totalCount: number | undefined, pageSize: number) =>
+  totalCount ? Math.ceil(totalCount / pageSize) : 0
```
DanielSchiavini, line 3: undefined count isn't the same as 0 😢

apps/main/src/dex/features/pool-list/cells/PoolListBaseApyCell.tsx
```diff
@@ -0,0 +1,29 @@
...
+    <Typography component="span" variant="tableCellMBold">
+      {formatNumber(baseDailyApr, 'percent.rate')}
+    </Typography>
+  ) : (
+    <Tooltip title={<TooltipBaseApy baseDailyApr={baseDailyApr} baseWeeklyApr={pool.baseWeeklyApr} />}>
+      <Typography component="span" variant="tableCellMBold">
+        {formatNumber(baseDailyApr, 'percent.rate')}
+      </Typography>
+    </Tooltip>
+  )
```
DanielSchiavini:
```suggestion
  return maybe(getValue(), baseDailyApr =>  baseDailyApr > LARGE_APY ? (
    <ChipVolatileBaseApy />
  ) : (
    <Tooltip title={<TooltipBaseApy baseDailyApr={baseDailyApr} baseWeeklyApr={pool.baseWeeklyApr} />}>
      <Typography component="span" variant="tableCellMBold">
        {formatNumber(baseDailyApr, 'percent.rate')}
      </Typography>
    </Tooltip>
  ) ?? (
    <Typography component="span" variant="tableCellMBold">
      {formatNumber(null, 'percent.rate')}
    </Typography>
  )
```

DanielSchiavini (dismissed):

apps/main/src/dex/features/pool-list/hooks/usePoolListVisibilitySettings.tsx
0xAlunara, line 37: You can disable the pool name column lol

Image

0xAlunara (dismissed):

tests/cypress/e2e/dex/dex-markets.cy.ts
DanielSchiavini: Indeed, [it's broken](https://github.com/curvefi/curve-frontend/actions/runs/27786976287/job/82226040784?pr=2644#step:12:350)

OnlyJousting: I will fix it in this PR! thanks for pointing it out pearce

packages/curve-ui-kit/src/shared/ui/DataTable/hooks/useTableScroll.ts
OnlyJousting: @0xPearce might be interested in this too

packages/curve-ui-kit/src/utils/pagination.ts
OnlyJousting: undefined = we don't know yet

tests/cypress/support/helpers/dex-pool-list-mocks.ts
```diff
@@ -0,0 +1,249 @@
...
+  page: number
+  pagination: number
+  search: string
+  poolType: string | null
+  sortBy: V2PoolSortField
+  sortDirection: SortDirection
+}
+
+const mockAddress = (chainId: number, index: number, offset = 0): Address =>
+  `0x${(BigInt(chainId) * 1_000_000n + BigInt(index) * 10n + BigInt(offset)).toString(16).padStart(40, '0')}`
```
DanielSchiavini, line 42: We already have `oneAddress` helper, we don't need to generate something deterministic
```suggestion
```

tests/cypress/support/helpers/dex-pool-list-mocks.ts
```diff
@@ -0,0 +1,249 @@
...
+
+const createCoins = (chainId: number, index: number) =>
+  range(2).map(poolIndex => ({
+    pool_index: poolIndex,
+    symbol: `MOCK${poolIndex + 1}`,
+    address: mockAddress(chainId, index, poolIndex + 1),
+    name: `Mock Token ${poolIndex + 1}`,
+    decimals: 18,
+  }))
```
DanielSchiavini, line 51: We should reuse `oneToken` instead of reinventing the wheel
```suggestion
```

tests/cypress/support/helpers/dex-pool-list-mocks.ts
```diff
@@ -0,0 +1,249 @@
...
+  crv_apr: null,
+  crv_apr_boosted: null,
+  extra_rewards_apr: [],
+  vyper_version: null,
+  gauges: [],
+})
+
+type MockPool = ReturnType<typeof createPool>
+
+const createMockPools = (chainId: number): MockPool[] => [
```
DanielSchiavini, line 89: It would be nice to generate some random data instead of hardcoding it

tests/cypress/support/helpers/dex-pool-list-mocks.ts
```diff
@@ -0,0 +1,249 @@
...
+    : []),
+  ...range(POOL_COUNT).map(index => createPool({ chainId, index })),
+]
+
+const MOCK_POOLS = [...createMockPools(ChainId.Ethereum), ...createMockPools(ChainId.Arbitrum)]
+
+const parsePositiveInt = (value: string | null, fallback: number) => {
+  const parsed = Number(value)
+  return Number.isInteger(parsed) && parsed > 0 ? parsed : fallback
+}
```
DanielSchiavini, line 165: We should not introduce these checks, we should just make sure the numbers are always valid
```suggestion
const parsePositiveInt = (value: string | null, fallback: number) => maybe(value, Number) ?? fallback
```

DanielSchiavini (approved):
I left a few remarks on the testing practices, they can be fixed separately this PR is way too large

0xAlunara (approved):

tests/cypress/support/helpers/dex-pool-list-mocks.ts
OnlyJousting: I'll make a follow up PR fixing these

apps/main/src/dex/features/pool-list/PoolListTable.tsx
```diff
@@ -1,97 +1,129 @@
...
+  )
 
-  const { globalFilter, setGlobalFilter, columnFilters, columnFiltersById, setColumnFilter, resetFilters } = useFilters(
-    {
-      columns: PoolColumnId,
-    },
+  const {
+    data: poolList,
+    isPlaceholderData,
+    isPending,
```
0xPearce, line 43: is there a reason to use `isPending` instead of `isLoading`?

DanielSchiavini: isLoading is false initially until the query gets enabled

DanielSchiavini: I hate the inconsistency, but I'm trying to find a better solution in #2729

PR #2646, opened 2026-06-04
fix: keep filters popover mounted

Pearce542:
Found an [edge case](https://github.com/curvefi/curve-frontend/actions/runs/26944738445/job/79494773452?pr=2645) in the tests where the market list table width changed depending on the applied filter → the market list header switched from sticky to non-sticky → the table children, including the table filters, were unmounted and remounted → the popover was unmounted → the test failed.

- split the filters trigger chip from the filters overlay
- render the overlay outside `DataTable` children to avoid remounts when sticky header layout changes


0xtutti (approved):

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersOverlay.tsx
```diff
@@ -0,0 +1,99 @@
...
+      <DrawerHeader title={t`Filter markets`}>
+        <Stack>{resetButton}</Stack>
+      </DrawerHeader>
+      <DrawerItems data-testid="drawer-filter-menu-lamalend-markets">{content}</DrawerItems>
+    </SwipeableDrawer>
+  ) : (
+    <Popover
+      open={open}
+      onClose={() => setOpen(false)}
+      anchorEl={() => anchorRef.current}
```
DanielSchiavini, line 66: why do we pass anchorEl and not anchorRef?

DanielSchiavini (approved):

OnlyJousting (approved):

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersOverlay.tsx
Pearce542: MUI Popover does not have an `anchorRef` prop, it only accepts `anchorEl`

PR #2660, opened 2026-06-06
refactor: remove setCustomFee call for ethereum mainnet

0xAlunara:
This only removes the call for mainnet. I've asked Denis if we can remove it for L2s too, but I'm not too optimistic (no pun intended).


0xtutti (approved):

DanielSchiavini (approved):
too optimistic 🤣

PR #2663, opened 2026-06-08
fix: stale new range in candle chart

OnlyJousting:
Closing a LlamaLend position makes the repay preview query resolve to `null`, meaning there is no new liquidation range to show. `useCallbackSync` previously skipped `null`, so the old preview range stayed in parent state and the candle chart could keep showing a stale “new range”.

This updates `useCallbackSync` to treat `null` as a resolved empty value and clear the synced callback state.

### DIFF
--- a/packages/curve-ui-kit/src/features/forms/useCallbackSync.ts
+++ b/packages/curve-ui-kit/src/features/forms/useCallbackSync.ts
@@ -1,11 +1,13 @@
 import { useEffect } from 'react'
-import { maybe } from '@primitives/objects.utils'
 import type { Query } from '@ui-kit/types/util'
 
 /**
- * Syncs `data` to `callback` whenever it changes to a valid value. Clears it when unmounting.
+ * Syncs `data` to `callback` whenever it changes to a resolved value. Keeps the previous value while loading, and
+ * clears it when the query explicitly resolves to null or unmounts.
  */
 export function useCallbackSync<T>({ data }: Query<T | null>, callback: (data: T | undefined) => void): void {
-  useEffect(() => maybe(data, data => callback(data)), [callback, data]) // keep parent in sync, keep stale while revalidating
+  useEffect(() => {
+    if (data !== undefined) callback(data ?? undefined)
+  }, [callback, data])
   useEffect(() => () => callback(undefined), [callback]) // clear stale data only when unmounting
 }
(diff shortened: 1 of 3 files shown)


packages/curve-ui-kit/src/features/forms/useCallbackSync.ts
```diff
@@ -1,11 +1,13 @@
...
 
 /**
- * Syncs `data` to `callback` whenever it changes to a valid value. Clears it when unmounting.
+ * Syncs `data` to `callback` whenever it changes to a resolved value. Keeps the previous value while loading, and
+ * clears it when the query explicitly resolves to null or unmounts.
  */
 export function useCallbackSync<T>({ data }: Query<T | null>, callback: (data: T | undefined) => void): void {
-  useEffect(() => maybe(data, data => callback(data)), [callback, data]) // keep parent in sync, keep stale while revalidating
+  useEffect(() => {
+    if (data !== undefined) callback(data ?? undefined)
```
DanielSchiavini, line 11: This change doesn't make any sense, it doesn't change anything in the runtime and the comment was removed

OnlyJousting: the problem was that updates passing `null` (like [useRepayPrices](https://github.com/curvefi/curve-frontend/blob/6d66fab27f407585f2ff21731c39525c98554714/apps/main/src/llamalend/queries/repay/repay-prices.query.ts#L42)) was not being propagated through `useCallbackSync` because `maybe` swallows it. this effectively allows `null` values to be synced but if you want to do it in a different way that's no problem

DanielSchiavini: Argh I get it, you wanna distinguish between `undefined` and `null`.
Then the callback should receive `T|null` not `T|undefined`. And we need to add a comment to explain why `null` is allowed to pass and `undefined` is blocked.

0xtutti (approved):

0xPearce (approved):

packages/curve-ui-kit/src/features/forms/useCallbackSync.ts
DanielSchiavini: Feel free to leave it to another PR, now this one is approved

OnlyJousting: I will leave it and merge then 🙏

PR #2665, opened 2026-06-08
fix: use useMarketOraclePrice query in candle-charts

OnlyJousting:
The ohlc chart state was still relying on an old zustand store data source for the latest oracle price while the rest of the UI had moved on

Changes:
- Use `useMarketOraclePrice` query for llamalend ohlc candle-chart latest oracle price data


0xAlunara (approved):
Nice, now when we finally delete the legacy loan forms we can also remove `pricesMapper`

0xPearce (approved):

DanielSchiavini (approved):

DanielSchiavini:
> Nice, now when we finally delete the legacy loan forms we can also remove `pricesMapper`

We can remove the whole store 🔥

PR #2666, opened 2026-06-08
feat: new scrvUSD forms

DanielSchiavini:
- implement new scrvUSD forms with the new components in beta mode
- add RPC tests


apps/main/src/loan/components/PageCrvUsdStaking/Statistics/index.tsx
```diff
@@ -39,24 +40,49 @@ const chartSelections: ChartSelections<StatisticsChart>[] = [
   { activeTitle: t`Historical Distributions`, label: chartLabels.distributions, key: 'distributions' },
 ]
 
+const useStatisticsChartControls = () => {
+  const useNewForms = useScrvUsdNewForms()
```
DanielSchiavini, line 43: We might not need this flag here, I don't see why we wanna save this stuff in a store at all

0xtutti (dismissed):

apps/main/src/loan/components/PageCrvUsdStaking/hooks/useScrvUsdDepositForm.ts
```diff
@@ -0,0 +1,58 @@
...
+  const form = useForm<ScrvUsdDepositForm>({
+    defaultValues: emptyDepositForm(),
+    validation: scrvUsdDepositFormValidationSuite,
+  })
+  const depositAmount = form.watchValue('depositAmount')
+  const approveInfinite = !!form.watchValue('approveInfinite')
+  const [params, isDebouncing] = useFormDebounce(
+    useMemo(() => ({ chainId, userAddress, depositAmount }), [chainId, depositAmount, userAddress]),
+  )
+  const userBalances = useScrvUsdUserBalances({ chainId, userAddress }, !!userAddress)
```
0xAlunara: Passing `!!userAddress` to `enabled` feels odd as the validation should already disable the query when `userAddress` is `undefined`.

apps/main/src/loan/components/PageCrvUsdStaking/CrvUsdStakingPage.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const [isChartExpanded = false, , minimizeChart, toggleChartExpanded] = useSwitch(false)
+  const { llamaApi: lendApi = null } = useCurve()
+  const { address, isConnecting } = useConnection()
+
+  const {
+    data: userScrvUsdBalance,
+    isFetching: isUserScrvUsdBalanceFetching,
+    isFetched: isUserScrvUsdBalanceFetched,
+    refetch: refetchUserScrvUsdBalance,
+  } = useScrvUsdUserBalances({ userAddress: address })
```
0xPearce: Why aren't we passing the `chainId` here like in the deposit form? 

Also checking old code we don't pass it there neither

apps/main/src/loan/components/PageCrvUsdStaking/CrvUsdStakingPage.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const inputAmount = useStore(state => state.scrvusd.inputAmount)
+  const fetchExchangeRate = useStore(state => state.scrvusd.fetchExchangeRate)
+  const fetchCrvUsdSupplies = useStore(state => state.scrvusd.fetchCrvUsdSupplies)
+  const stakingModule = useStore(state => state.scrvusd.stakingModule)
+  const chainId = lendApi?.chainId
+
+  useEffect(() => {
+    // eslint-disable-next-line @typescript-eslint/require-await -- Existing violation before enabling this rule.
+    const fetchData = async () => {
+      if (!enabled || !lendApi || !address) return
```
0xPearce, line 47: I think you meant this?
```suggestion
      if (enabled || !lendApi || !address) return
```

0xPearce (commented):
1. The gas estimation doesn't seem to be working for deposit. Also there is a "-" next to the infinte allowance`ActionInfo`, but maybe not outside the scope of this PR
Image
 
2. The chart doesn't take the available widthImage

3. Not sure if it is normal, but when withdrawing and `isFull` is true, my wallet still uses the "withdraw" method. SHouldn't it use "redeem"?

apps/main/src/loan/components/PageCrvUsdStaking/CrvUsdStakingPage.tsx
OnlyJousting: it's because scrvusd minting only exists on ethereum

DanielSchiavini: It should be passing it though!

apps/main/src/loan/components/PageCrvUsdStaking/CrvUsdStakingPage.tsx
DanielSchiavini: actually it should be `const enabled = !useScrvUsdNewForms()` :see_no_evil:

DanielSchiavini:
> 1. The gas estimation doesn't seem to be working for deposit. Also there is a "-" next to the infinte allowance`ActionInfo`, but maybe not outside the scope of this PR

Good find, we were missing the max value in that query (it's something I added in the last moment)

> 2. The chart doesn't take the available width

Good one, fixed and removed some more legacy stuff :partying_face: 

> 3. Not sure if it is normal, but when withdrawing and `isFull` is true, my wallet still uses the "withdraw" method. SHouldn't it use "redeem"?

I have checked that we are calling llamalend's redeem method.  Which calls redeem at the contract level. My wallet still says 'description: Withdraw', but that also happens in the old forms

DanielSchiavini:
[RPC run](https://github.com/curvefi/curve-frontend/actions/runs/27196032669/job/80287783139)

image

0xtutti (dismissed):

apps/main/src/loan/components/PageCrvUsdStaking/Statistics/index.tsx
DanielSchiavini: Discussed with @OnlyJousting, this is not needed indeed. Removed in https://github.com/curvefi/curve-frontend/pull/2664/commits/5b99b8d86d6cb1bcb6c824c530e7dce55e75fcac

apps/main/src/loan/components/PageCrvUsdStaking/Statistics/index.tsx
```diff
@@ -39,24 +39,49 @@ const chartSelections: ChartSelections<StatisticsChart>[] = [
...
+      revenueChartTimeOption: newFormsRevenueChartTimeOption,
+      setRevenueChartTimeOption: setNewFormsRevenueChartTimeOption,
+    }
+  }
+
+  return { selectedStatisticsChart, setSelectedStatisticsChart, revenueChartTimeOption, setRevenueChartTimeOption }
+}
+
 type StatisticsProps = {
+  chainId: ChainId | undefined
```
0xAlunara, line 65: I don't think we need the `chainId` prop, afaik scrvusd is `Chain.Ethereum` only

apps/main/src/loan/components/PageCrvUsdStaking/ScrvUsdWithdrawInfoList.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const ScrvUsdWithdrawInfoList = ({ params, networks, form }: ScrvUsdWithdrawInfoListProps) => {
+  const isOpen = form.isTouched('withdrawAmount')
+  const expectedCrvUsd = useScrvUsdPreviewWithdraw(params, isOpen)
+  return (
+    <ActionInfoCollapse isOpen={isOpen} testId="scrvusd-withdraw-action-info-list">
+      <Stack sx={{ ...ACTION_INFO_GROUP_SX }}>
+        <ActionInfo
+          label={t`You receive`}
+          value={maybe(expectedCrvUsd.data, data => formatNumber(data, { abbreviate: false }))}
```
0xAlunara: `formatNumber(data, 'token.amount')`

apps/main/src/loan/components/PageCrvUsdStaking/ScrvUsdWithdrawInfoList.tsx
```diff
@@ -0,0 +1,41 @@
...
+  networks: Record<number, NetworkConfig>
+  form: UseFormReturn<ScrvUsdWithdrawForm>
+}
+
+export const ScrvUsdWithdrawInfoList = ({ params, networks, form }: ScrvUsdWithdrawInfoListProps) => {
+  const isOpen = form.isTouched('withdrawAmount')
+  const expectedCrvUsd = useScrvUsdPreviewWithdraw(params, isOpen)
+  return (
+    <ActionInfoCollapse isOpen={isOpen} testId="scrvusd-withdraw-action-info-list">
+      <Stack sx={{ ...ACTION_INFO_GROUP_SX }}>
```
0xAlunara: `sx={ACTION_INFO_GROUP_SX}`

apps/main/src/loan/components/PageCrvUsdStaking/ScrvUsdDepositInfoList.tsx
```diff
@@ -0,0 +1,61 @@
...
+  isApproved: boolean | undefined
+  approveInfinite: boolean
+  onApproveInfiniteToggle: () => void
+}
+
+const formatTokenAmount = (value: Decimal | undefined, symbol: string) =>
+  maybe(
+    value,
+    value =>
+      `${formatNumber(value, { minimumFractionDigits: 2, maximumFractionDigits: 4, abbreviate: false })} ${symbol}`,
```
0xAlunara: Prob legacy but yeah we should use a proper category for this

apps/main/src/loan/components/PageCrvUsdStaking/ScrvUsdDepositInfoList.tsx
```diff
@@ -0,0 +1,61 @@
...
+  isApproved,
+  approveInfinite,
+  onApproveInfiniteToggle,
+}: ScrvUsdDepositInfoListProps) => {
+  const expectedScrvUsd = useScrvUsdPreviewDeposit(params, isOpen)
+  const gas = useScrvUsdDepositEstimateGas(networks, params, isOpen)
+
+  return (
+    <ActionInfoCollapse isOpen={isOpen} testId="scrvusd-deposit-action-info-list">
+      <Stack sx={{ ...ACTION_INFO_GROUP_SX }}>
```
0xAlunara: `<Stack sx={ACTION_INFO_GROUP_SX}>`

0xAlunara (commented):
We should probably hide the position details if the user has none, otherwise it's a N/A barrage.

image

0xAlunara:
This chart expand/shrink button doesn't do anything. Might as well remove it?

image

apps/main/src/loan/components/PageCrvUsdStaking/Statistics/index.tsx
DanielSchiavini: We don't _need_ it but imo it's correct to add it anyway, just like everywhere else in the crvusd app

apps/main/src/loan/components/PageCrvUsdStaking/hooks/useScrvUsdDepositForm.ts
DanielSchiavini: yes sorry for the slop

apps/main/src/loan/components/PageCrvUsdStaking/ScrvUsdWithdrawInfoList.tsx
DanielSchiavini: yeah I moved the old code without thinking

0xAlunara (dismissed):

0xAlunara (approved):

DanielSchiavini:
[RPC test run](https://github.com/curvefi/curve-frontend/actions/runs/27277718200/job/80563354102)

0xtutti (approved):

PR #2672, opened 2026-06-09
fix: keep routes in cache while page is open

DanielSchiavini:
- The form kept a routeId, but its route payload could be garbage-collected.
- Render then tried to look up that missing payload and crashed.
- The fix pins the selected route payload while the form uses its routeId.
- It does not fetch anything or keep old unselected routes alive.

### DIFF
--- a/packages/curve-ui-kit/src/entities/router-api/router-api.query.ts
+++ b/packages/curve-ui-kit/src/entities/router-api/router-api.query.ts
@@ -19,7 +19,11 @@ import { routerApiValidation } from './router-api.validation'
 type RouteByIdQuery = { routeId: string }
 type RouteByIdParams = FieldsOf<RouteByIdQuery>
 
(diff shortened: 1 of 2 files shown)


packages/curve-ui-kit/src/entities/router-api/router-api.query.ts
```diff
@@ -34,6 +38,13 @@ const { getQueryData: getRouteQueryData, setQueryData: setRouteQueryData } = que
   category: 'global.routerApi',
```
0xAlunara, line 38: Why not set `gcTime` to `Infinity` to prevent it from being garbage collected? https://tanstack.com/query/v5/docs/framework/react/reference/useQuery

DanielSchiavini: we get 5 routes every 15 seconds, if we set this to infinity it will blow up the memory and local storage in no time

0xAlunara (approved):

0xtutti (approved):

PR #2678, opened 2026-06-09
feat: enable new market list Stable channel

0xPearce:
- enable new market list Stable channel

### DIFF
--- a/packages/curve-ui-kit/src/hooks/useFeatureFlags.ts
+++ b/packages/curve-ui-kit/src/hooks/useFeatureFlags.ts
@@ -37,7 +37,7 @@ export const useLLv2 = useBetaChannel
 export const isLLv2Enabled = (releaseChannel: ReleaseChannel) => releaseChannel === ReleaseChannel.Beta
 
 /** New market list and search layout */
-export const useNewMarketListLayout = useBetaChannel
+export const useNewMarketListLayout = useStableChannel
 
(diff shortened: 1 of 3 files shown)


0xAlunara (approved):

packages/curve-ui-kit/src/shared/ui/DataTable/DataRow.tsx
```diff
@@ -53,8 +48,6 @@ export const DataRow = <T extends TableItem>({
     [url, push, hasUrl],
   )
   const visibleCells = row.getVisibleCells()
-  // to do: remove shouldApplyStickyLastRow and sx code once LegacyDataTable is deleted
-  const shouldApplyStickyLastRow = !useNewMarketListLayout() && isLastRow && shouldStickLastRowToTop
```
DanielSchiavini, line 57: we should only change the feature flag, this needs to stay the same so it works the same in legacy mode

DanielSchiavini (changes requested):

DanielSchiavini (approved):

packages/curve-ui-kit/src/shared/ui/DataTable/DataRow.tsx
DanielSchiavini: I see the problem now, the legacy table is used elsewhere

PR #2682, opened 2026-06-10
refactor: update range to liquidation calculation and tooltip description

OnlyJousting:
Changes:
- Updates "range to liquidation" calculation to show "how much the current price can drop before reaching the LT" instead of "the distance between the current price and the LT"

### DIFF
--- a/apps/main/src/llamalend/features/market-position-details/tooltips/LiquidationThresholdMetricTooltipContent.tsx
+++ b/apps/main/src/llamalend/features/market-position-details/tooltips/LiquidationThresholdMetricTooltipContent.tsx
@@ -29,7 +29,7 @@ export const LiquidationThresholdTooltipContent = ({
       <TooltipDescription
         text={[
           t`The price at which your position enters the liquidation range  and your collateral starts to be eroded by LLAMMA.`,
-          t`The distance to LT indicates the distance between the current price and the LT.`,
+          t`The distance to LT indicates how much the current price can drop before reaching the LT.`,
         ].join(' ')}
       />
       <TooltipItems secondary>
--- a/apps/main/src/llamalend/features/market-position-details/tooltips/LiquidationThresholdMetricTooltipContent.tsx
+++ b/apps/main/src/llamalend/features/market-position-details/tooltips/LiquidationThresholdMetricTooltipContent.tsx
@@ -14,7 +14,7 @@ import { formatNumber } from '@ui-kit/utils'
 
 type LiquidationThresholdTooltipContentProps = {
   userPrices: QueryProp<Range<Decimal>>
-  rangeToLiquidation: QueryProp<number>
+  rangeToLiquidation: QueryProp<Decimal>
   params: UserMarketParams
 }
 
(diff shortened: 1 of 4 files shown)


apps/main/src/llamalend/queries/user/user-prices.query.ts
```diff
@@ -12,6 +12,9 @@ import { useLoanExists } from './user-loan-exists.query'
 type UserPricesQuery = UserMarketQuery & { loanExists: boolean }
 type UserPricesParams = FieldsOf<UserPricesQuery>
 
+const calculatePriceDropToLiquidationThreshold = (currentPrice: Decimal, liquidationThreshold: Decimal) =>
+  ((+currentPrice - +liquidationThreshold) / +currentPrice) * 100
```
DanielSchiavini:
```suggestion
  decimalMultiply(decimalDiv(decimalMinus(currentPrice, liquidationThreshold), currentPrice), 100)
```

apps/main/src/llamalend/queries/user/user-prices.query.ts
```diff
@@ -12,6 +12,9 @@ import { useLoanExists } from './user-loan-exists.query'
 type UserPricesQuery = UserMarketQuery & { loanExists: boolean }
 type UserPricesParams = FieldsOf<UserPricesQuery>
 
+const calculatePriceDropToLiquidationThreshold = (currentPrice: Decimal, liquidationThreshold: Decimal) =>
```
DanielSchiavini: I think it would be a bit more readable to pass both prices and extract in the parameters, but not really important
```suggestion
const calculatePriceDropToLiquidationThreshold = (currentPrice: Decimal, [,liquidationThreshold]: [Decimal, Decimal]) =>
```

DanielSchiavini (dismissed):

apps/main/src/llamalend/queries/user/user-prices.query.ts
OnlyJousting: I didn't realise we have our own BigNumber library 👀

0xtutti (dismissed):

apps/main/src/llamalend/features/market-position-details/tooltips/LiquidationThresholdMetricTooltipContent.tsx
```diff
@@ -29,12 +29,12 @@ export const LiquidationThresholdTooltipContent = ({
...
         text={[
           t`The price at which your position enters the liquidation range  and your collateral starts to be eroded by LLAMMA.`,
-          t`The distance to LT indicates the distance between the current price and the LT.`,
+          t`The distance to LT indicates how much the current price can drop before reaching the LT.`,
         ].join(' ')}
       />
       <TooltipItems secondary>
         <TooltipItem title={t`Distance to LT`} variant="independent">
-          {rangeToLiquidation ? formatNumber(rangeToLiquidation, 'percent.value') : UnavailableNotation}
+          {rangeToLiquidation != null ? formatNumber(rangeToLiquidation, 'percent.value') : UnavailableNotation}
```
DanielSchiavini, line 37: double negative 🤓  and also totally unnecessary, rangeToLiquidation is Decimal so it will only be falsey when null/undefined. '0' is truthy.
```suggestion
          {rangeToLiquidation ? formatNumber(rangeToLiquidation, 'percent.value') : UnavailableNotation}
```
```suggestion
          {maybe(rangeToLiquidation, range => formatNumber(range, 'percent.value')) ?? UnavailableNotation}
```

DanielSchiavini (approved):

0xtutti (approved):

PR #2657, opened 2026-06-05
feat: disable zapv2 routers when unavailabie

DanielSchiavini:
- disable routers when we don't have a zap address
- allow a router to be selected even when there is no route found yet

PR #2683, opened 2026-06-10
fix: route provider story

DanielSchiavini:
- https://github.com/curvefi/curve-frontend/pull/2657 broke the route provider story. Due to the gas calculation, now we need wallet and query providers in the story.
- Also added a story with disabled cards


0xAlunara (approved):
When you refresh the routes the component suddenly grows in width 😄 

https://github.com/user-attachments/assets/47a32e80-4ad7-4525-bd13-cdc00f934d4d

DanielSchiavini:
> When you refresh the routes the component suddenly grows in width 😄
> 20260615-1718-07.1823942.mp4

Haha in practice it will already fit the container

0xtutti (approved):

PR #2685, opened 2026-06-10
fix: activity table empty state

0xPearce:
- separate empty state from error state in the Activity Table
- fix pools error shown after hydration

example new V2 markets, activity tab: https://www.curve.finance/lend/optimism/markets/0x745422BF49f3F6e4A8E12E4abD19339E7910F8C9

### DIFF
--- a/apps/main/src/llamalend/features/llamma-activity/hooks/useLlammaActivityTradesConfig.ts
+++ b/apps/main/src/llamalend/features/llamma-activity/hooks/useLlammaActivityTradesConfig.ts
@@ -67,5 +67,5 @@ export const useLlammaActivityTradesConfig = ({
     ...getTableOptions(tradesWithUrls),
   })
 
-  return { table, isLoading, isError, emptyMessage: t`No swap data found.` }
+  return { table, isLoading, isError, emptyMessage: t`No swap data found.`, errorMessage: t`Could not load swap data.` }
 }
--- a/apps/main/src/llamalend/features/llamma-activity/hooks/useLlammaActivityTradesConfig.ts
+++ b/apps/main/src/llamalend/features/llamma-activity/hooks/useLlammaActivityTradesConfig.ts
@@ -54,9 +54,6 @@ export const useLlammaActivityTradesConfig = ({
     [tradesData?.trades, networkConfig, network],
   )
 
-  const isLoading = isTradesLoading || !isHydrated || !isMarketAvailable
-  const isError = isTradesError && isMarketAvailable && isHydrated
-
   const table = useTable({
     data: tradesWithUrls,
     columns: LLAMMA_TRADES_COLUMNS,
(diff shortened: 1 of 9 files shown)


apps/main/src/llamalend/features/llamma-activity/hooks/useLlammaActivityTradesConfig.ts
```diff
@@ -67,5 +64,11 @@ export const useLlammaActivityTradesConfig = ({
     ...getTableOptions(tradesWithUrls),
   })
 
-  return { table, isLoading, isError, emptyMessage: t`No swap data found.` }
+  return {
+    table,
+    isLoading: isTradesLoading || !isHydrated || !isMarketAvailable,
+    isError: isTradesError && isHydrated && isMarketAvailable,
```
0xPearce, line 70: is waiting for hydration really necessary here?

0xPearce:
will make another PR to have consistent empty card, we have at least 3 implementations (`NoPosition`, `EmptyStateCard`, `ErrorMessage`)

0xtutti (approved):

packages/curve-ui-kit/src/features/activity-table/ActivityTable.tsx
```diff
@@ -27,29 +29,26 @@ export const ActivityTable = <TData extends TableItem>({
...
-        loading={isLoading}
-        maxHeight={height}
-        size="small"
-        expandedPanel={expandedPanel}
-      />
-    </Box>
-  )
-}
+}: ActivityTableProps<TData>) => (
+  <Box sx={{ minHeight: height }}>
```
0xAlunara, line 36: Think you could easily get rid of the scrollbar when there's no data? Having a scrollbar while there's no data is a bit janky 😅 

Image

0xAlunara (approved):

apps/main/src/llamalend/features/llamma-activity/hooks/useLlammaActivityTradesConfig.ts
0xAlunara: You can argue hydration is part of the loading process, but I think errors aren't necessarily tied to hydration

DanielSchiavini: There is no hydration anymore for llamalend lol

DanielSchiavini: https://github.com/curvefi/curve-frontend/blob/f25794b9a1419edbca81d4dd1fc38366dd10551e/apps/main/src/routes/RootLayout.tsx#L49

packages/curve-ui-kit/src/features/activity-table/ActivityTable.stories.tsx
```diff
@@ -322,7 +326,15 @@ const LoadingStateComponent = () => {
     columns: POOL_TRADES_COLUMNS,
     ...getTableOptions([]),
   })
-  return <ActivityTable table={table} isLoading={true} isError={false} emptyMessage="Loading trades..." />
+  return (
+    <ActivityTable
+      table={table}
+      isLoading={true}
```
DanielSchiavini, line 332: I think it's better to make the prop optional than faking values - why don't we have proper loading and error states?

apps/main/src/llamalend/features/llamma-activity/hooks/useLlammaActivityTradesConfig.ts
DanielSchiavini: I guess it gets called from lend/loan. I think it only makes sense to check the market object being available

0xPearce:
I will also make another PR for the addressing the last comments

PR #2692, opened 2026-06-11
feat: scroll to table top on filter and pagination change

0xPearce:
- scroll to top of the table instead of the page when filters or pagination change
next https://github.com/curvefi/curve-frontend/pull/2712


packages/curve-ui-kit/src/shared/ui/DataTable/hooks/useTableScroll.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Scrolls to the top of the table whenever the column filters change.
+ */
+export function useScrollToTopOnFilterChange<T extends TableItem>(
+  table: TanstackTable<T>,
+  tableTopRef: RefObject<HTMLElement | null>,
+) {
+  const { columnFilters } = table.getState()
+  useEffect(() => {
+    if (columnFilters.length) {
```
DanielSchiavini: This doesn't match the jsdoc, I'd expect to scroll to the top after clearing all filters, too.
The scroll does seem to work when clearing filters - maybe because we have some default filters?
So it's not clear why this `if` is necessary

packages/curve-ui-kit/src/shared/ui/DataTable/hooks/useTableScroll.ts
```diff
@@ -0,0 +1,39 @@
...
+) {
+  const { pageIndex } = table.getState().pagination
+  const lastPageIndexRef = useRef(pageIndex)
+  useEffect(() => {
+    // Avoid scrolling on the initial render, only scroll after the user changes pages.
+    if (lastPageIndexRef.current !== pageIndex) {
+      scrollToTableTop(tableTopRef)
+    }
+    lastPageIndexRef.current = pageIndex
+  }, [pageIndex, tableTopRef])
```
DanielSchiavini:
```suggestion
  const { pageIndex } = table.getState().pagination
  const lastPageIndexRef = useLatestValueRef(pageIndex)
  useEffect(() => {
    // scroll after the user changes pages
    if (lastPageIndexRef.current !== pageIndex) scrollToTableTop(tableTopRef)
  }, [pageIndex, tableTopRef])
```

DanielSchiavini (dismissed):
It works super nice! Feels very fast 🔥

OnlyJousting:
It's not working on search (another type of filter) atm, it feels great when paginating and using filters, when paginating in inline tables it scrolls up but on the market page the top of the table gets covered by the position absolute header 🤔

0xAlunara (dismissed):
Already broken in legacy, so up to you if you want to fix it or not, but when disabling the chain filter it won't scroll back up 🤔 

https://github.com/user-attachments/assets/3797db93-507a-45ab-9970-9ed88b58c144

packages/curve-ui-kit/src/shared/ui/DataTable/hooks/useTableScroll.ts
0xPearce: Right this check doesn't seems to make any sense, I just copied it from the `LegacyDataTable`. Don't know why the scroll was working for you when clearing the filters, it shouldn't. Now I removed the empty filter check, should work fine

0xPearce:
Fixed scroll on filter change so it only scrolls when filters actually change, then updated it to scroll the table container when the table has a max height, as noticed by @OnlyJousting

Will update all tables to the new `DataTable` in a next PR

OnlyJousting:
> Fixed scroll on filter change so it only scrolls when filters actually change, then updated it to scroll the table container when the table has a max height, as noticed by @OnlyJousting
> 
> Will update all tables to the new `DataTable` in a next PR

Works perfectly 🙏

OnlyJousting (approved):

0xAlunara (approved):

DanielSchiavini (approved):

PR #2690, opened 2026-06-11
feat: add new pool page header

0xAlunara:
1. Replace `usePoolAdvancedDetails` feature flag with `usePoolFreshup` as it no longer pertains just the advanced details section anymore.
2. Generalizes the market page header into a ui-kit widget.
3. Use that very same ui-kit page header for both market page and savings crvusd.
4. Implement a pool page header and also have it replace the "manage pool" page header.
5. Remove volume and tvl from new advanced details given it's now always visible in the sticky header.

**Pool page**
image

**Market page**
image

**Savings crvusd**
image

Remark: when a pool has more than 2 tokens and there's 2 rows of token icons, the header increases in height. Will create a horizontal stack variant of `TokenIcons` in a follow up PR.

PR #2702, opened 2026-06-12
feat: stack pool page header token icons horizontally

0xAlunara:
Depends on https://github.com/curvefi/curve-frontend/pull/2690

For pools with more than 2 tokens we want to stack the token icons horizontally, as using the `TokenIcons` component would increase the header height outside the norm.

Initially (as you can see in the first commit), I tried to generalize the existing `StackedChainIcon` into a more general `StackedIcons` and two specific implementations called `StackedChainIcon` and `StackedTokenIcons`, but my gut told me it was too complex for such a simple task.

In the final result I've decided to go with a simple non-generic `IconStack` that merely applies the overlapping effect via the `'& > *'` CSS selector on all its children. This means the actual icons are passed as children however you please, rather than a (render) component property of `IconStack` itself; feels more composable to me too.

The only downside is that you might have to specifiy `iconSize` of `IconStack` on top of the actual icon size property of whatever icons you pass as children, but that slight inconvenience outweights the added complexity of trying to remedy that.

**EURe pool on Gnosis**
image


packages/curve-ui-kit/src/shared/ui/IconStack.tsx
```diff
@@ -0,0 +1,42 @@
...
+export type IconStackProps = {
+  children: ReactNode
+  /** The intended size of icons that will be used in the children for correct margin calculations */
+  iconSize?: IconStackSize
+  /** Percentage of the icon width to overlap */
+  overlap?: number
+}
+
+// Builds responsive overlap offsets for each icon size, signed by direction.
+const overlapBySize = (direction: 'left' | 'right', overlap: number) => {
```
0xAlunara, line 12: Copied and untouched from the original implementation of the `StackedChainIcons` component. Well, I tried to simplify some stuff but it kept breaking visually in some way or another, so I gave up as it's not in scope.

packages/curve-ui-kit/src/widgets/PageHeader/PageHeader.stories.tsx
```diff
@@ -51,7 +51,13 @@ type Story = StoryObj<typeof PageHeader>
 export const DexPool: Story = {
   args: {
     title: 'crvUSD/USDC/USDT',
-    icon: <TokenIcons blockchainId="ethereum" tokens={[TOKENS.crvUSD, TOKENS.usdc, TOKENS.usdt]} />,
+    icon: (
+      <IconStack iconSize="lg">
+        {[TOKENS.crvUSD, TOKENS.usdc, TOKENS.usdt].map(({ address, symbol }) => (
+          <TokenIcon key={`${address}-${symbol}`} blockchainId="ethereum" address={address} tooltip={symbol} />
```
DanielSchiavini, line 57: Address should be enough in the same pool
```suggestion
          <TokenIcon key={address} blockchainId="ethereum" address={address} tooltip={symbol} />
```

DanielSchiavini (approved):
Nice! I think we could in the future merge with `TokenIcons`
Basically we should be able to configure vertical/horizontal overlapping and number of rows

0xtutti (approved):

PR #2703, opened 2026-06-13
refactor: metric query prop

DanielSchiavini:
- replace `value: Amount, loading: boolean, error: Error` prop of the `Metric` component by a `QueryProp<Amount>`
- update usages


apps/main/src/llamalend/features/market-advanced-information/AdvancedDetails.tsx
```diff
@@ -42,51 +42,50 @@ export const AdvancedDetails = ({ chainId, marketId, market, marketType }: Advan
...
+                    value: decimal(collateral.data?.totalBorrowed),
+                    symbol: collateral.data?.borrowedSymbol ?? undefined,
                   },
-                  { value: decimal(collateral?.totalBorrowed), symbol: collateral?.borrowedSymbol ?? undefined },
                 )
           }
           valueTooltip={{
             title: t`Total Collateral`,
-            body: <TotalCollateralTooltip {...collateral} />,
+            body: <TotalCollateralTooltip {...collateral.data} />,
```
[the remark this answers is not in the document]
DanielSchiavini: nonsense it doesn't throw

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -91,13 +99,14 @@ const notionalsToString = (notionals: MetricProps['notional']) =>
 /** At the moment of writing the default formatter already formats to 2 decimals, but I really want to make this explicit for potential future changes. */
 const formatChange = (value: number): string => defaultNumberFormatter(value, { decimals: 2 })
 
-type MetricValueProps = Pick<MetricProps, 'value' | 'valueOptions' | 'change' | 'testId'> & {
+type MetricValueProps = Pick<MetricProps, 'valueOptions' | 'change' | 'testId'> & {
+  value: Amount | undefined
```
0xAlunara: Not wrong, but I'd still prefer to extract the value rather than redefine the type: `value: MetricProps['value']['data']`

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -51,6 +52,13 @@ const MetricChangeSize = {
   extraLarge: 'highlightM',
 } as const satisfies Record<string, TypographyVariantKey>
 
+const MetricButtonSize = {
+  small: 'extraSmall',
```
0xAlunara, line 56: Error buttons are slightly larger now than in prod. No biggie (and no pun intended) I think.

packages/curve-ui-kit/src/widgets/PageHeader/PageHeader.stories.tsx
```diff
@@ -114,9 +115,32 @@ export const Loading: Story = {
...
-        <Metric label="24h volume" value={null} loading valueOptions={{ unit: 'dollar' }} alignment="end" />
+        <Metric
+          label="24h volume"
+          value={q({ isLoading: true, data: undefined, error: null })}
+          valueOptions={{ unit: 'dollar' }}
+          alignment="end"
+        />
+        <Metric
+          label="TVL"
+          value={q({ isLoading: true, data: undefined, error: null })}
```
0xAlunara, line 132: Thought maybe it's nice to have `constQLoading` and `constQError` but might be overkill

packages/curve-ui-kit/src/lib/queries/types.ts
```diff
@@ -1,4 +1,4 @@
-import { UseQueryOptions, UseQueryResult } from '@tanstack/react-query'
+import { type UseQueryOptions, type UseQueryResult } from '@tanstack/react-query'
```
0xAlunara:
```suggestion
import type { UseQueryOptions, UseQueryResult } from '@tanstack/react-query'
```

packages/curve-ui-kit/src/lib/queries/combine.ts
```diff
@@ -1,5 +1,6 @@
+import { useMemo } from 'react'
 import type { Decimal } from '@primitives/decimal.utils'
-import type { UseQueryOptions } from '@tanstack/react-query'
+import { type UseQueryOptions } from '@tanstack/react-query'
```
0xAlunara: Just commented about this in `types.ts`, but I think it's generally prefered to keep the `type` keyword outside of the curly braces when possible. But perhaps that's better delegated to a new linting PR.

packages/curve-ui-kit/src/lib/queries/combine.ts
```diff
@@ -15,14 +16,30 @@ type QueriesData<TQueries extends Queries> = {
   [K in keyof TQueries]: TQueries[K] extends Query<infer TData> ? Exclude<TData, undefined> : never
 }
 
+function combineQueryData<const TQueries extends Queries, TResult>(
+  queries: TQueries,
+  selector: (...data: QueriesData<TQueries>) => TResult | null | undefined,
+) {
+  return queries.some(({ data }) => data === undefined)
```
0xAlunara: No arrow body notation?

packages/curve-ui-kit/src/lib/queries/combine.ts
```diff
@@ -15,14 +16,30 @@ type QueriesData<TQueries extends Queries> = {
...
+    ? undefined
+    : selector(...(queries.map(({ data }) => data) as QueriesData<TQueries>))
+}
+
 export const combineQueries = <const TQueries extends Queries, TResult>(
```
0xAlunara, line 27: Not important for this PR, but I wonder if there's still a use for this now there's also `useCombinedQueries` with built-in `useMemo`. Using this variant smells like an anti-pattern.

packages/curve-ui-kit/src/features/integrations/components/IntegrationsList.tsx
```diff
@@ -60,16 +71,7 @@ export const IntegrationsList = ({ networkId, searchText }: { networkId?: string
   const isLoading = integrationsLoading || integrationsTagsLoading
   const isMobile = useIsMobile()
 
-  // Collect the unique list of networks from the integrations response
-  const networksQuery = useMappedQuery(integrationsQuery, data => [
-    ...new Set(
```
0xAlunara, line 65: Thanks; I wish there'd be a way to lint this such that the given `transform` is either wrapped in `useCallback` or a global module level function. Easy to forget about it an inline a function, which makes the internal `useMemo` useless.

apps/main/src/loan/components/PagePegKeepers/components/peg-status.util.ts
```diff
@@ -7,18 +7,18 @@ export type PegStatus = {
...
- *   - 'loading' if rate is undefined
+ *   - 'loading' if rate data is undefined
  *   - 'pegged' if within 2% tolerance (±0.02)
  *   - 'overpegged' if rate > 1.02 (above peg)
  *   - 'underpegged' if rate < 0.98 (below peg)
  *   Color severity: 'warning' if within 5% tolerance, 'alert' if beyond
  */
 export function pegStatus(rate: PegKeeperDetails['rate']): PegStatus {
-  if (!rate) return { label: 'loading', color: 'unselected' }
+  if (!rate.data) return { label: 'loading', color: 'unselected' }
```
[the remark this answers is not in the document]
0xAlunara: I mean, since it's now a QueryProp I suppose we could use `data.isLoading`?

apps/main/src/loan/components/PagePegKeepers/components/PegKeeperAdvancedDetails.tsx
```diff
@@ -52,8 +52,9 @@ export const PegKeeperAdvancedDetails = ({
 
     <ActionInfo
       label={t`Est. rebalance profit`}
-      loading={estCallerProfit == null}
-      value={formatNumber(amount(estCallerProfit), { decimals: 5, abbreviate: false, fallback: '-' })}
+      loading={estCallerProfit.isLoading}
```
0xAlunara, line 55: Next step is to make the same `QueryProp` PR but for `ActionInfo`? 😄

apps/main/src/llamalend/features/market-advanced-information/hooks/useAdvancedDetailsData.ts
```diff
@@ -31,85 +33,67 @@ export const useAdvancedDetailsData = ({
...
-    },
-    maxLeverage: {
-      value: maxLeverageData,
-      loading: !market || maxLeverageLoading,
-    },
-    availableLiquidity: {
-      available: capAndAvailable?.available,
-      totalAssets: capAndAvailable?.totalAssets,
-      borrowCap: capAndAvailable?.borrowCap,
-      loading: !market || capAndAvailableLoading,
```
0xAlunara, line 96: Not that it matters in this specific hook, as it'll get removed relatively soon anyway, but there's been a few cases where we have `!market` in the loading prop, because the queries are technically not loading and thus `isLoading` returns false, when in reality the market itself is being loaded and `isLoading` should return true. Might not be a problem anymore though given how hydration has gotten a revamp.

apps/main/src/llamalend/features/market-position-details/BorrowInformation.tsx
```diff
@@ -29,33 +28,27 @@ type BorrowInformationProps = {
   tokens: Partial<MarketTokens>
 }
 
-const useCollateralValue = ({
+const getCollateralValue = ({
```
0xAlunara: Shouldn't we go with `useCombinedQueries`, as otherwise we're losing the `useMemo`?

0xAlunara (commented):
Peg stability reserves seem to have some problems 🤔 

Image

packages/curve-ui-kit/src/shared/ui/Metric.tsx
DanielSchiavini: that requires opening another file and finding the type in order to understand the code :disappointed:

packages/curve-ui-kit/src/widgets/PageHeader/PageHeader.stories.tsx
DanielSchiavini: yeah this is a story

packages/curve-ui-kit/src/lib/queries/combine.ts
DanielSchiavini: I think this one is preferred! Most of the time the calculation is cheap and not worth memoizing. I only useMemo if we have a loop

packages/curve-ui-kit/src/lib/queries/combine.ts
DanielSchiavini: omg I hoped the build tools would be able to convert from one to the other. I generally like the type inside the braces because it makes easier to add other imports, but not a big issue

packages/curve-ui-kit/src/shared/ui/Metric.tsx
DanielSchiavini: I'll make it a bit smaller, but we don't have any buttons smaller than extraSmall

packages/curve-ui-kit/src/features/integrations/components/IntegrationsList.tsx
DanielSchiavini: I had initially added a lint rule to ignore memoization but I [reverted](https://github.com/curvefi/curve-frontend/commit/f79c74d0d546a9be476d3b0ae94fe5ec19627b65#diff-1685a9e933ca08d3fd9db88e6ec3519e37a0d7c0816a4e7a50a5ad86ae4bd27c) it because it can lead to errors.
We can add a lint rule no problem.

apps/main/src/loan/components/PagePegKeepers/components/PegKeeperAdvancedDetails.tsx
DanielSchiavini: Yes :pray:

apps/main/src/llamalend/features/market-position-details/BorrowInformation.tsx
DanielSchiavini: multiply & sum doesn't need memo, memoization is likely more expensive than calculation

apps/main/src/llamalend/features/market-advanced-information/hooks/useAdvancedDetailsData.ts
DanielSchiavini: It can definitely be a problem, but I'd rather solve it in a global level than manually adding !market to every query. I have a task (currently in this sprint, but dunno if I have time) to improve the loading states of the LL pages

0xPearce (commented):
As Alu mentioned, something wrong with the pegkeepers
Image

apps/main/src/llamalend/features/market-advanced-information/hooks/useAdvancedDetailsData.ts
OnlyJousting: the only reason those checks are there is to prevent `N/A` from flashing in metrics while the market object is initializing which is happening now again

DanielSchiavini: I'll think of a solution. Maybe the market should be aware of the query. Or we could introduce a market-aware Metric. Or we let the Metric show loading initially. But I definitely don't wanna repeat that for all queries

apps/main/src/loan/components/PagePegKeepers/components/PegKeeperAdvancedDetails.tsx
DanielSchiavini: But we have ~100 usages I'll try to make it so that we don't migrate everything at once

0xAlunara (dismissed):

apps/main/src/llamalend/features/market-advanced-information/hooks/useAdvancedDetailsData.ts
DanielSchiavini: - See https://github.com/curvefi/curve-frontend/pull/2703/commits/db7731c14a44de94bfcc68d9ade3c4c0340acac5 for a quick-fix and #2729 for some WIP

0xAlunara (approved):

OnlyJousting (approved):

0xPearce (approved):

PR #2705, opened 2026-06-13
styling: disable skeleton growth for market composition table

0xAlunara:
This should prevent the table for growing endlessly like in pic below

image

### DIFF
--- a/apps/main/src/dex/features/advanced-details/components/market-composition/index.tsx
+++ b/apps/main/src/dex/features/advanced-details/components/market-composition/index.tsx
@@ -2,6 +2,7 @@ import type { ChainId, PoolDataCacheOrApi } from '@/dex/types/main.types'
 import type { Pool as PricesApiPool } from '@curvefi/prices-api/pools'
 import CardHeader from '@mui/material/CardHeader'
 import Stack from '@mui/material/Stack'
+import { DEFAULT_INCREASING_LENGTH } from '@ui-kit/hooks/useIncreasingLength'
 import { t } from '@ui-kit/lib/i18n'
 import { getTableOptions, useTable } from '@ui-kit/shared/ui/DataTable/data-table.utils'
 import { DataTable } from '@ui-kit/shared/ui/DataTable/DataTable'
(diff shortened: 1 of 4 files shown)


packages/curve-ui-kit/src/hooks/useIncreasingLength.ts
```diff
@@ -7,14 +7,20 @@ export type IncreasingLengthOptions = {
...
+
 /**
  * Returns a length value that starts at `initialLength` and increases by one every `increaseEveryMs` until it reaches
  * `maxLength`.
  */
 export const useIncreasingLength = ({
-  initialLength = 3,
-  increaseEveryMs = 5000,
-  maxLength = 10,
+  initialLength = DEFAULT_INCREASING_LENGTH.initialLength,
```
0xAlunara, line 21: I tried many variations of getting this to work nicely (like destructuring inside the body rather than the in function parameters), but gippety kept overcomplicating things and breakings things too. So I kept this for now as I got tired of it.

apps/main/src/dex/features/advanced-details/components/market-composition/index.tsx
```diff
@@ -36,11 +37,12 @@ export const MarketComposition = ({
...
     <Stack>
       <CardHeader title={t`Market Composition`} size="small" />
-      <LegacyDataTable<MarketCompositionRow>
+      <DataTable<MarketCompositionRow>
         table={table}
         size="small"
-        loading={isLoading}
+        isLoading={isLoading}
         disableStickyHeader
+        increasingLengthOptions={{ ...DEFAULT_INCREASING_LENGTH, maxLength: DEFAULT_INCREASING_LENGTH.initialLength }}
```
0xPearce, line 45: `maxLength: DEFAULT_INCREASING_LENGTH.initialLength` this looks weird, why the max length should be equal to the default initial? I think the best approach here is to create once again categories. This could be reused for the [user position tables](https://github.com/curvefi/curve-frontend/blob/57605aa8a03935a22d980c590cbc1f568ae19cf9/apps/main/src/llamalend/features/market-list/UserPositionsMarketRateTable.tsx#L76-L77) as well

DanielSchiavini: I agree this reads a bit odd. Or just disable increasing
```suggestion
        increasingLengthOptions={{ increaseEveryMs: null }}
```

DanielSchiavini (approved):
It could be better but not a biggie

0xtutti (approved):

0xAlunara:
Created a ticket to work with categories instead, it does feel a bit messy rn

PR #2697, opened 2026-06-12
feat: use list endpoint for supply positions

DanielSchiavini:
- stop calling the earnings endpoint
- Make the lending data available at the market level
- cleanup usages and remove unused stuff

PR #2707, opened 2026-06-14
fix: market earnings convex validation

0xPearce:
Fixes a lend supplied amount cell error caused by the prices api returning `null` for Convex earnings fields where zod expected `strings`.

This is a temporary fix that normalizes those optional Convex fields to `0` so user position tables keep rendering correctly


0xPearce:
Fixed here: https://github.com/curvefi/curve-frontend/pull/2697

PR #2724, opened 2026-06-16
chore: leverage mocked tests

DanielSchiavini:
- update the mocked component tests so they support leveraging
- this is not the kind of test I would write by hand, this is only useful with AI-generated stuff
  - that's because the tests do not guarantee the functionality itself, but they can help us easily find regressions as we refactor the code
  - the RPC tests guarantee a much higher level of confidence but they cost $ and are slow


0xAlunara (approved):
If the tests pass I approve 🫠

0xtutti (approved):

PR #2523, opened 2026-05-14
refactor: implement tanstack form

DanielSchiavini:
- replace react-hook-form by tanstack/form

PR #2727, opened 2026-06-16
refactor: external links pt2

0xPearce:
This is just a refactor, some links are broken and fixed here: https://github.com/curvefi/curve-frontend/pull/2728

- refactor the external links
- remove dune analytics link


packages/ui/src/utils/utilsConstants.ts
```diff
@@ -7,6 +9,15 @@ export const CURVE_LOGO_URL = `${CURVE_ASSETS_URL}/branding/logo.png`
 export const CURVE_LOGO_GRAYSCALE_URL = `${CURVE_ASSETS_URL}/branding/logo-bw.svg`
 export const ERROR_IMAGE_URL = `${CURVE_ASSETS_URL}/branding/four-oh-llama.jpg`
 
+/** Dot-path union of every URL leaf in EXTERNAL_LINKS, e.g. "brand.assets". */
```
DanielSchiavini: Couldn't we use something like [`FieldPath`](https://github.com/curvefi/curve-frontend/blob/147d983ade00730b80fc5c4e37a9d72cca8f0928/packages/curve-ui-kit/src/features/forms/form.types.ts#L7)? Anyway I think this type shouldn't be specific to `typeof EXTERNAL_LINKS`

DanielSchiavini (dismissed):

packages/ui/src/utils/utilsConstants.ts
0xPearce: I was wondering if there was already something similar, but `Path` comes from `react-hook-form` it’s better to define something independently

0xtutti (approved):

packages/ui/src/utils/utilsConstants.ts
```diff
@@ -36,38 +50,60 @@ export const EXTERNAL_LINKS = {
...
   },
 } as const
 
+/** Dot-path union of every URL leaf in EXTERNAL_LINKS, e.g. "brand.assets". */
+export type ExternalLinkKey = DotPathByValue<typeof EXTERNAL_LINKS, string>
+
+/** Resolves a typed EXTERNAL_LINKS dot-path to its URL. */
+export const getExternalLink = (link: ExternalLinkKey) => {
+  const value = get(EXTERNAL_LINKS, link)
+  return assert(typeof value === 'string' && value, `Unknown external link: ${link}`)
```
0xAlunara, line 104: What's the point of assertion when the `ExternalLinkKey` type is already derived from `typeof EXTERNAL_LINKS`? I don't see how this would fail without a type error

0xAlunara (approved):

OnlyJousting (approved):

DanielSchiavini (approved):

packages/ui/src/utils/utilsConstants.ts
DanielSchiavini: I'd just use the react form one. #2523 already replaces it by tanstack

packages/ui/src/utils/utilsConstants.ts
DanielSchiavini: I guess `get` doesn't use the same type

PR #2729, opened 2026-06-16
refactor: llama markets loading state

DanielSchiavini:
- update llamalend market pages to stop flashing loading states
  - wait for loan state
  - when loan exists, wait for soft liquidation state
- remove the loading skeleton completely, instead reuse the existing curve loading spinner


packages/curve-ui-kit/src/features/candle-chart/hooks/useLiquidationRange.ts
```diff
@@ -25,24 +27,30 @@ export const useLiquidationRange = ({
   fallbackData = [],
   currentPrices,
   newPrices,
-}: UseLiquidationRangeProps): LiquidationRanges =>
-  useMemo(() => {
-    const timeSeriesData: TimeSeriesData = chartData.length > 0 ? chartData : fallbackData
+}: UseLiquidationRangeProps) =>
+  useMappedQuery(
+    currentPrices,
+    useCallback(
```
0xPearce, line 36: For some reason the new liquidations ranges seems to never be executed

<table>
  <tr>
    <th>Before


</th>
    <th>After</th>
  </tr>
  <tr>
    <td>

Image

</td>
    <td>

Image

</td>
  </tr>
</table>

0xPearce: you can test when opening a new position, for example here: lend/ethereum/markets/0x3DE37c38739dFb83b7A902842bF5393040f7BF50

DanielSchiavini: Thank you! While creating loans the prevRanges aren't allowed to be null rn

0xPearce (approved):

packages/primitives/src/string.utils.ts
```diff
@@ -38,3 +38,5 @@ export const joinButtonText = (...texts: (string | Falsy)[]) =>
   notFalsy(...texts)
     .map((t, i, all) => (i ? `${i === all.length - 1 ? ' & ' : ', '}${t}` : t))
     .join('')
+
+export const capitalize = (string: string) => string.charAt(0).toUpperCase() + string.slice(1)
```
0xAlunara, line 42: Why not use lodash for this? (Assuming doing it by CSS with `text-transform: capitalize` wasn't an option, too)

0xAlunara (approved):

packages/primitives/src/string.utils.ts
DanielSchiavini: Sure I can do it in the follow-up #2847

PR #2733, opened 2026-06-17
chore: avoid checking css props in tests

DanielSchiavini:
- Firefox can (sometimes?) report the `none` transform as `matrix(1, 0, 0, 1, 0, 0)` - which is equivalent.
- Instead of checking css properties, we should just make the testId depend on the transition state. That should be always correct.

### DIFF
--- a/apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersOverlay.tsx
+++ b/apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersOverlay.tsx
@@ -1,4 +1,4 @@
-import type { RefObject } from 'react'
+import { useState, type RefObject } from 'react'
 import { LlamaMarket } from '@/llamalend/queries/market-list/llama-markets'
 import Button from '@mui/material/Button'
 import IconButton from '@mui/material/IconButton'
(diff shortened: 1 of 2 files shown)


0xPearce (dismissed):

apps/main/src/llamalend/features/market-list/filters/LlamaTableFiltersOverlay.tsx
```diff
@@ -71,9 +72,10 @@ export const LlamaTableFiltersOverlay = ({
         paper: {
           sx: { backgroundColor: t => t.design.Layer[3].Fill, width: Width.modal.md },
         },
+        transition: { onEntered: () => setTestId('table-filters-popover'), onExit: () => setTestId(null) },
```
0xPearce, line 76: maybe we could have added a comment to remember why we did that

0xPearce: sorry forgot to send it

DanielSchiavini:
```suggestion
        // set the testid once transition is ready to avoid flaky tests trying to click during transition
        transition: { onEntered: () => setTestId('table-filters-popover'), onExit: () => setTestId(null) },
```

0xAlunara (approved):

0xtutti (approved):

PR #2741, opened 2026-06-17
feat: add and use new api for merkl proxy

0xAlunara:
Creates a new `app` which is just a proxy server for fetching Merkl opportunities with an optional API key.

The infra and setup is basically a copy paste of the router-api and how it uses Fastify.

@DanielSchiavini When this passes review and gets deployed I need some help on how to configure the `MERKL_API_KEY` environment variable on Vercel. I'm not even sure if I have sufficient rights and access to add, remove or modify environment variables in the first place. I've already generated a key fwiw.

EDIT: there's a `MERKLE_API_KEY` env variables on Vercel now, but it'll only work for preview builds, as I don't have sufficient rights to also enable it for production.


apps/merkl-api/test/opportunities.test.ts
```diff
@@ -0,0 +1,122 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { createMerklServer } from '../src/server'
+
+const MERKL_API_URL = 'https://merkl.test'
+
+const jsonResponse = (payload: unknown, init: ResponseInit = {}) =>
```
0xAlunara, line 6: Admittedly most of it is generated by gippety 😄 
Looked ok to me

apps/main/_api/merkl.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const server = createMerklServer()
+
+/**
+ * Vercel handler for API routes, using fastify under the hood. This is only used when deployed to Vercel.
+ * While in development, vite's dev server redirects /api/merkl/* calls to the merkl dev server.
+ * This file is automatically detected by vite-plugin-vercel because it is in the "_api" folder.
+ */
+export default async function handler(request: IncomingMessage, response: ServerResponse) {
+  const start = Date.now()
```
DanielSchiavini, line 12: this implementation is duplicate, it should be moved to some shared file imo

apps/merkl-api/src/routes/opportunities.ts
```diff
@@ -0,0 +1,48 @@
+import type { FastifyReply, FastifyRequest } from 'fastify'
+import type { OpportunitiesQuery } from './opportunities.schemas'
+
+const DEFAULT_MERKL_API_URL = 'https://api.merkl.xyz'
+
+type MerklConfig = {
+  MERKL_API_URL?: string
+  MERKL_API_KEY?: string
```
DanielSchiavini: optional?

apps/merkl-api/src/routes/opportunities.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export const getOpportunities =
+  ({ MERKL_API_URL = DEFAULT_MERKL_API_URL, MERKL_API_KEY }: MerklConfig = {}) =>
+  async (request: FastifyRequest<{ Querystring: OpportunitiesQuery }>, reply: FastifyReply) => {
+    const url = buildMerklOpportunitiesUrl(MERKL_API_URL, request.query)
+    const headers: Record<string, string> = { accept: 'application/json' }
+    if (MERKL_API_KEY) headers['X-API-Key'] = MERKL_API_KEY
+
+    const response = await fetch(url, { method: 'GET', headers })
```
DanielSchiavini, line 24: how about fetchJson from @primitives/fetch?

apps/merkl-api/src/dev.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Starts the Fastify server and sets up graceful shutdown handlers.
+ * This function is invoked when running the application locally.
+ * On Vercel, we use the server instance directly in the API route handler of the main app.
+ */
+async function start(): Promise<void> {
+  const server = createMerklServer()
+
+  const stopServer = (signal: NodeJS.Signals) => {
```
DanielSchiavini: another duplicate 🤨

apps/merkl-api/src/routes/opportunities.schemas.ts
```diff
@@ -0,0 +1,29 @@
+export const OpportunitiesPath = '/api/merkl/v1/opportunities'
+
+const opportunitiesQuerySchema = {
+  type: 'object',
+  additionalProperties: false,
+  properties: {
+    mainProtocolId: { type: 'string', minLength: 1 },
```
DanielSchiavini: we should hardcode only the protocols we use so our proxy isn't just used for anybody who wants a free API key
```suggestion
    mainProtocolId: { type: 'string', minLength: 1, enum: ['curve', 'llamalend'] },
```

apps/merkl-api/src/routes/opportunities.schemas.ts
```diff
@@ -0,0 +1,29 @@
...
+    test: { type: 'boolean' },
+    status: { type: 'string', minLength: 1 },
+    action: { type: 'string', minLength: 1 },
+    items: { type: 'integer', minimum: 1 },
+    page: { type: 'integer', minimum: 0 },
```
DanielSchiavini: we should hardcode only the stuff we use so our proxy isn't just used for anybody who wants a free API key
```suggestion
    mainProtocolId: { type: 'string', minLength: 1, enum: ['curve', 'llamalend'] },
    test: { type: 'boolean },
    status: { type: 'string', minLength: 1, enum: ['LIVE'] },
    action: { type: 'string', minLength: 1 },
    items: { type: 'integer', minimum: 1 },
    page: { type: 'integer', minimum: 0 },
```

apps/merkl-api/src/routes/opportunities.ts
0xAlunara: Thought I could get away with not depending on any package, even primitives, given it's literally just a proxy API.

apps/merkl-api/src/routes/opportunities.ts
0xAlunara: Technically not yet required, but yeah might as well make it required 🤔 
Url was optional for mocking / testing, but not sure if that's the case anymore after some modifications

apps/main/_api/merkl.ts
0xAlunara: I've created a shared `api-package` package. Didn't want to re-use primitives as I don't want to pollute it with API specific code.

apps/merkl-api/test/opportunities.test.ts
```diff
@@ -0,0 +1,155 @@
...
+  new Response(JSON.stringify(payload), {
+    status: 200,
+    headers: { 'content-type': 'application/json' },
+    ...init,
+  })
+
+const getFetchCall = (fetchMock: ReturnType<typeof vi.fn<typeof fetch>>) => {
+  expect(fetchMock).toHaveBeenCalled()
```
DanielSchiavini, line 14: imo we shouldn't be mocking anything but testing with real API keys to make sure it actually works

0xPearce (approved):
👍

packages/api-server/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "lint": "eslint --cache --cache-location ../../.cache/eslint/api-server --cache-strategy content .",
+    "lint:fix": "eslint --cache --cache-location ../../.cache/eslint/api-server --cache-strategy content . --fix",
+    "typecheck": "tsc --tsBuildInfoFile ../../.cache/typecheck/packages-api-server.tsbuildinfo",
+    "test": "vitest run",
+    "test:watch": "vitest"
+  },
+  "main": "src/index.ts",
+  "types": "src/index.ts",
+  "dependencies": {
+    "fastify": "*"
```
DanielSchiavini, line 19: The version could be here instead of in the root json

DanielSchiavini (approved):

packages/api-server/package.json
0xAlunara: Do peerDependencies work like that? Note that I didn't make `api-server` a proper dependency for `merkl-api` and `router-api`, we're just aliasing imports (this is how we do stuff apparantly)

PR #2748, opened 2026-06-18
feat: llama market pages without wallet

DanielSchiavini:
- make market pages available without a wallet connected by using API data
- the code isn't the cleanest but I think it suffices while we come up with a cleaner way to handle fallbacks
- depends on #2774 and #2775
- I'm also working on #2729 so I'm trying to avoid even more changes


apps/main/src/llamalend/widgets/MarketRateCurveChart.tsx
```diff
@@ -203,7 +212,10 @@ export const MarketRateCurveChart = ({
             size="medium"
             label={t`Total collateral`}
             {...tokenMetric({
-              value: combinedCollateral,
+              value: fallbackQ(
+                combinedCollateral,
+                mapQuery(apiMarket, m => m.totalCollateralUsd),
```
0xPearce: Here we need to display the token amount instead of the USD value. Also we recently decided to combine the collateral AND the converted borrow token into a metric that is collateral denominated
<table>
  <tr>
    <th>Before



</th>
    <th>After</th>
  </tr>
  <tr>
    <td>Image


</td>
    <td>Image
</td>
  </tr>
</table>

0xPearce (commented):
Feels like a dream! Also the token icons on the form are not showing without a wallet

Image

apps/main/src/llamalend/widgets/MarketRateCurveChart.tsx
DanielSchiavini: Do you know how to map that to the API data?

DanielSchiavini:
> Feels like a dream! Also the token icons on the form are not showing without a wallet

Yes that's why I'm removing the market object in the upcoming PR (#2781)

apps/main/src/llamalend/widgets/MarketRateCurveChart.tsx
0xPearce: We calculate it on the frontend, the api doesn't expose that. Here is the [helper](https://github.com/curvefi/curve-frontend/blob/b6f0a3ce19ca2f531ce53d9502eb00affe653f5b/apps/main/src/llamalend/widgets/MarketRateCurveChart.tsx#L60) and here an [example](https://github.com/curvefi/curve-frontend/blob/b6f0a3ce19ca2f531ce53d9502eb00affe653f5b/apps/main/src/llamalend/widgets/MarketRateCurveChart.tsx#L122)

0xtutti (dismissed):

apps/main/src/llamalend/widgets/MarketRateCurveChart.tsx
```diff
@@ -203,7 +212,12 @@ export const MarketRateCurveChart = ({
             size="medium"
             label={t`Total collateral`}
             {...tokenMetric({
-              value: combinedCollateral,
+              value: fallbackQ(
+                combinedCollateral,
+                combineQueries([apiMarket, collateralUsdRate], (market, collateralUsdRate) =>
+                  collateralUsdRate ? decimal(market.totalCollateralUsd / collateralUsdRate) : undefined,
```
0xPearce, line 218: mmh didn't think about calculating the total collateral like this, we should replace the calculation for the wallet connected data as well it's a lot simpler. Also it should be done in the same helper imo so we don't forget to change one if we change the other. I can do it in another PR

tests/cypress/e2e/all/header.cy.ts
```diff
@@ -33,14 +33,17 @@ const expectedFooterMaxWidth = 1536
...
+
   describe('Desktop', () => {
     let route: AppRoute
 
     beforeEach(() => {
       ;[width, height] = oneDesktopViewport()
       cy.viewport(width, height)
       dismissPhishingWarningBanner()
-      route = oneAppRoute()
+      route = 'llamalend/ethereum/markets' as const
```
0xPearce: why hardcode the llamalend route?

apps/main/src/llamalend/widgets/MarketRateCurveChart.tsx
0xPearce: Also we do not feed the values to the tooltip when wallet isn't connected
Image

apps/main/src/llamalend/widgets/page-header/hooks/usePageHeader.ts
```diff
@@ -127,29 +131,58 @@ export const useBorrowRate = ({
...
+    useApiMarket,
+  )
+  const apiBorrowRate = mapQuery(apiMarket, d => ({
+    rate: d.rates.borrowApr,
+    averageRate: averageApr,
+    averageCategory: 'llamalend.marketList.rate' as const,
+    rebasingYield: d.rates.borrowApr,
+    totalBorrowRate: d.rates.borrowTotalApr,
+    totalAverageBorrowRate: totalAverageBorrowApr,
+    extraRewards: d.rewards,
```
0xPearce: Also noticed in the page header some values are not matching: like the average window and the tooltip's rebasing yield




<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>Screenshot 2026-06-24 at 12 21 02</td>
    <td>
Image
</td>
  </tr>
</table>


```suggestion
    averageCategory: RATE_CATEGORY,
    rebasingYield: d.assets.collateral.rebasingYieldApr,
    totalBorrowRate: d.rates.borrowTotalApr,
    totalAverageBorrowRate: totalAverageBorrowApr,
    extraRewards: useFilteredRewards(d.rewards, marketType, MarketRateType.Borrow),
```

apps/main/src/llamalend/widgets/page-header/hooks/usePageHeader.ts
```diff
@@ -164,27 +197,59 @@ export const useSupplyRate = ({
...
+        rewardsApr: rates.incentives.map(({ percentage, title, address }) => ({
+          apy: percentage,
+          symbol: title,
+          tokenAddress: address,
+        })),
+      },
+      lendingSnapshots: apiSupplySnapshots.snapshots ?? undefined,
+      campaigns: apiSupplyCampaigns,
+      blockchainId,
+      category: 'llamalend.marketList.rate',
```
0xPearce:
```suggestion
      category: RATE_CATEGORY,
```

0xPearce (commented):
One last thing, the form's connect wallet button is one size smaller than the other forms' button. Something we can fix in the followup PR

<table>
  <tr>
    <th>Not connected
</th>
    <th>Connected</th>
  </tr>
  <tr>
    <td>Image</td>
    <td>Image
</td>
  </tr>
</table>

tests/cypress/e2e/all/header.cy.ts
DanielSchiavini: Oh sorry that was just to reproduce an issue

apps/main/src/llamalend/widgets/page-header/hooks/usePageHeader.ts
DanielSchiavini: Good find thank you

apps/main/src/llamalend/widgets/MarketRateCurveChart.tsx
DanielSchiavini: The tooltip is part of #2781 because it requires a lot of changes

> replace the calculation for the wallet connected data

The implementation only uses API data when the wallet is disconnected. I'm not sure I follow

0xPearce (approved):

0xtutti (approved):

PR #2750, opened 2026-06-18
feat: support merkle campaigns in borrow rates

0xAlunara:
This adds merkl campaign support merkl rates in the borrow and supply data for markets.

Instead of a `multiplier` property, there's now a proper union type for different kind of rewards in `CampaignRewards`.

### DIFF
--- a/apps/main/src/llamalend/rates.utils.ts
+++ b/apps/main/src/llamalend/rates.utils.ts
@@ -1,6 +1,7 @@
 import { sumBy } from 'lodash'
 import type { Decimal } from '@primitives/decimal.utils'
 import { maybe, maybes, notFalsy } from '@primitives/objects.utils'
+import type { CampaignRewards } from '@ui-kit/entities/campaigns'
 import type { CrvUsdSnapshot } from '@ui-kit/entities/crvusd-snapshots'
 import type { LendingSnapshot } from '@ui-kit/entities/lending-snapshots'
 import type { ExtraIncentive } from '@ui-kit/types/market'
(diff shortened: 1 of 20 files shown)


0xtutti (dismissed):

0xtutti (approved):

0xPearce (approved):

apps/main/src/llamalend/widgets/tooltips/RewardTooltipItems.tsx
```diff
@@ -43,16 +43,14 @@ export const RewardsTooltipItems = ({
...
+      {extraRewards.map(
+        (r, i) =>
           r.action === tooltipType && (
-            // eslint-disable-next-line @eslint-react/no-array-index-key -- Existing violation before enabling this rule.
             <TooltipItem
               variant="subItem"
+              // eslint-disable-next-line @eslint-react/no-array-index-key -- Existing violation before enabling this rule.
               key={i}
-              title={rewardType === 'apr' ? t`APR` : t`Points`}
+              title={r.reward?.type === 'apr' ? r.symbol || '' : t`Points`}
```
DanielSchiavini, line 53: `|| ''` is incorrect, we should alllow `title` to receive `null` or `undefined`!

apps/main/src/llamalend/rates.utils.ts
```diff
@@ -77,6 +82,22 @@ type OnChainSupplyRewardApr = { apy: number; symbol: string; tokenAddress: strin
...
+export const sumCampaignsApr = (campaigns: CampaignRewards[] | undefined) =>
+  campaigns && campaigns.length > 0
+    ? sumBy(
+        campaigns.filter(c => c.reward?.type === 'apr'),
+        c => c.reward?.value ?? 0,
+      )
+    : null
+
+export const sumCampaignsApy = (campaigns: CampaignRewards[] | undefined) =>
+  campaigns && campaigns.length > 0
```
DanielSchiavini, line 94: `campaigns.length == 0` should return `0` sum not `null`

0xAlunara: You want null because we don't want to render anything at all; otherwise we'd see 0% APR everywhere when campaigns are not applicable.

apps/main/src/llamalend/widgets/tooltips/RewardTooltipItems.tsx
0xAlunara: I think these will be replaced in the future anyway 😄

apps/main/src/llamalend/rates.utils.ts
DanielSchiavini: I could say this logic belongs in the component not the helper but this isn't important 😁

PR #2754, opened 2026-06-18
fix: denominate util breakdown with correct tokens

0xAlunara:
1. Total supplied shold be denominated in collateral token
2. Available to borrow should be denominated in borrow token
3. Debt ceiling should be pinned to crvUSD as it's only applicable to mint markets

No dollar values allowed!

I also put the token icon on the right for better alignment

**Old and incorrect**
image

**New and correct**
image

EDIT: reverted back to borrowed token for supplied amount
image

### DIFF
--- a/apps/main/src/llamalend/features/market-list/cells/UtilizationCell.tsx
+++ b/apps/main/src/llamalend/features/market-list/cells/UtilizationCell.tsx
@@ -60,7 +60,8 @@ const UtilizationTooltipContent = ({
         <TooltipItem variant="primary" title={t`Utilization breakdown`} />
         {type === LlamaMarketType.Lend && (
           <TooltipItem title={t`Total supplied`}>
-            <Currency {...collateral} />
+            {/* The supplied token is the same as the token people borrow, even though we use the collateral balance in this case */}
+            <Currency {...borrowed} balance={collateral.balance} />
           </TooltipItem>
         )}
         <TooltipItem title={t`Total borrowed`}>
--- a/apps/main/src/llamalend/features/market-list/cells/UtilizationCell.tsx
+++ b/apps/main/src/llamalend/features/market-list/cells/UtilizationCell.tsx
@@ -15,8 +15,6 @@ import { formatNumber } from '@ui-kit/utils'
 
 const { Spacing } = SizesAndSpaces
 
-const isUsd = (symbol: string) => symbol === 'USD' || symbol === 'crvUSD' // show crvUSD as USD
-
 const Currency = ({
   balance,
   symbol,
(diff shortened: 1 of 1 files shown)


0xtutti (dismissed):

apps/main/src/llamalend/features/market-list/cells/UtilizationCell.tsx
```diff
@@ -74,19 +60,22 @@ const UtilizationTooltipContent = ({
         <TooltipItem variant="primary" title={t`Utilization breakdown`} />
         {type === LlamaMarketType.Lend && (
           <TooltipItem title={t`Total supplied`}>
-            {/* as we are displaying the utilization breakdown, display everything as borrowed token */}
-            <Currency {...borrowed} balance={collateral.balanceUsd} />
+            <Currency {...collateral} />
```
0xPearce: What? Why are showing the collateral token for the total supplied?? 
Users will think that there is 463k WBTC supplied. All three values of the breakdown should borrow token denominated, or am I missing something?

Image

0xPearce (dismissed):

apps/main/src/llamalend/features/market-list/cells/UtilizationCell.tsx
```diff
@@ -74,19 +60,22 @@ const UtilizationTooltipContent = ({
...
           </TooltipItem>
         )}
         <TooltipItem title={t`Total borrowed`}>
           <Currency {...borrowed} />
         </TooltipItem>
-        <TooltipItem title={t`Available to borrow`}>
-          <Currency {...borrowed} balance={liquidityUsd} />
-        </TooltipItem>
+        {/** liquidityUsd is as the name says, in usd, but we want it denominated in the borrow token, and we have enough info to deduce the borrow token price directly (if both are > 0) */}
+        {borrowed.balance && borrowed.balanceUsd && (
```
DanielSchiavini, line 71: These are numbers not Decimal I think? Then you have the risk of printing `0`.
Better use `maybes`

apps/main/src/llamalend/features/market-list/cells/UtilizationCell.tsx
DanielSchiavini: I don't think we should convert supplied values :thinking:

0xAlunara: I confused myself by thinking "supplied" is what people put up as collateral, but we call that "deposit". Changed the comment slightly and reverted back to the borrowed token, but with the collateral value except non-usd this time.

image

apps/main/src/llamalend/features/market-list/cells/UtilizationCell.tsx
0xAlunara: The type of these values is `number | null`, and I explicitly only want to show the line when both values > 0

DanielSchiavini (approved):

OnlyJousting (approved):

PR #2760, opened 2026-06-19
fix: campaign tooltip fixed position

0xPearce:
This fixes the bug where the tooltip in the campaign banner jumps around on hover. It is now fixed to the left


0xAlunara (approved):

DanielSchiavini (approved):
Looks like a workaround rather than a fix.. 🤓

0xPearce:
> Looks like a workaround rather than a fix.. 🤓

100% a workaround because all the campaign's related components are legacy, it should be refactored

PR #2763, opened 2026-06-19
feat: denominate bands in borrow token

OnlyJousting:
Updates the LlamaLend bands chart to denominate band liquidity, shares, and tooltip values in the market borrow token instead of USD, while simplifying the chart data flow around total band values.

Changes:
- Calculate collateral value, total band liquidity, and user liquidity in borrow-token terms.
- Use precomputed band totals for chart sizing, zoom bounds, and tooltip percentages.
- Replace USD notional tooltip values with borrow-token liquidity formatting.
- Simplify band balance processing with Decimal values and shared query types.
- Add decimalSqrt utility with unit coverage.

### DIFF
--- a/apps/main/src/llamalend/features/bands-chart/TooltipContent.tsx
+++ b/apps/main/src/llamalend/features/bands-chart/TooltipContent.tsx
@@ -5,12 +5,11 @@ import { t } from '@ui-kit/lib/i18n'
 import { formatChartAxisNumber } from '@ui-kit/shared/ui/Chart'
 import { LegendBox } from '@ui-kit/shared/ui/Chart/LegendSet'
 import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
-import { decimalDiv, decimalGreaterThan, decimalMultiply, formatNumber } from '@ui-kit/utils'
+import { decimalDiv, decimalGreaterThan, decimalMultiply, formatNumber, ZERO } from '@ui-kit/utils'
 import { useBandsChartPalette } from './hooks/useBandsChartPalette'
 import type { BandsChartToken, ChartDataPoint } from './types'
 
 const { Spacing } = SizesAndSpaces
-const ZERO: Decimal = '0'
 
 type TooltipContentProps = {
   data: ChartDataPoint
--- a/apps/main/src/llamalend/features/bands-chart/TooltipContent.tsx
+++ b/apps/main/src/llamalend/features/bands-chart/TooltipContent.tsx
@@ -26,10 +26,14 @@ const calculateBandShare = (numerator: Decimal | undefined, denominator: Decimal
     : '?'
 
 const formatAbbreviatedNumber = (value: Decimal | undefined): string =>
-  value != null ? `${formatNumber(value, { abbreviate: true })}` : '?'
+  formatNumber(value, { abbreviate: true, fallback: '?' })
 
-const formatBorrowTokenValue = (value: Decimal | undefined, borrowTokenSuffix: string): string =>
-  value != null ? `${formatNumber(value, { abbreviate: true })}${borrowTokenSuffix}` : '?'
+const formatBorrowTokenValue = (value: Decimal | undefined, borrowTokenSymbol: string | undefined): string =>
+  formatNumber(value, {
+    abbreviate: true,
+    fallback: '?',
+    unit: { symbol: borrowTokenSymbol ? ` ${borrowTokenSymbol}` : '', position: 'suffix' },
+  })
 
 export const TooltipContent = ({ data, collateralToken, borrowToken }: TooltipContentProps) => {
   const palette = useBandsChartPalette()
@@ -72,7 +76,7 @@ export const TooltipContent = ({ data, collateralToken, borrowToken }: TooltipCo
               </TooltipItem>
             </TooltipItems>
             <TooltipItem variant="primary" title={t`Your band liquidity`}>
-              {formatBorrowTokenValue(data.userBandTotalValue, borrowTokenSuffix)}
+              {formatBorrowTokenValue(data.userBandTotalValue, borrowToken?.symbol)}
             </TooltipItem>
           </Stack>
         )}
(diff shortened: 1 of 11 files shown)


0xtutti (dismissed):

apps/main/src/llamalend/features/bands-chart/hooks/useDerivedChartData.ts
```diff
@@ -1,13 +1,18 @@
 import { useMemo } from 'react'
-import { ChartDataPoint, DerivedChartData } from '../types'
+import type { Decimal } from '@primitives/decimal.utils'
+import { decimalSum } from '@ui-kit/utils/decimal'
+import type { ChartDataPoint, DerivedChartData } from '../types'
+
+const ZERO: Decimal = '0'
```
0xAlunara: probably better suited inside `decimal.ts`?

apps/main/src/llamalend/features/bands-chart/hooks/useDerivedChartData.ts
```diff
@@ -16,25 +21,24 @@ import { ChartDataPoint, DerivedChartData } from '../types'
...
 
     for (const d of chartData) {
       // Use median price for y-axis labels
       yAxisData.push(d.pUpDownMedian)
 
-      const totalBandValue = (d.bandCollateralValueUsd ?? 0) + (d.bandBorrowedValueUsd ?? 0)
-      const userCollateralValue = d.userBandCollateralValueUsd ?? 0
-      const userBorrowedValue = d.userBandBorrowedValueUsd ?? 0
-      const marketBandValue = totalBandValue - userCollateralValue - userBorrowedValue
+      const userCollateralValue = d.userBandCollateralValue ?? ZERO
```
0xAlunara: I guess it might already be numbers, but it's a bit inconsistent to be using `toNumber` for the decimals and here we copy paste `?? ZERO`.

packages/curve-ui-kit/src/utils/decimal.spec.ts
```diff
@@ -77,3 +77,11 @@ describe('amount', () => {
     },
   )
 })
+
+describe('decimalSqrt', () => {
+  it('returns square roots as Decimal strings', () => {
+    expect(decimalSqrt('4')).toBe('2')
+    expect(decimalSqrt('2.25')).toBe('1.5')
+    expect(decimalSqrt('0.000000000000000001')).toBe('0.000000001')
```
0xAlunara, line 85: I wonder what it does for negative numbers

apps/main/src/llamalend/features/bands-chart/TooltipContent.tsx
```diff
@@ -1,40 +1,46 @@
...
 import { LegendBox } from '@ui-kit/shared/ui/Chart/LegendSet'
 import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
-import { formatNumber } from '@ui-kit/utils'
+import { decimalDiv, decimalGreaterThan, decimalMultiply, formatNumber } from '@ui-kit/utils'
 import { useBandsChartPalette } from './hooks/useBandsChartPalette'
-import { BandsChartToken, ChartDataPoint } from './types'
+import type { BandsChartToken, ChartDataPoint } from './types'
 
 const { Spacing } = SizesAndSpaces
+const ZERO: Decimal = '0'
```
0xAlunara: Duplicate const, guess that's why it's preferable inside `decimal.ts`

apps/main/src/llamalend/features/bands-chart/TooltipContent.tsx
```diff
@@ -1,40 +1,46 @@
...
 
-const calculateBandShare = (numerator: number | undefined, denominator: number | undefined): string =>
-  typeof denominator === 'number' && denominator > 0 && typeof numerator === 'number'
-    ? `${formatNumber((numerator / denominator) * 100, 'percent.rate')}`
+const isPositiveDecimal = (value: Decimal | undefined): boolean => value != null && decimalGreaterThan(value, ZERO)
+
+const calculateBandShare = (numerator: Decimal | undefined, denominator: Decimal | undefined): string =>
+  numerator != null && denominator != null && decimalGreaterThan(denominator, ZERO)
+    ? `${formatNumber(decimalMultiply(decimalDiv(numerator, denominator), '100'), 'percent.rate')}`
     : '?'
```
0xAlunara, line 25: We not going with the default formatter fallback of `-`?

I guess not since we already override the fallback in `formatBorrowTokenValue `

OnlyJousting: I just kept the behaviour that was already in use tbh

apps/main/src/llamalend/features/bands-chart/hooks/useDerivedChartData.ts
DanielSchiavini: It's unnecessary, ZERO is longer than '0'

apps/main/src/llamalend/features/bands-chart/hooks/useDerivedChartData.ts
DanielSchiavini: I know we already had all those null=0 here, but as a bit of boy-scouting, can't we handle null/undefined properly instead? Do we get them regularly? If we really need 0, please explain why in comments (it could make sense in a chart context)

packages/curve-ui-kit/src/utils/decimal.spec.ts
OnlyJousting: ah, BigNumber returns NaN for negative numbers so adding a guard, good catch ser

apps/main/src/llamalend/features/bands-chart/hooks/useDerivedChartData.ts
DanielSchiavini: I stand corrected, sometimes '0' as Decimal is necessary

DanielSchiavini (approved):

apps/main/src/llamalend/features/bands-chart/hooks/useDerivedChartData.ts
OnlyJousting: I think `0` is correct here. These derived arrays are chart geometry, not display values. The missing values are expected after merging market bands with user bands, they mean “this segment does not exist for this band.” ECharts needs a numeric width, and `0` avoids `NaN` in width stacking and x-axis bounds. I'll add some comments and improve the helper names so maybe it's a little more clear

0xAlunara (approved):

PR #2765, opened 2026-06-19
fix: points campaigns mixup in new advanced details tables

0xAlunara:
Fixes two bugs:

1. Merkl points campaigns would show up in the yield breakdown, as Merkl points campaigns have an APR of 0 as its only distinction. Very annoying.
2. Points campaigns would not show non-multiplier Merkl campaigns. I know the column is called 'Multiplier' and it doesn't make sense for non-multipliers, but these are edge cases and I don't want to rename the column again.

As a bonus it also removes the aggregation footer row in yield breakdown if there's only 1 row, making aggregation useless.

`0x317837aed98bea887074d1f97fd3c83ebca6905b`
image

`0xf4d0cf32908b2c7f1021339c43df0f77f06896d7`
image

`0xefc6516323fbd28e80b85a497b65a86243a54b3e`
image

### DIFF
--- a/apps/main/src/dex/features/advanced-details/hooks/usePointsCampaigns.tsx
+++ b/apps/main/src/dex/features/advanced-details/hooks/usePointsCampaigns.tsx
@@ -28,9 +28,9 @@ export const usePointsCampaigns = ({
   const rows = useMemo(
     () =>
       campaigns
-        .filter(({ reward }) => reward?.type === 'points')
+        .filter(({ reward, symbol }) => reward?.type === 'points' || (!reward?.type && symbol))
         .map(
-          ({ dashboardLink, reward, platform, platformImageId }): PointsCampaignsRow => ({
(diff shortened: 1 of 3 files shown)


apps/main/src/dex/features/advanced-details/hooks/usePointsCampaigns.tsx
```diff
@@ -28,9 +28,9 @@ export const usePointsCampaigns = ({
   const rows = useMemo(
     () =>
       campaigns
-        .filter(({ reward }) => reward?.type === 'points')
+        .filter(({ reward, symbol }) => reward?.type === 'points' || (!reward?.type && symbol))
```
0xAlunara, line 31: The type of these campaigns isn't technically points, but for now we want to show them in this table regardless as we have no better alternative.

DanielSchiavini (approved):

0xtutti (approved):

PR #2767, opened 2026-06-19
fix: token symbol and notional metrics

0xPearce:
- created an helper to have consistent token symbol and notional value for a metric
- fixed market page's borow cap and total collateral metrics

Screenshot 2026-06-19 at 19 10 31
Screenshot 2026-06-19 at 19 10 54

### DIFF
--- a/apps/main/src/llamalend/llama.utils.ts
+++ b/apps/main/src/llamalend/llama.utils.ts
@@ -13,12 +13,13 @@ import { getUserMarketCollateralEvents as getLendUserMarketCollateralEvents } fr
 import type { BadDebt } from '@curvefi/prices-api/liquidations'
 import { type Address, Hex } from '@primitives/address.utils'
 import type { Amount, Decimal } from '@primitives/decimal.utils'
-import { assert, maybe, notFalsy, objectKeys } from '@primitives/objects.utils'
+import { assert, maybe, maybes, notFalsy, objectKeys } from '@primitives/objects.utils'
 import { getLib, requireLib, type Wallet } from '@ui-kit/features/connect-wallet'
-import { isZapV2Enabled } from '@ui-kit/hooks/useFeatureFlags'
 import { t } from '@ui-kit/lib/i18n'
+import { MetricProps } from '@ui-kit/shared/ui/Metric'
 import { LlamaMarketType, LlamaMarketVersion } from '@ui-kit/types/market'
-import { CRVUSD, decimalMinus, decimalSum, formatNumber } from '@ui-kit/utils'
+import { QueryProp } from '@ui-kit/types/util'
+import { CRVUSD, decimal, decimalMinus, decimalMultiply, decimalSum, formatNumber } from '@ui-kit/utils'
 import { SOLVENCY_THRESHOLDS } from './llama-markets.constants'
 
 /**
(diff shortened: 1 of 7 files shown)


0xtutti (dismissed):

apps/main/src/llamalend/llama.utils.ts
```diff
@@ -438,3 +439,51 @@ export const lowSolvencyDeprecatedMessage = (solvencyPercent: number | null) =>
...
+
+/** Builds Metric props for a token-denominated value with the matching USD notional shown below it. */
+export const tokenMetric = ({
+  value,
+  symbol,
+  usdRate,
+}: {
+  value: MetricProps['value']
+  symbol: string | null | undefined
+  usdRate: QueryProp<Amount | null | undefined>
```
0xAlunara: Not that it matters much, but `QueryProp` already implies `undefined`

apps/main/src/llamalend/llama.utils.ts
```diff
@@ -438,3 +439,51 @@ export const lowSolvencyDeprecatedMessage = (solvencyPercent: number | null) =>
...
+  const notionalValue = maybes([decimal(value.data), usdRate.data], ([value, usdRate]) =>
+    decimalMultiply(value, usdRate),
+  )
+  return {
+    value,
+    valueOptions: {
+      abbreviate: true,
+      unit: maybe(symbol, symbol => ({ symbol, position: 'suffix' as const })),
+    },
+    notional: maybe(notionalValue, value => ({ value, unit: 'dollar' as const })),
```
0xAlunara:
```suggestion
  notional: maybes([decimal(value.data), usdRate.data], ([value, usdRate]) => ({
    value: decimalMultiply(value, usdRate),
    unit: 'dollar' as const,
  })),
```
And then we can use arrow-body notation to remove `return`

apps/main/src/llamalend/llama.utils.ts
```diff
@@ -438,3 +439,51 @@ export const lowSolvencyDeprecatedMessage = (solvencyPercent: number | null) =>
...
+
+/**
+ * Returns the total collateral expressed in collateral-token units.
+ * The borrowed-token portion can appear in collateral after soft liquidation, so it must be converted through USD rates
+ * before being added to the native collateral amount.
+ *
+ * Example: 10 WETH collateral + 4,000 crvUSD borrowed collateral, with WETH at $2,000 and crvUSD at $1,
+ * becomes 12 WETH: 10 + (4,000 * 1 / 2,000).
+ */
+export const calculateCombinedCollateral = ({
```
0xAlunara: I'd keep this local to the only user `MarketRateCurveChart.tsx` as not to pollute the global util scope

apps/main/src/llamalend/widgets/tooltips/TotalCollateralTooltip.tsx
```diff
@@ -38,7 +38,9 @@ export const TotalCollateralTooltip = ({
 
   return (
     <TooltipWrapper>
-      <TooltipDescription text={t`The total USD value of all collateral assets deposited in this market.`} />
+      <TooltipDescription
+        text={t`The total collateral deposited and converted in this market, shown in the collateral token.`}
```
0xAlunara:
```suggestion
        text={t`The total collateral deposited and converted in this market, all denominated in the collateral token.`}
```

0xAlunara (commented):
In the future we will go from token symbols to using a `TokenIcon` as `valueRight` and you'll all praise Aludamus!

It also contradicts the standard usage of tokens elsewhere like in the util tooltip and table columns
image

image

DanielSchiavini (approved):

0xtutti (approved):

PR #2777, opened 2026-06-23
chore: add missing MERKL_API_KEY to e2e github action

0xAlunara:
Also had to add support for `.env.local` in case you want to use the Merkl API key without having to commit to production `.env`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -200,6 +200,8 @@ jobs:
       - name: Start merkl api server
         run: yarn dev 2>&1 | tee ../../e2e-api-${{ matrix.browser }}-${{ matrix.count }}.log &
         working-directory: apps/merkl-api
+        env:
+          MERKL_API_KEY: ${{ secrets.MERKL_API_KEY }}
 
       - name: e2e tests
         working-directory: tests
--- a/packages/api-server/src/index.ts
+++ b/packages/api-server/src/index.ts
@@ -6,6 +6,14 @@ type ApiServerEnv = Partial<
 
 type DevServerEnv = Partial<Record<'HOST' | 'PORT', string | undefined>>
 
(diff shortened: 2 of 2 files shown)


packages/api-server/src/index.ts
```diff
@@ -6,6 +6,14 @@ type ApiServerEnv = Partial<
 
 type DevServerEnv = Partial<Record<'HOST' | 'PORT', string | undefined>>
 
+const loadEnvFile = (path?: string) => {
+  try {
+    process.loadEnvFile(path)
+  } catch (error) {
+    if ((error as NodeJS.ErrnoException).code !== 'ENOENT') throw error
```
DanielSchiavini, line 13: probably simpler to check if the file exists

.github/workflows/ci.yaml
```diff
@@ -200,6 +200,8 @@ jobs:
       - name: Start merkl api server
         run: yarn dev 2>&1 | tee ../../e2e-api-${{ matrix.browser }}-${{ matrix.count }}.log &
         working-directory: apps/merkl-api
+        env:
+          MERKL_API_KEY: ${{ secrets.MERKL_API_KEY }}
```
DanielSchiavini, line 204: How do we set this up?

DanielSchiavini (approved):

0xPearce (approved):

.github/workflows/ci.yaml
0xAlunara: This is a normal github secret like the Enso API key; I've filled it in already

packages/api-server/src/index.ts
0xAlunara: The data has to be loaded and parsed as `process.env` anyway with this function that we must call, so you get the file check for free by the error being thrown.

PR #2778, opened 2026-06-23
refactor: v2 pool list updates

OnlyJousting:
Changes:
- Do APR to APY conversion in front-end (was pending to be added as a field on the API)
- Remove "pool" from pool-list visibility toggles
- Inline `TooltipBaseApy` in cell


0xtutti (dismissed):

apps/main/src/dex/features/pool-list/hooks/usePoolListVisibilitySettings.tsx
```diff
@@ -8,7 +8,7 @@ import { POOL_LIST_COLUMNS, POOL_LIST_COLUMN_OPTIONS, PoolListColumnId, type Poo
 import type { PoolListSorting } from './usePoolListSorting'
 
 const migration: MigrationOptions<Record<PoolListColumnVariant, VisibilityGroup<PoolListColumnId>[]>> = {
-  version: 1,
+  version: 2,
```
0xPearce, line 11: doesn't matter right now because it's only in Beta, but in the future we should keep the user preferences like we do in the Llama market list when changing version

apps/main/src/dex/components/TooltipBaseApy.tsx
```diff
@@ -8,11 +8,11 @@ import { formatNumber } from '@ui-kit/utils'
...
 
 type TooltipBaseApyProps = {
-  baseDailyApr: V2Pool['baseDailyApr']
-  baseWeeklyApr: V2Pool['baseWeeklyApr']
+  baseDailyApy: PoolListItem['baseDailyApy']
+  baseWeeklyApy: PoolListItem['baseWeeklyApy']
 }
 
-export const TooltipBaseApy = ({ baseDailyApr, baseWeeklyApr }: TooltipBaseApyProps) => (
+export const TooltipBaseApy = ({ baseDailyApy, baseWeeklyApy }: TooltipBaseApyProps) => (
```
0xPearce: I don't think we should be passing APY. IMO, we should pass the APR as far as possible since it is the base from which we calculate the APY.

This makes it much easier to switch between APY and APR if needed, without having to change all the prop drilling. Also, if we add a setting in the future to switch between APR and APY, we should be able to rely on APR everywhere.

TL;DR: pass the APR and only convert to APY when displaying

0xPearce: wdyt?

0xPearce (changes requested):

apps/main/src/dex/features/pool-list/hooks/usePoolListVisibilitySettings.tsx
OnlyJousting: yea that nobody should have had time to set preferences was my thinking when trying to just sneak this change in, especially that nobody would have toggled the pool names off (as it blows up the table 😅)

apps/main/src/dex/components/TooltipBaseApy.tsx
OnlyJousting: yea no problem! I made the update to run the calculation just before displaying, it would have been ok to just rename the prop to `baseDailyRate` and control what value to display in `getPoolListItem` too to not have to touch as many fields though

apps/main/src/dex/features/pool-list/poolList.utils.ts
```diff
@@ -20,6 +22,8 @@ export const getPoolIdByAddressEntries = (poolMapper: PoolIdByAddressSource | un
 export const getCurvePoolIdByAddressEntries = (curve: CurveApi) =>
   curve.getPoolList().map(poolId => [normalizeAddress(curve.getPool(poolId).address), poolId] as const)
 
+export const getPoolYieldApy = (apr: number | null | undefined) => aprToApy(apr, POOL_YIELD_COMPOUND_WINDOW)
```
0xPearce, line 25:
```suggestion
export const getPoolYieldApy = (apr: number | null | undefined) => aprToApy(apr, AVERAGE_CATEGORIES['dex.poolYield.compoundRate'].window)
```

In the future I want to add the category directly in the helper 

```suggestion
export const getPoolYieldApy = (apr: number | null | undefined) => aprToApy(apr, 'dex.poolYield.compoundRate')
```

0xPearce (dismissed):

apps/main/src/dex/features/pool-list/columns/column.definitions.tsx
```diff
@@ -48,6 +48,7 @@ export const POOL_LIST_TITLES: Record<PoolListColumnId, string> = {
 export const POOL_LIST_COLUMNS = [
   accessor(PoolListColumnId.PoolName, 'name', {
     cell: PoolListTitleCell,
+    enableHiding: false,
```
DanielSchiavini: This is weird we originally just removed non-hidable columns from the visibility groups - why is this needed?

0xtutti (dismissed):

apps/main/src/dex/features/pool-list/columns/column.definitions.tsx
OnlyJousting: you're right, this was redundant

0xPearce (approved):

0xAlunara (approved):

DanielSchiavini (approved):

PR #2781, opened 2026-06-23
refactor: market props

DanielSchiavini:
- prop-drill market details instead of retrieving everything from the market object
- this allows the forms to show as much information as they possibly can
- all the queries are still disabled since marketId is not set from API data
- I'm not super happy with the amount of props we are passing - however, each form/page requires a slightly different set of props - and creating an abstraction would necessarily mean passing more than they need
- Unfortunately, the market object itself is kept since we use it to reset the LLjs cache ([useOnChangeCallback usages](https://github.com/search?q=repo%3Acurvefi/curve-frontend%20useOnChangeCallback&type=code))
- depends on #2748


0xtutti (dismissed):

apps/main/src/llamalend/features/borrow/hooks/useCreateLoanForm.tsx
```diff
@@ -181,19 +189,19 @@ export function useCreateLoanForm<ChainId extends LlamaChainId>({
...
     routes: useMarketRoutes({
       chainId,
-      marketAddress: getAmmAddress(market),
+      marketAddress: ammAddress,
       tokenIn: borrowToken,
       tokenOut: collateralToken,
       amountIn: decimalSum(params.debt, params.userBorrowed),
       ...pick(params, 'slippage'),
-      enabled: params.leverageEnabled && !!market && hasZapV2(market),
+      enabled: params.leverageEnabled && !!zapAddress,
```
0xPearce, line 190: is this `zapAddress` really equivalent to this  `!!market && hasZapV2(market)`?

0xPearce (commented):
ChatGPT Image Jun 26, 2026, 09_45_45 AM

apps/main/src/llamalend/features/borrow/hooks/useCreateLoanForm.tsx
DanielSchiavini: https://github.com/curvefi/curve-frontend/pull/2781/commits/c4d5a195d92d7966afe2157ccafb7235dbc4777c

packages/curve-ui-kit/src/features/forms/FormButton.tsx
```diff
@@ -34,7 +34,7 @@ export const FormButton = ({
         type="submit"
         disabled={disabled}
         fullWidth={fullWidth}
-        loading={loading ? true : pending}
+        loading={loading ?? pending}
```
0xPearce, line 37: It will never fallback to `pending` if `loading` is `false` by default?

0xPearce (approved):

OnlyJousting (approved):

packages/curve-ui-kit/src/features/forms/FormButton.tsx
DanielSchiavini: It's more like the pending state always shows loading.
This should be || though
It used to be, but it generated a lint warning which made me make this mistake. I think given the usages it should be ok

0xAlunara (approved):

PR #2785, opened 2026-06-24
chore: remove legacy market table

0xPearce:
- remove legacy market table code

Wanted to remove the `LegacyDataTable` as well but the dex still uses it


OnlyJousting (approved):
❤️‍🔥

DanielSchiavini (approved):

PR #2789, opened 2026-06-25
feat: new market fields from API

DanielSchiavini:
- add new mint fields available from the api
- make the market page display available market parameters
- depends on #2781


0xPearce (commented):
Image

I was trying locally and the market list page crashed as soon as I changed to this branch. I think the "mint-markets" cache is the problem, we should bump that version

0xtutti (dismissed):

0xtutti (approved):

0xPearce (approved):

PR #2790, opened 2026-06-25
feat: split up advanced details

0xAlunara:
Splits up the advanced details feature into two new smaller components:

1. Pool information, showing some metrics, composition, yield breakdown and points campaigns
2. Advanced details (but this time, it's just a bunch of action infos showing parameters and prices)

There's a new 'View more' button that's used for expanding cards, which resembles an accordion, but is different.

image


apps/main/src/dex/features/advanced-details/components/Info.tsx
```diff
@@ -31,61 +34,53 @@ const getPoolType = ({
   return pool.implementation
 }
 
-export const useParameters = ({
```
0xAlunara, line 34: I've removed the single `useParameters` hook as the data fetched from this hook got split over multiple sub components. There's a bit of repetition regarding
```tsx
  const { pool } = poolDataCacheOrApi
  const poolAddress = pool.address as Address
  const { data: network } = useNetworkByChain({ chainId })
  const chain = network.networkId as BlockchainId
```
I tried a small hook that would remove this repetition among the components, but it felt like an anti pattern. We'll have to revisit when we refactor away `poolDataCacheOrApi` in the future.

apps/main/src/dex/features/advanced-details/components/Parameters.tsx
```diff
@@ -31,26 +30,29 @@ export const Parameters = ({
...
-  } = useParameters({ chainId, poolDataCacheOrApi, poolId })
+    future_A,
+    future_A_time,
+    gamma,
+    adminFee = '',
+    fee,
+  } = parameters ?? {}
+  const formatADisplay = (a: number | string | undefined) =>
+    formatNumber(amount(!isFxSwap || a == null ? a : formatCryptoA(a, FXSWAP)), { abbreviate: false, fallback: '-' })
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
```
0xAlunara, line 54: Pre-existing

0xtutti (dismissed):

DanielSchiavini (approved):

apps/main/src/dex/features/advanced-details/components/Info.tsx
DanielSchiavini: I think we could inline network if it's not used elsewhere. Otherwise it makes sense
Also please use getBlockchainId instead of casting to Chain

0xtutti (approved):

PR #2792, opened 2026-06-25
feat: new liquidation buffer & health bar

0xPearce:
Implemented the new Liquidation Buffer and Health bar.

This introduces a new way of displaying Llamalend positions, closer to the model users may already know from other lending platforms, while still accounting for LLAMMA’s liquidation protection range. The previous single health bar is split into two separate metrics: Health and Liquidation Buffer.

1. **Health bar**: shows how far the position is from entering liquidation protection / soft liquidation. It is calculated as `healthFull - healthNotFull`, representing the above-band cushion before the position reaches the liquidation range.
2. **Liquidation Buffer bar**: maps to healthNotFull. It represents the remaining buffer before hard liquidation once the position is evaluated against the liquidation protection bands. 

PS: This is still a work in progress. A lot of health-related copy doesn't match this new implementation and will need follow-up refinement.

Screenshot 2026-06-26 at 15 33 32
Screenshot 2026-06-26 at 15 33 55


packages/curve-ui-kit/src/utils/number.ts
```diff
@@ -227,6 +227,7 @@ const NUMBER_FORMAT_CATEGORIES = {
     maximumSignificantDigits: undefined,
     fallback: '-',
   },
+  'number.compact': { abbreviate: true, fallback: '-' },
```
0xAlunara: Is the same as `token.compact` but diff context I suppose

apps/main/src/llamalend/features/market-position-details/tooltips/LiquidationBufferTooltipContent.tsx
```diff
@@ -0,0 +1,9 @@
+import { TooltipDescription, TooltipWrapper } from '@/llamalend/widgets/tooltips/TooltipComponents'
+import { t } from '@ui-kit/lib/i18n'
+
+export const LiquidationBufferTooltipContent = () => (
+  <TooltipWrapper>
+    <TooltipDescription text={t`Liquidation Buffer shows the remaining buffer before hard liquidation can occur.`} />
+    <TooltipDescription text={t`When at 0, hard liquidation can occur.`} />
```
0xAlunara: `may occur` sounds friendler

apps/main/src/llamalend/features/market-position-details/tooltips/LegacyHealthTooltipContent.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { t, Trans } from '@ui-kit/lib/i18n'
+
+export const LegacyHealthTooltipContent = () => (
+  <TooltipWrapper>
+    <TooltipDescription text={t`Health shows your cushion before hard liquidation.`} />
+    <TooltipDescription
+      text={
+        <Trans>
+          Higher is safer;{' '}
+          <Box component="span" sx={{ color: 'text.highlight' }}>
```
0xAlunara, line 12: I suppose it works but I'd expect `<Typography component='span'>` here

EDIT: nvm i guess it's legacy shit

apps/main/src/llamalend/features/market-position-details/health/HealthDetails.tsx
```diff
@@ -0,0 +1,54 @@
...
+      health: { tooltip: HEALTH_TOOLTIP, value: mapQuery(healthQuery, d => d.health) },
+      liquidationBuffer: {
+        tooltip: LIQUIDATION_BUFFER_TOOLTIP,
+        value: mapQuery(healthQuery, d => d.liquidationBuffer),
+      },
+    } satisfies Record<HealthType, HealthMetricConfig>
+  )[type]
+
+  return (
+    <Stack>
```
0xAlunara: What's the purpose of a Stack with 1 element?

apps/main/src/llamalend/features/market-position-details/health/HealthDetails.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <Stack sx={{ gap: Spacing.xs }}>
```
0xAlunara: Another Stack with 1 element, wouldn't that also make the gap useless?

apps/main/src/llamalend/features/market-position-details/health/HealthDetails.tsx
```diff
@@ -0,0 +1,54 @@
...
+        <Stack direction="row" sx={{ alignItems: 'flex-end', gap: Spacing.md.mobile }}>
```
0xAlunara: Do we really need `.mobile`? What's wrong with just `Spacing.md`

apps/main/src/llamalend/features/market-position-details/health/HealthAndBufferBar.tsx
```diff
@@ -0,0 +1,158 @@
...
+      </WithWrapper>
+    </Grid>
+  )
+}
+
+export const HealthAndBufferBar = ({ healthQuery, sx }: { healthQuery: HealthQuery; sx?: SxProps }) => {
+  const { state, type: activeType } = getHealthDetailsState(healthQuery.data)
+
+  return (
+    <Grid container columnSpacing={DIVIDER_SPACING} sx={applySxProps(sx)}>
```
0xAlunara: No need for `applySxProps` right? Can just pass `sx={sx}`

apps/main/src/llamalend/features/market-position-details/health/HealthAndBufferBar.tsx
```diff
@@ -0,0 +1,158 @@
...
+  },
+}
+
+const DashedDivider = () => (
+  <Box
+    component="span"
+    sx={{
+      alignSelf: 'stretch',
+      borderRight: theme => `2px dashed ${theme.design.Color.Neutral[400]}`,
+      pointerEvents: 'none',
```
0xAlunara: I don't think `pointerEvents: 'none'` does anything for a simple `Box`?

apps/main/src/llamalend/features/market-position-details/health/HealthDetails.tsx
0xAlunara: On desktop and dark mode it also makes it look too dense imo

apps/main/src/llamalend/features/market-position-details/health/HealthDetails.tsx
```diff
@@ -0,0 +1,54 @@
...
+          <Metric
```
0xAlunara: Do we still want this Metric? We're showing duplicate numbers (either buffer or health, in this case 4.16)

0xAlunara (dismissed):

apps/main/src/llamalend/features/market-position-details/health/HealthAndBufferBar.tsx
```diff
@@ -0,0 +1,158 @@
...
+        Wrapper={Stack}
+        sx={{ flex: 1, alignItems: 'stretch' }}
+        spacing={DIVIDER_SPACING}
+        direction="row"
+      >
+        {withDivider && <DashedDivider />}
+        <Tooltip title={tooltip.title} body={tooltip.body}>
+          <Stack sx={{ flex: 1 }}>
+            <Typography variant="bodyXsRegular" color="textTertiary">
+              {/* TODO: add the parenthesis as a number unit? */}
```
DanielSchiavini: Parenthesis isn't a number unit - it could be a category but I don't think it makes sense, either.
I do miss a fallback value here though, does it make sense to hide parenthesis completely while loading?

apps/main/src/llamalend/features/market-position-details/health/HealthBar.stories.tsx
```diff
@@ -1,7 +1,7 @@
 import Box from '@mui/material/Box'
 import Stack from '@mui/material/Stack'
 import type { Meta, StoryObj } from '@storybook/react-vite'
-import { HealthBar } from './'
+import { HealthBar } from '..'
```
DanielSchiavini: The stories file is usually co-located with the component

apps/main/src/llamalend/features/market-position-details/health/HealthDetails.tsx
```diff
@@ -0,0 +1,54 @@
...
+            label={tooltip.shortTitle}
+            value={value}
+            valueOptions={{
+              unit: 'none',
+              // TODO: fix metric color (either accept all string or return primitives "caution", "success" ...)
```
DanielSchiavini: `MetricProps.valueOptions.color` is just `TypographyProps['color']` right now which is
`OverridableStringUnion<'primary' | 'secondary' | 'success' | 'error' | 'info' | 'warning' | `text${Capitalize<keyof TypeText>}`, TypographyPropsColorOverrides> | (string & {}) | undefined`

So string is in there already 🤔

apps/main/src/llamalend/features/market-position-details/health/utils.ts
```diff
@@ -0,0 +1,128 @@
...
+} as const
+
+const LIQUIDATION_BUFFER_THRESHOLDS: Record<Exclude<LiquidationBufferState, 'risky'>, number> = {
+  /** Below this value the position is hard liquidated */
+  hardLiquidation: 0,
+  /** Below this value the position is critical. Above it is risky */
+  critical: 10,
+} as const
+
+export const clampPercentage = (health: number | undefined | null): number => Math.max(0, Math.min(health ?? 0, 100))
```
DanielSchiavini, line 38: Is this really correct? Aren't we trying to show something as percentage that's not actually 0~100 range?

apps/main/src/llamalend/features/market-position-details/health/utils.ts
```diff
@@ -0,0 +1,128 @@
...
+
+export const getHealthState = (health: number): HealthState => {
+  if (health > HEALTH_THRESHOLDS.good) return 'pristine'
+  if (health > HEALTH_THRESHOLDS.tight) return 'good'
+  if (health > HEALTH_THRESHOLDS.softLiquidation) return 'tight'
+  return 'softLiquidation'
+}
```
DanielSchiavini:
```suggestion
export const getHealthState = (health: number): HealthState =>
  recordEntries(HEALTH_THRESHOLDS).find(([, threshold]) => health <= threshold)?.[0] ?? 'pristine'
```

apps/main/src/llamalend/features/market-position-details/health/utils.ts
```diff
@@ -0,0 +1,128 @@
...
+
+export const getLiquidationBufferState = (liquidationBuffer: number): LiquidationBufferState => {
+  if (liquidationBuffer <= LIQUIDATION_BUFFER_THRESHOLDS.hardLiquidation) return 'hardLiquidation'
+  if (liquidationBuffer < LIQUIDATION_BUFFER_THRESHOLDS.critical) return 'critical'
+  return 'risky'
+}
```
DanielSchiavini:
```suggestion
export const getLiquidationBufferState = (liquidationBuffer: number): LiquidationBufferState =>
  recordEntries(LIQUIDATION_BUFFER_THRESHOLDS).find(([, threshold]) => liquidationBuffer <= thresold)?.[0] ?? 'risky'
```

apps/main/src/llamalend/features/market-position-details/health/utils.ts
```diff
@@ -0,0 +1,128 @@
...
+export const getHealthPercent = (state: HealthAndBufferState | undefined, health: Decimal | null | undefined) => {
+  if (health == null || state == null) return 0
+  switch (state) {
+    case 'pristine':
+    case 'hardLiquidation':
+      return 100
+    case 'good':
+    case 'tight':
+      // TODO: get the last thresholds rather then hardcoding it
+      return clampPercentage((+health / HEALTH_THRESHOLDS.good) * 100)
```
DanielSchiavini:
```suggestion
      return clampPercentage((+health / recordValues(HEALTH_THRESHOLDS).at(-1)) * 100)
```

apps/main/src/llamalend/features/market-position-details/health/utils.ts
```diff
@@ -0,0 +1,128 @@
...
+    default:
+      return 100
+  }
+}
+
+export const getHealthDetailsState = (healthData: QueryData<typeof useUserHealthValue> | undefined) => {
+  const { health, liquidationBuffer } = healthData ?? {}
+  // it returns the current type of the position, to either show the "health" or the "liquidationBuffer"
+  const type: HealthType =
+    liquidationBuffer != null && health != null
```
DanielSchiavini: maybe use `maybes`?

apps/main/src/llamalend/features/market-position-details/BorrowPositionDetails.tsx
```diff
@@ -20,6 +22,8 @@ export const BorrowPositionDetails = ({
   const statusContent =
     liquidationStatus.data &&
     getPositionStatusContent(collateralToken?.symbol, borrowToken?.symbol)[liquidationStatus.data]
+  const softLiquidation = mapQuery(liquidationStatus, positionStatus => positionStatus === 'softLiquidation')
```
DanielSchiavini, line 22: I'd keep this inside the LegacyHealthDetails branch

packages/curve-ui-kit/src/utils/number.ts
DanielSchiavini: This is not a valid category name imo, it says nothing about what kind of number it is
```suggestion
  'health': { abbreviate: true, fallback: '-' },
```

OnlyJousting:
I think it would help to make the component less busy by cutting down the dashed splitline to just the bar height or maybe just remove it all together 🤔

SCR-20260629-nzao

apps/main/src/llamalend/features/market-position-details/health/utils.ts
0xPearce: Exactly, we’re clamping the health and liquidation buffer values and treating them as percentages below a certain threshold in order to display them in a bar, even though they are not actually percentages.

apps/main/src/llamalend/features/market-position-details/health/HealthDetails.tsx
0xPearce: right, the actual "bug" is that MUI Typography color={...} only resolves registered variant color names at runtime. Also the metric component seems already a little broken when using the variant color names, so I prefer fixing that in a separate task 
Screenshot 2026-06-29 at 15 21 08

apps/main/src/llamalend/features/market-position-details/BorrowPositionDetails.tsx
0xPearce: why? I have to pass that or the `liquidationStatus`

apps/main/src/llamalend/features/market-position-details/health/HealthDetails.tsx
0xPearce: forgot to remove it, there was another child

apps/main/src/llamalend/features/market-position-details/health/HealthDetails.tsx
0xPearce: no this one has two children: the metric and the bars

apps/main/src/llamalend/features/market-position-details/health/HealthDetails.tsx
0xPearce: nope, copied from the Legacy implementation but didn't noticed

apps/main/src/llamalend/features/market-position-details/health/HealthDetails.tsx
0xPearce: I removed the numbers with parenthesis

0xPearce:
Screenshot 2026-06-29 at 16 25 34

@0xtutti wdyt?

apps/main/src/llamalend/features/market-position-details/health/utils.ts
```diff
@@ -0,0 +1,142 @@
+import { useUserHealthValue } from '@/llamalend/queries/user/user-health.query'
+import type { Theme } from '@mui/material/styles'
+import { Decimal } from '@primitives/decimal.utils'
+import { maybe, maybes, recordEntries, recordValues } from '@primitives/objects.utils'
+import { QueryData } from '@ui-kit/lib'
+
+const HEALTH_UPPER_BOUND_STATE = 'pristine' as const
+const LIQ_BUFFER_UPPER_BOUND_STATE = 'light' as const
```
0xPearce, line 8: I refactored using:
1.  this boundary states
2. refactored the helpers below using an object that satisfies a Record rather than using a switch
to make it easier to add a state because Typescript will complain.

apps/main/src/llamalend/features/market-position-details/health/utils.ts
```diff
@@ -0,0 +1,142 @@
...
+    pristine: Layer.Feedback.Success,
+    good: Layer.Feedback.Success,
+    caution: Layer.Feedback.Caution,
+    tight: Layer.Feedback.Warning,
+    softLiquidation: undefined,
+    light: undefined,
+    risky: undefined,
+    critical: undefined,
+    hardLiquidation: Layer.Feedback.Error,
+  } satisfies Record<HealthAndBufferState, string | undefined>
```
DanielSchiavini, line 59: the switch wasn't wrong but this is nicer imo 🤓

apps/main/src/llamalend/features/market-position-details/tooltips/LiquidationBufferTooltipContent.tsx
```diff
@@ -0,0 +1,9 @@
+import { TooltipDescription, TooltipWrapper } from '@/llamalend/widgets/tooltips/TooltipComponents'
+import { t } from '@ui-kit/lib/i18n'
+
+export const LiquidationBufferTooltipContent = () => (
```
DanielSchiavini, line 4: too simple to deserve its own file

DanielSchiavini (approved):
Nice!

apps/main/src/llamalend/features/market-position-details/health/HealthAndBufferBar.tsx
```diff
@@ -0,0 +1,132 @@
...
+type HealthQuery = QueryProp<QueryData<typeof useUserHealthValue>>
+
+const STATE_LABEL: Record<HealthAndBufferState, string> = {
+  pristine: t`Pristine`,
+  good: t`Good`,
+  caution: t`Caution`,
+  tight: t`Tight`,
+  softLiquidation: t`Soft Liquidation`,
+  light: t`At risk`,
+  risky: t`At risk`,
```
0xAlunara, line 36: Duplicate 'at risk`?

0xAlunara (approved):

apps/main/src/llamalend/features/market-position-details/health/HealthAndBufferBar.tsx
0xPearce: We ran out of words 😆 . No they are basically the same state, we have two of them because one remains between the Liquidation Buffer bar limit. It's like `good` and `pristine`, but we left the same name for the label

PR #2793, opened 2026-06-26
feat: add more pool list v2 filter options

OnlyJousting:
Adds API-backed filter controls to the DEX pool list, including TVL, volume, Base vAPY, and pool type filtering with URL persistence, active filter chips, reset behavior, and broader Cypress coverage. Also refactors pool-list table/query code to reduce duplicated logic and improve render performance.

Changes:
- Added min/max TVL, volume, APY, and API-only creation-date params to the pool-list API client and query keys.
- Added pool-list range filters for TVL, volume, and Base vAPY.
- Applies a hidden default $10k minimum TVL without showing it as an active filter.
- Added URL parsing, serialization, cleanup, and reset handling for pool-list filters.
- Added a pool-list refresh action and reload support for error states.
- Reduced pool-list render work by memoizing heavier cells and fetching campaign rewards once per table.
- Updated Cypress mocks and tests for range filters, default TVL behavior, active chips, sorting, reloads, and resets.


apps/main/src/dex/features/pool-list/cells/PoolListRewards.tsx
```diff
@@ -1,16 +1,20 @@
...
 import { t } from '@ui-kit/lib/i18n'
 import { Tooltip } from '@ui-kit/shared/ui/Tooltip'
 import { formatNumber } from '@ui-kit/utils'
 import type { PoolListItem } from '../poolList.types'
-import { normalizeAddress } from '../poolList.utils'
 
 type CrvRewards = Pick<PoolListItem, 'crvApr' | 'crvAprBoosted'>
+type PoolListRewardsProps = {
+  campaigns: CampaignRewards[]
+  mobile?: boolean
```
DanielSchiavini: We usually prefix with is

apps/main/src/dex/features/pool-list/drawers/PoolListSortDrawer.tsx
```diff
@@ -1,70 +1,19 @@
...
-                {sortField === id && <CheckIcon sx={{ marginLeft: Spacing.sm }} />}
-              </MenuItem>
-            </InvertOnHover>
-          ))}
-        </MenuList>
-      </DrawerItems>
-    </SwipeableDrawer>
-  )
-}
+export const PoolListSortDrawer = ({ onSortingChange, options, sortField }: Props) => (
```
DanielSchiavini: this component is unnecessary, it's too simple and only used once

apps/main/src/dex/features/pool-list/filters/PoolListFilters.tsx
```diff
@@ -0,0 +1,88 @@
...
+  tvlRange,
+  volumeRange,
+}: PoolListFilterProps) => {
+  const onPoolTypeChange = (_: MouseEvent<HTMLElement>, value: PoolTypeFilterValue | null) =>
+    setPoolType(value && value !== ALL_FILTER_VALUE ? value : null)
+  const poolTypeOptions: readonly TableFilterButtonOption<PoolTypeFilterValue>[] = [
+    { value: ALL_FILTER_VALUE, label: t`All` },
+    ...poolTypeFilters.map(({ key, label }) => ({ value: key, label })),
+  ]
+  const numberRangeFilters: readonly NumberRangeFilter[] = [
```
DanielSchiavini: this list is created just to map into components, it would be simpler to simply create the components directly

apps/main/src/dex/features/pool-list/filters/PoolListFilters.tsx
```diff
@@ -0,0 +1,88 @@
...
+          <RangeFilter {...filter} />
+        </TableFilterItem>
+      ))}
+      <TableFilterButtonGroup
+        title={t`Pool type`}
+        value={poolType ?? ALL_FILTER_VALUE}
+        onChange={onPoolTypeChange}
+        ariaLabel={t`Pool type filter`}
+        options={poolTypeOptions}
+        testIdSuffix="pool-filter-type"
```
DanielSchiavini: As much as I don't like using the empty string to mean an empty value, in this case I think it's better than a random const. Also, there is no point in declaring onPoolTypeChange and poolType options to use them once. They can be inlined
```suggestion
        value={poolType ?? ''}
        onChange={(_: MouseEvent<HTMLElement>, value: PoolTypeFilterValue | null) => setPoolType(value || null)}
        ariaLabel={t`Pool type filter`}
        options={[{ value: '', label: t`All` }, ...poolTypeFilters.map(({ key, label }) => ({ value: key, label }))]}
        testIdSuffix="pool-filter-type"
```

apps/main/src/dex/features/pool-list/filters/PoolListFiltersOverlay.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+type PoolListFiltersOverlayProps = {
+  anchorRef: RefObject<HTMLDivElement | null>
+  hasActiveFilters: boolean
+  open: boolean
+  resetFilters: () => void
+  setOpen: (open: boolean) => void
+} & PoolListFilterProps
+
+export const PoolListFiltersOverlay = ({
```
DanielSchiavini: I don't think we need this component, either

packages/curve-ui-kit/src/shared/ui/DataTable/hooks/useDebouncedTableRangeFilter.ts
```diff
@@ -0,0 +1,80 @@
...
+  range: T
+  setRange: (range: T) => void
+  sanitize?: (range: T) => T
+}
+
+/**
+ * Debounced adapter for API-backed table filters that write to URL/API state.
+ * Keep this hook in a stable table owner, outside transient drawers/popovers, so pending edits survive overlay close.
+ */
+export const useDebouncedTableRangeFilter = <T extends TableRangeFilter>({
```
DanielSchiavini: We already have range filters. What's wrong with useRangeFilter?
I think we shouldn't couple debouncing+draft stuff with range, because then you need to write draft/debouncing logic for every single type of filter
They are 3 separate concerns. We have hooks to take care of debouncing (for the whole popover), draft state (for the whole popover) and then one just to handle ranges.

packages/curve-ui-kit/src/shared/ui/DataTable/filters.tsx
```diff
@@ -24,9 +24,9 @@ export const parseRangeFilter = (serialized: string | undefined) =>
 export const getRangeFilterLabel = (
   [min, max]: Range<number | null>,
   unit?: Unit,
-  { defaultMin = 0 }: { defaultMin?: number } = {},
+  { defaultMin = 0 }: { defaultMin?: number | null } = {},
 ) => {
-  const hasMin = min != null && min !== defaultMin
+  const hasMin = min != null && (defaultMin == null || min !== defaultMin)
```
DanielSchiavini: hasMin will be true if defaultMin is null? Why 🤔

packages/curve-ui-kit/src/shared/ui/DataTable/urlFilter.utils.ts
DanielSchiavini, line 1: Why do we need all this new logic? We already have existing filters that handle URL just fine. See useFilters

packages/prices-api/src/pools/index.ts
```diff
@@ -33,6 +33,14 @@ export type ListPoolsParams = {
   pagination?: number
   searchString?: string
   poolType?: Schema.V2PoolFilterType
+  minTvl?: number
```
DanielSchiavini, line 36: are these really optional?

tests/cypress/support/helpers/dex-pool-list-mocks.ts
```diff
@@ -170,7 +176,13 @@ const createMockPools = (chainId: MockChainId): MockPool[] => [
 
 const MOCK_POOLS = MOCK_CHAIN_IDS.flatMap(createMockPools)
 
-const parseNumberParam = (value: string | null, fallback: number) => maybe(value, Number) ?? fallback
+const parseOptionalNumberParam = (value: string | null) => {
+  const parsed = maybe(value, Number)
+
+  return parsed != null && Number.isFinite(parsed) ? parsed : undefined
+}
```
DanielSchiavini: calling maybe and then comparing to null kind of defeats the purpose 😆 
why would we expect infinite/invalid values in our tests? I don't see the use case.
I'm not even sure we need  to build this mock-backend here, we could just assert the backend is called with the expected parameters and that the new backend data is shown

apps/main/src/dex/features/pool-list/poolListFilterQuery.ts
```diff
@@ -0,0 +1,179 @@
...
+
+const serializeRangeUpdate = <T extends string | number>(field: RangeQueryField, value: UrlRange<T>) => ({
+  [field]: serializeRangeFilter(value),
+})
+
+const serializeDefaultNumberRangeUpdate = (
+  field: DefaultNumberRangeQueryField,
+  value: PoolListNumberRange,
+  defaults: RangeFilterDefaults<number>,
+) => serializeRangeUpdate(field, normalizeRangeFilterDefaults(value, defaults))
```
DanielSchiavini: do we really need to normalize ranges?

apps/main/src/dex/features/pool-list/poolListFilterQuery.ts
```diff
@@ -0,0 +1,179 @@
...
+  const tvlQueryValue = searchParams.get(tvl)
+  const volumeQueryValue = searchParams.get(volume)
+  const apyQueryValue = searchParams.get(apy)
+  const parsedPoolType = getPoolTypeFromQuery(poolTypeQueryValue)
+  const tvlRangeState = parseNumberUrlRange(tvlQueryValue, POOL_LIST_DEFAULT_TVL_RANGE)
+  const volumeRangeState = parseNumberUrlRange(volumeQueryValue, POOL_LIST_DEFAULT_NON_NEGATIVE_RANGE)
+  const apyRangeState = parseNumberUrlRange(apyQueryValue, POOL_LIST_DEFAULT_NUMBER_RANGE)
+
+  return {
+    cleanup: {
```
DanielSchiavini: I don't see why we would need a cleanup object

apps/main/src/dex/features/pool-list/hooks/usePoolListFilters.ts
```diff
@@ -31,34 +39,64 @@ export const getPoolTypeFromQuery = (value: string | null): PoolListPoolType | u
...
+  // URL params are untrusted: users can manually edit or share stale links.
+  // Drop invalid filter values here so chips, active counts, and API params all derive from the same valid state.
+  const filterQueryState = useMemo(() => getPoolListFilterQueryState(searchParams), [searchParams])
+  const { cleanup, poolType, ranges, searchText } = filterQueryState
+  const { apy: apyRange, tvl: tvlRange, volume: volumeRange } = ranges
+
+  useEffect(() => {
+    if (!isEmpty(cleanup)) {
+      // Canonicalize manually edited/bookmarked URLs without pushing history or resetting pagination.
+      searchNavigate(cleanup, { replace: true })
```
DanielSchiavini: we should not be changing urls in dex-specific code. We have useFilters hook that already handles URL navigation

apps/main/src/dex/features/pool-list/hooks/usePoolListFilters.ts
```diff
@@ -31,34 +39,64 @@ export const getPoolTypeFromQuery = (value: string | null): PoolListPoolType | u
...
+    (value: PoolListNumberRange) => updateQueryAndResetPage(getPoolListApyRangeUpdate(value)),
+    [updateQueryAndResetPage],
+  )
+  const resetFilters = useCallback(
+    () => updateQueryAndResetPage(getPoolListResetFiltersUpdate()),
+    [updateQueryAndResetPage],
+  )
+  const hasActiveFilters = hasPoolListActiveFilters(filterQueryState)
+  const apiParams: PoolListApiParams = getPoolListApiParams(filterQueryState)
+  const filterProps: PoolListFilterProps = {
```
DanielSchiavini: we should inline these variables if they are only used once

apps/main/src/dex/features/pool-list/hooks/usePoolListFilters.ts
```diff
@@ -31,34 +39,64 @@ export const getPoolTypeFromQuery = (value: string | null): PoolListPoolType | u
...
-    (value: string) => updateQueryAndResetPage({ [SEARCH_QUERY_FIELD]: value || null }),
+    (value: string) => updateQueryAndResetPage(getPoolListSearchUpdate(value)),
     [updateQueryAndResetPage],
   )
   const setPoolType = useCallback(
-    (value: PoolListPoolType | null) => updateQueryAndResetPage({ [POOL_TYPE_QUERY_FIELD]: value }),
+    (value: PoolListPoolType | null) => updateQueryAndResetPage(getPoolListPoolTypeUpdate(value)),
+    [updateQueryAndResetPage],
+  )
+  const setTvlRange = useCallback(
```
DanielSchiavini: the set functions are full of repetition, they are asking for some abstraction

apps/main/src/dex/features/pool-list/filters/PoolListFilters.tsx
```diff
@@ -0,0 +1,88 @@
...
+    },
+  ]
+
+  return (
+    <Stack spacing={Spacing.sm} sx={{ padding: Spacing.sm }}>
+      {numberRangeFilters.map(({ title, ...filter }) => (
+        <TableFilterItem key={filter.id} title={title}>
+          <RangeFilter {...filter} />
+        </TableFilterItem>
+      ))}
```
DanielSchiavini:
```suggestion
    <TableFilterItem title={t`TVL`}>
      <RangeFilter id="tvl" range={tvlRange} setRange={setTvlRange} adornment="dollar" min={0} />
    </TableFilterItem>
    <TableFilterItem title={t`Volume`}>
      <RangeFilter id="volume" range={volumeRange} setRange={setVolumeRange} adornment="dollar" min={0} />
    </TableFilterItem>
    <TableFilterItem title={t`Base vAPY`}>
      <RangeFilter id="apy" range={apyRange} setRange={setApyRange} adornment="percentage" />
    </TableFilterItem>
```

packages/curve-ui-kit/src/shared/ui/DataTable/filters.tsx
OnlyJousting: I made that more explanatory

packages/prices-api/src/pools/index.ts
OnlyJousting: yea they can be omitted when calling the API endpoint

tests/cypress/support/helpers/dex-pool-list-mocks.ts
OnlyJousting: fair enough, I'll simplify

apps/main/src/dex/features/pool-list/PoolListTable.tsx
```diff
@@ -103,22 +67,23 @@ export const PoolListTable = ({ network }: { network: NetworkConfig }) => {
     manualPagination: true,
     manualSorting: true,
     manualFiltering: true,
+    meta: { campaignsByAddress } satisfies PoolListTableMeta,
```
DanielSchiavini: Why would we pass this via the table meta instead of attaching to the applicable pools at `getPoolListItem`?

apps/main/src/dex/features/pool-list/poolListFilterQuery.ts
```diff
@@ -0,0 +1,59 @@
...
+export const POOL_LIST_SEARCH_QUERY_FIELD = 'search'
+
+export enum PoolListFilterId {
+  PoolType = 'filter',
+  Tvl = 'tvl',
+  Volume = 'volume',
+  Apy = 'apy',
+}
+
+export const POOL_LIST_FILTER_CHANGE_UPDATE = { page: null } as const
```
DanielSchiavini: This should be the default behavior for any backend-paginated table tbh

apps/main/src/llamalend/features/market-list/filters/LlamaTableActiveFiltersChip.tsx
```diff
@@ -71,37 +68,20 @@ export const LlamaTableActiveFiltersChip = <T extends TableItem>({
...
+            key: `selected-chip-${id}`,
+            labels,
+            onRemove: isRangeFilterFn ? () => setColumnFilter(id, null) : removeClickedValue,
+            formatLabel: label => formatLabel(label, id),
+            renderChips:
+              id === LlamaMarketColumnId.Chain
+                ? ({ labels, onRemove }) => (
+                    <ChainFilterChips chainsQuery={constQ(labels)} selectedChains={labels} toggleChain={onRemove} />
+                  )
+                : undefined,
```
DanielSchiavini:
```suggestion
            ...(id === LlamaMarketColumnId.Chain && {
              renderChips: ({ labels, onRemove }) => (
                <ChainFilterChips chainsQuery={constQ(labels)} selectedChains={labels} toggleChain={onRemove} />
              ),
            })
```

The prop shouldn't need to be called render though, it can be just a normal component.
```suggestion
            ...(id === LlamaMarketColumnId.Chain && { Chips: ChainFilterLabelChips })
```

packages/curve-ui-kit/src/shared/ui/DataTable/hooks/useFilters.tsx
```diff
@@ -45,21 +47,34 @@ function parseFilters<TColumnId extends string>(
...
+  const getScopedUpdate = useCallback(
+    (filters: PartialRecord<TColumnId, string | null>) =>
+      fromEntries(Object.entries(filters).map(([id, value]) => [scopedKey(scope, id), value])) as SearchParamsUpdate,
+    [scope],
+  )
+  const updateColumnFilters = useCallback(
+    (filters: PartialRecord<TColumnId, string | null>) =>
+      searchNavigate({ ...getScopedUpdate(filters), ...extraParamsOnFilterChange }, { replace: true }),
+    [extraParamsOnFilterChange, getScopedUpdate, searchNavigate],
+  )
```
DanielSchiavini: getScopedUpdate is not necessary, it's only used once
```suggestion
  const updateColumnFilters = useCallback(
    (filters: PartialRecord<TColumnId, string | null>) =>
      searchNavigate(
        {
          ...fromEntries(recordEntries(filters).map(([id, value]) => [scopedKey(scope, id), value])),
          ...extraParamsOnFilterChange,
        },
        { replace: true },
      ),
    [extraParamsOnFilterChange, scope, searchNavigate],
  )
```

packages/curve-ui-kit/src/shared/ui/DataTable/hooks/useFilters.tsx
```diff
@@ -109,11 +128,11 @@ export const useFilters = <TColumnId extends string>({
...
-     * relative to that same stale snapshot, effectively re-adding whatever the first
-     * call had just cleared.
+     * Clearing global and column filters with separate navigations would not work
+     * because each navigation would be backed by a `searchNavigate` closure that
+     * captures a snapshot of `URLSearchParams` at render time. The second call would
+     * navigate relative to that same stale snapshot, effectively re-adding whatever
+     * the first call had just cleared.
      *
      * To avoid this, `resetFilters` owns its own `searchNavigate` and merges all keys
      * to clear into one update, so only a single navigation occurs.
```
DanielSchiavini: Too much text + line length is messed up
```suggestion
     * Clears all filters (both global search and column filters) in a single navigation. Multiple navigations
     * do not work because each navigation captures its own `searchNavigate` closure at render time.
     * To avoid this, `resetFilters` merges all keys into one update, so only a single navigation occurs.
```

packages/curve-ui-kit/src/shared/ui/DataTable/hooks/useFilters.tsx
```diff
@@ -93,11 +108,15 @@ function useGlobalFilter(key = DEFAULT_SEARCH_KEY) {
...
  */
 export const useFilters = <TColumnId extends string>({
+  extraParamsOnFilterChange,
   searchKey = DEFAULT_SEARCH_KEY,
   ...columnFilterOptions
-}: Parameters<typeof useColumnFilters<TColumnId>>[0] & { searchKey?: string }) => {
-  const globalFilter = useGlobalFilter(searchKey)
-  const columnFilters = useColumnFilters(columnFilterOptions)
+}: Parameters<typeof useColumnFilters<TColumnId>>[0] & {
+  extraParamsOnFilterChange?: SearchParamsUpdate
```
DanielSchiavini: We should try to keep consistency - only offer flexibility when really necessary
```suggestion
  resetPageOnChange?: boolean
```

packages/curve-ui-kit/src/shared/ui/DataTable/hooks/useRangeFilter.ts
```diff
@@ -9,22 +9,66 @@ import {
 import { Range } from '@ui-kit/types/util'
 
 export const useRangeFilter = <TColumnId extends string>({
```
DanielSchiavini, line 11: Why is this rewritten? It was working fine until now.

packages/curve-ui-kit/src/shared/ui/DataTable/hooks/useRangeFilter.ts
```diff
@@ -9,22 +9,66 @@ import {
...
+  debounceMs,
```
DanielSchiavini: the debounce should be handled at the form level?

packages/curve-ui-kit/src/shared/ui/DataTable/filters.tsx
```diff
@@ -24,14 +24,15 @@ export const parseRangeFilter = (serialized: string | undefined) =>
 export const getRangeFilterLabel = (
   [min, max]: Range<number | null>,
   unit?: Unit,
-  { defaultMin = 0 }: { defaultMin?: number } = {},
+  { defaultMin = 0 }: { defaultMin?: number | null } = {},
 ) => {
-  const hasMin = min != null && min !== defaultMin
+  // APY can be negative, so it has no default lower bound; a selected `0` should still show as `>0%`.
```
DanielSchiavini: This should be documented at the parameter level, it's simply an optional parameter
```suggestion
```

packages/curve-ui-kit/src/shared/ui/DataTable/TableActiveFilterGroups.utils.tsx
```diff
@@ -0,0 +1,59 @@
...
+  key: string
+  labels: readonly string[] | null | undefined
+  onRemove: TableActiveFilterGroupRemove
+  renderChips?: (props: { labels: string[]; onRemove: TableActiveFilterGroupRemove }) => ReactNode
+  testId?: string
+  title: ReactNode
+}
+
+/** Creates an active-filter group with standard chip overflow and removal behavior. */
+export const createTableActiveFilterGroup = ({
```
DanielSchiavini: This function is completely unnecessary, it seems like a large workaround to avoid creating proper components.
The whole `<>...</>` part can just be the default component. renderChips is just a component.
All we need to do is change `TableActiveFilterGroup` from `chips: ReactNode` to `Chips: FunctionComponent<{ labels: string[]; onRemove: TableActiveFilterGroupRemove }>` and pass the two extra props with it.

tests/cypress/e2e/dex/dex-markets.cy.ts
```diff
@@ -89,46 +140,169 @@ describe('DEX Pools', () => {
...
+
+    it('resets page when search changes', () => {
+      expectPageResetAfter(() => {
+        cy.get('[data-testid="table-text-search-dex-pool-list"] input').type(DEX_POOL_LIST_SEARCH)
+      })
+      cy.location('search').should('include', `search=${DEX_POOL_LIST_SEARCH}`)
+    })
+
+    it('resets page when a filter changes', () => {
+      const poolType = 'crypto'
```
DanielSchiavini: ideally all these tests should be picking a random valid value from a range
this works and it's still a good test, but at least we should create constants

tests/cypress/e2e/dex/dex-markets.cy.ts
```diff
@@ -89,46 +140,169 @@ describe('DEX Pools', () => {
...
+
+      expectPageResetAfter(() => {
+        clickFilterChip(poolType)
+      }, false)
```
DanielSchiavini: boolean parameters aren't readable when passed like this. The `{}` are not necessary.
```suggestion
      expectPageResetAfter(() => clickFilterChip(poolType), { waitForRequest: false })
```

apps/main/src/dex/features/pool-list/PoolListTable.tsx
DanielSchiavini: see https://github.com/curvefi/curve-frontend/blob/4792bdcec4c097e12bba912a3ed18500079d7477/apps/main/src/llamalend/queries/market-list/llama-markets.ts#L463

apps/main/src/dex/features/pool-list/poolListFilterQuery.ts
OnlyJousting: yea it will be needed for all backend paginated tables

apps/main/src/dex/features/pool-list/PoolListTable.tsx
OnlyJousting: I was trying to optimize but the llama-markets example is better so will update

apps/main/src/dex/features/pool-list/poolListFilterQuery.ts
```diff
@@ -0,0 +1,57 @@
...
+  const [minVolume, maxVolume] = parsePoolListRangeFilter(columnFiltersById[PoolListFilterId.Volume])
+
+  return {
+    maxApy: maxApy ?? undefined,
+    maxTvl: maxTvl ?? undefined,
+    maxVolume: maxVolume ?? undefined,
+    minApy: minApy ?? undefined,
+    minTvl: minTvl ?? POOL_LIST_DEFAULT_TVL_MIN,
+    minVolume: minVolume ?? undefined,
+    poolType: columnFiltersById[PoolListFilterId.PoolType] || undefined,
```
0xPearce, line 41: we should probably make a migration and normalize the pool type in order to prevent any type of filter to be fetched `/dex/ethereum/pools?filter=any`:

Image

OnlyJousting: I had sanitization for all inputs that could come in ugly via the URL but dani was against it

packages/curve-ui-kit/src/shared/ui/DataTable/hooks/useRangeFilter.ts
OnlyJousting: it just needed a different min to display vs what was passed as default when resetting, I went a bit overboard but honed it back in

tests/cypress/e2e/dex/dex-markets.cy.ts
OnlyJousting: adding constants

apps/main/src/dex/features/pool-list/poolListFilterQuery.ts
DanielSchiavini: we can use query validation

DanielSchiavini: but we shouldn't change the user given values

OnlyJousting: I think it's nice to remove invalid inputs from the URL but am fine either way

apps/main/src/dex/features/pool-list/poolListFilterQuery.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+}
+
+export const hasPoolListActiveFilters = (columnFiltersById: PoolListColumnFilters) =>
```
DanielSchiavini: this should not be necessary, we should just do `equals(filters, defaultFilters)` (I believe we never save default values)

packages/curve-ui-kit/src/shared/ui/DataTable/hooks/useRangeFilter.ts
DanielSchiavini: Why would we want a different min to display? It's not clear from the code. It's confusing that we would then pass that 'display' value to `defaultValue` in the debounce.

tests/cypress/support/helpers/dex-pool-list-mocks.ts
DanielSchiavini, line 1: nice to have quite a few new test cases

packages/curve-ui-kit/src/shared/ui/DataTable/hooks/useRangeFilter.ts
DanielSchiavini: See also [this comment](https://github.com/curvefi/curve-frontend/pull/2793#discussion_r3519094039) - I don't understand why we need custom debounce implementation just for ranges - they should work the same for any kind of filter

apps/main/src/dex/features/pool-list/poolListFilterQuery.ts
0xAlunara: There's a big chance this filter goes away in favor of proper token filtering, so we can prob ignore for now

apps/main/src/dex/features/pool-list/poolListFilterQuery.ts
0xAlunara: Removed this in favor of `const hasActiveFilters = !!table.getState().columnFilters.length` and setting the `columnFilters` and `globalFilter` inside `state: { expanded, sorting, pagination, columnVisibility, columnFilters, globalFilter }` the same way we do for llama markets.

apps/main/src/dex/features/pool-list/poolList.utils.ts
```diff
@@ -2,14 +2,15 @@ import type { CurveApi, NetworkConfig, PoolData } from '@/dex/types/main.types'
...
 import { recordValues } from '@primitives/objects.utils'
+import type { CampaignRewards } from '@ui-kit/entities/campaigns'
 import { DEX_ROUTES } from '@ui-kit/shared/routes'
 import { AVERAGE_CATEGORIES, aprToApy } from '@ui-kit/utils'
 import { isVyperVulnerablePool } from './poolList.alerts'
-import type { PoolListCampaignsByAddress, PoolListItem } from './poolList.types'
+import type { PoolListItem } from './poolList.types'
 
 const POOL_YIELD_COMPOUND_WINDOW = AVERAGE_CATEGORIES['dex.poolYield.compoundRate'].window
-const EMPTY_CAMPAIGN_REWARDS: PoolListCampaignsByAddress[string] = []
```
DanielSchiavini, line 11: I guess this was here to avoid extra renders

apps/main/src/dex/hooks/useTokenAlert.tsx
```diff
@@ -228,9 +197,7 @@ const TOKEN_ALERTS: Record<string, PoolAlert> = {
       <div>
         <Trans>
           The StablR has suffered an exploit.{' '}
-          <StyledExternalLink href="https://x.com/StablREuro/status/2058520949075386683?s=20">
-            Learn more here
-          </StyledExternalLink>
+          <InlineLink to="https://x.com/StablREuro/status/2058520949075386683?s=20">Learn more here</InlineLink>
```
DanielSchiavini, line 200: Shouldn't we move this to EXTERNAL_LINKS?

DanielSchiavini (approved):

0xAlunara (approved):

PR #2794, opened 2026-06-26
refactor: extract reusable table filter controls

OnlyJousting:
Extract the generic parts of the LlamaLend market table filters into shared DataTable components so the same filter drawer, chips, range inputs, active filter groups, hidden selected items, and mobile sort UI can be reused by the pool-list table.

Changes:
- Moved the filter overlay drawer/popover into shared DataTable UI
- Moved active/selected filter chip rendering into shared DataTable UI
- Moved hidden inline selected-item rendering into shared DataTable UI
- Moved the button group filter control into shared DataTable UI
- Moved the range filter input and range filter hook into shared DataTable UI
- Moved the mobile sort drawer into shared DataTable UI
- Updated LlamaLend markets to use the shared components while keeping existing behavior

### DIFF
--- a/packages/curve-ui-kit/src/shared/ui/DataTable/HiddenInlinedItems.utils.ts
+++ b/packages/curve-ui-kit/src/shared/ui/DataTable/HiddenInlinedItems.utils.ts
@@ -5,7 +5,7 @@ const MAX_VISIBLE_INLINED_ITEMS = 4
 
 /**
  * Split an array between visible and hidden items to maintain inline compaction.
- * Any remaining items are collapsed into a single "+N" item.
+ * Any remaining items are collapsed into a single "+N" item, , if more than 1 item is hidden
  */
 export const getInlinedItemsVisibility = (
   selectedItems: string[] | undefined,
--- a/packages/curve-ui-kit/src/shared/ui/DataTable/HiddenInlinedItems.utils.ts
+++ b/packages/curve-ui-kit/src/shared/ui/DataTable/HiddenInlinedItems.utils.ts
@@ -5,7 +5,7 @@ const MAX_VISIBLE_INLINED_ITEMS = 4
 
 /**
  * Split an array between visible and hidden items to maintain inline compaction.
- * Any remaining items are collapsed into a single "+N" item, if more than 1 item is hidden
+ * Any remaining items are collapsed into a single "+N" item.
  */
 export const getInlinedItemsVisibility = (
   selectedItems: string[] | undefined,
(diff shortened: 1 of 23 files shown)


packages/curve-ui-kit/src/shared/ui/DataTable/TableActiveFiltersBar.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+type TableActiveFiltersBarProps = {
+  children: ReactNode
+  endSlot?: ReactNode
+  hasActiveFilters: boolean
+  resetFilters: () => void
+  testId: string
+}
+
+/** Collapsible active-filter row for DataTable screens. Extracted from the LlamaLend markets list. */
```
0xPearce: There is no need to add where it comes from IMO. Also wdym by "screens"
```suggestion
/** Collapsible active-filter row for DataTable. */
```

packages/curve-ui-kit/src/shared/ui/DataTable/RangeFilter.tsx
```diff
@@ -3,35 +3,26 @@ import { useCallback } from 'react'
...
-type RangeFilterProps<TColumnId extends string> = FilterProps<TColumnId> & {
-  id: TColumnId
+type InputIndex = 0 | 1
+
+type RangeFilterProps = {
+  id: string
+  range: Range<number | null>
+  setRange: (range: Range<number | null>) => void
   adornment?: NumericTextFieldProps['adornment']
-  isLoading?: boolean
```
0xPearce, line 18: We shouldn't get rid of the `isLoading` since we are passing the loading state, it was used temporary to disable the range while loading. But anyway I have a task to properly display the loading to all filters components

packages/curve-ui-kit/src/shared/ui/DataTable/HiddenInlinedItems.utils.ts
```diff
@@ -5,7 +5,7 @@ const MAX_VISIBLE_INLINED_ITEMS = 4
 
 /**
  * Split an array between visible and hidden items to maintain inline compaction.
- * Any remaining items are collapsed into a single "+N" item, if more than 1 item is hidden
```
0xPearce, line 8: Why remove it? To illustrate this, here are 4 items selected:

Image

And here 5 items selected:

Image

As you can see we never display the "+1" chip.

0xPearce (dismissed):
Nice!

DanielSchiavini (dismissed):

packages/curve-ui-kit/src/shared/ui/DataTable/RangeFilter.tsx
OnlyJousting: nice, I'll restore it as well as the comments that were changed

DanielSchiavini (approved):

packages/curve-ui-kit/src/shared/ui/DataTable/HiddenInlinedItems.utils.ts
```diff
@@ -5,7 +5,7 @@ const MAX_VISIBLE_INLINED_ITEMS = 4
...
+ * Any remaining items are collapsed into a single "+N" item, , if more than 1 item is hidden
```
0xAlunara, line 8: `, ,`?

0xAlunara (approved):

PR #2799, opened 2026-06-29
refactor: `maybes` arguments

DanielSchiavini:
- slightly simplify the `maybes` helper so we don't need to use a tuple for the 2nd argument
- this makes the helper more consistent with e.g. `combineQueries`


0xPearce (approved):
Nice, wanted to implement it as well

0xtutti (approved):

PR #2802, opened 2026-06-29
chore: update llamalend.js to 2.3

DanielSchiavini:
- update llamalend.js to 2.3.1
  - https://github.com/curvefi/curve-llamalend.js/pull/117
  - https://github.com/curvefi/curve-llamalend.js/pull/118
  - https://github.com/curvefi/curve-llamalend.js/pull/119
  - https://github.com/curvefi/curve-llamalend.js/pull/120
- remove userBorrowed field from borrow more zapv2
- check userBorrowed isn't passed to any zapv2 functions
- re-enable zapv2 only for users that have changed local storage
- remove legacy route details component


apps/main/src/llamalend/queries/repay/repay-prices.query.ts
```diff
@@ -50,11 +51,11 @@ export const { useQuery: useRepayPrices, invalidate: invalidateRepayPrices } = q
     // it looks like all implementations have the same signature, but `args` is typed differently for each
     switch (type) {
       case 'zapV2':
+        assert(!+userBorrowed, `Unsupported userBorrowed for zapv2: ${userBorrowed}`)
```
0xAlunara: This assertion is all over the place with no explanation for why it's there. Is it possible to move this inside `getRepayImplementation` with an explanation? That function already receives `userBorrowed` anyway.

EDIT: okay just noticed it's not passed as argument for all cases, like in `getCreateLoanImplementation`, but I'd still prefer to assert at that level. Basically when there's a `userBorrowed` there's no valid implementation available for `zapV2`.

DanielSchiavini: Yeah makes sense, I centralized the implementation choices for a reason after all

apps/main/src/llamalend/features/borrow/components/AdvancedCreateLoanOptions.tsx
```diff
@@ -46,10 +47,10 @@ export const AdvancedCreateLoanOptions = ({
           loading={routeImageLoading || expectedCollateralLoading}
           swapFrom={borrowToken}
           swapTo={collateralToken}
-          swapFromAmounts={[
-            { value: `${debt ?? 0}`, label: t`Debt` },
-            { value: `${userBorrowed ?? 0}`, label: t`Wallet` },
-          ]}
+          swapFromAmounts={notFalsy(
+            debt && { value: debt, label: t`Debt` },
```
0xPearce: this crashes if both aren't available, I think the component expects a tuple

Image

0xPearce (changes requested):
When I try to repay using the borrowed token , I get this error: `Borrow amount is not supported for unleveragedLend`

0xAlunara (dismissed):

0xAlunara (dismissed):

0xAlunara (approved):

apps/main/src/llamalend/widgets/RouteDetails/index.tsx
```diff
@@ -1,114 +0,0 @@
-import type { ComponentPropsWithRef } from 'react'
```
0xPearce, line 1: 👋

0xPearce (approved):

PR #2804, opened 2026-06-29
feat: price impact calculation

DanielSchiavini:
- implement generic price impact calculation that works the same for all routers
  - we request a small amount from the same router to compare
- price impact is always displayed in the action info, but it only blocks transactions >$1k
- refactor routes backend to distinguish between user and zap addresses
- depends on #2802


apps/main/src/llamalend/hooks/usePriceImpact.ts
```diff
@@ -0,0 +1,92 @@
...
+}) => {
+  const amountIn = fromWei(selectedAmountIn, tokenInDecimals)
+  const amountOut = fromWei(selectedAmountOut, tokenOutDecimals)
+  const referenceOut = fromWei(referenceAmountOut, tokenOutDecimals)
+  const referenceIn = fromWei(referenceAmountIn, tokenInDecimals)
+  const selectedRate = decimalDiv(amountOut, amountIn)
+  const referenceRate = decimalDiv(referenceOut, referenceIn)
+  const ratio = decimalDiv(selectedRate, referenceRate)
+  const priceImpact = decimalMax('0', decimalMultiply(decimalMinus('1', ratio), '100'))
+  return decimal(priceImpact)
```
0xPearce: is `decimal` needed here if it is returned by `decimalMax`?

apps/main/src/llamalend/features/borrow/hooks/useCreateLoanForm.tsx
```diff
@@ -146,7 +146,7 @@ export function useCreateLoanForm<ChainId extends LlamaChainId>({
 
   useCallbackSync(useCreateLoanPrices(params), onPricesUpdated)
 
-  const priceImpact = q(useCreateLoanPriceImpact(params, values.leverageEnabled))
+  const priceImpact = q(useCreateLoanPriceImpact(params, !zapAddress))
```
0xPearce: I think this change is disabling the button below, for market that have a `zapAddress`

0xPearce: It may be the same for borrow more and repay

DanielSchiavini: yeah sorry I'm not sure why this came back, I'm sure I moved the check to the form tsx

apps/main/src/llamalend/hooks/usePriceImpact.ts
DanielSchiavini: no, you are right

0xPearce (dismissed):

0xAlunara (approved):

OnlyJousting (approved):

0xPearce (approved):

PR #2805, opened 2026-06-30
feat: add Metric orientation and categories 

0xPearce:
next: https://github.com/curvefi/curve-frontend/pull/2810

This PR adds the horizontal orientation to the Metric component from Figma and introduces categories as well. Categories are needed because, for the same metric, we can have different orientations (horizontal/vertical) or different sizes across breakpoints.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
Screenshot 2026-06-30 at 16 21 03
</td>
    <td>
Screenshot 2026-06-30 at 16 20 40
</td>
  </tr>
</table>

Screenshot 2026-06-30 at 16 19 55


packages/curve-ui-kit/src/hooks/useBreakpoints.ts
```diff
@@ -8,3 +8,11 @@ export const useIsTiny = () => useMediaQuery('(max-width:400px)')
 export const useIsMobile = () => useMediaQuery(isTabletDown)
 export const useIsDesktop = () => useMediaQuery(isDesktopUp)
 export const useIsTablet = () => ![useIsDesktop(), useIsMobile()].includes(true)
+
+// Returns the current app breakpoint bucket for responsive category lookups.
+export const useBreakpoint = (): Breakpoint => {
```
0xPearce, line 13: Does this hook already exists?

packages/curve-ui-kit/src/shared/ui/stories/Metric.stories.tsx
```diff
@@ -63,10 +75,18 @@ const meta: Meta<typeof Metric> = {
     label: 'Metrics label',
     copyText: 'Copied metric value',
   },
+  render: args => (
+    <Stack sx={{ alignItems: 'center', gap: Spacing.lg, width: '440px' }}>
```
0xPearce, line 79: Needed to display the Vertical and Horizontal orientation side by side for each Story

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -99,9 +99,16 @@ const notionalsToString = (notionals: MetricProps['notional']) =>
 /** At the moment of writing the default formatter already formats to 2 decimals, but I really want to make this explicit for potential future changes. */
 const formatChange = (value: number): string => defaultNumberFormatter(value, { decimals: 2 })
 
+/**
+ * MUI Typography resolves the `color` prop through registered theme color names.
+ * Hex values need to be applied as CSS to avoid being treated as unresolved theme colors.
+ */
+const getTypographyColorProps = (color: TypographyProps['color']) =>
+  typeof color === 'string' && color.startsWith('#') ? { sx: { color } } : { color }
```
0xPearce, line 140: Open to suggestions. The problem I want to solve is being able to pass a hex string as a color, as well as a standard color variant like `textPrimary`, etc.

packages/curve-ui-kit/src/shared/ui/metric-categories.ts
```diff
@@ -0,0 +1,94 @@
...
+    mobile: { size: 'small', orientation: 'horizontal' },
+    tablet: { size: 'medium', orientation: 'vertical' },
+    desktop: { size: 'medium', orientation: 'vertical' },
+  },
+  mobileStandardRow: {
+    mobile: { size: 'medium', orientation: 'horizontal' },
+    tablet: { size: 'medium', orientation: 'vertical' },
+    desktop: { size: 'medium', orientation: 'vertical' },
+  },
+  compactMobileRow: {
```
0xAlunara: I'm okay with categories, but one thing I dislike is that `compactMobileRow` for example also sets sizes and orientations for both `tablet` and `desktop`. It's not a deal breaker, but it feels we're coupling too tightly to different breakpoints. What if we want a compact mobile row, except on desktop it should be horizontal too? Would that be `compactMobileRowAndDesktopHorizontal`?

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -184,13 +202,13 @@ export type MetricProps = {
   /** Notional values give extra context to the metric, like underlying value */
   notional?: number | string | Notional | Notional[]
 
-  /** Optional content to display to the right of the value */
-  rightAdornment?: ReactNode
+  /** Optional icon shown after the value in vertical orientation and before the label in the horizontal orientation. */
+  leadingIcon?: ReactNode
```
0xAlunara: Calling it `leadingIcon` feels a bit off when it's still at the end for vertical orientation. Maybe `adornment` or `icon` suffices? There's only one anyway, and no symmetric counterparty.

packages/curve-ui-kit/src/hooks/useBreakpoints.ts
0xAlunara: Don't think so, the three breakpoints are our own construct and not a MUI default. We got away with needing a hook like this because of `handleBreakpoints` in `basic-theme.ts`. This way we got to leverage the different breakpoint behavior by relying on native CSS through `sx`, but in your case you've implemented the orientation styling completely through javascript rather than CSS.

You could try the same approach by adding adding a [custom CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascading_variables/Using_custom_properties) called `--orientation` on the root `Stack` of the `Metric` component, depending on the breakpoint, and adjust the layout using [container style queries ](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_size_and_style_queries), similarly to the example they give wrt device orientation but with our custom `--orientation` prop:
```css
@property --orientation {
  syntax: "horizontal | vertical";
  inherits: true;
  initial-value: horizontal;
}

.metric {
  container-name: metric;
  --orientation: horizontal; /** or vertical depending on the breakpoint, use a media query for this or javascript */
}

@container metric style(--orientation: horizontal) {
  .some-div {
    flex-direction: row;
  }
}

@container metric style(--orientation: vertical) {
  .some-div {
    flex-direction: column;
  }
}
```

But that would require a total revamp on all your effort so far, probably not worth it at this stage.

packages/curve-ui-kit/src/shared/ui/Metric.tsx
0xAlunara: I don't see us using hex strings anywhere though, not in prod and not in this PR. Using hex colors is a code smell anyway and we could reduce complexity by just enforcing the use of the standard color variants like `textPrimary`.

0xAlunara (commented):
When looking at the advanced details of a market page, you can see a slight discrepancy in spacing between the metric values above the TVL header and all the action infos below. Naturally there's a difference between the two components, but either should make the metric look more like an action info in mobile, or perhaps change the action infos to metrics as well (and use the horizontal orientation for all breakpoints?). I think adjusting the spacing is easier than repurposing all abused action infos as metrics.

Image

Not a biggie (I don't know if we're using this or not), but the leading icon might have additional padding in the icon itself, making it appear not aligned with the label in horizontal orientation.

Image

Might want to use horizontal orientation for savings crvusd metrics on mobile as well (test position for `0xE13aDC278c252e04DCBdca8eDced2C973Db994fA`)

Image

apps/main/src/dao/components/PageAnalytics/CrvStats/index.tsx
```diff
@@ -19,6 +19,7 @@ const WEEKS_PER_YEAR = 52
 const VECRV_APR_AVERAGE_WEEKS = 4
 const VECRV_APR_FEE_WEEKS = VECRV_APR_AVERAGE_WEEKS + 1
 const PROPOSAL_MIN_VECRV = 2500
+const METRIC_CATEGORY = 'dao.crvStats'
```
DanielSchiavini, line 22: this is just `string` how is typescript happy?

packages/curve-ui-kit/src/hooks/useBreakpoints.ts
```diff
@@ -8,3 +8,11 @@ export const useIsTiny = () => useMediaQuery('(max-width:400px)')
 export const useIsMobile = () => useMediaQuery(isTabletDown)
 export const useIsDesktop = () => useMediaQuery(isDesktopUp)
 export const useIsTablet = () => ![useIsDesktop(), useIsMobile()].includes(true)
+
+// Returns the current app breakpoint bucket for responsive category lookups.
```
DanielSchiavini: jsdoc pls

packages/curve-ui-kit/src/shared/ui/metric-categories.ts
```diff
@@ -0,0 +1,94 @@
...
+  },
+  mobileCompactRow: {
+    mobile: { size: 'small', orientation: 'horizontal' },
+    tablet: { size: 'medium', orientation: 'vertical' },
+    desktop: { size: 'medium', orientation: 'vertical' },
+  },
+  mobileStandardRow: {
+    mobile: { size: 'medium', orientation: 'horizontal' },
+    tablet: { size: 'medium', orientation: 'vertical' },
+    desktop: { size: 'medium', orientation: 'vertical' },
```
DanielSchiavini: We might need one more level of categorization to fix these duplicated objects 🤓

packages/curve-ui-kit/src/shared/ui/metric-categories.ts
DanielSchiavini: I agree, we should try to keep all names business/design-related and not tight to the implementation.
These names are not clear and don't tell me which one to choose when implementing a new feature

packages/curve-ui-kit/src/shared/ui/Metric.tsx
```diff
@@ -225,13 +246,46 @@ export const Metric = ({
...
+    >
+      <WithWrapper
+        shouldWrap={isHorizontal}
+        Wrapper={Stack}
+        direction="row"
+        sx={{ alignItems: 'baseline', flexShrink: 0 }}
+      >
+        {isHorizontal && leadingIcon}
+        <Typography
+          variant={isHorizontal ? 'bodyMRegular' : 'bodyXsRegular'}
```
DanielSchiavini: looks like we should be using a Record<Orientation, Variant>

packages/curve-ui-kit/src/hooks/useBreakpoints.ts
DanielSchiavini: I agree it would be great to fix this with fixed css, the good thing is that it'll be easy to change after this PR lands! A simple helper should be able to convert the category config into all the necessary sx props

packages/curve-ui-kit/src/shared/ui/Metric.tsx
0xPearce: copied form Figma, but yeah icon makes more sense

packages/curve-ui-kit/src/hooks/useBreakpoints.ts
0xPearce: Thanks for the detailed feedback Alu, will make a small task to fix it

packages/curve-ui-kit/src/shared/ui/Metric.tsx
0xPearce: It's used [here](https://github.com/curvefi/curve-frontend/pull/2792/changes#diff-38432f6cc01816d478f5c269782e6aa613f0e08ff9a761f25dbc8e0751864794R48-R49) in the #2792 new health PR, to display the metric with the same color as the bar.

apps/main/src/dao/components/PageAnalytics/CrvStats/index.tsx
0xPearce: it's not a string, it's a 'dao.crvStats'

0xPearce: image
'dao.crvStats'

DanielSchiavini: whaat I assumed you'd need `as const` there

packages/curve-ui-kit/src/shared/ui/metric-categories.ts
```diff
@@ -7,88 +7,94 @@ export type MetricLayout = {
...
+    desktop: 'vertical',
   },
-} as const satisfies Record<string, MetricViewportLayout>
+  inline: breakpointValues('horizontal'),
+} as const satisfies Record<string, Record<Breakpoint, MetricLayout['orientation']>>
+
+const metricLayout = (
+  emphasis: keyof typeof METRIC_EMPHASIS,
+  presentation: keyof typeof METRIC_PRESENTATION,
+): MetricViewportLayout => ({
```
DanielSchiavini: mapRecord pls

packages/curve-ui-kit/src/shared/ui/metric-categories.ts
```diff
@@ -7,88 +7,94 @@ export type MetricLayout = {
...
+  detail: {
+    mobile: 'horizontal',
+    tablet: 'vertical',
+    desktop: 'vertical',
   },
-} as const satisfies Record<string, MetricViewportLayout>
+  inline: breakpointValues('horizontal'),
+} as const satisfies Record<string, Record<Breakpoint, MetricLayout['orientation']>>
+
+const metricLayout = (
```
DanielSchiavini, line 49: METRIC_TYPES should call metricLayout imo, not be deleted

0xPearce:
@0xAlunara 

> When looking at the advanced details of a market page, you can see a slight discrepancy in spacing between the metric values above the TVL header and all the action infos below. Naturally there's a difference between the two components, but either should make the metric look more like an action info in mobile, or perhaps change the action infos to metrics as well (and use the horizontal orientation for all breakpoints?). I think adjusting the spacing is easier than repurposing all abused action infos as metrics.

Thanks for spotting that, I added a mapped `minHeight` (like it is in Figma). Now both components have the same height

> Not a biggie (I don't know if we're using this or not), but the leading icon might have additional padding in the icon itself, making it appear not aligned with the label in horizontal orientation.

Exactly, the icon is aligned with the baseline, as per the design. We can deal with the icon padding another time 

> Might want to use horizontal orientation for savings crvusd metrics on mobile as well (test position for 0xE13aDC278c252e04DCBdca8eDced2C973Db994fA)

Fixed 

Screenshot 2026-07-01 at 13 32 22

0xAlunara (approved):

DanielSchiavini (approved):

PR #2806, opened 2026-06-30
fix: hide bands chart when disconnected

DanielSchiavini:
- hide the bands chart when the wallet is disconnected, as it currently keeps loading forever
- replace chip toggle with a proper ToggleButton, fixing existing TODO

### DIFF
--- a/packages/curve-ui-kit/src/shared/ui/Chart/ToggleBandsChartButton.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/Chart/ToggleBandsChartButton.tsx
@@ -1,38 +1,31 @@
-import type { ReactNode } from 'react'
+import { ReactNode } from 'react'
 import Stack from '@mui/material/Stack'
+import ToggleButton from '@mui/material/ToggleButton'
 import { EyeClosed } from '@ui-kit/shared/icons/EyeClosed'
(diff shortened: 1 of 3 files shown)


OnlyJousting (approved):

packages/curve-ui-kit/src/shared/ui/Chart/ToggleBandsChartButton.tsx
```diff
@@ -1,38 +1,31 @@
...
+  isVisible,
+  toggle,
+  tooltip,
+}: {
+  label: string
+  isVisible: boolean
+  toggle: () => void
+  tooltip?: ReactNode
+}) => (
+  <ToggleButton size="extraSmall" onClick={toggle} value={true}>
```
0xAlunara, line 23: It's a bit weird to have a toggle button always set to `true`, as it now also looks toggled on when it's actually turned off. 

Image

Image

0xAlunara (approved):

packages/curve-ui-kit/src/shared/ui/Chart/ToggleBandsChartButton.tsx
DanielSchiavini: - it actually looks deactivated, because we are missing a ToggleButtonGroup. See https://github.com/curvefi/curve-frontend/pull/2809

PR #2807, opened 2026-06-30
refactor: ActionInfo query support

DanielSchiavini:
- add support to query props in action info
- do not update usages yet, that can be done in separate PRs


packages/curve-ui-kit/src/shared/ui/ActionInfo/ActionInfo.tsx
```diff
@@ -122,25 +140,32 @@ const ValueDecorator = (props: ValueDecoratorProps) => (
...
-  size = DEFAULT_SIZE,
-  copyValue,
-  copiedTitle,
-  loading = false,
-  error = false,
-  testId = 'action-info',
-  sx,
-}: ActionInfoProps) => {
+const isQuery = (value: ActionInfoProps['value']): value is QueryProp<ReactNode> =>
+  value != null && typeof value === 'object'
```
0xAlunara: I don't think this would guarantee it being a `QueryProp`, as it would pass even when `value` is a non query prop object. Checking if `value.data` exists would be stricter.

0xAlunara (dismissed):

0xPearce (approved):

0xAlunara (approved):

PR #2811, opened 2026-06-30
refactor: form submission void

0xAlunara:
The return type of form submissions is `() => Promise<void> | void`, but we want to treat all submissions as if they're just `void` and with an optional `events` param. This PR gets rid of all silenced lint errors and makes sure we standardize and `void` all submission calls inside `Form.tsx`.

### DIFF
--- a/apps/main/src/dao/components/PageGauges/CreateVote/CreateVoteModal.tsx
+++ b/apps/main/src/dao/components/PageGauges/CreateVote/CreateVoteModal.tsx
@@ -52,7 +52,6 @@ export const CreateVoteModal = ({ isOpen, onClose }: CreateVoteModalProps) => {
       compact
       open={isOpen}
       onClose={onClose}
-      // eslint-disable-next-line @typescript-eslint/no-misused-promises -- Existing violation before enabling this rule.
       formProps={{ onSubmit }}
       footer={
         <Button
@@ -68,8 +67,7 @@ export const CreateVoteModal = ({ isOpen, onClose }: CreateVoteModalProps) => {
       }
     >
       <FormProvider {...form}>
-        {/* eslint-disable-next-line @typescript-eslint/no-misused-promises -- Existing violation before enabling this rule. */}
-        <Stack component="form" spacing={Spacing.lg} onSubmit={onSubmit}>
+        <Stack spacing={Spacing.lg}>
           <Stack spacing={Spacing.xs}>
             <Typography variant="bodySBold">{t`Requirements:`}</Typography>
 
--- a/apps/main/src/dex/features/add-gauge-reward-token/ui/AddRewardToken.tsx
+++ b/apps/main/src/dex/features/add-gauge-reward-token/ui/AddRewardToken.tsx
@@ -74,8 +74,7 @@ export const AddRewardToken = ({ chainId, poolId }: AddRewardTokenProps) => {
 
   return (
     <FormProvider {...form}>
-      {/* eslint-disable-next-line @typescript-eslint/no-misused-promises -- Existing violation before enabling this rule. */}
-      <form onSubmit={handleSubmit(onSubmit)}>
+      <form onSubmit={event => void handleSubmit(onSubmit)(event)}>
         <FormContainer>
           <FormFieldsContainer>
             <FlexContainer>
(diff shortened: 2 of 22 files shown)


apps/main/src/dao/components/PageGauges/CreateVote/CreateVoteModal.tsx
```diff
@@ -68,8 +67,7 @@ export const CreateVoteModal = ({ isOpen, onClose }: CreateVoteModalProps) => {
       }
     >
       <FormProvider {...form}>
-        {/* eslint-disable-next-line @typescript-eslint/no-misused-promises -- Existing violation before enabling this rule. */}
-        <Stack component="form" spacing={Spacing.lg} onSubmit={onSubmit}>
+        <Stack spacing={Spacing.lg}>
```
0xAlunara, line 70: `ModalDialog` with `formProps` already makes sure the root component is a `<form>`

apps/main/src/dex/features/add-gauge-reward-token/ui/AddRewardToken.tsx
```diff
@@ -74,8 +74,7 @@ export const AddRewardToken = ({ chainId, poolId }: AddRewardTokenProps) => {
 
   return (
     <FormProvider {...form}>
-      {/* eslint-disable-next-line @typescript-eslint/no-misused-promises -- Existing violation before enabling this rule. */}
-      <form onSubmit={handleSubmit(onSubmit)}>
+      <form onSubmit={event => void handleSubmit(onSubmit)(event)}>
```
0xAlunara, line 77: Not using `Form.tsx` so we have to wrap it in a `void` manually, same for the `ActionStepper.tsx` calls

packages/curve-ui-kit/src/shared/ui/ModalDialog.tsx
```diff
@@ -79,7 +84,9 @@ type ModalDialogProps = {
  * When we have a form inside the modal, we can use this component to wrap it with a form element,
  * so that we can keep the form submission in the footer.
  */
-const Form = (props: FormHTMLAttributes<HTMLFormElement>) => <form {...props} />
+const Form = ({ onSubmit, ...props }: ModalFormProps) => (
+  <form {...props} onSubmit={onSubmit ? event => void onSubmit(event) : undefined} />
```
0xAlunara, line 88: Gippety told me `useCallback` isn't worth the hassle (same for `Form.tsx`)

EDIT: this got me thinking why do we have a custom `<form>` inside this component instead of re-using `Form.tsx`. But that's for another PR.

packages/curve-ui-kit/src/features/report-error/ErrorReportModal.tsx
```diff
@@ -62,8 +62,7 @@ export const ErrorReportModal = ({ isOpen, onClose, context }: ErrorReportModalP
       footer={
         <Button
           fullWidth
-          // eslint-disable-next-line @typescript-eslint/no-misused-promises -- Existing violation before enabling this rule.
-          onClick={onSubmit}
+          onClick={event => void onSubmit(event)}
```
DanielSchiavini, line 65: voiding like this everywhere is still misusing promises and might miss exceptions. It's not much different than ignoring the lint rule

DanielSchiavini (approved):
I'm approving because the original was already not ideal but I think we should properly handle the exceptions instead of voiding everywhere

packages/curve-ui-kit/src/features/report-error/ErrorReportModal.tsx
0xAlunara: I'm not sure if there's a way around this, because the type of `onSubmit` for `<form`> is `SubmitEventHandler<HTMLFormElement> | undefined`, which implies only sync callbacks are accepted. But both react-hook-forms as Tanstack Forms allow `onSubmit` to be async. 

When looking at [Tanstack Form](https://tanstack.com/form/v1/docs/framework/react/guides/submission-handling), their example has an async `onSubmit`, but they call `form.handleSubmit()` without `void`. I don't know the exact type, but perhaps the example would cause lint errors for us too if it's a promise that they're not `void`ing. If it's sync then the migration to Tanstack Forms might solve this issue.

EDIT: looks to be async too...
image

image

DanielSchiavini: We should just handle the errors in our useForm hook. Thats where the promise should become a void, and errors should be returned to the caller.
Anyway that can be done after this PR

0xPearce (approved):

PR #2819, opened 2026-07-01
fix: use static Vyper vulnerability pool lookup

OnlyJousting:
Beta pool list was showing the Vyper vulnerability deposit warning based only on vyperVersion, which creates false positives for pools compiled with affected Vyper versions but not using vulnerable implementations. This caused pools like Ethereum stETH/frxETH to show an incorrect “do not deposit” warning.

Changes:
- Replace the vyperVersion based alert check with a static vulnerable pool address lookup.
- Add poolList.alerts.ts with static vulnerable pool addresses keyed by chainId.


apps/main/src/dex/features/pool-list/poolList.alerts.ts
```diff
@@ -0,0 +1,99 @@
+/**
+ * Static lookup for pools that should show the Vyper reentrancy vulnerability warning.
+ *
+ * This is checked in as data because the incident set is historical and closed; it
```
DanielSchiavini: These comments should take the whole line width of 120.
Also, it's way too long I think we can describe this all in a much more succinct way

apps/main/src/dex/features/pool-list/poolList.alerts.ts
```diff
@@ -0,0 +1,99 @@
...
+  },
+  // Avalanche (43114)
+  43114: {
+    // factory-v2-33 - Topshelf AVAXL - 0x64448b78561690b70e17cbe8029a3e5c1bb7136e
+    '0x445109bcf194c538b73ea60c71bb773849bc9b79': true,
+  },
+}
+
+export const isVyperVulnerablePool = (chainId: number | undefined, poolAddress: string | undefined) =>
+  chainId != null &&
```
DanielSchiavini: this will incorrectly return false when it should be undefined. That's why we introduced the `maybes` helper

OnlyJousting: I think false is perfect here if there isn't a chainId present, maybes just increases complexity?

DanielSchiavini: if there isn't a chainId then we don't know if there is a vulnerability, saying there is no vulnerability is incorrect

OnlyJousting: we're not gonna show a `?` state for all alerts, if there is no chainId there is no pool list either

0xtutti (dismissed):

apps/main/src/dex/features/pool-list/poolList.alerts.ts
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * `pool.hasVyperVulnerability()` delegates to curve-js
+ * `checkVyperVulnerability(chainId, poolId, implementation)`, so this list is
+ * based on known affected pool ids/implementations rather than `vyperVersion`.
+ * Inline comments keep the curve-js pool id and display name for auditing.
+ */
+
+export const VYPER_VULNERABLE_POOL_ADDRESS_LOOKUP: Partial<Record<number, Readonly<Record<string, true>>>> = {
+  // Ethereum (1)
+  1: {
```
0xAlunara: You can remove the comments and just use `[Chain.Ethereum]: {` I think (which is preferred over hardcoded numbers anyway (they've changed before like for testnets))

0xAlunara (dismissed):

apps/main/src/dex/features/pool-list/poolList.alerts.ts
DanielSchiavini: network.chainId and pool.address is never null anyway, otherwise you'd have to call network?.chainId and pool?.address

OnlyJousting: I'll update it

apps/main/src/dex/features/pool-list/poolList.utils.ts
```diff
@@ -31,7 +31,7 @@ export const getPoolListItem = (
 ): PoolListItem => ({
   ...pool,
   hasPosition,
-  hasVyperVulnerability: pool.vyperVersion != null && VYPER_EXPLOIT_VERSIONS.has(pool.vyperVersion),
+  hasVyperVulnerability: isVyperVulnerablePool(network.chainId, pool.address),
```
DanielSchiavini, line 34: It looks like isVyperVulnerablePool doesn't need to support null/undefined at all!

DanielSchiavini (approved):

0xPearce (approved):

PR #2834, opened 2026-07-02
refactor: delete unused code with knip

0xAlunara:
I'm strongly in favor of being more strict on disallowing unused exports, because that way linters and tools can more easily detect whether something is unused, because exports could exist for the sake of outside consumers, even without internal consumers.


knip.json
```diff
@@ -0,0 +1,5 @@
+{
```
0xAlunara, line 1: Not the perfect config, but was a good base which I don't want to lose. I did not add `knip` as a dependency though, I don't see the need. You can just install it globally on your local development machine.

DanielSchiavini (approved):
Great!

0xPearce (approved):
YESSS

PR #2821, opened 2026-07-01
feat: metric notional queries

DanielSchiavini:
- convert notional props to query object
- update usages
- extract duplicated calculations to decimalPercent helper

PR #2837, opened 2026-07-03
fix: no pool composition for empty pool

0xAlunara:
Filtered a bit too aggressively.

**Before**
image

**After**
image

### DIFF
--- a/apps/main/src/dex/features/pool-information/hooks/useMetrics.ts
+++ b/apps/main/src/dex/features/pool-information/hooks/useMetrics.ts
@@ -21,8 +21,8 @@ export const useMetrics = ({
   const staked = usePoolTotalStaked(poolDataCacheOrApi)
 
   return {
-    gaugeTotalSupply: maybe(staked?.gaugeTotalSupply, x => +x),
-    totalStakedPercent: maybe(staked?.totalStakedPercent, x => +x),
+    gaugeTotalSupply: maybe(staked?.gaugeTotalSupply, x => (x === 'N/A' ? 0 : +x)),
+    totalStakedPercent: maybe(staked?.totalStakedPercent, x => (x === 'N/A' ? 0 : +x)),
     liquidityUtilization: maybes(
       [tvlFromCurve ?? pricesApiPoolData?.tvlUsd, volumeFromCurve ?? pricesApiPoolData?.tradingVolume24h],
       (tvl, volume) => +tvl && (+volume / +tvl) * 100,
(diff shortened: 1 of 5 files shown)


apps/main/src/dex/features/pool-information/hooks/useMetrics.ts
```diff
@@ -21,8 +21,8 @@ export const useMetrics = ({
   const staked = usePoolTotalStaked(poolDataCacheOrApi)
 
   return {
-    gaugeTotalSupply: maybe(staked?.gaugeTotalSupply, x => +x),
-    totalStakedPercent: maybe(staked?.totalStakedPercent, x => +x),
+    gaugeTotalSupply: maybe(staked?.gaugeTotalSupply, x => (x === 'N/A' ? 0 : +x)),
```
0xAlunara, line 24: It looks a bit stupid but this is what the legacy hook returns for lite networks. Refactoring is out of scope.

mimaklas (dismissed):

apps/main/src/dex/features/pool-information/hooks/useMetrics.ts
DanielSchiavini: I made this a query and used decimal in [this PR](https://github.com/curvefi/curve-frontend/pull/2821/changes#diff-312e82e607262dc519f199154538e3807b767734b61e74f06658130e568863e5)
```suggestion
    gaugeTotalSupply: decimal(staked?.gaugeTotalSupply),
```

apps/main/src/dex/features/pool-information/hooks/usePoolComposition.ts
```diff
@@ -42,32 +42,31 @@ export const usePoolComposition = ({
...
+    }
+  })
 
   return {
     error: null, // TODO: correctly handle error and loading state
     // this isn't a proper loading check, but we need a bigger refactor for that later on
     isLoading: usePricesApiReserves ? !pricesApiPoolData?.balances.length : !currencyReserves,
     rows,
     totalUsd: usePricesApiReserves ? pricesApiTotalUsd?.toString() : currencyReserves?.totalUsd,
+    hasBalance: sum(rows.map(row => row.amount ?? 0)) > 0,
```
DanielSchiavini: No need to calculate the whole thing just to throw the result away
```suggestion
    hasBalance: rows.some(row => row.amount),
```
or if we wanna keep it
```suggestion
    total: sum(rows.map(row => row.amount ?? 0)),
```

DanielSchiavini (dismissed):

apps/main/src/dex/features/pool-information/hooks/usePoolComposition.ts
0xAlunara: Ah right, even moved it outside of the hook now so i can just inline it in the prop, even easier

mimaklas (approved):

0xPearce (approved):

PR #2838, opened 2026-07-03
fix: negative base APY wasn't shown in yield breakdown

0xAlunara:
Base APY can be negative due to rebases and such

Pool: `0xeC19D5f427c56e5A2Df1c620539ECd20a4D0a419`
image

### DIFF
--- a/apps/main/src/dex/features/pool-information/hooks/useYieldBreakdown.tsx
+++ b/apps/main/src/dex/features/pool-information/hooks/useYieldBreakdown.tsx
@@ -119,7 +119,7 @@ export const useYieldBreakdown = ({
         iconPosition: 'left',
         primary: t`Base APY`,
       },
-      apy: maybe(rewardsApy?.base?.day, x => (+x > 0 ? +x : undefined)), // already APY, no need to calculate
+      apy: maybe(rewardsApy?.base?.day, x => +x), // already APY, no need to calculate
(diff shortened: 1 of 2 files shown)


pilotvietnam (dismissed):

apps/main/src/dex/features/pool-information/hooks/useYieldBreakdown.tsx
```diff
@@ -119,7 +119,7 @@ export const useYieldBreakdown = ({
         iconPosition: 'left',
         primary: t`Base APY`,
       },
-      apy: maybe(rewardsApy?.base?.day, x => (+x > 0 ? +x : undefined)), // already APY, no need to calculate
+      apy: maybe(rewardsApy?.base?.day, x => +x), // already APY, no need to calculate
```
DanielSchiavini, line 122: I'd use decimal, no reason to convert to number in this hook
```suggestion
      apy: decimal(rewardsApy?.base?.day), // already APY, no need to calculate
```

DanielSchiavini (approved):

apps/main/src/dex/features/pool-information/hooks/useYieldBreakdown.tsx
0xAlunara: The type of `apy` is `number | undefined` though, as the other rows get their apy that was as well.

0xPearce (approved):

PR #2840, opened 2026-07-03
feat: llamalend forms drawer

0xPearce:
- depends on #2729, next https://github.com/curvefi/curve-frontend/pull/2853
- moved the llamalend market page action forms into a drawer for mobile
- add an action bar sticky at the button of the market page screen
- update tests

Screenshot 2026-07-07 at 13 59 33
Screenshot 2026-07-07 at 14 00 08
Screenshot 2026-07-07 at 14 00 23


packages/curve-ui-kit/src/features/forms/FormButton.tsx
```diff
@@ -16,6 +18,8 @@ export type FormButtonProps = Pick<ButtonProps, 'disabled' | 'fullWidth' | 'load
   testId?: string
 }
 
+const fixedBottomSx = { position: 'fixed', bottom: 0, left: 0, right: 0 } as const
```
0xPearce, line 22: The form button is rendered at the bottom of the drawer when the placement is of type `mobile-drawer`

Image

packages/curve-ui-kit/src/shared/ui/SwipeableDrawer/SwipeableDrawer.tsx
```diff
@@ -29,9 +31,12 @@ export const SwipeableDrawer = ({ button, children, open, setOpen, paperSx }: Pr
       open={open}
       onClose={() => setOpen(false)}
       onOpen={() => setOpen(true)}
-      disableSwipeToOpen={false}
-      keepMounted={false}
-      slotProps={{ paper: { sx: paperSx } }}
+      // if false, add an invisible bottom hit area (~20px height) to swipe-open the drawer
+      disableSwipeToOpen={true}
```
0xPearce, line 35: This was the "bug" that @0xAlunara found yesterday during the sync. It wasn't a bug, it's the drawer hit-area needed to open the drawer by swiping up when it's closed. You can test it on production in the market list or any page that has a drawer, you can swipe up from the bottom and open it. I disabled this feature everywhere because I don't think it's useful, we usually have multiple drawers on pages

packages/curve-ui-kit/src/widgets/DetailPageLayout/form-context/FormPlacementContext.tsx
```diff
@@ -0,0 +1,16 @@
+import { createContext, use } from 'react'
+import { useIsMobile } from '@ui-kit/hooks/useBreakpoints'
+import type { FormPlacement } from '../types'
+
+export const FormPlacementContext = createContext<FormPlacement | undefined>(undefined)
```
0xPearce, line 5: I figured it's better a context than prop drilling the form placement down to every form

packages/curve-ui-kit/src/widgets/DetailPageLayout/form-context/FormPlacementContext.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+const useFormPlacement = () => {
+  const placement = use(FormPlacementContext)
+  if (!placement) throw new Error('useFormPlacement must be used within FormPlacementProvider')
+  return placement
+}
+
+export const getIsMobileFormDrawer = (placement: FormPlacement, isMobile: boolean) =>
+  placement === 'mobile-drawer' && isMobile
```
0xPearce, line 14: reused outside the context, in case we change from mobile to tablet we don't forget to update

packages/curve-ui-kit/src/widgets/DetailPageLayout/FormTabs.tsx
```diff
@@ -95,23 +101,36 @@ type FormTabsProps<T extends object> = UseFormTabOptions<T> & {
...
+            overflow={overflow}
+          />
+        )}
+        {subTab && subTabs.length > 1 && (
+          <TabsSwitcher
+            variant="underlined"
+            value={subTab.value}
+            options={subTabs}
+            overflow="fullWidth"
+            sx={applySxProps(!isMobileDrawer && { backgroundColor: t => t.design.Layer[1].Fill })}
```
0xAlunara, line 142: I don't think you need applySxProps here?

packages/curve-ui-kit/src/widgets/DetailPageLayout/FormTabs.tsx
```diff
@@ -95,23 +101,36 @@ type FormTabsProps<T extends object> = UseFormTabOptions<T> & {
...
-        />
-      )}
-      <WithWrapper shouldWrap={shouldWrap} Wrapper={FormContent}>
-        {content}
-      </WithWrapper>
-    </Stack>
+    <WithWrapper shouldWrap={isMobileDrawer} Wrapper={MobileFormTabsDrawer} tabs={tabs} onSelectTab={onChangeTab}>
+      <Stack sx={{ marginInline }}>
+        {isMobileDrawer ? (
+          <CardHeader title={tabOption.label} size="small" data-testid="mobile-form-active-action" />
```
0xAlunara: If we're only using the label here we might as well only return `const label = tabOption.label` in `useFormTabs`?

packages/curve-ui-kit/src/widgets/DetailPageLayout/DetailPageLayout.tsx
```diff
@@ -67,6 +80,8 @@ export const DetailPageLayout = ({
   const headerRef = useRef<HTMLDivElement>(null)
   // page header metrics's notionals lazy rendering make the height change by 9px so we need a smaller threshold
   const [, pageHeaderHeight = 0] = useResizeObserver(headerRef, { threshold: 5 })
+  const placement: FormPlacement = formTabs?.placement ?? 'inline'
```
0xAlunara: Don't need the explicit typing as it's implicit from the type of `formTabs?.placement`

packages/curve-ui-kit/src/widgets/DetailPageLayout/DetailPageLayout.tsx
```diff
@@ -75,35 +90,40 @@ export const DetailPageLayout = ({
...
+        container
+        data-testid={testId ?? 'detail-page-layout'}
+        spacing={PAGE_SPACING}
+        sx={{ ...PAGE_MARGIN, ...(!header && { marginBlockStart: Spacing.xl }) }}
+        direction="row-reverse" // direction is only used when size<12 (on mobile, form shows first, otherwise children first)
+      >
+        {isMobile && <Grid size={12}>{headerStack}</Grid>}
+        {/* In Figma, columns are 12/4/3, but too small around breakpoints. I've added one extra column.
+            Ultrawide isn't a breakpoint yet, use maxWidth so it's not too large. */}
+        {formTabs && !showMobileDrawer && (
```
0xAlunara: Shouldn't we compare with `formTabs?.content` here as was the original intent?

packages/curve-ui-kit/src/widgets/DetailPageLayout/FormTabs.tsx
0xPearce: no but it's more readable than `sx={!isMobileDrawer ? { backgroundColor: t => t.design.Layer[1].Fill } : undefined}` or alternatives imo

packages/curve-ui-kit/src/widgets/DetailPageLayout/DetailPageLayout.tsx
0xPearce: no but we should check `formTabs !== null` actually, then if the `formTabs?.content` it's not ready it loads the form skeleton

0xAlunara (dismissed):

packages/curve-ui-kit/src/widgets/DetailPageLayout/form-context/FormPlacementContext.tsx
DanielSchiavini: Context is a good solution sometimes, but I feel like this small value shouldn't justify a whole new context - however, I see we don't wanna pass the placement to every tab - and FormContext is already inside the tabs. Maybe it could be fixed with css? Like `display: none` when the parent has a specific class. Or the tabs **already know** which placement they have - the only places with branching placement are branching on the beta flag which can be accessed anywhere.

packages/curve-ui-kit/src/widgets/DetailPageLayout/FormTabs.tsx
```diff
@@ -74,10 +79,20 @@ function useFormTabs<T extends object>({ menu, params }: UseFormTabOptions<T>) {
 
   const Component = components[0] || Skeleton // skeleton just for mui Tab validation, won't be rendered due to href
   const content = <Component {...params} />
-  return { tab, tabs, subTabs, subTab, content, onChangeTab, onChangeSubTab }
+  return {
+    tab,
+    tabLabel: findTab(tabs, tabKey).label,
```
DanielSchiavini: It's not clear why the `tabLabel` isn't just `tab.label`

packages/curve-ui-kit/src/widgets/DetailPageLayout/MobileFormTabsDrawer.tsx
```diff
@@ -0,0 +1,68 @@
...
+const { Spacing, ButtonSize } = SizesAndSpaces
+
+type MobileFormTabsDrawerProps = {
+  children: ReactNode
+  tabs: readonly TabOption<string>[]
+  onSelectTab: (value: string) => void
+}
+
+export const MobileFormTabsDrawer = ({ children, tabs, onSelectTab }: MobileFormTabsDrawerProps) => {
+  const [open, , , , setOpen] = useSwitch(false)
```
DanielSchiavini:
```suggestion
  const [open, setOpen] = useState(false)
```

DanielSchiavini (commented):
Fancy stuff 🔥

packages/curve-ui-kit/src/widgets/DetailPageLayout/MobileFormTabsDrawer.tsx
0xAlunara: I can't believe you're the one suggesting not to use `useSwitch` 🤣

DanielSchiavini: Haha it would make sense if we use the other callbacks 😂

apps/main/src/dex/features/manage-pool/components/RefuelForm.tsx
```diff
@@ -88,7 +88,6 @@ export const RefuelForm = ({ chainId, blockchainId, poolAddress }: RefuelFormPar
 
       <FormButton
         pending={isPending}
-        size="large"
```
0xPearce, line 91: hope this wasn't a must, it's the only form button with size `large`, made it consistent with the others @0xAlunara

DanielSchiavini (dismissed):

apps/main/src/lend/components/PageVault/VaultTabs.tsx
```diff
@@ -24,7 +24,13 @@ const VaultMenu = [
       { value: 'unstake', label: t`Unstake`, component: props => <UnstakeForm networks={networks} {...props} /> },
     ],
   },
-  { value: 'claim', label: t`Claim`, component: props => <ClaimTab networks={networks} {...props} /> },
+  {
+    value: 'claim',
+    label: t`Claim`,
+    // claim tab handles its own buttons because it has multiple actions that are not submit buttons.
+    withFormButton: false,
```
0xPearce, line 31: Needed to avoid this `marginBlockEnd` 
Image
Didn't want to use `FormButton` for the claim tab, because then on mobile we would have two fixed buttons on the bottom and it implies more custom logic. this was the easiest

packages/curve-ui-kit/src/features/activity-table/panels/PoolLiquidityExpandedPanel.tsx
```diff
@@ -27,11 +27,7 @@ export const PoolLiquidityExpandedPanel: ExpandedPanel<PoolLiquidityRow> = ({
     <Stack>
       <Stack sx={{ paddingTop: Spacing.md, gap: Spacing.xs }}>
         {nonZeroAmounts.map(({ amount, token }, index) => (
-          <Stack
-            key={token?.address ?? index}
```
0xPearce, line 31: not sure why I got this key-index lint error

0xAlunara (dismissed):

DanielSchiavini (approved):

0xAlunara (approved):

PR #2841, opened 2026-07-03
feat: add shrink availability check

0xAlunara:
Implements the stub check that looks at reset availability first before anything else.

image

### DIFF
--- a/apps/main/package.json
+++ b/apps/main/package.json
@@ -18,7 +18,7 @@
     "postpublish": "pinst --enable"
   },
   "dependencies": {
-    "@curvefi/llamalend-api": "2.3.1",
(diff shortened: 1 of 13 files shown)


DanielSchiavini (approved):
Shouldn't we just hide the tab? It's not great UX to have an error as the first thing

0xAlunara:
Hiding is even worse imo, because it doesn't tell the user why it's suddenly gone.

0xPearce (approved):
Agree better to keep the tab and show an error, same for the llamalend supply claim tab. Do not like very much the "Please correct the errors" because you can't correct this one 😅

DanielSchiavini:
I agree hiding sucks, but maybe we should just move it to the last tab then

PR #2846, opened 2026-07-06
chore: enforce upper-cased global consts

DanielSchiavini:
- add new eslint rule to enforce strings, numbers and booleans to be upper-cased
- add new eslint rule to enforce functions to be camelCase or TitleCase (for components)
- rename the violations


0xPearce (approved):
Nice

0xtutti (approved):

PR #2850, opened 2026-07-06
feat: default to position history on market page for users with no active position but has historic ones

OnlyJousting:
Changes:
- Set tab to `Activity` in user position details if a user doesn't have an open borrow position but has activity history.

### DIFF
--- a/apps/main/src/llamalend/features/market-position-details/hooks/usePositionDetailsTabs.tsx
+++ b/apps/main/src/llamalend/features/market-position-details/hooks/usePositionDetailsTabs.tsx
@@ -1,4 +1,4 @@
-import { type ReactNode, useMemo } from 'react'
+import { type ReactNode, useEffect, useMemo } from 'react'
 import { UserPositionHistory } from '@/llamalend/features/user-position-history'
 import type { ParsedUserCollateralEvent } from '@/llamalend/features/user-position-history/hooks/useUserCollateralEvents'
 import Stack from '@mui/material/Stack'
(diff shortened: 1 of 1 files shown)


apps/main/src/llamalend/features/market-position-details/hooks/usePositionDetailsTabs.tsx
```diff
@@ -46,5 +46,12 @@ export const usePositionDetailsTabs = ({
   )
 
   const { tab = DEFAULT_TAB, onTabChange } = useTabs(tabOptions, DEFAULT_TAB)
+
+  useEffect(() => {
+    if (hasPosition === false && events.data?.length) {
+      onTabChange('activity')
```
DanielSchiavini, line 52: setting state on effect is an anti-pattern, could we maybe just change the DEFAULT_TAB?

DanielSchiavini (approved):

0xPearce (approved):
I don't really like this feature :/

OnlyJousting:
> I don't really like this feature :/

it can be a bummer to always have to be reminded of your mistakes if the last position didn't go well heh

apps/main/src/llamalend/features/market-position-details/hooks/usePositionDetailsTabs.tsx
OnlyJousting: it would also change the tab for users that has an active position unfortunately

DanielSchiavini: Why? `defaultTab: hasPosition === false && events.data?.length ? 'activity' : 'borrowDetails'`

PR #2861, opened 2026-07-08
feat: liq. buffer and health improvements

0xPearce:
- added back the "%" symbol for the new Health and Liq. Buffer
- the liq. buffer bar no longer capped at 25% but goes to 100%
- show inactive liq. buffer status rather then showing green and 100%
- update the tooltips with the formulas

Screenshot 2026-07-08 at 18 24 24


0xtutti (dismissed):

0xAlunara (dismissed):
Not a fan of the formula formatting haha
Image

Also I'm not sure if showing a value in the buffer bar is meaningful, it raises even more questions for me as a user.

1. Is it bad if the buffer is low even when it's inactive?
2. How do I raise it?
3. Do I even care when my loan is currently healthy?

Showing less might be better

Image

apps/main/src/llamalend/features/market-position-details/tooltips/HealthTooltipContent.tsx
```diff
@@ -18,5 +18,6 @@ export const HealthTooltipContent = () => (
     <TooltipDescription
       text={t`Once Health reaches 0, monitor the liquidation buffer to track your position before hard liquidation.`}
     />
+    <TooltipDescription text={t`Calculated as: (oracle_price - top_band_price) * collateral / debt`} />
```
DanielSchiavini: Like alu said, the display sucks - maybe we can use `<pre>` here
```suggestion
    <TooltipDescription text={
      <Box>
        {t`Calculated as: `}
        <pre>{t`(oracle_price - top_band_price) * collateral / debt`}</pre>
      </Box>
    } />
```

apps/main/src/llamalend/features/market-position-details/tooltips/LiquidationBufferTooltipContent.tsx
```diff
@@ -5,5 +5,6 @@ export const LiquidationBufferTooltipContent = () => (
   <TooltipWrapper>
     <TooltipDescription text={t`Liquidation Buffer shows the remaining buffer before hard liquidation can occur.`} />
     <TooltipDescription text={t`When at 0, hard liquidation may occur.`} />
+    <TooltipDescription text={t`Calculated as: get_x_down(user) * (1 - liquidation_discount) / debt - 1`} />
```
DanielSchiavini: same here

DanielSchiavini (dismissed):

0xPearce:
Didn't really found a solution to display the formulas without overflowing or making the tooltip significantly wider, so I removed them for now and created a task for this.

0xAlunara (approved):

DanielSchiavini (approved):

PR #2872, opened 2026-07-10
refactor: avoid react anti-patterns when reloading market tables

0xAlunara:
Found this overcomplicated reload functionality with `useState` and `useEffect` anti-pattern while working on JJ's new pool list filter PR which also contains (new) reload logic.

The gist is that resetting the given queries already resets `useLlamaMarkets`, which causes it to refetch and set `isPending` to true, indicating data is being (re)loaded. Note that `isPending` does not get set to true during stale cache refetching.

### DIFF
--- a/apps/main/src/llamalend/features/market-list/LlamaMarketsList.tsx
+++ b/apps/main/src/llamalend/features/market-list/LlamaMarketsList.tsx
@@ -6,12 +6,12 @@ import { useUserProfileStore } from '@ui-kit/features/user-profile'
 import { q } from '@ui-kit/types/util'
 import { ListPageWrapper } from '@ui-kit/widgets/ListPageWrapper'
 import {
-  invalidateUserLendingSupplies,
-  invalidateAllUserLendingVaults,
+  resetUserLendingSupplies,
+  resetAllUserLendingVaults,
   resetLendingVaults,
 } from '../../queries/market-list/lending-vaults'
 import { useLlamaMarkets } from '../../queries/market-list/llama-markets'
-import { invalidateAllUserMintMarkets, resetMintMarkets } from '../../queries/market-list/mint-markets'
+import { resetAllUserMintMarkets, resetMintMarkets } from '../../queries/market-list/mint-markets'
 import { LendTableFooter } from './LendTableFooter'
 import { LlamaMarketsTable } from './LlamaMarketsTable'
 import { UserPositionsTables } from './UserPositionsTables'
@@ -29,9 +29,9 @@ const useTableLlamaMarkets = (address: Address | undefined) => {
       resetLendingVaults({}),
       resetMintMarkets({}),
       resetBadDebtMarkets(),
-      invalidateAllUserLendingVaults(address),
-      invalidateUserLendingSupplies({ userAddress: address }),
-      invalidateAllUserMintMarkets(address),
+      resetAllUserLendingVaults(address),
+      resetUserLendingSupplies({ userAddress: address }),
+      resetAllUserMintMarkets(address),
     ])
   }, [address])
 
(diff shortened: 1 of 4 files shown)


apps/main/src/llamalend/queries/market-list/lending-vaults.ts
```diff
@@ -76,6 +78,22 @@ export async function invalidateAllUserLendingVaults(userAddress: Address | null
   await Promise.all(invalidateContracts)
 }
 
+export async function resetAllUserLendingVaults(userAddress: Address | null | undefined) {
+  await resetUserLendingVaults({ userAddress })
+
+  const resetContracts = recordEntries(getCurrentUserLendingVaults({ userAddress }) ?? {}).flatMap(
```
0xAlunara, line 84: A bit of code repetition but i didn't want to overengineer. We still need the invalidate variant (likewise of the mint markets) as these are used elsewhere too. They're primarily used to invalidate data after submitting a contract like in `useLlamaMutation`. If you're ok with resetting rather than invalidating I'm fine with that too. This would mean loading states will be shown after a contract execution rather than values suddenly updating when reloading is done.

apps/main/src/llamalend/features/market-list/LlamaMarketsList.tsx
```diff
@@ -1,68 +1,43 @@
...
+      resetAllUserLendingVaults(address),
+      resetUserLendingSupplies({ userAddress: address }),
+      resetAllUserMintMarkets(address),
+    ])
+  }, [address])
 
   return {
     onReload,
-    tableQuery: q({ ...query, isLoading: loading }),
+    tableQuery: q({ ...query, isLoading: query.isPending }), // isPending so that the table shows a loading state on initial load and manual reload, but not on cache refetches
```
0xAlunara, line 40: `isPending` mimics the old behavior where `isLoading` would only be true on the initial load and manual reload. Manual reload function now resets rather than invalidates so that `isPending` is true when the queries are manually reloaded, but not when they refetch due stale caches.

0xtutti (approved):

apps/main/src/llamalend/features/market-list/LlamaMarketsList.tsx
DanielSchiavini: I don't like to use isPending in some places and isLoading elsewhere. However, using isPending everywhere is not an option because some queries are just never enabled. We should find a pattern that works everywhere instead of introducing these kinds of inconsistencies

apps/main/src/llamalend/queries/market-list/lending-vaults.ts
DanielSchiavini: That's fine, they call different stuff, this is simple enough. I don't see why we need the const though, we could just pass it straight to promise.all

DanielSchiavini (approved):

apps/main/src/llamalend/queries/market-list/lending-vaults.ts
0xAlunara: I think it was done for readability 🤔

apps/main/src/llamalend/features/market-list/LlamaMarketsList.tsx
0xAlunara: I think it's best if we just add `isPending?` to `QueryProp`, as it's somethig completely different from `isLoading` or `isFetching`. Truth be told, we already abuse the notation by often passing `isFetching` to props we call `isLoading`. 🙈

DanielSchiavini: Yeah but I want to have consistent loading states and behavior. Think of components like action infos.
It's only safe to use isFetching when you know the query will be valid

PR #2874, opened 2026-07-13
feat: chain switcher sort by tvl + lite section

DanielSchiavini:
- sort the networks by tvl (pool or lending tvl, depending on the app)
- separate the lite networks in another header

### DIFF
--- a/packages/curve-ui-kit/src/features/switch-chain/ui/ChainList.tsx
+++ b/packages/curve-ui-kit/src/features/switch-chain/ui/ChainList.tsx
@@ -4,6 +4,7 @@ import Alert from '@mui/material/Alert'
 import AlertTitle from '@mui/material/AlertTitle'
 import Box from '@mui/material/Box'
 import MenuList from '@mui/material/MenuList'
+import { recordEntries } from '@primitives/objects.utils'
 import type { NetworkDef } from '@ui/utils'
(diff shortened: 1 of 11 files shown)


0xtutti (approved):

packages/curve-ui-kit/src/features/switch-chain/ui/ChainList.tsx
```diff
@@ -13,41 +14,50 @@ import { MenuItem } from '@ui-kit/shared/ui/MenuItem'
...
     () =>
       lodash.groupBy(
         options.filter(o => o.name.toLocaleLowerCase().includes(searchValue.toLocaleLowerCase())),
-        o => (o.isTestnet ? ChainType.test : ChainType.main),
-      ),
-    [options, searchValue],
+        o =>
+          o.isTestnet
+            ? ChainType.test
+            : o.isLite || (tvls && tvls[o.id] === undefined) // flag chains not supported by prices API as lite
```
0xAlunara, line 54: Doens't this mean that while TVLs are still loading chains could be incorrectly categorized as lite? Not sure this is a robust test

0xAlunara (approved):

packages/curve-ui-kit/src/features/switch-chain/ui/ChainList.tsx
DanielSchiavini: The other way around. It's only lite when we have some data and there is no data for that chain

PR #2875, opened 2026-07-14
fix: keep OHLC pages immutable when flattening

DanielSchiavini:
- fix a production-only candle chart crash when scaling/panning into historical OHLC data.
- replace in-place page reversal with an immutable reverse so React Query’s cached pages array is not mutated during chart data flattening.
- prevent lightweight-charts from receiving non-ascending duplicate timestamp data after historical pagination and re-renders.


0xAlunara (approved):
Maybe a good idea to have a lint rule to prevent all mutable array methods in favor of immutable ones if they exist?

0xtutti (approved):

PR #2871, opened 2026-07-10
refactor: available liquidity column with amount

0xPearce:
- llamalend available liquidity column with amount and notional

Screenshot 2026-07-10 at 18 20 39
Screenshot 2026-07-10 at 18 24 52
Screenshot 2026-07-10 at 17 50 53

PR #2877, opened 2026-07-14
refactor: format token helpers

DanielSchiavini:
- depends on #2871 (follow-up to [this review comment](https://github.com/curvefi/curve-frontend/pull/2871#discussion_r3569948188))
- create functions to format tokens consistently
- replace usages


0xtutti (dismissed):

packages/curve-ui-kit/src/utils/tokens.ts
```diff
@@ -0,0 +1,14 @@
+import type { Amount } from '@primitives/decimal.utils'
+import { formatNumber } from './number'
+
+/** Formats a token amount with compact suffixes for dense displays, e.g. "1.23k CRV". */
+export const formatTokenCompact = <T extends Amount | null | undefined>(value: T, symbol: string) =>
```
0xAlunara: Not gonna block this PR over it, but it'd be more ergonomic if we had just one function with an additional parameter called `category` which is just a subset of the token categories, so `compact | amount | balance`.

0xAlunara (dismissed):

0xtutti (approved):

0xAlunara (approved):

PR #2881, opened 2026-07-14
chore: test llamalend table disconnected

DanielSchiavini:
- run the llamalend table tests with and without wallet, randomly


0xtutti (dismissed):

tests/cypress/support/e2e.ts
```diff
@@ -15,7 +15,7 @@ Cypress.on(
  * However, in some cases we want to test functionality without a wallet connected.
  */
 Cypress.Commands.add('visitWithoutTestConnector', (route: AppRoute, options?: Partial<Cypress.VisitOptions>) =>
-  cy.visit(`/${route}`, {
+  cy.visit(`/${route.replace(/^\//, '')}`, {
```
DanielSchiavini, line 18: I don't like to add this replace, but we should actually keep this consistent with cy.visit tbh (and update usages)
```suggestion
  cy.visit(route, {
```

0xAlunara (approved):

0xtutti (approved):

PR #2887, opened 2026-07-15
fix: flaky tests

DanielSchiavini:
- fix create-loan component test flakyness due to token balances being refreshed mid-test
  - replaced custom query options with preset categories
  - add categories for token balance(s) & simplify comments
  - overwrite query type constant so the refresh interval is disabled during mocked tests
- fix "btn-close-filters" being clicked during animation (mostly on firefox) by also gating it with `onEntered`
- fix dex-market assert waiting for a 2nd route
  - this was due to an assertion already done in beforeEach
  - therefore I removed the beforeEach and do the visit on each test
- fix loan buttons not being found as visible due to being out of the viewport
  - added scroll into view
- fix soft liquidation flakyness
  - code was not being able to reach the target price sometimes
  - instead, the code calculates a target price on-the-fly
  - rpc test [run 1](https://github.com/curvefi/curve-frontend/actions/runs/29423064277/job/87378267776) | [run 2](https://github.com/curvefi/curve-frontend/actions/runs/29473805372/job/87542381388) | [run 3](https://github.com/curvefi/curve-frontend/actions/runs/29483736933/job/87573206151)
- fix query factory dependencies
  - we now check whether dependencies are !== undefined, ie. whether the query ran
  - this broke the shrink tests, which skipped some queries when the result was `false`
- change cleanup to beforeEach instead of afterEach, to make sure the afterEach isn't running as the new test starts


0xtutti (approved):

packages/curve-ui-kit/src/lib/model/query/query-categories.ts
```diff
@@ -12,6 +12,8 @@ export const QUERY_CATEGORIES = {
   // Global / shared queries
   'global.gasInfo': marketDetail,
   'global.tokenRate': marketDetail,
+  'global.tokenBalance': user, // for selected tokens, update balances quite often
```
0xAlunara, line 15: I'd make this urgent so that forms are always up to date in case the user has changed their balance by using their wallet somewhere else.

tests/cypress/support/helpers/llamalend/soft-liq-setup.helpers.ts
```diff
@@ -88,6 +89,38 @@ const isSoftLiquidationState = ({ borrowed, debt }: SoftLiquidationState) => deb
...
+  band,
+  client,
+}: Pick<SoftLiquidationReadParams, 'client' | 'ammAddress'> & { band: bigint }) => {
+  const [pUp, pDown] = await Promise.all([
+    client.readContract({ address: ammAddress, abi: AMM_ABI, functionName: 'p_oracle_up', args: [band] }),
+    client.readContract({ address: ammAddress, abi: AMM_ABI, functionName: 'p_oracle_down', args: [band] }),
+  ])
+  const [lower, upper] = pDown < pUp ? [pDown, pUp] : [pUp, pDown]
+
+  return lower + (upper - lower) / 2n
```
0xAlunara, line 103: Wouldn't `(pUp + pDown) / 2n` yield the same result?

0xAlunara (approved):

PR #2916, opened 2026-07-22
feat(llamalend): add native market section navigation

0xtutti:
next: https://github.com/curvefi/curve-frontend/pull/2952
## Intent

Give the Beta LlamaLend market page a clear, deep-linkable document structure so users can move between the major market sections and understand which section is currently in view.

## Implementation plan

The current patch follows this outline. Maintainers may adjust the component boundaries during review as long as the behavior and scope remain intact.

- add sections to the market page with ID
- add an optional navigation node to render in the `DetailPageLayout`
- add navigation bar where clicking an item updates the hash and uses native scrolling
- split the Beta price chart and market activity into separate sections


0xPearce (approved):

0xPearce:
closed in favor of https://github.com/curvefi/curve-frontend/pull/2952

0xPearce (dismissed):

0xPearce (dismissed):

apps/main/src/llamalend/widgets/market-section-nav/getMarketSections.ts
```diff
@@ -0,0 +1,32 @@
...
+const PARAMETERS_SECTION = {
+  value: 'market-parameters',
+  label: { default: t`Advanced details`, short: t`Advanced` },
+} as const
+const FAQ_SECTION = { value: 'faqs', label: { default: t`FAQs` } } as const
+const OVERVIEW_SECTION = { value: 'market-overview', label: { default: t`Overview` } } as const
+
+export const getMarketSections = ({ rateType }: { rateType: MarketRateType }): readonly MarketSectionOption[] => {
+  const sections = {
+    [MarketRateType.Supply]: [POSITION_SECTION, OVERVIEW_SECTION, RATES_SECTION, PARAMETERS_SECTION, FAQ_SECTION],
```
0xPearce: I know that the position section is not rendered on the Supply tab when the user doesn't have a position opened, creating a dead link, but we should render the card instead like on the borrow tab so it's always there. I will change it in another PR, that's not a problem since the navigation is behind Beta

packages/curve-ui-kit/src/widgets/DetailPageLayout/DetailPageLayout.tsx
```diff
@@ -83,22 +85,54 @@ export const DetailPageLayout = ({
...
+        <Stack ref={headerRef} sx={pageNavigation ? undefined : stickyHeaderSx(navHeight)}>
+          {header}
+        </Stack>
+      )}
+      {pageNavigation && (
+        <Stack
+          sx={{
+            backgroundColor: theme => theme.palette.background.default,
+            position: { tablet: 'sticky' },
+            // -1 to hide the top border behind the page headers and not have two borders when sticky
```
0xAlunara, line 103: Shouldn't we use `BorderWidth.thin` then?

apps/main/src/index.css
```diff
@@ -14,6 +14,10 @@ html {
   scrollbar-gutter: stable;
 }
 
+html:has([data-testid='market-section-nav']) {
+  scroll-behavior: smooth;
```
0xAlunara, line 19: Why is this here rather than inside the `sx` of the `MarketSectionNav` component?

apps/main/src/loan/components/MarketInformationComposite.tsx
```diff
@@ -29,17 +30,32 @@ export const MarketInformationComposite = ({ previewPrices }: MarketInformationC
...
+        <Stack sx={{ gap: PAGE_SPACING }}>
+          <ChartAndActivityComp previewPrices={previewPrices} />
+          <CrvUsdPriceChart />
+        </Stack>
+      </MarketSection>
+      <MarketSection id="historical-rates" ariaLabel={t`Rates`}>
+        <MarketHistoricalRatesChart rateMode={MarketRateType.Borrow} />
+      </MarketSection>
+      {isNewLlamaMarketDetailPage && (
+        <MarketSection id="market-activity" ariaLabel={t`Market activity`}>
```
0xAlunara: Aria label is pure gippety slop lol, seeing it all over the PR.

apps/main/src/llamalend/widgets/market-section-nav/MarketSection.tsx
```diff
@@ -0,0 +1,23 @@
+import type { ReactNode } from 'react'
+import Box from '@mui/material/Box'
+import type { MarketSectionId } from './types'
+
+export const MarketSection = ({
```
0xAlunara: Not really market related and the widget could be moved to ui-kit so it can be used for the pool page as well

apps/main/src/llamalend/widgets/market-section-nav/MarketSectionNav.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Order: top, right, bottom, left. The negative margins leave a 10%-high activation strip between 20% and 30% from the
+ * top of the viewport, so the section crossing that strip is considered active.
+ */
+const ACTIVE_SECTION_ROOT_MARGIN = '-20% 0px -70% 0px'
+
+export const MarketSectionNav = ({ sections }: { sections: readonly MarketSectionOption[] }) => {
+  const isMobile = useIsMobile()
+  const [activeSection, setActiveSection] = useState<MarketSectionId | undefined>(() => {
```
0xAlunara: Not something we could more easily manage with Tanstack Router?

EDIT: I give gippety a quick look at it:

Yes—use a hybrid approach. TanStack Router should own the URL fragment, but not the scroll-spy.

- Hash clicks are already routed through TanStack: [MarketSectionNav.tsx](/C:/Users/micro/Source/curve-frontend/apps/main/src/llamalend/widgets/market-section-nav/MarketSectionNav.tsx:62) → [TabsSwitcher.tsx](/C:/Users/micro/Source/curve-frontend/packages/curve-ui-kit/src/shared/ui/Tabs/TabsSwitcher.tsx:165) → [RouterLink.tsx](/C:/Users/micro/Source/curve-frontend/packages/curve-ui-kit/src/shared/ui/RouterLink.tsx:8).
- Replace the direct `window.location.hash` read with the reactive `useLocation().hash` wrapper. This also handles back/forward hash changes.
- TanStack supports hash scrolling, and this app already enables `scrollRestoration` in [routes/index.tsx](/C:/Users/micro/Source/curve-frontend/apps/main/src/routes/index.tsx:30). [TanStack documents built-in hash scrolling](https://tanstack.com/router/latest/docs/guide/scroll-restoration#hashtop-of-page-scrolling).
- Keep the `IntersectionObserver`. Router active matching—including `includeHash`—tracks URL state, not which DOM section is crossing the viewport activation strip.
- Don’t model these sections as child routes; they are simultaneously rendered document fragments.

I would not blindly delete the `window.location.replace()` workaround: the app can delay rendering `<Outlet>`, so Router may attempt its initial hash scroll before the target exists. If hard-load deep links need a fallback, directly call `document.getElementById(hash)?.scrollIntoView()` after mounting instead of navigating to the same URL again.

One adjacent issue: the supply navigation always includes `position-details`, but [PageVault/Page.tsx](/C:/Users/micro/Source/curve-frontend/apps/main/src/lend/components/PageVault/Page.tsx:83) only renders that section when `hasPosition` is true. That option should be conditional; Router cannot fix a missing anchor target.

apps/main/src/llamalend/widgets/market-section-nav/getMarketSections.ts
```diff
@@ -0,0 +1,32 @@
...
+      OVERVIEW_SECTION,
+      { value: 'price-chart', label: { default: t`Risk & Liquidation`, short: t`Risk` } },
+      RATES_SECTION,
+      { value: 'market-activity', label: { default: t`Market activity`, short: t`Activity` } },
+      PARAMETERS_SECTION,
+      FAQ_SECTION,
+    ],
+  } satisfies Record<MarketRateType, readonly MarketSectionOption[]>
+
+  return sections[rateType]
```
0xAlunara: If we return immediately in can be an arrow body function

0xAlunara (dismissed):
All comments are non-blocking, feel free to ignore or cherry-pick for a follow up PR.

apps/main/src/index.css
0xPearce: because we have to scroll on html, and must exist before initial hash navigation runs

apps/main/src/loan/components/MarketInformationComposite.tsx
0xPearce: yeah I guessed that Tutti maybe wanted to add it for SEO or smth, but it can be done in a dedicated PR, I removed it

apps/main/src/llamalend/widgets/market-section-nav/MarketSectionNav.tsx
0xPearce: I added the use of tanstack router, thanks

0xAlunara (approved):

mimaklas (approved):

PR #2917, opened 2026-07-22
fix(llamalend): align market chart cards

0xtutti:
## Intent

Make the Beta market analytics surfaces read as one card system: chart cards use the LlamaLend-local contained title treatment, and the standalone Swaps/Activity section has the same bounded height as the chart cards.

This PR should simplify the chart/activity composition without changing chart data, financial values, or activity behavior.

## Implementation plan

The current patch follows this outline. Maintainers may choose a different internal API if it keeps one state owner and preserves the same modes.

1. Apply the existing LlamaLend-local market card header to:
   - historical borrow and supply rate charts
   - interest-rate/utilization
   - historical crvUSD peg
2. Keep header actions, such as time-range selectors, in their existing positions and preserve heading semantics.
3. Support two explicit chart/activity modes through one shared state implementation:
   - combined mode starts on Chart and offers Chart, Swaps, and Activity
   - standalone mode starts on Swaps and offers only Swaps and Activity
4. Remove the duplicate standalone activity state/rendering path.
5. Size the standalone mode with the shared chart-height token and let its table content scroll internally.
6. Keep existing chart queries, activity queries, bands behavior, tooltips, formatters, and loading/error states unchanged.

## Reviewer focus

- combined and standalone modes have one clear state owner
- changing tabs does not remount or duplicate data-fetching unnecessarily
- standalone activity matches chart-card height and does not increase page height
- long activity tables scroll inside the card without horizontal page overflow
- card titles and header actions remain consistent across all market chart types
- lend and mint market implementations retain the same behavior

## Out of scope

- market section navigation; handled by #2916
- chart colors, series styling, legends, and financial calculations
- compact position cards; handled by #2918
- global Card or CardHeader theme changes

## Stack

Stacked on #2916. Part of the scoped replacement for #2866.

## Validation plan

- switch through Chart, Swaps, and Activity in combined mode
- switch between Swaps and Activity in standalone mode
- verify the correct initial tab for both modes
- verify the standalone card height and internal table scrolling on desktop and mobile
- verify historical-rate, rate-curve, and crvUSD-peg headers and their actions
- confirm existing chart/activity loading, empty, and error states remain intact

PR #2915, opened 2026-07-22
feat(llamalend): refresh market summary cards

0xtutti:
## Intent

Establish the feature-flagged Beta LlamaLend market-summary hierarchy without changing the legacy market page:

- keep the page header focused on the market identity and its primary rates
- move secondary market facts into a dedicated Overview card
- keep parameters and contracts in a separate Advanced details card
- apply the same composition consistently to lend, vault, and mint market pages

This is a presentation/composition change. It reuses the existing market context and query values; it does not add queries or change protocol behavior.

## Implementation plan

- [x] Gate the refreshed composition behind `useNewLlamaMarketDetailPage` and preserve the legacy header and combined Advanced details path.
- [x] Make the market header page-aware: show the borrow rate first on borrow pages and the supply rate first on supply pages.
- [x] Use a responsive two-column mobile/four-column desktop metric grid for the compact header and Overview summary.
- [x] Add LlamaLend-local card-header and metric-grid helpers using the semantic card spacing introduced in #2913.
- [x] Compose the Overview card from existing values for solvency, borrowers, deployment age, market tokens, prices, Max LTV, and max leverage.
- [x] Separate the Beta Advanced details card into parameters and contracts while retaining the existing combined card for the legacy path.
- [x] Share the existing Max LTV derivation between Overview and Advanced details and cover the pure calculation with a unit test.
- [x] Preserve token-symbol casing and expose the market page title as the page `h1`.
- [x] Update the existing LlamaLend market-detail test helpers for the new metric composition.

## Reviewer focus

- Beta versus legacy feature-flag composition
- borrow, supply/vault, and mint market variants
- primary-rate ordering and responsive metric layout
- existing loading, fallback, and tooltip behavior
- unchanged query execution and financial formatting

## Out of scope

- new market data sources or query work, including the future Total suppliers value
- market section navigation, charts/activity, and FAQ changes
- global Card or CardHeader theme changes
- transaction or protocol behavior

## Stack

Built on the semantic card padding tokens from #2913, which is now merged.

Part of the scoped replacement for #2866.

## Validation

- Max LTV unit coverage
- existing LlamaLend market-detail Cypress coverage updated for the new metric structure
- repository CI covers formatting, lint, typechecking, unit tests, and browser suites

PR #2866, opened 2026-07-09
Add LlamaLend market section navigation

0xtutti:
## Summary
- add a local LlamaLend market section nav using existing underlined TabsSwitcher
- wire borrow, mint, and supply detail pages with explicit section refs and anchors
- allow the top-level nav to switch the existing chart/activity tab without duplicating chart UI
- add focused Cypress helper assertions for the new nav behavior

## Replacement PRs
- [ ] #2910 — UI Kit: XS and contained tabs 
- [ ] #2913 — UI Kit: semantic card spacing tokens 
- [ ] #2915 — LlamaLend: market header and Overview cards 
- [ ] #2916 — LlamaLend: native market section navigation 
- [ ] #2917 — LlamaLend: aligned market chart cards 
- [ ] #2918 — LlamaLend: compact position cards 
- [ ] #2919 — LlamaLend: market FAQ card 

## Deferred
- [ ] Market detail query and derivation deduplication 

## Verification
- yarn workspace main eslint src/lend/components/ChartAndActivityComp/index.tsx src/lend/components/MarketInformationComposite.tsx src/lend/components/PageLendMarket/LendMarketPage.tsx src/lend/components/PageVault/Page.tsx src/llamalend/widgets/ChartAndActivityLayout.tsx src/llamalend/widgets/market-section-nav/MarketSection.tsx src/llamalend/widgets/market-section-nav/MarketSectionNav.tsx src/llamalend/widgets/market-section-nav/types.ts src/loan/components/ChartAndActivityComp.tsx src/loan/components/MarketInformationComposite.tsx src/loan/components/PageMintMarket/MintMarketPage.tsx
- yarn workspace tests eslint cypress/support/helpers/llamalend/market-details.helpers.ts
- git diff --check

Cypress was not run.

PR #2918, opened 2026-07-22
feat(llamalend): add compact position cards

0xtutti:
## Intent

Adapt the position-details surface to the denser Beta market-page composition while preserving the established stable layout and all existing financial behavior.

The Beta page should label the section as “Your position,” use the shared responsive metric rhythm, and keep empty states compact. The stable page should continue to use its current labels and density.

## Implementation plan

The current patch follows this outline. Maintainers may refine the prop or composition shape, but the feature-flag decision should remain at the page boundary rather than being rediscovered deep in the position components.

1. Select the position-card mode at each market-page entry point:
   - Beta borrow/mint: compact mode with “Your position”
   - stable borrow/mint: existing density with “Borrow Details”
   - Beta supply: “Your position”
   - stable supply: “Supply Details”
2. Pass the selected label and compact mode explicitly through the position-details composition.
3. Reuse the shared market metric grid for borrow-position metrics:
   - compact mode follows the Beta two-column mobile/four-column desktop rhythm
   - stable mode retains its existing one/four/five-column behavior
4. Keep every existing metric, tooltip, unit, query wrapper, and conditional leverage value unchanged.
5. Reduce empty-position padding and empty-state size only in compact mode.
6. Preserve the existing activity tab and health/liquidation details.
7. Update stories and component-test call sites so the intended mode is explicit.

## Reviewer focus

- compact/stable behavior is selected once and cannot drift between nested components
- all existing position metrics and financial formatting are unchanged
- optional leverage does not break the grid rhythm
- position activity remains reachable when events exist
- connected, disconnected, loading, no-position, and active-position states retain their behavior
- supply and borrow labels are correct in both release channels

## Out of scope

- market summary/header metrics; handled by #2915
- section navigation and chart/activity layout; handled by #2916 and #2917
- new position queries, health calculations, liquidation behavior, or transactions
- a general redesign of shared Metric or EmptyState components

## Stack

Stacked on #2917. Part of the scoped replacement for #2866.

## Validation plan

- Beta and stable borrow pages with and without a position
- disconnected and loading empty states
- leveraged and non-leveraged metric layouts
- position activity when collateral events exist
- Beta and stable supply labels
- desktop and 375px layouts with no horizontal overflow
- existing position-status component coverage and stories


0xPearce:
closed in favor of https://github.com/curvefi/curve-frontend/pull/2952

PR #2922, opened 2026-07-22
feat: new liquidation buffer formula and health bars

0xPearce:
depends on https://github.com/curvefi/curve-frontend/pull/2924 llamalend.js 2.3.12 https://github.com/curvefi/curve-llamalend.js/pull/133
next: https://github.com/curvefi/curve-frontend/pull/2926

- implement the new liquidation buffer formula as 
`Liquidation Buffer = health(full=False) / (loan_discount - liquidation_discount)`
- display health and liquidation buffer as separate stacked bars
- update state colors and badges for soft and hard liquidation

PR #2926, opened 2026-07-24
feat: new health factor metric

0xPearce:
depends on https://github.com/curvefi/curve-frontend/pull/2922, next https://github.com/curvefi/curve-frontend/pull/2937

- display health (the above-band range cushion) as a factor, where `1.00` represents the entry point into soft liquidation, instead of displaying it as `0.00%`.
- clamp negligible differences between full and non-full health to zero
- keep positive bar values visible with a `minWidth` of `1px`
- add development-only health diagnostics

Screenshot 2026-07-24 at 15 48 20
Screenshot 2026-07-24 at 16 03 29
Screenshot 2026-07-24 at 16 04 53


0xtutti (approved):

0xPearce:
opened this instead: https://github.com/curvefi/curve-frontend/pull/2940

PR #2928, opened 2026-07-24
chore: remove odos

0xAlunara:
Odos is winding down and cannot support routes any longer. It needs to be removed from the frontend. A vote to disable it on the zap contract will come up shortly.

I've just searched for the string "odos" and deleted anything that popped up.

[llamalend v2.4.0](https://github.com/curvefi/curve-llamalend.js/pull/134#event-28602064154)


0xPearce (approved):

0xPearce:
this instead: https://github.com/curvefi/curve-frontend/pull/2944

PR #2940, opened 2026-07-28
refactor: beta health and liquidation buffer 

0xPearce:
next: https://github.com/curvefi/curve-frontend/pull/2942

- implement the new liquidation buffer formula as
`Liquidation Buffer = health(full=False) / (loan_discount - liquidation_discount)`
- display health (the above-band range cushion) as a factor, where 1.00 represents the entry point into soft liquidation, instead of displaying it as 0.00%.
- clamp negligible differences between full and non-full health to zero
- keep positive bar values visible with a minWidth of 1px
- add development-only health diagnostics
- refactor the health details layout to a 5 columns grid layout
- implemented the new health and liquidation buffer design

Screenshot 2026-07-28 at 17 01 41
Screenshot 2026-07-28 at 17 03 07
Screenshot 2026-07-28 at 17 02 25


0xPearce:
@0xtutti  made the health bar and liquidation buffer slightly higher to match the `Metric` height

apps/main/src/llamalend/features/market-position-details/tooltips/HealthTooltipContent.tsx
```diff
@@ -2,21 +2,26 @@ import { TooltipDescription, TooltipWrapper } from '@/llamalend/widgets/tooltips
...
+  const liquidationBufferBoundary = variant === 'metric' ? t`Health reaches 1` : t`the bar is empty`
+
+  return (
+    <TooltipWrapper>
+      <TooltipDescription text={t`Health shows the cushion before your position enters liquidation protection.`} />
+      <TooltipDescription
+        text={
+          <Trans>
+            Higher is safer;{' '}
+            <Box component="span" sx={{ color: 'text.highlight' }}>
```
0xAlunara, line 16: Already the case in legacy, but im not a fan of using `text.highlight`. Makes it looks like a clickable link

0xAlunara (approved):
Looking good!

0xtutti (approved):

PR #2948, opened 2026-07-29
feat(ui-kit): show mobile tooltips in drawers

0xtutti:
## Goal

Make shared `curve-ui-kit` tooltips usable and discoverable on mobile by presenting their complex content in a bottom drawer, while leaving tablet and desktop tooltips unchanged.

## What changed

- On beta-channel mobile, the shared `Tooltip` keeps MUI activation plumbing but opens the existing `SwipeableDrawer` immediately on touch.
- Reuses `SwipeableDrawer` and `TooltipContent`; no new drawer component, public tooltip prop, dependency, or caller migration.
- Keeps controlled tooltips and non-mobile presentations on the existing MUI path.
- Makes drawer content fill the available width without changing desktop tooltip constraints.
- For `Metric`, mobile beta taps open the drawer instead of copying; an explicit full-width **Copy value** action appears after the tooltip content.
- Gates both the drawer and metric interaction change behind the existing release-channel feature-flag system so stable mode retains current behavior while spacing is refined.
- Leaves legacy and chart tooltip implementations unchanged.

## Why

Mobile has no reliable hover interaction, and long-press tooltips are difficult to discover and too constrained for richer metric explanations. A bottom drawer provides space and context while reusing the established UI-kit primitive and its dismissal, focus, scroll-lock, puller, and height behavior.

Moving copy into an explicit drawer action prevents one tap from both opening explanatory context and mutating the clipboard. Keeping the behavior in the shared UI kit applies it consistently to modern tooltips, while the beta gate allows spacing follow-ups before release.

## Validation

- `yarn format:check`
- `yarn lint`
- `yarn workspace curve-ui-kit typecheck`
- `yarn workspace tests typecheck`
- `yarn build`
- Focused Cypress component coverage added; not run locally because Cypress runs on deployment

Full root `yarn typecheck` still reports pre-existing `ProcessEnv` errors in API-server consumers; both affected workspaces pass their typechecks.

image


mimaklas (dismissed):

packages/curve-ui-kit/src/shared/ui/TooltipComponents.tsx
```diff
@@ -14,7 +15,7 @@ import { applySxProps } from '@ui-kit/utils'
 const { Spacing } = SizesAndSpaces
 
 export const TooltipWrapper = ({ children }: { children: ReactNode }) => (
-  <Stack sx={{ gap: Spacing.sm, maxWidth: '20rem' }}>{children}</Stack>
+  <Stack sx={{ gap: Spacing.sm, ...(!useIsMobile() && { maxWidth: '20rem' }) }}>{children}</Stack> // in mobile we show the tooltip in a drawer, so we don't need to limit the width
```
0xPearce: are you sure all tooltips are rendered inside a drawer on mobile? Only the Metric's tooltips right?

0xPearce (dismissed):

packages/curve-ui-kit/src/shared/ui/TooltipComponents.tsx
0xAlunara: Hmm good point, but I also found out that even though we use drawers for the metric ones, a tooltip is still shown as well when you hold and press it. Need to remove that too.

0xAlunara: Rather than using the breakpoint hook I'm now using pure CSS to see if there's a MUI drawer parent or not when applying the `maxWidth`.

packages/curve-ui-kit/src/shared/ui/Tooltip.tsx
```diff
@@ -37,20 +39,40 @@ const TooltipContent = (
...
-      {children}
-    </MuiTooltip>
+export const Tooltip = ({ title, body, clickable, children, slotProps, ...props }: TooltipProps) => {
+  const isMobile = useIsMobile()
+  const [drawerOpen, openDrawer, , , setDrawerOpen] = useSwitch(false)
+
+  return title || body ? (
+    <>
+      <MuiTooltip
+        key={`${isMobile}`} // force remount when switching so we don't change from uncontrolled to controlled; internal mui tooltip shenanigans
```
0xAlunara, line 49: Shorter than `.toString()` 🤓

0xAlunara (approved):

mimaklas (approved):

PR #2951, opened 2026-07-30
test: seed Cypress randomness

0xPearce:
This lays the groundwork for an AI-assisted workflow that runs tests, identifies failing cases, proposes and pushes fixes to a branch, then reads the resulting GitHub Actions output and iterates. In order to do that, reproducible test data and configurable actions are necessary for the agent to reliably replay failures and verify its changes.

New workflows can only work on the main branch, so to test we have to merge this branch first.

- seed Cypress test data so failures can be replayed locally and across browsers
- add a manual workflow for repeated, targeted flake detection runs
- collect failed job logs alongside uploaded screenshots and server logs

### DIFF
--- a/tests/cypress/support/seed.ts
+++ b/tests/cypress/support/seed.ts
@@ -1,0 +1,27 @@
+/** Produces replayable Cypress randomness by combining the logged run seed with the current spec path. */
+
+const hashSeed = (seed: string) => {
+  let hash = 2166136261
+
+  for (let index = 0; index < seed.length; index++) {
+    hash ^= seed.charCodeAt(index)
+    hash = Math.imul(hash, 16777619)
+  }
+
+  return hash >>> 0
+}
+
+export const createSeededRandom = (seed: string) => {
+  let state = hashSeed(seed)
+
+  return () => {
+    state = (Math.imul(state, 1664525) + 1013904223) >>> 0
+    return state / 0x100000000
+  }
+}
+
+export const getTestSeed = () => {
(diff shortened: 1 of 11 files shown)


tests/cypress/support/seed.ts
```diff
@@ -0,0 +1,27 @@
+/** Produces replayable Cypress randomness by combining the logged run seed with the current spec path. */
+
+const hashSeed = (seed: string) => {
```
0xAlunara, line 3: You might want to re-use the logic inside `seeds.ts` in the `prices-api` folder, as I've used test seeding there as well. You might also notice a helper functions there like `once` and `requireSeed`.

0xAlunara (approved):

mimaklas (approved):

tests/cypress/support/seed.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  return () => {
+    state = (Math.imul(state, 1664525) + 1013904223) >>> 0
+    return state / 0x100000000
+  }
+}
+
+export const getTestSeed = () => {
+  const seed = process.env.TEST_SEED
+  if (typeof seed !== 'string' || seed.length === 0) throw new Error('Missing Cypress TEST_SEED')
```
0xLokiB, line 25: AFAIK env vars are always string
```suggestion
  const seed = assert(process.env.TEST_SEED, 'Missing Cypress TEST_SEED')
```

PR #2959, opened 2026-07-30
feat: sort llamalend user positions

OnlyJousting:
Makes the LlamaLend user-position tables sortable, with borrowing positions defaulting to highest Borrow Amount first.
This required more than updating the column definitions because position values were previously fetched inside individual cells. TanStack Table can only sort values available through row accessors, so the position stats and token prices needed to be resolved at the table level and shared with the cells and summary.

Changes:
- Added sortable accessors for:
  - Borrow Amount, using debt value in USD.
  - Collateral Amount, including collateral converted into the borrow token during soft liquidation.
  - LTV and Health, using the same values displayed in their cells.
  - Collateral/Borrow pairs, sorted alphabetically.
- Changed the default borrowing sort to `Borrow Amount` descending.
- Batched position-stat and token-price queries at the market-table boundary.
- Enriched table rows with the resulting query states so rows re-sort as asynchronous data becomes available.
- Reused the enriched data across cells, status indicators, and the position summary, removing duplicate query orchestration.
- Kept loading, missing, and errored values at the end of sorted results while preserving cell-level states.
- Updated the table row types throughout the table, cell, and filter components to support the enriched row data.


0xtutti (dismissed):

0xAlunara (approved):

apps/main/src/llamalend/features/market-list/cells/HealthCell.tsx
```diff
@@ -1,37 +1,36 @@
...
+  const { data: { status } = {}, error } = row.original.positionQueries.stats
+  const health = getValue()
   const content = status ? getPositionStatusContent(assets.collateral.symbol, assets.borrowed.symbol)[status] : null
-  return health == null ? (
-    error && <ErrorCell error={error} />
-  ) : (
+
+  if (error) return <ErrorCell error={error} />
+
+  return maybe(health, health => (
```
0xPearce, line 24:
```suggestion
  return error ? <ErrorCell error={error} /> :
  maybe(health, health => (
```

0xPearce (approved):

PR #2981, opened 2026-08-05
feat: add sorting for net apy, base daily apy and crv apy

0xAlunara:
Add sorting for net apy, base daily apy and crv apy in the new pools list.

Base weekly apr, rewards apy and age sorting has yet to be implemented on the server-side before we can re-enable that too here.

In addition to adding sorting, I've also simplified `usePoolsSorting` a bit; it was a bit convoluted.


mimaklas (dismissed):

0xPearce (commented):
Sorting by Net APY the first one is 5k% and the second one 6k% ?

Image

Also for the CRV APY sorting, some have the `-` value but still get sorted?
Image

0xPearce (dismissed):
many failing tests 🤔

0xAlunara:
Merged with main to fix the flaky tests and fixed 1 broken test case (volatile chip now shows up correctly!)

mimaklas (approved):

0xPearce (approved):

PR #2982, opened 2026-08-05
refactor: health fallback value

0xPearce:
- change the health fallback value form `∞` to `-` so that users do not mistake an undefined value with Infinity


0xAlunara (approved):

mimaklas (approved):

apps/main/src/llamalend/widgets/action-card/LoanActionInfoList.tsx
```diff
@@ -160,16 +160,14 @@ export const LoanActionInfoList = ({
...
+            value={prevHealth ? mapQuery(prevHealth, data => formatNumber(data, 'health.compact')) : DISABLED_Q}
             futureValue={
-              // todo: do not ignore loading state for health - some forms/tests expect ∞ when the query is disabled
-              isFullRepay || health?.data === undefined
+              // todo: do not ignore loading state for health - some forms/tests expect the fallback when the query is disabled
+              isFullRepay
                 ? '∞'
-                : mapQuery(health, data => formatNumber(data, { abbreviate: true, fallback: '∞' }))
+                : health?.data === undefined
+                  ? '-'
```
0xLokiB, line 169: I don't think we need this, since we have a fallback? 🤔 might as well update the tests.
```suggestion
```

PR #2983, opened 2026-08-05
chore: replace pool list v2 tooltip e2e tests with component tests

0xAlunara:
I kinda 'abused' e2e tests to test the shape / content of the tooltips, when a component test would be sufficient, quicker and less flaky. So this PR moves tooltip tests out of e2e and into a dedicated component test so that should get rid of the flakiness.

### DIFF
--- a/apps/main/src/dex/features/pool-list/cells/RewardIcons.tsx
+++ b/apps/main/src/dex/features/pool-list/cells/RewardIcons.tsx
@@ -35,7 +35,7 @@ const ExtraRewardTooltipBody = ({ reward }: { reward: ExtraReward }) => (
   </Stack>
 )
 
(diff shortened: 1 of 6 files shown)


0xPearce (approved):
So are we no longer testing the actual tooltip integration? It’s happened to me a couple of times that a tooltip wasn’t rendered because of an issue with a ref or a wrapper

0xAlunara:
We're still testing the tooltip content, which is where the bulk of the logic resides. Only thing we're now not testing is the tooltip trigger itself. Imo the flakyness is not worth it. If there's a display bug with the hover trigger itself it'll be fairly obvious and consistent.

mimaklas (approved):

PR #2991, opened 2026-08-07
fix: use provider price impact when available

0xPearce:


0xPearce:
in favor of https://github.com/curvefi/curve-frontend/pull/2994

PR #2993, opened 2026-08-07
chore: temporarily disable router tests

0xAlunara:
Skip these tests for now as they're very flaky and blocking development. Besides, it's not really a 'unit test' as they depend on 3rd-party services


mimaklas (approved):

0xPearce (approved):
let's make a task to reenable them or fix it properly

PR #2995, opened 2026-08-07
fix: band range denom should be collateral token rather than borrowed token

0xAlunara:
**Before**
image

**After**
image

### DIFF
--- a/apps/main/src/llamalend/features/bands-chart/TooltipContent.tsx
+++ b/apps/main/src/llamalend/features/bands-chart/TooltipContent.tsx
@@ -40,7 +40,6 @@ export const TooltipContent = ({ data, collateralToken, borrowToken }: TooltipCo
   const palette = useBandsChartPalette()
   const hasMarketData = isPositiveDecimal(data.bandTotalValue)
   const hasUserData = isPositiveDecimal(data.userBandTotalValue)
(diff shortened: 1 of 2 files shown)


mimaklas (dismissed):

apps/main/src/llamalend/features/bands-chart/TooltipContent.tsx
```diff
@@ -40,8 +40,7 @@ export const TooltipContent = ({ data, collateralToken, borrowToken }: TooltipCo
   const palette = useBandsChartPalette()
   const hasMarketData = isPositiveDecimal(data.bandTotalValue)
   const hasUserData = isPositiveDecimal(data.userBandTotalValue)
-  const borrowTokenSuffix = borrowToken?.symbol ? ` ${borrowToken.symbol}` : ''
-  const bandRange = `${formatChartAxisNumber(data.p_down, { abbreviateFrom: false })} - ${formatChartAxisNumber(data.p_up, { abbreviateFrom: false })}${borrowTokenSuffix}`
+  const bandRange = `${formatChartAxisNumber(data.p_down, { abbreviateFrom: false })} - ${formatChartAxisNumber(data.p_up, { abbreviateFrom: false })}${collateralToken?.symbol ? ` ${collateralToken.symbol}` : ''}`
```
0xPearce: It should be denominated with the price pair `collateral token symbol / borrow token symbol` convention like we do in the liquidation threshold metric, the price chart etc
Image

0xPearce (changes requested):
Also you could maybe try to fix the bands tooltip being cropped: 

Image

mimaklas (approved):

0xPearce (approved):

PR #2998, opened 2026-08-07
fix: use tradable coins instead of literally all coins for new pools list

0xAlunara:
Adds the `tradeable_coins` pools v2 API property and uses it for token icons to fix metapools showing up.

**Before**
image

**After**
image

### DIFF
--- a/apps/main/src/dex/features/pool-list/cells/PoolTitleCell/PoolBadges.tsx
+++ b/apps/main/src/dex/features/pool-list/cells/PoolTitleCell/PoolBadges.tsx
@@ -51,7 +51,7 @@ const AlertBadge = ({ alert, source }: { alert: PoolAlert; source: 'pool' | 'tok
 
 /** Displays classification, status, pool alert, and token alert badges for a pool. */
 export const PoolBadges = ({ pool }: { pool: PoolRow }) => {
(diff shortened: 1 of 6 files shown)


mimaklas (dismissed):

apps/main/src/dex/features/pool-list/cells/PoolTitleCell/PoolBadges.tsx
```diff
@@ -51,7 +51,7 @@ const AlertBadge = ({ alert, source }: { alert: PoolAlert; source: 'pool' | 'tok
 
 /** Displays classification, status, pool alert, and token alert badges for a pool. */
 export const PoolBadges = ({ pool }: { pool: PoolRow }) => {
-  const tokenAddresses = useMemo(() => pool.coins.map(({ address }) => address), [pool.coins])
+  const tokenAddresses = useMemo(() => pool.tradeableCoins.map(({ address }) => address), [pool.tradeableCoins])
```
0xPearce: Why don't we want to display alerts for the other coins as well?

0xAlunara: I don't think we have alerts for wrapper coins but sure I've reverted it

mimaklas (approved):

0xPearce (approved):

PR #2999, opened 2026-08-07
feat: add sorting by 7d base apy, rewards apy and age

0xAlunara:
Prices v2 API just added supporting for these sorting methods


mimaklas (approved):

0xPearce (commented):
Sorting by ascending order seems to correctly display negative values, but then undefined values before positive values (example 7D column, but also in the base APY column and maybe other columns as well)

Image

0xAlunara:
That's because these values are not undefined, they're zero, but we have a rule where we don't show zero but a dash.

0xPearce (approved):

PR #3000, opened 2026-08-07
feat: add support for lite networks in new pools list

0xAlunara:
Adds support (in beta mode) for lite networks in the new pools list. 

Main idea is to have the table have its own (normalized) view model and map both pool list responses of Prices API and Lite API into that view model. Prices API and Lite API are both only invoked depending on if the network is lite or not, and if the chain is actually supported by the new API. Thus, the fallback to legacy for unsupported chains has been removed and we show an error message instead.

Lite networks don't support server-side pagination, filtering and sorting, so it all happens client-side on a single giant page (like the legacy pool list, basically). There's also no base APYs and Volume column support (yet). Sorting Net APY, CRV APY and Rewards APY for Lite is not yet supported either.

The following networks are expected to not be supported: Aurora, Moonbeam, Mega Testnet, Neon (Testnet)
The following networks will work in the future but not now: Celo

### DIFF
--- a/apps/main/src/dex/features/pool-list/columns/column.definitions.tsx
+++ b/apps/main/src/dex/features/pool-list/columns/column.definitions.tsx
@@ -28,7 +28,7 @@ type PoolColumnOptions = Omit<PoolColumn, 'id' | 'header'>
 
 const columnHelper = createColumnHelper<PoolRow>()
 
-// TanStack requires an accessorFn for sorting controls, even when it's a computed value rather than a direct property; manualSorting leaves ordering to the v2 pools API.
+// TanStack requires an accessorFn for sorting controls, even when it's a computed value rather than a direct property; manualSorting leaves ordering to the v2 pools API on full networks.
 const serverSortableAccessor = () => 0
 
 const createTooltip = (id: keyof typeof POOL_TITLES, body: ReactNode): Tooltip => ({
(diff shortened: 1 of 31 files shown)


apps/main/src/dex/features/pool-list/columns/column.definitions.tsx
```diff
@@ -53,78 +53,85 @@ const accessor = (
...
-  accessor(PoolColumnId.Age, 'creationDate', {
-    cell: AgeCell,
-    meta: {
-      type: 'numeric',
-      tooltip: createTooltip(PoolColumnId.Age, <AgeHeaderTooltipContent />),
-    },
-    sortUndefined: 'last',
-  }),
-] satisfies PoolColumn[]
+const createPoolColumns = ({ isLite }: { isLite: boolean }) =>
```
0xAlunara, line 56: Advice: enable 'Hide whitespace' when reviewing on Github

mimaklas (approved):

0xPearce (approved):

apps/main/src/dex/features/pool-list/hooks/usePoolsGlobalFilter.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const POOL_SEARCH_KEYS = [
+  'name',
+  { name: 'symbols', getFn: ({ coins }) => cleanValue(coins.map(({ symbol }) => symbol)) },
+  {
+    name: 'addresses',
+    getFn: ({ address, gauge, gauges, coins }) => [
+      ...new Set([
+        address,
+        ...(gauge ? [gauge.address] : []),
```
0xLokiB, line 13:
```suggestion
        ...notFalsy(gauge?.address),
```

PR #3002, opened 2026-08-11
fix: buffer leveraged max borrow amount

0xPearce:
- apply a `99.999%` of the max borrowable when leveraging when clicking the max borrowable or 100% chip, so that if the max debt changes in the meantime it doesn't throw validation errors

### DIFF
--- a/apps/main/src/llamalend/widgets/action-card/LoanFormTokenInput.tsx
+++ b/apps/main/src/llamalend/widgets/action-card/LoanFormTokenInput.tsx
@@ -131,10 +131,7 @@ export const LoanFormTokenInput = <
     },
     [name, onValueChange, updateForm],
   )
-  const onMaxBalance = useCallback(
-    (v?: Decimal) => onBalance(onMax ? onMax(v) : v),
-    [onBalance, onMax],
-  )
+  const onMaxBalance = useCallback((v?: Decimal) => onBalance(onMax ? onMax(v) : v), [onBalance, onMax])
   return (
     <LargeTokenInput
       name={name}
--- a/apps/main/src/llamalend/widgets/action-card/LoanFormTokenInput.tsx
+++ b/apps/main/src/llamalend/widgets/action-card/LoanFormTokenInput.tsx
@@ -29,7 +29,10 @@ export type LoanFormTokenInputProps<
    * Optional max-value query for this field, including loading and error state.
    * When present, it also carries an optional related max-field name whose errors should be reflected here.
    */
-  max?: QueryProp<Decimal> & { fieldName: TMaxFieldName }
+  max?: QueryProp<Decimal> & {
+    fieldName: TMaxFieldName
+    onMax?: NonNullable<LargeTokenInputProps['maxBalance']>['onMax']
+  }
   name: TFieldName
   form: UseFormReturn<TFieldValues> // the form, used to set the value and get errors
   testId: string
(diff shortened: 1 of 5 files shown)


apps/main/src/llamalend/widgets/action-card/LoanFormTokenInput.tsx
```diff
@@ -145,13 +150,13 @@ export const LoanFormTokenInput = <
...
       {...(!hideBalance && { walletBalance })}
-      maxBalance={max && { chips: 'range', balance: max }}
+      maxBalance={max && { chips: 'range', balance: max, onMax: max.onMax }}
       inputBalanceUsd={decimal(usdRate && usdRate * +(value ?? 0))}
       message={errorMessage ? undefined : message}
       disabled={disabled}
     >
       {maxMessage && !errorMessage && <HelperMessage onNumberClick={onBalance} message={maxMessage} />}
-      {errorMessage && <HelperMessage message={errorMessage} onNumberClick={onBalance} isError />}
+      {errorMessage && <HelperMessage message={errorMessage} onNumberClick={onMaxBalance} isError />}
```
0xAlunara, line 159: I'm not sure if we should change this. My assumption here is that if you have $10 and want to max leverage $200 and it throws an error, that same clickable $200 would be in the error.

But my worry is more that if the error is $10 and you're not maxing out, clicking the $10 will still lead you to $200 simply because an `onMax` callback is defined.

packages/curve-ui-kit/src/shared/ui/LargeTokenInput/LargeTokenInput.tsx
```diff
@@ -35,14 +35,23 @@ type InputChip = {
...
   /** The function that returns the new input amount, possibly based on the max balance. */
   newBalance: (() => void) | ((maxBalance?: Decimal) => Decimal | undefined)
+  /** Whether the chip selects the maximum balance. */
+  isMax?: boolean
 }
 
+const [MIN_PERCENTAGE, MAX_PERCENTAGE] = [0, 100]
+
+/** Splits a maximum percentage into evenly spaced percentage steps. */
+const createPercentageSteps = (maxPercentage: number, count = 4) =>
```
0xAlunara, line 45: Technically correct but imo I was fine with just ` [25, 50, 75, 100]` lol

packages/curve-ui-kit/src/shared/ui/LargeTokenInput/LargeTokenInput.tsx
```diff
@@ -88,6 +97,8 @@ export type LargeTokenInputProps = {
     showSlider?: boolean
     /** Custom or preset chips to show. */
     chips?: ChipsPreset | InputChip[]
+    /** Optional override for the amount selected by a Max-style chip. */
+    onMax?: (maxBalance?: Decimal) => Decimal | undefined
```
0xAlunara, line 101: Referring back to my other comment, I feel an override by callback introduces the chance of quirky (edge case) behavior. Wouldn't it be easier if we just let the consumer of this component modify the max debt query value that goes into `maxBalance.balance`?

In other words, rather than `...q(maxDebt)` we could do something like `...mapQuery(maxDebt, debt => getMaxDebt(debt))`. This keeps the (generic) component lean and puts the edge case responsibility onto the parent.

0xAlunara (changes requested):

packages/curve-ui-kit/src/shared/ui/LargeTokenInput/LargeTokenInput.tsx
0xPearce: agree but I didn't want to have three separate `100` magic numbers or constants for the max percentage

0xAlunara (dismissed):

0xAlunara (approved):

mimaklas (approved):

0xLokiB (commented):
Shouldn't the max be sticky when clicking on it? 99% is basically always wrong

PR #3004, opened 2026-08-11
chore: set router fee to zero

0xPearce:
- set router fee to zero

### DIFF
--- a/apps/router-api/src/router-fees.ts
+++ b/apps/router-api/src/router-fees.ts
@@ -2,7 +2,8 @@ import type { Address } from '@primitives/address.utils'
 import type { Decimal } from '@primitives/decimal.utils'
 import { Chain } from '@primitives/network.utils'
 
-export const ROUTER_FEE_BPS: Decimal = '8' // note: no fractions allowed by enso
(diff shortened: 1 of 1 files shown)


apps/router-api/src/router-fees.ts
```diff
@@ -2,7 +2,8 @@ import type { Address } from '@primitives/address.utils'
 import type { Decimal } from '@primitives/decimal.utils'
 import { Chain } from '@primitives/network.utils'
 
-export const ROUTER_FEE_BPS: Decimal = '8' // note: no fractions allowed by enso
+// Fee is set to 0 because previous value (8) was too important for big leverages (like x30)
```
0xAlunara, line 5: Fwiw Enso team themselves said it should only be taken once, not per swap; so 20x or 30x shouldn't really matter? I'm also not sure if 0 is the correct amount; surely somebody needs to get paid?

0xAlunara (approved):

mimaklas (approved):

apps/router-api/src/router-fees.ts
michwill: Yeah good q: does that work with 0? Should we ask Enso?

PR #3012, opened 2026-08-12
fix: improve form error alerts

0xPearce:
- dismissible form alert boxes
- add copy button for the content of the error
- add 5 line clamp to the form alert so the message is truncated

Screenshot 2026-08-12 at 16 10 31

### DIFF
--- a/packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.stories.tsx
+++ b/packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.stories.tsx
@@ -1,11 +1,11 @@
+import { range } from '@primitives/objects.utils'
 import type { Meta, StoryObj } from '@storybook/react-vite'
 import { FormAlerts } from './FormAlerts'
 
 const LONG_ERROR = new Error(
-  Array.from(
-    { length: 20 },
-    (_, index) => `Transaction error detail ${index}: reverted call data 0x${index.toString(16).padStart(8, '0')}`,
-  ).join(' '),
+  range(20)
+    .map(index => `Transaction error detail ${index}: reverted call data 0x${index.toString(16).padStart(8, '0')}`)
+    .join(' '),
 )
 
 const meta: Meta<typeof FormAlerts> = {
--- a/packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.stories.tsx
+++ b/packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.stories.tsx
@@ -1,0 +1,48 @@
+import type { Meta, StoryObj } from '@storybook/react-vite'
+import { FormAlerts } from './FormAlerts'
+
+const LONG_ERROR = new Error(
+  Array.from(
+    { length: 20 },
+    (_, index) => `Transaction error detail ${index}: reverted call data 0x${index.toString(16).padStart(8, '0')}`,
+  ).join(' '),
+)
+
+const meta: Meta<typeof FormAlerts> = {
+  title: 'UI Kit/Widgets/Detail Page/Form Alerts',
+  component: FormAlerts,
+  args: { formErrors: [], handledErrors: [] },
+  decorators: [Story => <div style={{ width: 'min(480px, 100vw)' }}>{Story()}</div>],
+}
+
+type Story = StoryObj<typeof FormAlerts>
+
+export const Default: Story = {
+  args: { error: new Error('Transaction reverted') },
+}
+
+export const ValidationErrors: Story = {
+  args: {
+    formErrors: [
+      ['collateral', 'Collateral amount exceeds your balance'],
+      ['debt', 'Debt amount is below the minimum'],
+    ],
+  },
+}
+
+export const RejectedTransaction: Story = {
+  args: { error: new Error('User rejected the request') },
+}
+
+export const LongSubmissionError: Story = {
+  args: { error: LONG_ERROR },
+}
+
+export const ValidationAndSubmissionErrors: Story = {
+  args: {
+    error: new Error('Unable to estimate gas'),
+    formErrors: [['debt', 'Enter a valid debt amount']],
+  },
+}
+
+export default meta
--- a/packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.tsx
+++ b/packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.tsx
@@ -1,13 +1,17 @@
+import { useState } from 'react'
+import CloseIcon from '@mui/icons-material/Close'
 import Alert from '@mui/material/Alert'
 import AlertTitle from '@mui/material/AlertTitle'
 import Box from '@mui/material/Box'
 import Button from '@mui/material/Button'
+import IconButton from '@mui/material/IconButton'
 import Stack from '@mui/material/Stack'
 import type { Decimal } from '@primitives/decimal.utils'
 import { ErrorReportModal } from '@ui-kit/features/report-error'
 import { usePreviousValue } from '@ui-kit/hooks/usePreviousValue'
 import { useSwitch } from '@ui-kit/hooks/useSwitch'
 import { t } from '@ui-kit/lib/i18n'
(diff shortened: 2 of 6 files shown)


packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.tsx
```diff
@@ -45,19 +52,46 @@ export const FormAlerts = <Field extends string>({ error, formErrors, handledErr
...
           <AlertTitle>{t`An error occurred`}</AlertTitle>
           <Stack sx={{ gap: Spacing.xs, width: '100%' }}>
-            <span>{getErrorMessage(error)}</span>
-            <Button color="ghost" size="extraSmall" sx={{ alignSelf: 'flex-end' }} onClick={openReportModal}>
-              {t`Submit error report`}
-            </Button>
+            <Box
+              component="span"
+              data-testid="loan-alert-error-message"
+              sx={{ display: '-webkit-box', WebkitBoxOrient: 'vertical', WebkitLineClamp: 5, overflow: 'hidden' }}
```
0xAlunara, line 79: Lmao I love how ellipsis clamping in 2026 with non-vendor CSS is still an unsolved problem.

packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.tsx
```diff
@@ -45,19 +52,46 @@ export const FormAlerts = <Field extends string>({ error, formErrors, handledErr
...
           variant="outlined"
           severity="error"
           sx={{ overflowWrap: 'anywhere' /* break anywhere as there is often JSON in the error breaking the design */ }}
           data-testid="loan-alert-error"
+          action={
+            <IconButton
+              color="ghost"
+              size="extraSmall"
+              title={t`Dismiss error`}
+              aria-label={t`Dismiss error`}
```
0xAlunara: hello gippety 😄

packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.tsx
```diff
@@ -34,7 +38,10 @@ const { Spacing } = SizesAndSpaces
 
 export const FormAlerts = <Field extends string>({ error, formErrors, handledErrors }: FormAlertProps<Field>) => {
   const [isReportOpen, openReportModal, closeReportModal] = useSwitch(false)
+  const [dismissedError, setDismissedError] = useState<Error | null>(null)
```
0xAlunara, line 42: I'm not sure if this could happen in prod, but we're only allowing one error at a time to be dismissed.

In the hypothetical case where we re-use the same component for two errors that happen simultaneously (like `<FormAlert error={firstError && secondError} />`) dismissing one alert would make the other one pop up again if it was previously dismissed.

packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.tsx
```diff
@@ -45,19 +52,46 @@ export const FormAlerts = <Field extends string>({ error, formErrors, handledErr
...
+            <Box
+              component="span"
+              data-testid="loan-alert-error-message"
+              sx={{ display: '-webkit-box', WebkitBoxOrient: 'vertical', WebkitLineClamp: 5, overflow: 'hidden' }}
+            >
+              {errorMessage}
+            </Box>
+            <Stack direction="row" sx={{ alignSelf: 'flex-end', alignItems: 'center', gap: Spacing.xxs }}>
+              <CopyIconButton
+                copyText={errorMessage}
```
0xAlunara, line 85: I don't believe `useCopyToClipboard` supports it yet, but right now it shows the copied text verbatim in the toast body. This means that copying large values (and most prominently on mobile) can block the entire screen, especially given the toasts are not dismissible. Would be a better UX if either the toast becomes dismissible, or the hook/component allows a different toast body to what's actually being copied. Or both.

Image

packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.stories.tsx
```diff
@@ -0,0 +1,48 @@
+import type { Meta, StoryObj } from '@storybook/react-vite'
+import { FormAlerts } from './FormAlerts'
+
+const LONG_ERROR = new Error(
+  Array.from(
+    { length: 20 },
+    (_, index) => `Transaction error detail ${index}: reverted call data 0x${index.toString(16).padStart(8, '0')}`,
+  ).join(' '),
```
0xLokiB: see [range](https://github.com/curvefi/curve-frontend/blob/72864e560a4b13c45221922fa573e4847e042594/packages/primitives/src/objects.utils.ts#L61)
```suggestion
  range(20).map(index => `Transaction error detail ${index}: reverted call data  0x${index.toString(16).padStart(8, '0')}`).join(' '),
```

packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.tsx
```diff
@@ -34,7 +38,10 @@ const { Spacing } = SizesAndSpaces
 
 export const FormAlerts = <Field extends string>({ error, formErrors, handledErrors }: FormAlertProps<Field>) => {
   const [isReportOpen, openReportModal, closeReportModal] = useSwitch(false)
+  const [dismissedError, setDismissedError] = useState<Error | null>(null)
   const unhandledErrors = formErrors.filter(([field]) => !handledErrors.includes(field))
+  const visibleError = error === dismissedError ? null : error
+  const errorMessage = visibleError ? getErrorMessage(visibleError) : ''
```
0xLokiB:
```suggestion
  const visibleError = error !== dismissedError && error
  const errorMessage = maybe(visibleError, getErrorMessage)
```

0xLokiB (dismissed):

packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.tsx
0xPearce: tried `maxHeight` but wasn't satisfied, but this looks and works nice, also it is supported everywhere

packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.tsx
0xPearce: Agree, do you have any idea how to solve this? Having a Set that remembers the dismissed error I fear that if an error is thrown a second time, it is considered as already dismissed, which shouldn't happen

mimaklas (approved):

packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.tsx
0xAlunara: Errors are class objects so I think it's compared by reference rather than value. So if the same error throws again it'll be thrown by a new instance of the `Error` class.

0xPearce: you are right, but wait just noticed something: we only pass one error to the `FormAlert`, so if the first one is dismissed in your previous example (`<FormAlert error={firstError && secondError} />`), we don't show the second one because the component never sees it

0xPearce: I think it's good like this for now, if we want the `FormAlert` to manage multiple errors I can make a ticket and fix that seperately

0xAlunara (approved):

packages/curve-ui-kit/src/widgets/DetailPageLayout/FormAlerts.tsx
```diff
@@ -85,6 +85,7 @@ export const FormAlerts = <Field extends string>({ error, formErrors, handledErr
                 copyText={errorMessage}
                 label={t`Copy error message`}
                 confirmationText={t`Error message copied to clipboard`}
+                confirmationMessage="" // confirmation title is enough
```
0xLokiB, line 88: Should the prop be optional, then?

PR #3013, opened 2026-08-12
fix: truncate wei conversion

0xPearce:
- truncate instead of rounding wei conversion

This was causing errors because llamalend.j truncates and we were rouding

### DIFF
--- a/packages/curve-ui-kit/src/utils/decimal.ts
+++ b/packages/curve-ui-kit/src/utils/decimal.ts
@@ -1,5 +1,5 @@
 import { BigNumber } from 'bignumber.js'
-import { formatUnits, parseUnits } from 'viem'
+import { formatUnits } from 'viem'
 import type { Amount, Decimal } from '@primitives/decimal.utils'
 import { maybe } from '@primitives/objects.utils'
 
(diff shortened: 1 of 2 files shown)


packages/curve-ui-kit/src/utils/decimal.ts
```diff
@@ -71,5 +71,11 @@ export const decimalDiv = (first: Decimal, second: Decimal) =>
 export const decimalPercent = (part: Decimal, total: Decimal): Decimal =>
   +total ? decimalMultiply(decimalDiv(part, total), '100') : '0'
 
-export const toWei = (n: string, decimals: number) => decimal(parseUnits(n || '0', decimals))!
+// Viem parseUnits rounds excess decimals, but we must truncate to match llamalend.js transaction amounts.
```
0xAlunara, line 74: I'm shocked viem does this. If your balance is 1.01 ETH and it gets somehow rounded to 1010000000000000001 wei you'd get a contract error as obviously that's bigger than your actual balance. It's better to have some dust than fail on a rounding error I think.

0xAlunara (approved):

mimaklas (approved):

PR #3015, opened 2026-08-13
chore: mobile tooltip drawer opt in

0xAlunara:
Turns out turning every tooltip into a drawer on mobile fails given we apply tooltip on interactive controls like buttons as well. This causes blocking drawers to pop up on mobile when you're just trying to use the app. After weighing multiple options I've decided to just make mobile tooltip drawers opt-in for now, starting with the `Metric` and `ActionInfo` components. Worst case scenario is some tooltips might not have a drawer on mobile, but they didn't have working tooltips before anyway so it's not like that's a bad thing.

In addition this also fixes the issue of clicking a metric on a mobile table cell. The intention is often to expand the row, not trigger a drawer.

PR also contains a (hopeful) fix for the flaky table filter popover tests, as setting testids via transition is wonky.

RPC tests: [1](https://github.com/curvefi/curve-frontend/actions/runs/31696232235/job/94434527821)

### DIFF
--- a/packages/curve-ui-kit/src/shared/ui/Tooltip.tsx
+++ b/packages/curve-ui-kit/src/shared/ui/Tooltip.tsx
@@ -1,5 +1,5 @@
 import lodash from 'lodash'
-import type { ReactNode } from 'react'
+import type { ReactNode, SyntheticEvent } from 'react'
 import Box from '@mui/material/Box'
 // eslint-disable-next-line no-restricted-imports
 import MuiTooltip, { TooltipProps as MuiTooltipProps } from '@mui/material/Tooltip'
@@ -13,6 +13,8 @@ import { InvertTheme } from './ThemeProvider'
 export type TooltipProps = MuiTooltipProps & {
   body?: ReactNode
   clickable?: boolean
+  /** Show the tooltip content in a drawer on mobile. Mobile tooltips are otherwise disabled. */
+  mobileDrawer?: boolean
 }
 
 const { MaxHeight, Spacing } = SizesAndSpaces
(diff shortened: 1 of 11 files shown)


mimaklas (dismissed):

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
```diff
@@ -219,12 +219,22 @@ testCases.forEach(([width, height, breakpoint]) => {
     it('should allow filtering by chain', firefoxRetry, () => {
       const chains = objectKeys(vaultData)
       const chain = oneOf(...chains)
-      withFilters(breakpoint, () => cy.get(`[data-testid="chip-chain-${chain}"]`).click())
+      const selectChain = (chain: Chain) =>
+        withFilters(breakpoint, () => {
+          cy.get(`[data-testid="chip-chain-${chain}"]`).click()
+          if (breakpoint === 'mobile') {
```
0xPearce: What's the purpose of this? this part was flaky on mobile?

0xPearce (dismissed):

tests/cypress/e2e/llamalend/llamalend-markets.cy.ts
0xAlunara: Ah shit forgot to remove this. This was a flaky test fix before the opt-in was decided and buttons would still pop up drawers.

mimaklas (approved):

0xPearce (approved):

packages/curve-ui-kit/src/shared/ui/Tooltip.tsx
```diff
@@ -39,39 +41,48 @@ const TooltipContent = ({ title, children }: { title?: ReactNode; children?: Rea
...
+  body,
+  clickable,
+  mobileDrawer = false,
+  children,
+  slotProps,
+  ...props
+}: TooltipProps) => {
   const isMobile = useIsMobile()
   const [drawerOpen, openDrawer, , , setDrawerOpen] = useSwitch(false)
+  const showMobileDrawer = isMobile && mobileDrawer
```
0xLokiB, line 55: Isn't isMobile the screen size, so it doesn't mean it's actually a touch screen?
Also, the tooltips were supposed to work on touch screens by using the long touch.

0xAlunara: Long touch technically worked but people wouldn't know about it (it's not intuitive), so tutti changed his mind. It's really a screen estate problem, so we only want these drawers on mobile. Touch screens are often big enough to show actual tooltips.

PR #3016, opened 2026-08-13
feat: whitelist leverage providers by market

0xPearce:
- whitelist leverage providers for each market
- disable leverage for markets without approved providers
- show only approved providers in leverage routes

[RPC run](https://github.com/curvefi/curve-frontend/actions/runs/31791727827)


packages/curve-ui-kit/src/widgets/RouteProvider/RouteProviderCard.tsx
```diff
@@ -52,83 +51,76 @@ export const RouteProviderCard = ({
   const Icon = RouteProviderIcons[router]
   const disabledTooltip = t`${RouteProviderLabels[router]} is unavailable on ${networks[chainId].name}.`
   const onClick = useCallback(() => (enabled ? onSelect(router) : undefined), [onSelect, router, enabled])
-  const is0xEnabled = use0xRouter()
-  const isCurveEnabled = useCurveRouter()
-  const isCurveSolverEnabled = useCurveSolverRouter()
   return (
-    (is0xEnabled || router !== '0x') &&
```
0xPearce, line 59: No longer decides if providers are available using the release-channel, it trusts the provider list supplied by the market flow

packages/curve-ui-kit/src/entities/router-api/router-api.utils.ts
```diff
@@ -49,54 +48,41 @@ export const parseMutationRoute = (
...
-  isCurveRouterEnabled(releaseChannel)
-    ? isCurveSolverRouterEnabled(releaseChannel) && SOLVER_CHAINS.includes(chainId)
-      ? 'curve-solver'
-      : 'curve'
-    : 'enso'
-
 /**
  * This function can be used as a callback for curve-js calldata methods or llamalend.js leverageZapV2 methods.
  */
-export const getExpectedFn = ({
```
0xPearce, line 68: There is a bug here, if the curve solver doesn't return a route (example syrup market) it blocks the other providers as well and it is not possible to levereage even though the other providers are ok

0xPearce: right now it is not super important because we can whitelist only the providers that works

apps/main/src/llamalend/markets.constants.ts
```diff
@@ -401,34 +402,145 @@ export const NO_LEVERAGE_LEND: PartialRecord<ApiChain, Address[]> = {
...
+      slippage: SLIPPAGE.stable.default,
+    },
     // sUSDe-crvUSD
-    '0xB536FEa3a01c95Dd09932440eC802A75410139D6': { slippage: SLIPPAGE.stable.default },
+    '0xB536FEa3a01c95Dd09932440eC802A75410139D6': {
+      ...DEFAULT_ZAPV2_MARKET_CONFIG,
+      slippage: SLIPPAGE.stable.default,
+    },
+    // Deprecated ZapV2 markets remain enabled so existing positions can deleverage.
+    '0x4f87158350c296955966059C50263F711cE0817C': DEFAULT_ZAPV2_MARKET_CONFIG,
```
0xPearce: probably a lot of these could be removed if not all of them

tests/cypress/support/helpers/llamalend/mocks/shared.mocks.ts
```diff
@@ -77,32 +77,31 @@ export const routeMutationMeta = {
...
-              protocol: 'curve',
-              action: 'swap',
-              args: {},
-              chainId,
-            },
-          ],
-          tx: { to: ROUTER_ADDRESS, data: ROUTER_CALLDATA, from: TEST_ADDRESS, value: '0' },
-        },
-      ],
+      body: routers.map(router => ({
```
0xAlunara:
```suggestion
      body: toArray(router).map(router => ({
```

apps/main/src/llamalend/markets.constants.ts
```diff
@@ -401,34 +402,110 @@ export const NO_LEVERAGE_LEND: PartialRecord<ApiChain, Address[]> = {
   ],
 }
 
-type MarketLeverageConfig = { slippage?: Decimal }
+export type ZapV2MarketConfig = { providers: readonly RouteProvider[]; slippage?: Decimal }
```
0xAlunara: unnecessary `export`

apps/main/src/llamalend/markets.constants.ts
```diff
@@ -401,34 +402,110 @@ export const NO_LEVERAGE_LEND: PartialRecord<ApiChain, Address[]> = {
...
 
-export const MARKET_LEVERAGE: PartialRecord<number, Record<Address, MarketLeverageConfig>> = {
+// Defaults means the most used
+const DEFAULT_ZAPV2_CONFIG = { providers: ['enso'] } satisfies ZapV2MarketConfig
+const DEFAULT_ZAPV2_STABLE_CONFIG = {
+  ...DEFAULT_ZAPV2_CONFIG,
+  slippage: SLIPPAGE.stable.default,
+}
+
+export const ZAPV2_MARKET_CONFIG: PartialRecord<number, Record<Address, ZapV2MarketConfig>> = {
```
0xAlunara: I know internally it's all called `ZapV2` but honestly it's such a shitty name because zap can mean anything. Should we call this `MARKETS_LEVERAGE_CONFIG`?

apps/main/src/llamalend/queries/create-loan/create-loan-query.helpers.ts
```diff
@@ -1,6 +1,13 @@
 import { getMarket, hasLegacyMintLeverage, hasZapV2 } from '@/llamalend/llama.utils'
 import type { MarketTemplate } from '@/llamalend/llamalend.types'
 import { LendMarketTemplate } from '@curvefi/llamalend-api/lib/lendMarkets'
+import type { RouteProvider } from '@primitives/router.utils'
+
+/** Checks whether leverage may be enabled for a create loan. */
+export const isLeverageCreateLoanSupported = (
```
0xAlunara: Its only consumer is `useCreateLoanForm.tsx` so I'd just move to there so it's more local

apps/main/src/llamalend/queries/borrow-more/borrow-more-query.helpers.ts
```diff
@@ -71,4 +72,7 @@ export const isLeverageBorrowMore = (
  * Checks whether leverage may be enabled for a given market.
  * This is used to determine whether to show the leverage toggle in the UI.
  */
-export const isLeverageBorrowMoreSupported = (market?: MarketTemplate) => !!market && hasZapV2(market)
+export const isLeverageBorrowMoreSupported = (
```
0xAlunara: Same here, I'd just move it to `useBorrowMoreForm.tsx`

apps/main/src/llamalend/queries/borrow-more/borrow-more-max-receive.query.ts
```diff
@@ -85,7 +89,8 @@ export const { useQuery: useBorrowMoreMaxReceive, invalidate: invalidateBorrowMo
               userAddress,
               zapAddress: getZapAddress(market),
               slippage,
-              ...(routeId && { router: getRouteById(routeId).router }),
+              router:
+                selectedProvider && leverageProviders.includes(selectedProvider) ? selectedProvider : leverageProviders,
```
0xAlunara: A bit confused by the typing here; `getRouteById` already asserts `routeId` so at this point the null check feels redundant

apps/main/src/llamalend/mutations/repay.mutation.ts
```diff
@@ -88,13 +94,23 @@ const repay = async (
...
+  leverageProviders,
+  ...props
+}: RepayOptions) => {
   const config = useConfig()
   const { mutate, error, isPending } = useMarketMutation<RepayMutation>({
     network,
     marketId,
     mutationKey: [...rootKeys.userMarket({ chainId, marketId, userAddress }), 'repay'] as const,
     mutationFn: async (variables, { market }) => {
+      if (getRepayImplementationType(market, variables) === 'zapV2') {
```
0xAlunara: Given this is a validation of some kind, I'd expect it to be handled by the validation suite rather than the `mutationFn`. But perhaps there's a good reason to keep this inside `mutationFn` that I don't yet know about. This holds for all three mutations `repay`, `create-loan` and `borrow-more`.

apps/main/src/llamalend/features/borrow/hooks/useCreateLoanForm.tsx
```diff
@@ -76,6 +78,13 @@ export function useCreateLoanForm<ChainId extends LlamaChainId>({
   }
   const form = useForm<CreateLoanForm>(formOptions)
   useSyncMarketLeverageSlippage(form, defaultSlippage)
+  const isLeverageSupported = isLeverageCreateLoanSupported(market, leverageProviders)
+  const { update: formUpdate } = form
+  useEffect(() => {
```
0xAlunara: Think `useFormSync` would be a better (more readable) fit here?

0xAlunara (dismissed):

apps/main/src/llamalend/queries/borrow-more/borrow-more-max-receive.query.ts
0xPearce: routeId is optional for the max-receive query because it can run before the user selects a route. And getRouteById asserts by throwing an error when called without an route ID

apps/main/src/llamalend/features/borrow/hooks/useCreateLoanForm.tsx
```diff
@@ -41,6 +42,11 @@ const validation = createLoanQueryValidationSuite({
   collateralRequired: true,
 })
 
+const isLeverageCreateLoanSupported = (
+  market: MarketTemplate | undefined,
+  leverageProviders: readonly RouteProvider[],
+) => !!market && (hasLegacyMintLeverage(market) || (hasZapV2(market) && leverageProviders.length > 0))
```
0xLokiB:
```suggestion
) => maybe(market, market => (hasLegacyMintLeverage(market) || (hasZapV2(market) && leverageProviders.length > 0)))
```

apps/main/src/llamalend/features/borrow/hooks/useCreateLoanForm.tsx
```diff
@@ -76,6 +83,8 @@ export function useCreateLoanForm<ChainId extends LlamaChainId>({
   }
   const form = useForm<CreateLoanForm>(formOptions)
   useSyncMarketLeverageSlippage(form, defaultSlippage)
+  const isLeverageSupported = isLeverageCreateLoanSupported(market, leverageProviders)
+  useFormSync(form, { leverageEnabled: false, routeId: undefined }, !!market && !isLeverageSupported)
```
0xLokiB: It's not clear why this statement hardcodes leverageEnabled to false

apps/main/src/llamalend/features/manage-loan/hooks/useBorrowMoreForm.ts
```diff
@@ -92,6 +107,11 @@ const isRouteRequired = (market: MarketTemplate | undefined, leverageEnabled: bo
   return !!implementation && isRouterRequired(implementation)
 }
 
+const isLeverageBorrowMoreSupported = (
+  market: MarketTemplate | undefined,
+  leverageProviders: readonly RouteProvider[],
+) => !!market && hasZapV2(market) && leverageProviders.length > 0
```
0xLokiB:
```suggestion
) => maybe(market, market => && hasZapV2(market) && leverageProviders.length > 0)
```

apps/main/src/llamalend/features/borrow/types.ts
```diff
@@ -22,7 +23,8 @@ type CalculatedValues = { maxDebt: Decimal | undefined; maxCollateral: Decimal |
 export type CreateLoanForm = MakeOptional<CompleteCreateLoanForm, 'debt' | 'userCollateral'> & CalculatedValues
 
 /** Full query type for create loan queries, including pool identification and all form fields */
-export type CreateLoanFormQuery<T = IChainId> = MarketQuery<T> & CompleteCreateLoanForm
+export type CreateLoanFormQuery<T = IChainId> = MarketQuery<T> &
+  CompleteCreateLoanForm & { leverageProviders: readonly RouteProvider[] }
```
0xLokiB: Do we really need to add the providers to queries? Or can we derive them from the other params?

apps/main/src/llamalend/features/manage-loan/hooks/useMaxBorrowMoreValues.ts
```diff
@@ -49,7 +49,9 @@ export function useMaxBorrowMoreValues<ChainId extends LlamaChainId>({
   useFormSync(form, { maxBorrowed: maxUserBorrowed.data })
   useFormSync(form, { maxDebt: maxReceive.data?.maxDebt })
   // the leverage checkbox only shows after this value is known, purposefully override the value if the backend changes
-  useFormSync(form, { leverageEnabled: events && isPositionLeveraged(events.originalLeverage) })
+  useFormSync(form, {
+    leverageEnabled: !!params.leverageProviders?.length && events && isPositionLeveraged(events.originalLeverage),
```
0xLokiB: Is this correct when we don't know the leverage providers yet?

apps/main/src/llamalend/features/manage-loan/hooks/useRepayTokens.ts
```diff
@@ -37,14 +40,16 @@ const getRepayTokenOptions = ({
       field: 'userBorrowed',
     },
     collateralToken &&
-      canRepayFromStateCollateral(market) && {
+      canRepayFromStateCollateral(market) &&
+      (!market || !hasZapV2(market) || hasLeverageProvider) && {
```
0xLokiB: looks like `canRepayFromStateCollateral` should be updated to include `hasLeverageProvider`
and that `hasZapV2` should be using `maybe`

apps/main/src/llamalend/features/manage-loan/hooks/useRepayTokens.ts
```diff
@@ -78,5 +90,5 @@ export const useRepayTokens = ({
     // eslint-disable-next-line @eslint-react/set-state-in-effect -- Existing violation before enabling this rule.
     if (defaultToken) setToken(defaultToken)
   }, [defaultToken])
-  return { tokens, token: token ?? tokens[0], onToken: setToken }
+  return { tokens, token: token && tokens.includes(token) ? token : tokens[0], onToken: setToken }
```
0xLokiB: Do we need this change? If there are legitimate cases where it's needed, a comment would be nice

apps/main/src/llamalend/queries/borrow-more/borrow-more-max-receive.query.ts
0xLokiB: How could a provider be selected when it's not in leverageProviders, tho?

apps/main/src/llamalend/queries/validation/repay.validation.ts
```diff
@@ -67,6 +70,7 @@ const validateRepayFieldsForMarket = (
       })
     const swapRequired = stateCollateral || userCollateral || routeId
     validateRoute(routeId, !!(type && swapRequired && isRouterRequired(type)))
+    validateRouteProvider(routeId, leverageProviders ?? [], !!leverageProviders && type === 'zapV2')
```
0xLokiB: `?? []` and `!!leverageProviders` are being repeated in calls to `validateRouteProvider`

apps/main/src/llamalend/llama.utils.ts
```diff
@@ -40,9 +41,22 @@ export const tryGetMarket = (marketId: MarketTemplate | string | null | undefine
...
+ * Resolves leverage providers from the market whitelist: approved markets get every provider on Beta and only their
+ * explicitly tested providers on Stable/Legacy, while unlisted markets remain disabled.
+ */
+export const getMarketLeverageProviders = (
+  chainId: number,
+  controllerAddress: Address | undefined,
+  releaseChannel: ReleaseChannel,
+): readonly RouteProvider[] => {
+  const config = controllerAddress && MARKETS_LEVERAGE_CONFIG[chainId]?.[getAddress(controllerAddress)]
+  return config ? (releaseChannel === ReleaseChannel.Beta ? RouteProviders : config.providers) : []
```
0xLokiB: Should we use maybe instead of defaulting to []?

packages/curve-ui-kit/src/entities/router-api/router-api.utils.ts
0xLokiB: Should your PR comment be in the code, though?

apps/main/src/llamalend/features/manage-loan/hooks/useBorrowMoreForm.ts
0xLokiB: actually do you think the maybe could be inside `hasZapV2` (you'll need `extends`) and `leverageProviders` could be nullable?

apps/main/src/llamalend/features/borrow/types.ts
0xPearce: leverage providers also depends on release channel and should affect the query key. But I scoped it to the actual queries that consumes them

apps/main/src/llamalend/features/manage-loan/hooks/useRepayTokens.ts
0xPearce: agree but adding it to `canRepayFromStateCollateral` would disable legacy V1 deleverage, which does not require a provider

apps/main/src/llamalend/features/borrow/hooks/useCreateLoanForm.tsx
0xPearce: It’s needed because leverage providers can differ between release channels, so leverage may be available in Beta but not Stable/Legacy. In that case we need to reset it to false

apps/main/src/llamalend/llama.utils.ts
0xPearce: Changed it to use maybe but validation is a bit heavier now because providers can be `undefined`

apps/main/src/llamalend/queries/borrow-more/borrow-more-max-receive.query.ts
0xPearce: this prevents using a stale route if we change release channel

0xAlunara (dismissed):

0xtutti (dismissed):

0xtutti (dismissed):

tests/cypress/component/llamalend/loan-action-info-layout.cy.tsx
```diff
@@ -117,10 +117,9 @@ describe('route provider allowlist', () => {
   it('renders only providers supplied by the market policy', () => {
     cy.mount(
       <ComponentTestWrapper config={mockedWagmiConfig}>
-        <LoanActionInfoList
-          isOpen
-          oraclePrice={q({ data: '1', isLoading: false, error: null })}
-          gas={q({ data: null, isLoading: false, error: null })}
+        <LoanActionSettings
+          slippage={undefined}
```
0xLokiB: This deserves a comment - shouldn't the prop be optional?

apps/main/src/llamalend/features/borrow/hooks/useCreateLoanForm.tsx
```diff
@@ -44,8 +44,8 @@ const validation = createLoanQueryValidationSuite({
 
 const isLeverageCreateLoanSupported = (
```
0xLokiB: btw if we change the `market` type to `T`, the return type gets more specific (it can only be `undefined` when the input can also be `undefined`)
```suggestion
const isLeverageCreateLoanSupported = <T extends MarketTemplate | undefined>(
```

apps/main/src/llamalend/features/manage-loan/hooks/useMaxBorrowMoreValues.ts
```diff
@@ -50,7 +51,10 @@ export function useMaxBorrowMoreValues<ChainId extends LlamaChainId>({
   useFormSync(form, { maxDebt: maxReceive.data?.maxDebt })
   // the leverage checkbox only shows after this value is known, purposefully override the value if the backend changes
   useFormSync(form, {
-    leverageEnabled: !!params.leverageProviders?.length && events && isPositionLeveraged(events.originalLeverage),
+    leverageEnabled: maybe(
+      params.leverageProviders,
+      providers => !!providers.length && events && isPositionLeveraged(events.originalLeverage),
```
0xLokiB:
```suggestion
      providers => providers.length > 0 && events && isPositionLeveraged(events.originalLeverage),
```

apps/main/src/llamalend/features/borrow/hooks/useCreateLoanForm.tsx
```diff
@@ -41,6 +42,11 @@ const validation = createLoanQueryValidationSuite({
   collateralRequired: true,
 })
 
+const isLeverageCreateLoanSupported = (
+  market: MarketTemplate | undefined,
+  leverageProviders: readonly RouteProvider[] | undefined,
+) => maybe(market, market => hasLegacyMintLeverage(market) || (hasZapV2(market) && !!leverageProviders?.length))
```
0xLokiB, line 48:
```suggestion
) => maybes([market, leverageProviders], (market, leverageProviders) => hasLegacyMintLeverage(market) || (hasZapV2(market) && leverageProviders.length > 0))
```

apps/main/src/llamalend/features/market-context/MarketContextProvider.tsx
```diff
@@ -24,6 +25,7 @@ export const MarketContextProvider = <ChainId extends LlamaChainId>({
 }) => {
   const { address: userAddress } = useConnection()
   const { llamaApi: api = null } = useCurve()
+  const [releaseChannel] = useReleaseChannel()
```
0xLokiB, line 28: I don't think this needs to be part of the market context

apps/main/src/llamalend/markets.constants.ts
```diff
@@ -401,34 +402,110 @@ export const NO_LEVERAGE_LEND: PartialRecord<ApiChain, Address[]> = {
...
+    '0xf8C786b1064889fFd3c8A08B48D5e0c159F4cBe3': DEFAULT_LEVERAGE_CONFIG, // cbBTC Mint
+    '0x2fb54c8eae57767A9A509A395b9C4FA0702e2675': DEFAULT_STABLE_LEVERAGE_CONFIG, // syrupUSDC-crvUSD
+    '0x3cD4d86a2c65e57ce4b4121b67E2D2224BA41bbe': DEFAULT_STABLE_LEVERAGE_CONFIG, // sfrxUSD-crvUSD v2
+    '0xC77d97cF01737EB7aCE46cAb7cd9F60eC51a40c0': DEFAULT_STABLE_LEVERAGE_CONFIG, // sDOLA-crvUSD v2
+    '0x4F79Fe450a2BAF833E8f50340BD230f5A3eCaFe9': DEFAULT_STABLE_LEVERAGE_CONFIG, // sreUSD-crvUSD
+    '0x8035b16053560b3C351b665b10f6C7dBDb6A1E05': DEFAULT_STABLE_LEVERAGE_CONFIG, // fxSAVE-crvUSD
+    '0x2dA313f6DCEE04BA46466E100c4656618E5d3dDd': DEFAULT_STABLE_LEVERAGE_CONFIG, // sUSDS-crvUSD
+    '0x3DE37c38739dFb83b7A902842bF5393040f7BF50': DEFAULT_STABLE_LEVERAGE_CONFIG, // sfrxUSD-crvUSD
+    '0xB536FEa3a01c95Dd09932440eC802A75410139D6': DEFAULT_STABLE_LEVERAGE_CONFIG, // sUSDe-crvUSD
+    // Deprecated ZapV2 markets remain enabled so existing positions can deleverage.
```
0xLokiB, line 432: Is there a reason why we just don't default any markets not listed to DEFAULT_LEVERAGE_CONFIG? Let's document that in a comment.

apps/main/src/llamalend/features/manage-loan/hooks/useMaxBorrowMoreValues.ts
0xPearce: is it always better to use `.length > 0` instead of `!!` ?

apps/main/src/llamalend/features/borrow/hooks/useCreateLoanForm.tsx
0xPearce: but legacy Mint leverage does not use route providers

apps/main/src/llamalend/markets.constants.ts
0xPearce: yes, we don't want to enable leverage to new markets

apps/main/src/llamalend/features/manage-loan/hooks/useMaxBorrowMoreValues.ts
0xLokiB: `!!` supports `number|null|undefined`, `>0` only supports `number`
Since this is inside a `maybe` call, it's fine to use `>0`

apps/main/src/llamalend/features/borrow/hooks/useCreateLoanForm.tsx
0xLokiB: ah so they are always null, got it (but wasn't clear from the code, maybe they should have length == 0?)

0xAlunara (dismissed):

apps/main/src/llamalend/markets.constants.ts
```diff
@@ -411,6 +411,7 @@ const DEFAULT_STABLE_LEVERAGE_CONFIG = {
   slippage: SLIPPAGE.stable.default,
 }
 
+// This is a leverage allowlist: unlisted markets remain disabled until their leverage routes are tested and approved
```
0xLokiB, line 414:
```suggestion
/** This is a leverage allowlist: unlisted markets remain disabled until their leverage routes are tested and approved */
```

0xLokiB (approved):

0xAlunara (approved):

PR #3019, opened 2026-08-14
refactor: snapshots query rename and return type edit

0xAlunara:
1. Renames `llama-snapshots` query to `market-snapshots` conform the standard.
2. Rename `useMarketSnapshots` to `useMarketRateHistory` to prevent name conflict with the new snapshots query and also make its use case more clear.
3. Expand on the return type of `market-snapshots` so you can actually access all properties of either mint or lend markets with a type guard.


0xPearce (approved):
Nice

mimaklas (approved):

PR #3022, opened 2026-08-18
refactor: move loan action settings into forms

0xPearce:
- move slippage and route provider controls into the forms
- show and hide leverage settings with a transition
- preserve the selected range when switching to the custom loan preset

Screenshot 2026-08-18 at 09 06 41

### DIFF
--- a/apps/main/src/llamalend/features/borrow/components/LoanPresetSelector.tsx
+++ b/apps/main/src/llamalend/features/borrow/components/LoanPresetSelector.tsx
@@ -41,7 +41,7 @@ export const LoanPresetSelector = ({
         onChange={useCallback(
           (_: MouseEvent<HTMLElement>, p: LoanPreset) => {
             setPreset(p)
-            setRange(PRESET_RANGES[p])
+            if (p !== LoanPreset.Custom) setRange(PRESET_RANGES[p])
           },
           [setPreset, setRange],
         )}
--- a/apps/main/src/llamalend/features/manage-loan/components/BorrowMoreForm.tsx
+++ b/apps/main/src/llamalend/features/manage-loan/components/BorrowMoreForm.tsx
@@ -144,13 +144,12 @@ export const BorrowMoreForm = <ChainId extends IChainId>({
             onToggle={onLeverageToggle}
             maxLeverage={max.maxLeverage.data}
           />
-          {values.leverageEnabled && (
-            <LoanActionSettings
-              slippage={values.slippage}
-              onSlippageChange={slippage => updateForm({ slippage })}
-              routes={routes}
-            />
-          )}
+          <LoanActionSettings
+            show={values.leverageEnabled}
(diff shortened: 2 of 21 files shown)


apps/main/src/llamalend/features/borrow/components/LoanPresetSelector.tsx
```diff
@@ -41,7 +41,7 @@ export const LoanPresetSelector = ({
         onChange={useCallback(
           (_: MouseEvent<HTMLElement>, p: LoanPreset) => {
             setPreset(p)
-            setRange(PRESET_RANGES[p])
+            if (p !== LoanPreset.Custom) setRange(PRESET_RANGES[p])
```
0xPearce, line 44: For the custom preset we want to see the last band number number selected, instead of 10

packages/curve-ui-kit/src/features/forms/FormProvider.tsx
```diff
@@ -6,9 +6,13 @@ import { IS_DEVELOPMENT } from '@ui-kit/utils'
 import type { FieldValues, UseFormReturn } from './form.types'
 import { FormContext } from './useFormContext'
 
-export type FormProviderProps<T extends FieldValues> = UseFormReturn<T> & { children: ReactNode }
+export type FormProviderProps<T extends FieldValues> = UseFormReturn<T> & {
+  children: ReactNode
+  showFormState?: boolean // Show the development-only form state accordion.
```
0xPearce: The form provider is also used for the slippage modal, so it was showing the dev-only form state accordion in the modal as well, which was very annoying

0xtutti (approved):

apps/main/src/llamalend/features/manage-loan/components/BorrowMoreForm.tsx
```diff
@@ -138,12 +137,20 @@ export const BorrowMoreForm = <ChainId extends IChainId>({
...
-        />
+        <Stack>
+          <LeverageInput
+            checked={values.leverageEnabled}
+            leverage={leverage}
+            onToggle={onLeverageToggle}
+            maxLeverage={max.maxLeverage.data}
+          />
+          <LoanActionSettings
+            show={values.leverageEnabled === true}
```
0xAlunara, line 148: `!== true` not necessary I think

0xAlunara (approved):

0xAlunara:
UX feels a bit cramped and I think it needs a second iteration (@0xtutti), but not a blocker. IT's nice that this makes the info lists more lightweight too!

image

apps/main/src/llamalend/features/borrow/components/LoanPresetSelector.tsx
0xLokiB: This comment should probably be in the code

PR #3026, opened 2026-08-18
fix: only show max boost if not zero and deviates from total

0xAlunara:
Hides the max boost label if it doesn't deviate from the normal total (eg there's no crv emissions with boost!)

**Before**
image

**After**
image

The API tells us the type of `crvApyRange` is
```ts
{
    unboostedApy: number;
    maximumApy: number;
} | undefined
```
But that doesn't mean it can't be `{ unboostedApy: 0, maximumApy: 0 }`. Hence to prevent further confusion I made the type `number` by defaulting to `0` if it's `undefined` rather than keeping the possibility of both `undefined` and `0`. This way you can easily check if it's non-zero rather than having to check for both in `FooterRow`.

### DIFF
--- a/apps/main/src/dex/features/pool-information/components/yield-breakdown/FooterRow.tsx
+++ b/apps/main/src/dex/features/pool-information/components/yield-breakdown/FooterRow.tsx
@@ -1,7 +1,6 @@
 import type { ReactNode } from 'react'
 import TableCell from '@mui/material/TableCell'
 import Typography from '@mui/material/Typography'
-import { maybe } from '@primitives/objects.utils'
 import type { Column } from '@tanstack/react-table'
 import { t } from '@ui-kit/lib/i18n'
 import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
@@ -13,7 +12,7 @@ const { Spacing } = SizesAndSpaces
 
 type FooterRowProps = {
   visibleColumns: Column<YieldBreakdownRow, unknown>[]
-  maxBoostTotal: number | undefined
+  maxBoostTotal: number
   total: number
 }
 
@@ -29,11 +28,11 @@ const footerCellByColumnId: Record<YieldBreakdownColumnId, (props: FooterCellPro
   [YieldBreakdownColumnId.Apy]: ({ columnId, maxBoostTotal, total }: FooterCellProps) => (
     <TableCell key={columnId} sx={{ paddingInline: Spacing.md, paddingBlock: Spacing.sm, textAlign: 'right' }}>
       <Typography variant="tableCellMBold">{formatNumber(total, 'percent.rate')}</Typography>
-      {maybe(maxBoostTotal, x => (
+      {maxBoostTotal && maxBoostTotal != total && (
         <Typography variant="tableCellSRegular" color="textSecondary">
-          {t`Max boost ${formatNumber(x, 'percent.rate')}`}
+          {t`Max boost ${formatNumber(maxBoostTotal, 'percent.rate')}`}
         </Typography>
-      ))}
+      )}
     </TableCell>
   ),
 }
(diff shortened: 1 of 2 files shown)


apps/main/src/dex/features/pool-information/components/yield-breakdown/FooterRow.tsx
```diff
@@ -29,11 +28,11 @@ const footerCellByColumnId: Record<YieldBreakdownColumnId, (props: FooterCellPro
   [YieldBreakdownColumnId.Apy]: ({ columnId, maxBoostTotal, total }: FooterCellProps) => (
     <TableCell key={columnId} sx={{ paddingInline: Spacing.md, paddingBlock: Spacing.sm, textAlign: 'right' }}>
       <Typography variant="tableCellMBold">{formatNumber(total, 'percent.rate')}</Typography>
-      {maybe(maxBoostTotal, x => (
+      {maxBoostTotal && maxBoostTotal != total && (
```
0xAlunara, line 31: Technically the number can't be negative but adding `> 0` is longer 🤡

mimaklas (approved):

apps/main/src/dex/features/pool-information/components/yield-breakdown/FooterRow.tsx
0xPearce: maxBoostTotal can technically be 0, and React will return it

```suggestion
      {maxBoostTotal !== 0 && maxBoostTotal != total && (
```
or like you suggested:

```suggestion
      {maxBoostTotal > 0 && maxBoostTotal != total && (
```

0xtutti (approved):

apps/main/src/dex/features/pool-information/components/yield-breakdown/FooterRow.tsx
0xLokiB: Indeed, using && with numbers is tricky, but the options that @0xPearce gave aren't null-proof
```suggestion
      {!!maxBoostTotal && maxBoostTotal != total && (
```

apps/main/src/dex/features/pool-information/hooks/useYieldBreakdown.tsx
```diff
@@ -148,7 +148,7 @@ export const useYieldBreakdown = ({
   const total = useMemo(() => sum(rows.map(row => row.apy)), [rows])
 
   return {
-    maxBoostTotal: maybe(crvApyRange, ({ unboostedApy, maximumApy }) => total - unboostedApy + maximumApy),
+    maxBoostTotal: maybe(crvApyRange, ({ unboostedApy, maximumApy }) => total - unboostedApy + maximumApy) ?? 0,
```
0xLokiB, line 151: Should we really default boost to zero just because `crvApyRange` is `null`?

apps/main/src/dex/features/pool-information/components/yield-breakdown/FooterRow.tsx
0xPearce: > maxBoostTotal: number

`maxBoostTotal` shouldn't be null

apps/main/src/dex/features/pool-information/hooks/useYieldBreakdown.tsx
0xAlunara: Yeah, see PR description for the reasoning. Brings in more in line with `total` as well which can't be `undefined`

apps/main/src/dex/features/pool-information/components/yield-breakdown/FooterRow.tsx
0xAlunara: Damnit I keep falling for that `0` being rendered

PR #3031, opened 2026-08-19
feat: retrieve user positions by chain

0xLokiB:
- update datatable to show data, even in loading/error states.
  - during loading a single skeleton row is displayed
- replace user position endpoints to separate calls
  - the backend does real onchain calls, and some chains can be slow
- update calls to include a new list of mint/lend market chains
  - extract user position logic to a new hook `useUserLlamaPositions`
  - update `useLlamaMarkets` to use the more up-to-date helper `combineQueryState`
- [RPC test run](https://github.com/curvefi/curve-frontend/actions/runs/32272264922/job/96131303033)


0xtutti (approved):

packages/prices-api/src/llamalend/index.ts
```diff
@@ -70,9 +79,20 @@ export async function getAllUserMarkets(
   return Schema.getAllUserMarketsResponse.parse(response)
 }
 
-export async function getUserMarkets(userAddr: string, chain: Chain, options?: Options) {
+export async function getUserMarkets(
+  userAddr: string,
+  chain: Chain,
+  {
+    page = USER_MARKETS_FIRST_PAGE,
+    per_page = USER_MARKETS_DEFAULT_PER_PAGE,
```
0xPearce, line 87: should we paginate lending supply positions as well?

packages/prices-api/src/index.ts
```diff
@@ -42,6 +42,9 @@ export const chains = [
 
 export type Chain = (typeof chains)[number]
 
+export const LEND_CHAINS = ['ethereum', 'arbitrum', 'optimism', 'fraxtal', 'sonic'] as const satisfies Chain[]
```
0xPearce, line 45: Isn't this a duplicate? can't we reuse the same logic as the chain switcher for example?

0xPearce (approved):
Nice!

packages/prices-api/src/llamalend/index.ts
0xLokiB: Yeah apparently I forgot those. I'll do it in a separate PR.

packages/prices-api/src/index.ts
0xLokiB: The supported lend networks are different than the supported lend networks for the prices API :thinking: 
But maybe they are the same. I'll investigate in the PR I'll need to create for your previous comment.

packages/prices-api/src/llamalend/index.ts
0xLokiB: - #3050

PR #3033, opened 2026-08-20
fix: validate zapv2 calldata size

0xPearce:
- validate ZapV2 router calldata against the 9,400 byte limit
- block create loan, borrow more, and repay when calldata is too large
- add test coverage 

Screenshot 2026-08-20 at 10 32 25


packages/curve-ui-kit/src/entities/router-api/router-api.utils.ts
```diff
@@ -10,6 +10,19 @@ import type { RouteProvider } from '@primitives/router.utils'
...
+ * revert. LlamaLend.js also adds roughly 200 bytes of metadata, leaving 9,400 bytes safely available to routers.
+ */
+const MAX_ZAP_V2_ROUTER_CALLDATA_BYTES = 9_400
+
+/**
+ * Checks the ZapV2 router calldata limit. Hex calldata starts with `0x` and uses two hex characters per byte, so the
+ * prefix is removed before dividing the remaining character count by two.
+ */
+export const isZapV2RouterCalldataTooLarge = (calldata: string | undefined) =>
+  !!calldata && (calldata.length - 2) / 2 > MAX_ZAP_V2_ROUTER_CALLDATA_BYTES
```
0xLokiB, line 24: If we validate `!!calldata`, should the function name include 'TooLarge'? How about `isRouterCalldataValid`?

0xLokiB (approved):

0xtutti (approved):

PR #3034, opened 2026-08-20
test: re-enable router integration tests

0xPearce:
- re-enable router integration tests

### DIFF
--- a/apps/router-api/test/integration/routes.test.ts
+++ b/apps/router-api/test/integration/routes.test.ts
@@ -234,8 +234,7 @@ const failureCases: Record<string, FailureCase> = {
   },
 }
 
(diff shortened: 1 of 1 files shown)


0xLokiB (approved):
🙌

0xAlunara (approved):
Not sure if anything about the underlying failures have been fixed yet, but we'll see. I also still believe they're not actually unit tests and shouldn't be blocking a production deployment 😄

0xLokiB:
> Not sure if anything about the underlying failures have been fixed yet, but we'll see. I also still believe they're not actually unit tests and shouldn't be blocking a production deployment 😄

They are indeed not unit tests, but they also don't need cypress and it doesn't seem worth it to create a separate integration suite

PR #3036, opened 2026-08-20
fix: connect buttons flashing on scroll

0xLokiB:
- when the wallet is disconnected, the connect wallet buttons were flashing during scroll of the market page
- that was caused by the router detecting a route change and re-rendering the whole wallet context
- this PR changes the logic to overpass the router


0xLokiB:
superseded by #3048

PR #3037, opened 2026-08-20
fix: max debt loading forever without wallet

0xLokiB:
- when the wallet is disconnected, the marketId is not set, so the max debt stays loading forever
- instead, we should use the standard query combiners to create the query results


0xtutti (approved):

0xAlunara (approved):
`combineQueries` such a relief we don't need to destructure and re-combine all those individual queries any more 😌

PR #3039, opened 2026-08-20
refactor: rename curve-ui-kit to evm-ui

0xLokiB:
- rename curve-ui-kit folder to evm-ui
- note that storybook deployment fails because vercel configuration needs to be updated
- depends on #3038


apps/main/src/dao/components/PageProposal/ProposalHeader/index.tsx
```diff
@@ -1,4 +1,4 @@
-import { Chain } from 'curve-ui-kit/src/utils/network'
+import { Chain } from '@ui-kit/utils/network'
```
0xPearce: multiple lint warnings

0xAlunara (approved):

0xPearce (approved):

PR #3048, opened 2026-08-21
fix: root layout

0xLokiB:
- move the root layout to the actual root of the application so it doesn't get remounted on route changes
  - this fixes the connect button flashing whenever we change routes
  - also fixes #3036
- update error boundary tests
  - update sentry filters to make sure user reports are always sent and ignore the issue filters
  - add test to verify the error boundary outside of the wagmi provider
- avoid router links in the error page when rendered outside the router context


apps/main/src/routes/RootLayout.tsx
```diff
@@ -44,64 +44,66 @@ const useBreadcrumbs = (pathname: string, { origin, search } = window.location)
...
-    </WagmiProvider>
-  ) : (
-    <Loading />
+  return (
+    <>
+      {backendMaintenance.isMaintenanceMode ? (
+        <MaintenancePage />
+      ) : (
+        <CurveProvider app={currentApp} network={network} onChainUnavailable={onChainUnavailable} hydrate={hydrate}>
+          {network && networks ? (
```
0xAlunara, line 70: It's probably just a type-only SNAFU, but `NetworkAwareLayout` is already only shown by `WagmiConfigProvider` when `networks` is true, which makes this `networks` check a bit redundant.

It's also a bit awkward that not only do we call `useNetworksQuery` twice (not too important though), but `WagmiConfigProvider` is now network aware too despite living outside of `NetworkAwareLayout` as the name implies.

In addition, `<MaintenancePage>` and `useMaintenance` aren't required to be network aware and could maybe be moved up to the root layer.

It's all non blocking and the PR fixes the flashing button bug, but it's maybe worth considering giving this file another thought later on.

0xAlunara (approved):

0xtutti (approved):

apps/main/src/routes/RootLayout.tsx
0xLokiB: - NetworkAwareLayout is already only shown by WagmiConfigProvider when networks is true

The name is supposed to imply that we know _which_ network is selected. The wagmi provider in principle only has a list of networks. Do you have a better name perhaps?

- we call useNetworksQuery twice

Yes but that should be OK being a query. I'd like to get rid of the double call but I couldn't find a much better solution to pass the prop


- WagmiConfigProvider is now network aware too 

It used to be already, but it was remounted with every route change (which was causing the issue)

- <MaintenancePage> and useMaintenance aren't required to be network aware

True, but I chose to place it there because we pass the data to GlobalLayout

PR #3061, opened 2026-08-21
feat: add pool historical base rate chart

0xAlunara:
Adds a new historical base rate chart to the pool page if it's supported by the prices API.

image

### DIFF
--- a/apps/main/src/dex/features/PoolHistoricalBaseRateChart.tsx
+++ b/apps/main/src/dex/features/PoolHistoricalBaseRateChart.tsx
@@ -28,7 +28,7 @@ import {
 import { Metric } from '@ui-kit/shared/ui/Metric'
 import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
 import { mapQuery } from '@ui-kit/types/util'
-import { formatNumber, TIME_OPTION_MS } from '@ui-kit/utils'
+import { decimal, formatNumber, TIME_OPTION_MS } from '@ui-kit/utils'
 
 const { Height, Spacing } = SizesAndSpaces
 
@@ -152,7 +152,7 @@ export const PoolHistoricalBaseRateChart = ({
             series={series}
             visibleSeries={visibleSeries}
             xTickFormatter={value => formatDate(value)}
-            yTickFormatter={value => formatNumber(+value, 'percent.value')}
+            yTickFormatter={value => formatNumber(decimal(value), 'percent.value')}
             yPaddingRatio={0.05}
             renderTooltip={({ datum, visibleSeries: activeSeries }) => (
               <ChartTooltipShell title={formatDate(datum.timestamp, 'long')}>
--- a/apps/main/src/dex/features/PoolHistoricalBaseRateChart.tsx
+++ b/apps/main/src/dex/features/PoolHistoricalBaseRateChart.tsx
@@ -1,0 +1,178 @@
+import { useMemo, useState } from 'react'
+import type { Address } from 'viem'
+import { usePoolSnapshots } from '@/dex/entities/pool-snapshots.query'
+import { aprToPoolApy } from '@/dex/features/pool-list/cells/utils'
+import { usePoolsPricesApi } from '@/dex/queries/pools-prices-api.query'
+import type { Chain } from '@curvefi/prices-api'
+import { formatDate } from '@legacy-ui/utils'
+import Card from '@mui/material/Card'
+import CardContent from '@mui/material/CardContent'
+import CardHeader from '@mui/material/CardHeader'
+import Stack from '@mui/material/Stack'
+import { useTheme } from '@mui/material/styles'
+import { maybe } from '@primitives/objects.utils'
+import { t } from '@ui-kit/lib/i18n'
+import { type TimeOption, timeOptions } from '@ui-kit/lib/model/query/time-option-validation'
+import {
+  CHART_LINE_DASH_PATTERNS,
+  ChartFooter,
+  ChartStateWrapper,
+  ChartTooltipSeriesGroup,
+  ChartTooltipSeriesRow,
+  ChartTooltipShell,
+  EChartsLineChart,
+  type LegendItem,
+  type LineSeriesConfig,
+  SelectTimeOption,
+} from '@ui-kit/shared/ui/Chart'
+import { Metric } from '@ui-kit/shared/ui/Metric'
+import { SizesAndSpaces } from '@ui-kit/themes/design/1_sizes_spaces'
+import { mapQuery } from '@ui-kit/types/util'
+import { formatNumber, TIME_OPTION_MS } from '@ui-kit/utils'
+
+const { Height, Spacing } = SizesAndSpaces
+
+const METRIC_CATEGORY = 'dex.poolInformation'
+
+type BaseRateSeriesKey = 'dailyBaseApy' | 'weeklyBaseApy'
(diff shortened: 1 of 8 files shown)


apps/main/src/dex/features/PoolHistoricalBaseRateChart.tsx
```diff
@@ -0,0 +1,178 @@
...
+            options={timeOptions}
+            activeOption={timeOption}
+            setActiveOption={setTimeOption}
+            isLoading={snapshots.isLoading}
+          />
+        }
+      />
+      <CardContent component={Stack} sx={{ gap: Spacing.md }}>
+        <Stack
+          sx={{
```
0xAlunara, line 123: There's a ticket already to create a proper card metric grid component

apps/main/src/dex/features/PoolHistoricalBaseRateChart.tsx
```diff
@@ -0,0 +1,178 @@
...
+  dailyBaseApy: number | undefined
+  weeklyBaseApy: number | undefined
+}
+
+const SERIES_CONFIG = [
+  { key: 'dailyBaseApy', label: t`Daily Base APY` },
+  { key: 'weeklyBaseApy', label: t`Weekly Base APY`, dash: CHART_LINE_DASH_PATTERNS.tight },
+] as const
+
+export const PoolHistoricalBaseRateChart = ({
```
0xAlunara, line 49: This component is heavily based on `MarketHistoricalRatesChart.tsx`. So any review comment here I'll probably also have to apply there.

apps/main/src/dex/features/PoolHistoricalBaseRateChart.tsx
```diff
@@ -0,0 +1,178 @@
...
+}) => {
+  const [timeOption, setTimeOption] = useState<TimeOption>('1M')
+  const [visibleSeries, setVisibleSeries] = useState<BaseRateSeriesKey[]>(() => SERIES_CONFIG.map(({ key }) => key))
+
+  const chain = blockchainId as Chain
+  const [end] = useState(() => Math.floor(Date.now() / 1000))
+  const snapshots = usePoolSnapshots({
+    chain,
+    poolAddress,
+    start: end - TIME_OPTION_MS[timeOption] / 1000,
```
0xAlunara, line 64: There's a 300 snapshot limit enforced by the API. With 365 days per year it means the yearly time limit won't show an actual full year worth of data (more like 10 months). I need to implement proper snapshot pagination for that, but want to do that in a follow up PR.

apps/main/src/dex/features/PoolHistoricalBaseRateChart.tsx
0xLokiB: Isn't there some code we could share, then?

apps/main/src/dex/features/PoolHistoricalBaseRateChart.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const ratePoints = mapQuery(snapshots, snapshots =>
+    snapshots.toReversed().map(snapshot => ({
+      timestamp: snapshot.timestamp,
+      dailyBaseApy: maybe(snapshot.baseDailyApr, apr => aprToPoolApy(apr * 100)) ?? undefined,
+      weeklyBaseApy: maybe(snapshot.baseWeeklyApr, apr => aprToPoolApy(apr * 100)) ?? undefined,
+    })),
+  )
+
+  // Current metric values are based on pool list data to avoid mismatches, and is also updated more frequently than snapshots
+  const currentPool = mapQuery(usePoolsPricesApi({ blockchainId: chain }), pools => pools[poolAddress.toLowerCase()])
```
0xLokiB: toLowerCase scattered across components is a smell imo

apps/main/src/dex/features/PoolHistoricalBaseRateChart.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+  const {
+    design: { Color },
+  } = useTheme()
+
+  const seriesColors: Record<BaseRateSeriesKey, string> = useMemo(
+    () => ({ dailyBaseApy: Color.Primary[500], weeklyBaseApy: Color.Secondary[500] }),
```
0xLokiB, line 85: Shouldn't we have more specific (semantic) constants than just `Color`?

apps/main/src/dex/features/PoolHistoricalBaseRateChart.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+  const legendSets: LegendItem[] = useMemo(
+    () =>
+      SERIES_CONFIG.map(({ key, label, ...config }) => ({
+        label,
+        line: { lineStroke: seriesColors[key], ...config },
+        toggled: visibleSeries.includes(key),
+        onToggle: () =>
+          setVisibleSeries(previous =>
+            previous.includes(key) ? previous.filter(seriesKey => seriesKey !== key) : [...previous, key],
```
0xLokiB, line 102: This line is repeated in several places, is it worth a helper?

apps/main/src/dex/features/PoolHistoricalBaseRateChart.tsx
```diff
@@ -0,0 +1,178 @@
...
+          errorMessage={t`Unable to fetch historical base rate data.`}
+        >
+          <EChartsLineChart<BaseRateChartPoint, BaseRateSeriesKey, 'timestamp'>
+            data={ratePoints.data ?? []}
+            height={Height.shortChart}
+            xKey="timestamp"
+            series={series}
+            visibleSeries={visibleSeries}
+            xTickFormatter={value => formatDate(value)}
+            yTickFormatter={value => formatNumber(+value, 'percent.value')}
```
0xLokiB: formatNumber is supposed to support `Decimal`, why do we need `+`?

0xAlunara: Because the type of `value` given by the ECharts wrapper is `string | number`. I've wrapped both usages of market and pool in ` decimal` now.

apps/main/src/dex/features/PoolHistoricalBaseRateChart.tsx
0xAlunara: I'll create a proper hook for fetching a singular pool via `usePoolsPricesApi`. We have a few other usages like this that slightly deviate.

apps/main/src/dex/features/PoolHistoricalBaseRateChart.tsx
0xAlunara: Well that's what `ChartStateWrapper` and `EChartsLineChart` are for. So far we only have two cards like this, not enough yet to justify generalizing further

apps/main/src/dex/features/PoolHistoricalBaseRateChart.tsx
0xAlunara: I'll make a ticket for this; it seems we're using non-semantic chart line colors all over the place.

apps/main/src/dex/features/PoolHistoricalBaseRateChart.tsx
0xAlunara: Will create a ticket as it would expand the scope beyond just this card

0xLokiB (approved):

0xtutti (approved):

PR #3064, opened 2026-08-22
fix: add different test seeds per iteration

0xLokiB:
- starting parallel tests do not add much value if they all use the same seed
- therefore, also use the browser and the iteration to the test seed

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -101,7 +101,7 @@ jobs:
         env:
           ELECTRON_ENABLE_LOGGING: true # send console logs to stdout in electron browser
           NODE_ENV: test
-          TEST_SEED: ${{ github.run_id }}-${{ github.run_attempt }}-${{ matrix.count }}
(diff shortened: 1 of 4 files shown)


0xAlunara (commented):
Why wouldn't you want the same seed per browser? By giving each browser the same seed you can more easily rule out if an issue is browser related or not. With a diff seed per browser you don't know if it was either the browser or the test that failed, since each browser is running a different configuration of tests. If the same seed is used by all browser and the same test with the same seed only fails once out of 6 times you're more sure that it's just a fluke or a browser quirk.

0xLokiB:
> Why wouldn't you want the same seed per browser? By giving each browser the same seed you can more easily rule out if an issue is browser related or not

Sure, we can start with that. I also like covering more cases in each run. But for now I'll add only the iteration to the seed

0xAlunara (approved):

0xtutti (approved):

PR #3068, opened 2026-08-22
chore: form skeleton height improvements

0xLokiB:
- update the height of the form skeleton to match the simplest forms (a single input and a submit button)
- make sure the tab title and the token input match the expected sizes
- add component test

### DIFF
--- a/tests/cypress/component/form-skeleton.cy.tsx
+++ b/tests/cypress/component/form-skeleton.cy.tsx
@@ -1,0 +1,82 @@
+import { DepositForm } from '@/llamalend/features/supply/components/DepositForm'
+import { MockLoanTestWrapper } from '@cy/support/helpers/llamalend/MockLoanTestWrapper'
+import { createDepositScenario } from '@cy/support/helpers/llamalend/supply/supply-test-scenarios.helpers'
+import {
+  llamaNetworks,
+  resetLlamaTestContext,
+  setGasInfo,
+  setLlamaApi,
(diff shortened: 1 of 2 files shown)


tests/cypress/component/form-skeleton.cy.tsx
```diff
@@ -0,0 +1,82 @@
...
+    setGasInfo({ chainId: 1, networks: llamaNetworks })
+
+    cy.mount(
+      <MockLoanTestWrapper llamaApi={llamaApi} market={market}>
+        <Stack>
+          <Stack data-testid="skeleton-form">
+            <FormSkeleton />
+          </Stack>
+          <Stack data-testid="actual-form">
+            <FormTabs params={{}} menu={depositMenu} />
```
0xAlunara, line 60: Not blocking, but I think this test will be very brittle. The generic form skeleton's height is now 'magically' linked to the highly specific use case of depositing on llamalend. One change in desired behavior of either components will result in a failing test. Why not actually recreate the minimal form the skeleton is based on, independent of llamalend?

0xAlunara (approved):

tests/cypress/component/form-skeleton.cy.tsx
0xLokiB: I want to know if we add something to the llamalend forms, that might need to be added to the skeleton, too. But indeed, this is too specific

0xtutti (approved):

PR #3069, opened 2026-08-22
refactor: remove legacy query types

0xLokiB:
- Remove legacy helpers `createQueryStatus`, `QUERY_SUCCESS_BASE`, `RESOLVED_QUERY_RESULT`, `combineQueriesMeta`, `PartialQueryResult`. Use `combineQueryState` instead.
- Update `combineQueriesToObject` to use `Query<Record<K, TData>>` instead.

### DIFF
--- a/apps/main/src/dex/components/PagePool/Withdraw/components/FormWithdraw.tsx
+++ b/apps/main/src/dex/components/PagePool/Withdraw/components/FormWithdraw.tsx
@@ -1,8 +1,7 @@
 import lodash from 'lodash'
 import { ReactNode, useCallback, useEffect, useMemo, useRef, useState } from 'react'
-import { styled, css } from 'styled-components'
-import { useConnection, type Config } from 'wagmi'
-import { useConfig } from 'wagmi'
+import { css, styled } from 'styled-components'
+import { type Config, useConfig, useConnection } from 'wagmi'
 import { AlertFormError } from '@/dex/components/AlertFormError'
 import { AlertSlippage } from '@/dex/components/AlertSlippage'
 import { DetailInfoEstGas } from '@/dex/components/DetailInfoEstGas'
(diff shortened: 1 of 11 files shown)


apps/main/src/dex/components/PagePool/Withdraw/components/FormWithdraw.tsx
```diff
@@ -1,8 +1,7 @@
 import lodash from 'lodash'
 import { ReactNode, useCallback, useEffect, useMemo, useRef, useState } from 'react'
-import { styled, css } from 'styled-components'
-import { useConnection, type Config } from 'wagmi'
-import { useConfig } from 'wagmi'
+import { css, styled } from 'styled-components'
```
0xAlunara, line 3: All unrelated to the typing but lgtm 😄

apps/main/src/llamalend/hooks/useHealthQueries.ts
```diff
@@ -36,9 +35,9 @@ type HealthQueryResults = QueryResultsArray<readonly HealthQueryOptions[]>
  *   In or below soft-liq healthFull = healthNotFull, above soft-liq healthFull > healthNotFull
  *   When healthNotFull is below 0, the user is in soft-liq and we should return the corresponding metric.
  */
-const combineHealth = ([healthFull, healthNotFull]: HealthQueryResults) => ({
-  data: maybes([healthFull.data, healthNotFull.data], (full, notFull) => (+notFull < 0 ? notFull : full)),
-  ...combineQueriesMeta([healthFull, healthNotFull]),
+const combineHealth = ([healthFull, healthNotFull]: Query<Decimal>[]) => ({
```
0xAlunara: Could've been made `[Query<Decimal>, Query<Decimal>]` 🤓

packages/evm-ui/src/lib/queries/combine.ts
```diff
@@ -42,24 +41,18 @@ export const useCombinedQueries = <const TQueries extends Queries, TResult>(
...
-) => ({
-  // Using flatMap instead of map + filter(Boolean), because it's not correctly erasing | undefined from the Record value type
-  data: Object.fromEntries(results.flatMap(({ data }, index) => (data == null ? [] : [[keys[index], data]]))),
-  ...combineQueriesMeta(results),
-})
+/** Combines multiple queries into a query whose data is keyed by the matching key list. */
+export const combineQueriesToObject = <TData, K extends string = string>(results: Query<TData>[], keys: readonly K[]) =>
+  q<Record<K, TData>>({
+    data: (results.some(({ data }) => data != null)
+      ? Object.fromEntries(
```
0xAlunara: Maybe it doesn't matter, but perhaps the better type inference of `fromEntries` might simplify typing?

packages/evm-ui/src/lib/queries/combine.ts
```diff
@@ -42,24 +41,18 @@ export const useCombinedQueries = <const TQueries extends Queries, TResult>(
...
+          notFalsy(...results.map(({ data }, index) => data != null && ([keys[index], data] as const))),
+        )
+      : results.some(({ data }) => data === null)
```
0xAlunara: I'm not sure why this ternary is necessary. This differs from the original implementation. When all items are not nullish we default to `null` when there's at least one `null`. But isn't `null` a valid value? And when there's at least one non nullish value, we no longer return null but we filter it out from the resulting data set.

**Old comment**
1. When everything is nullish (undefined or null), the new data set might contain key-value pairs that might be `null`.
2. But when there is one data element not nullish, we no longer include key-value pairs that might be `null`, suddenly it's not a valid value anymore?

In other words, `{ 'foo': null, 'bar': undefined }` would be combined into `{ 'foo': null }`, but `{ 'foo': null, 'bar': 5 }` becomes `{ 'bar': 5 }`. Suddenly `'foo': null` is not inside anymore?

Maybe I'm misinterpreting what the function 100% does now but otoh it might be overly complex?

apps/main/src/dex/components/PagePool/Withdraw/components/FormWithdraw.tsx
0xLokiB: :see_no_evil:

packages/evm-ui/src/lib/queries/combine.ts
0xLokiB: yeah ofc I should've seen that :see_no_evil:

packages/evm-ui/src/lib/queries/combine.ts
0xLokiB: What I meant to write is: if any values are set, then we return an object.
Otherwise (i.e. all values are null or undefined): If any value is set to null, then we return null. Otherwise, undefined. 
But we might not need to distinguish between them at all

0xtutti (approved):

0xAlunara (approved):

PR #3076, opened 2026-08-24
fix: remove min chart length of 2 before fuzzy search kicks in

0xAlunara:
Fixes fuzzy search not kicking in when you're searching with just 1 character. This was a deliberate choice made by previous devs since the genesis of Curve front-end, but I don't see a need for it anymore.

**Before**
image

**After**
image


0xtutti (approved):

michwill (approved):

michwill:
Screenshots look epic!

PR #3080, opened 2026-08-24
chore: not falsy lint rule

0xAlunara:
Encourages the use of `notFalsy` over singleton stuff array ternaries like `x != null ? [x] : []` and `.filter(Boolean)`.

### DIFF
--- a/.eslint/use-not-falsy-pattern.rule.mjs
+++ b/.eslint/use-not-falsy-pattern.rule.mjs
@@ -1,0 +1,316 @@
+import ts from 'typescript'
+
+const EXPRESSION_WRAPPERS = new Set([
+  'ChainExpression',
+  'ParenthesizedExpression',
+  'TSAsExpression',
+  'TSNonNullExpression',
+  'TSTypeAssertion',
+])
+
+const unwrap = node => (EXPRESSION_WRAPPERS.has(node?.type) ? unwrap(node.expression) : node)
+const propertyName = node =>
+  node.property.type === 'Identifier' && !node.computed
+    ? node.property.name
+    : node.computed && node.property.type === 'Literal' && typeof node.property.value === 'string'
+      ? node.property.value
+      : null
+const unionMembers = type => (type.isUnion() ? type.types : [type])
+const isBoundedArrayLiteral = node =>
+  node.type === 'ArrayExpression' &&
+  node.elements.length > 0 &&
+  node.elements.every(element => element && element.type !== 'SpreadElement')
+const isFilterLength = node => {
+  while (EXPRESSION_WRAPPERS.has(node.parent?.type)) node = node.parent
+  return (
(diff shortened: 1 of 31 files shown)


.eslint/use-not-falsy-pattern.rule.mjs
```diff
@@ -0,0 +1,316 @@
+import ts from 'typescript'
```
0xAlunara, line 1: Admittedly this entire rule has been gippety generated based on wishes and example use cases. Shrunk it down by yelling at it a few times that it was too complex and didn't need features such as auto-fix.

apps/main/src/dex/features/manage-pool/components/RefuelPricesChart.tsx
```diff
@@ -48,7 +48,7 @@ const formatPrice = (value: number | null | undefined) =>
   formatNumber(value, { abbreviate: false, decimals: 6, fallback: '-' })
 
 // Helper functions and filters that don't type check nicely when inlined
-const toExportPoint = (time: number, value: number | null) => (value == null ? [] : [{ time, value }])
+const toExportPoint = (time: number, value: number | null) => notFalsy(value != null && { time, value })
```
0xAlunara, line 51: I suppose `0` is a valid result, but the linter yells at us to use `notFalsy` because of the empty array lol, so we need to use it with a `!= null` check.

packages/external-rewards/src/index.ts
```diff
@@ -28,6 +28,7 @@ export type RewardsAction = 'supply' | 'borrow' | 'lp'
 
 const parsedCampaignsJsons = campaignsJsons as Record<string, Campaign>
 
+// eslint-disable-next-line local/use-not-falsy-pattern
```
0xAlunara, line 31: Easier than excluding this project in the lint config or adding a dependency to ui-kit (evm-kit now?)

mimaklas (approved):

.eslint/use-not-falsy-pattern.rule.mjs
```diff
@@ -0,0 +1,316 @@
...
+const nullishMask = type =>
+  unionMembers(type).reduce(
+    (mask, member) =>
+      mask |
+      (member.flags & ts.TypeFlags.Null ? 1 : 0) |
+      (member.flags & (ts.TypeFlags.Undefined | ts.TypeFlags.Void) ? 2 : 0),
+    0,
+  )
+
+const isObjectsUtilsImport = source =>
```
0xLokiB, line 91: Could've ignored the rule in that file

.eslint/use-not-falsy-pattern.rule.mjs
0xLokiB: I think it's too complex but whatever if it works lol. It doesn't need maintenance

0xLokiB (approved):

0xtutti (approved):

PR #3081, opened 2026-08-24
fix: re-enable sorting for all lite network pool columns

0xAlunara:
A bit difficult to test perhaps as there's no pools with CRV APY, and the only lite networks with APYs is Monad.

### DIFF
--- a/apps/main/src/dex/features/pool-list/columns/column.definitions.tsx
+++ b/apps/main/src/dex/features/pool-list/columns/column.definitions.tsx
@@ -9,6 +9,7 @@ import { PointsCell } from '../cells/PointsCell'
 import { PoolTitleCell } from '../cells/PoolTitleCell'
 import { RewardsApyCell } from '../cells/RewardsApyCell'
 import { UsdCell } from '../cells/UsdCell'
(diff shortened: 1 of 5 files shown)


apps/main/src/dex/features/pool-list/columns/column.definitions.tsx
```diff
@@ -53,85 +51,79 @@ const accessor = (
...
-  ] satisfies PoolColumn[]
-
-export const POOL_COLUMNS = createPoolColumns({ isLite: false })
-export const LITE_POOL_COLUMNS = createPoolColumns({ isLite: true })
+export const POOL_COLUMNS = [
+  accessor(PoolColumnId.PoolName, 'name', {
+    cell: PoolTitleCell,
+    meta: { tooltip: createTooltip(PoolColumnId.PoolName, <PoolHeaderTooltipContent />) },
+  }),
+  accessor(PoolColumnId.NetApy, getNetApy, {
```
0xAlunara, line 59: These accessors aren't being used for sorting normal networks, because `manualSorting: !isLite`. But they still need an accessor defined so they remain (server-side) sortable by TanStack. So this is all good.

0xLokiB (approved):

mimaklas (approved):

PR #3082, opened 2026-08-24
feat: support multiple max receive providers

0xLokiB:
- Refactor max-receive queries to evaluate one router at a time.
- Add multi-router max selection for create-loan and borrow-more flows.
- Return the winning router with the max borrow result so the form can use the same route choice.
- Update max-button behavior to select the matching router when applying max borrow.
- Updates llamalend.js https://github.com/curvefi/curve-llamalend.js/pull/141

### DIFF
--- a/packages/evm-ui/src/lib/model/query/factory.ts
+++ b/packages/evm-ui/src/lib/model/query/factory.ts
@@ -171,13 +171,7 @@ export function queryFactory<
         ...options,
         staleTime: 0,
       }),
-    useQuery: (params: TParams, condition?: boolean) => {
-      const { enabled, ...options } = getQueryOptions(params, condition)
-      const query = useQuery({ enabled, ...options })
-      const validation = enabled ? {} : validate(validationSuite, params)
-      // eslint-disable-next-line @eslint-react/purity -- Preserve the query result object's getters and generic inference.
-      return Object.assign(query, { enabled, validation })
-    },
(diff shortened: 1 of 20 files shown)


packages/evm-ui/src/lib/model/query/factory.ts
```diff
@@ -171,13 +171,7 @@ export function queryFactory<
...
         staleTime: 0,
       }),
-    useQuery: (params: TParams, condition?: boolean) => {
-      const { enabled, ...options } = getQueryOptions(params, condition)
-      const query = useQuery({ enabled, ...options })
-      const validation = enabled ? {} : validate(validationSuite, params)
-      // eslint-disable-next-line @eslint-react/purity -- Preserve the query result object's getters and generic inference.
-      return Object.assign(query, { enabled, validation })
-    },
+    useQuery: (params: TParams, condition?: boolean) => useQuery(getQueryOptions(params, condition)),
```
0xAlunara, line 174: Huh why was this ever so complex to begin with? 🤔

0xAlunara (approved):

packages/evm-ui/src/lib/model/query/factory.ts
0xLokiB: `validation` was added for debugging purposes and `enabled` was added because we didn't know about `isEnabled`

0xtutti (approved):

PR #3083, opened 2026-08-24
fix: abbreviate y axis for historical rate charts

0xAlunara:
This obviously doesn't fix the bad data itself, but at least it doesn't mess up the rendering of the page

**Before**
image

**After**
image


mimaklas (approved):

0xLokiB (approved):
Image

PR #3086, opened 2026-08-25
fix: bridge alert overflow

0xLokiB:
- Fixes the overflow in the bridge alert
image

### DIFF
--- a/apps/main/src/bridge/features/bridge/components/BridgeInfoAlert.tsx
+++ b/apps/main/src/bridge/features/bridge/components/BridgeInfoAlert.tsx
@@ -29,7 +29,14 @@ export const BridgeInfoAlert = () => (
         {t`Bridges are not instant, and funds may take up to 15 minutes or more before arriving.`}
       </Typography>
 
-      <Stack direction="row" sx={{ justifyContent: 'space-between' }}>
+      <Stack
+        direction="row"
+        sx={{
+          justifyContent: 'space-between',
+          // MUI Button endIcon has a negative right margin, which creates a tiny horizontal scroll in Alert.
+          '& .MuiButton-endIcon': { marginRight: 0 },
+        }}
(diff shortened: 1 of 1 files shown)


apps/main/src/bridge/features/bridge/components/BridgeInfoAlert.tsx
```diff
@@ -29,7 +29,14 @@ export const BridgeInfoAlert = () => (
         {t`Bridges are not instant, and funds may take up to 15 minutes or more before arriving.`}
       </Typography>
 
-      <Stack direction="row" sx={{ justifyContent: 'space-between' }}>
+      <Stack
+        direction="row"
+        sx={{
+          justifyContent: 'space-between',
+          // MUI Button endIcon has a negative right margin, which creates a tiny horizontal scroll in Alert.
+          '& .MuiButton-endIcon': { marginRight: 0 },
```
0xAlunara, line 37: This solves the issue locally, but should we consider setting marginRight to 0 for all mui-button end icons? I know we have quite a few resets to 0 already for other component elements 🤔

0xAlunara (approved):

apps/main/src/bridge/features/bridge/components/BridgeInfoAlert.tsx
0xLokiB: If that's what the design system says @0xtutti

0xtutti (approved):

apps/main/src/bridge/features/bridge/components/BridgeInfoAlert.tsx
0xtutti: I'm unsure of the visual impact of this, but it could work indeed. Better to do such a fix globally than do it locally imho.

PR #3092, opened 2026-08-25
feat: new pool title tooltip

0xAlunara:
Adds a new tooltip to pool titles that give a better quick overview of the pool's tokens, contracts and badges.

This means that we no longer need the copy button in the pool title itself.

image

### DIFF
--- a/apps/main/src/dex/features/pool-list/cells/PoolTitleCell/LegacyPoolTitleCell.tsx
+++ b/apps/main/src/dex/features/pool-list/cells/PoolTitleCell/LegacyPoolTitleCell.tsx
@@ -42,7 +42,7 @@ export const LegacyPoolTitleCell = ({
         <Stack direction="column">
           <Stack direction="row" sx={{ alignItems: 'center', gap: Spacing.xs }}>
             <PoolAlertIcons poolAlert={poolAlert} tokenAlert={tokenAlert} />
-            <TableRowTitle id={pool.address} url={url} title={pool.name}></TableRowTitle>
+            <TableRowTitle url={url} title={pool.name} testId={pool.address} />
             <CopyIconButton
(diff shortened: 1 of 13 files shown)


apps/main/src/dex/features/pool-list/cells/PoolTitleCell/classifications.ts
```diff
@@ -0,0 +1,15 @@
+import type { PoolRow } from '../../types'
+
+type PoolType = NonNullable<PoolRow['poolType']>
+export type PoolClassification = 'stable' | 'volatile'
+
+export const poolTypeClassifications: Record<PoolType, PoolClassification> = {
```
0xAlunara, line 6: Simply extracted from `PoolBadges` so it can be re-used in `PoolTooltipContent`

apps/main/src/dex/features/pool-list/cells/PoolTitleCell/LegacyPoolTitleCell.tsx
```diff
@@ -41,7 +42,13 @@ export const LegacyPoolTitleCell = ({
         <Stack direction="column">
           <Stack direction="row" sx={{ alignItems: 'center', gap: Spacing.xs }}>
             <PoolAlertIcons poolAlert={poolAlert} tokenAlert={tokenAlert} />
-            <MarketTitle url={url} address={pool.address as Address} title={pool.name} />
+            <TableRowTitle url={url} title={pool.name} testId={pool.address} />
+            <CopyIconButton
```
0xAlunara, line 46: The `CopyIconButton` usage might look 'copied' twice now; here and in `MarketTitleCell`, but this is legacy functionality for the pool, so in the future it remains only for market rows.

0xtutti (approved):

0xLokiB (approved):

PR #3102, opened 2026-08-27
refactor: update tanstack table to v9

0xAlunara:
1. Updates `tanstack-query`, `tanstack-router` and `tanstack-table` (to v9) so that they all use the same underlying `tanstack-core` package.
2. Removed `getTableOptions` inside `data-table.utils.ts` in favor of registering an 'app table' with a common feature set and options. App tables are new in v9 and use a global set of features and options, kinda how we already used Tanstack Table. Technically it's possible for each table to have its own feature set and options for better tree-shaking, but it's a bit unpractical for the time being.
3. The `useTable` hook has been renamed to `useCurveTable` to more clearly denote it's our global Curve app table, which also takes a query as parameter rather than a direct row `data` array.

Because v9 has way better inferred typing, this PR also contains a lot of (generic) type removals in favor of implicitly inferred types. However, at the same time v9 requires the type of a table's feature set more often in its own type's generic parameters.

You'll see a lot of cases where we use have to import `CurveTableFeatures` and pass them as the first generic in a React Table type like `Column`, `CellContext`, `Header` etc. There's many types now requiring a features type as generic parameter, and to keep things modular and not overly typed I decided not to make Curve table specific type re-exports. So a big part of the diff is just importing the `CurveTableFeatures` type to pass it along a React Table generic parameter.

Finally, row navigation on click has also been moved a `meta` functionality on the table level, as that we we don't need to enforce a generic `CurveTableItem` everywhere, which simplifies typing quite a bit.


packages/evm-ui/src/shared/ui/DataTable/hooks/useCellSx.ts
```diff
@@ -0,0 +1,53 @@
...
+const emptyObject = {} satisfies SxProps<Theme>
+
+/**
+ * Creates the styles for the table cell, including handling sticky columns and collapse icon.
+ * @param column the tanstack column
+ * @param showCollapseIcon whether to show the collapse icon (for mobile last column)
+ * @param isSticky whether the column is sticky (first column on tablet)
+ * @returns an array with the cell sx and the wrapper sx (empty object if no wrapper needed)
+ */
+export function useCellSx<T extends RowData>({
```
0xAlunara: Moved out of `data-table.utils.ts`, figured it deserved its own hook file

packages/evm-ui/src/shared/ui/DataTable/categories.ts
```diff
@@ -0,0 +1,50 @@
...
+  disableStickyHeader?: boolean // can also be disabled by limited rows or table width overflow.
+  hideHeader?: boolean
+  increasingLength?: IncreasingLengthCategory
+  emptyStateSize?: NonNullable<EmptyStateCardProps['size']>
+  emptyStateRowSize?: EmptyStateRowSize
+}
+
+export type DataTableCategory = keyof typeof DATA_TABLE_CATEGORIES
+
+export const DATA_TABLE_CATEGORIES = {
```
0xLokiB, line 22: nice

packages/evm-ui/src/shared/ui/DataTable/data-table.utils.ts
```diff
@@ -40,133 +41,81 @@ export const CLICKABLE_IN_ROW_CLASS = 'clickable-in-row'
...
+  columnId: string,
+): (() => RowModel<TFeatures, TData>) => {
+  // This fixed feature set intentionally omits the type-only `tableMeta` slot, so the global generic augmentation wins.
+  const override = (table.options.meta as TableMeta<TFeatures, TData> | undefined)?.facetedRowModelFactory
+  return override ? override(table, columnId) : createFacetedRowModel<TFeatures, TData>()(table, columnId)
+}
 
-/** Required fields for the data in the table. */
-export type TableItem = { url?: string | null }
+/** Static feature set shared by every Curve DataTable. */
```
0xLokiB: I got that from the code, but why do we have a static set?

```suggestion
/** set of static features so that ....... */
```

packages/evm-ui/src/shared/ui/DataTable/data-table.utils.ts
```diff
@@ -40,133 +41,81 @@ export const CLICKABLE_IN_ROW_CLASS = 'clickable-in-row'
...
+  },
+})
 
-export type TanstackTable<T extends TableItem> = ReturnType<typeof useTable<T>>
+export type CurveTableFeatures = typeof features
 
-export type ColumnMeta = TanstackColumnMeta<TableItem, unknown>
+const options = tableOptions({
+  features,
+  getRowCanExpand: () => true, // Curve renders expanded content as sibling detail rows with its own expansion check, not TanStack subRows.
```
0xLokiB: again here, why don't we use subrows?

packages/evm-ui/src/shared/ui/DataTable/data-table.utils.ts
```diff
@@ -40,133 +41,81 @@ export const CLICKABLE_IN_ROW_CLASS = 'clickable-in-row'
...
-/**
- * In the figma design, the first and last columns seem to be aligned to the table title.
- * However, the normal padding causes them to be misaligned.
- */
-export const getExtraColumnPadding = <T>(column: Column<T, unknown>) => ({
-  ...(column.getIsFirstColumn() && { paddingInlineStart: Spacing.md }),
-  ...(column.getIsLastColumn() && { paddingInlineEnd: Spacing.md }),
+/** Query-aware wrapper around the Curve app table hook. */
+export const useCurveTable = <TData extends RowData>({
+  query,
```
0xLokiB:
```suggestion
  query: { data, isLoading, error },
```

packages/evm-ui/src/shared/ui/DataTable/data-table.utils.ts
```diff
@@ -40,133 +41,81 @@ export const CLICKABLE_IN_ROW_CLASS = 'clickable-in-row'
...
+export const getCellVariant = (variant?: TypographyVariantKey) => variant ?? 'tableCellMBold'
 
-export const getHiddenCount = <T>(table: Table<T>): number =>
-  table.getPreFilteredRowModel().rows.length - table.getFilteredRowModel().rows.length
+export const isSortedBy = <TFeatures extends TableFeatures, TData extends RowData>(
+  table: Pick<Table<TFeatures, TData>, 'getColumn'>,
+  columnId: string,
+) => {
+  const column = table.getColumn(columnId)
+  return column ? Boolean(column_getIsSorted(column)) : false
```
0xLokiB:
```suggestion
  return !!maybe(column, column_getIsSorted)
```

packages/evm-ui/src/shared/ui/DataTable/data-table.utils.ts
```diff
@@ -1,35 +1,36 @@
...
+  type RowModel,
+  type Table,
+  type TableFeatures,
+  type TableMeta,
+  type TableOptions,
 } from '@tanstack/react-table'
-import { RowData, type Table, TableOptions } from '@tanstack/table-core'
-import { EmptyStateCardProps } from '../EmptyStateCard'
-import { EmptyStateRowSize } from './EmptyStateRow'
+import { column_getIsSorted } from '@tanstack/react-table/static-functions'
```
0xLokiB, line 32: that is some unconventional naming convention

packages/evm-ui/src/shared/ui/DataTable/data-table.utils.ts
```diff
@@ -40,133 +41,81 @@ export const CLICKABLE_IN_ROW_CLASS = 'clickable-in-row'
...
-  // only pass the filtered model once loaded, it causes an error: https://github.com/TanStack/table/issues/5026
-  getFilteredRowModel: maybe(result, () => getFilteredRowModel<TData>()),
-  getFacetedRowModel: getFacetedRowModel<TData>(),
-  getFacetedUniqueValues: getFacetedUniqueValues<TData>(),
-  getFacetedMinMaxValues: getFacetedMinMaxValues<TData>(),
-  getExpandedRowModel: getExpandedRowModel<TData>(),
-  getPaginationRowModel: getPaginationRowModel<TData>(),
-  autoResetPageIndex: false, // autoreset causing stack too deep issues when receiving new data
-  maxMultiSortColCount: 3, // allow 3 columns to be sorted at once while holding shift
+export const getExtraColumnPadding = (columnId: string, visibleColumns: readonly { id: string }[]) => ({
```
0xLokiB: couldn't we use css for this?

packages/evm-ui/src/shared/ui/DataTable/DataTable.d.ts
```diff
@@ -1,19 +1,30 @@
...
+declare module '@tanstack/table-core' {
+  interface ColumnMeta<
+    in out TFeatures extends TableFeatures,
+    in out TData extends RowData,
+    TValue extends CellData = CellData,
+  > {
     type?: 'numeric' // aligns cell content to the right
     unit?: Unit // used when displaying the filter's serialized value
-    hidden?: boolean // todo: get rid of this property, use column visibility, it breaks e.g. column.getIsLastColumn()
+    hidden?: boolean // todo: get rid of this property; metadata and column visibility can diverge
```
0xLokiB, line 19: Do we have a task for this todo? It could lead to bugs

packages/evm-ui/src/shared/ui/DataTable/ExpandedPanelActions.tsx
```diff
@@ -10,16 +10,24 @@ import { DrawerItems } from '@evm-ui/shared/ui/SwipeableDrawer/DrawerItems'
...
 import IconButton from '@mui/material/IconButton'
 import Stack from '@mui/material/Stack'
 import { splitAt } from '@primitives/array.utils'
-import type { ExpandedPanelAction } from './data-table.utils'
 
 const { MaxHeight, Spacing } = SizesAndSpaces
 // number of button rendered before hiding the rest in a kebab menu
 const VISIBLE_ACTION_COUNT = 2
 
+type ExpandedPanelAction = Omit<ButtonProps, 'children' | 'component' | 'href' | 'ref'> & {
```
0xLokiB: nitpick: It's easier to read/refactor when we know which props are used i.e. only picking the used props

packages/evm-ui/src/shared/ui/DataTable/data-table.utils.ts
```diff
@@ -40,133 +41,81 @@ export const CLICKABLE_IN_ROW_CLASS = 'clickable-in-row'
...
-
-/**
- * In the figma design, the first and last columns seem to be aligned to the table title.
- * However, the normal padding causes them to be misaligned.
- */
-export const getExtraColumnPadding = <T>(column: Column<T, unknown>) => ({
-  ...(column.getIsFirstColumn() && { paddingInlineStart: Spacing.md }),
-  ...(column.getIsLastColumn() && { paddingInlineEnd: Spacing.md }),
+/** Query-aware wrapper around the Curve app table hook. */
+export const useCurveTable = <TData extends RowData>({
```
0xLokiB, line 100: This should have been in a separate PR IMO, this is much harder to review now it's mixing dependency updates with a large refactor.

apps/main/src/dex/features/pool-information/components/yield-breakdown/FooterRow.tsx
```diff
@@ -11,7 +12,7 @@ import { YieldBreakdownColumnId } from './columns/columns.enum'
 const { Spacing } = SizesAndSpaces
 
 type FooterRowProps = {
-  visibleColumns: Column<YieldBreakdownRow, unknown>[]
+  visibleColumns: Column<CurveTableFeatures, YieldBreakdownRow>[]
```
0xAlunara, line 15: You'll see a lot of cases where we use have to import `CurveTableFeatures` and pass them as the first generic in a React Table type like `Column`, `CellContext` etc. There's many types now requiring a features type as generic parameter, and to keep things modular and not overly typed I decided not to make Curve table specific type re-exports.

So a big part of the diff is just importing the `CurveTableFeatures` type to pass it along a React Table generic parameter.

apps/main/src/dex/features/pool-list/LegacyPoolsTable.tsx
```diff
@@ -46,24 +46,24 @@ export const LegacyPoolsTable = ({ network }: { network: NetworkConfig }) => {
   const [expanded, onExpandedChange] = useState<ExpandedState>({})
   const filterProps = { columnFiltersById, setColumnFilter }
 
-  const table = useTable({
+  const table = useCurveTable({
     columns: LEGACY_POOL_COLUMNS,
     query: q({ data, isLoading, error: null }),
+    meta: { getRowHref: ({ url }) => url },
```
0xAlunara, line 52: This is an easier solution for pure row navigation on click compared to enforcing a base `CurveTableItem` on all of our table rows. A lot less generic typing required this way.

packages/evm-ui/src/shared/ui/DataTable/categories.ts
```diff
@@ -0,0 +1,50 @@
+import type { IncreasingLengthCategory } from '@evm-ui/hooks/useIncreasingLength'
```
0xAlunara, line 1: File was extracted from `data-table.utils.ts`, just like `useCellSx.ts`.

packages/evm-ui/src/shared/ui/DataTable/HeaderCell.tsx
```diff
@@ -5,41 +5,51 @@ import { borderStyle } from '@evm-ui/utils'
...
-  getExtraColumnPadding,
-  type DataTableSize,
-  type TableItem,
-} from './data-table.utils'
+import { type CurveTableFeatures, getAlignment, getExtraColumnPadding, type DataTableSize } from './data-table.utils'
 
 const { Spacing, Sizing } = SizesAndSpaces
 
-function useHeaderSx<T extends TableItem>({
+const HeaderCellPaddingBlockEnd = {
```
0xAlunara, line 14: Extracted from `data-table.utils.ts` as they were the only consumer

packages/evm-ui/src/shared/ui/DataTable/data-table.utils.ts
0xAlunara: See PR description for clarification

packages/evm-ui/src/shared/ui/DataTable/data-table.utils.ts
0xAlunara: I'll expand the comment, hoped this single line would be sufficient 😄

packages/evm-ui/src/shared/ui/DataTable/data-table.utils.ts
0xAlunara: Just to reiterate from the Slack thread: the refactor's only possible thanks to the new and better typing of React Table v9, so you can't really pull the refactor out as a separate PR.

packages/evm-ui/src/shared/ui/DataTable/data-table.utils.ts
0xAlunara: It's because features are now modular, so accessing specific feature functionality is a bit more difficult. It's a bit code smelly, but this is only used for the sorting arrow in `LegacyPoolTable`. Since that table will be removed in the future I let it slide for now.

packages/evm-ui/src/shared/ui/DataTable/DataTable.d.ts
0xAlunara: Created a ticket! Only used once outside of the legacy pool table too, so might want to wait until that one's been removed

packages/evm-ui/src/shared/ui/DataTable/LegacyDataTable.tsx
```diff
@@ -49,7 +50,7 @@ const { Sizing } = SizesAndSpaces
 /**
  * DataTable component to render the table with headers and rows.
  */
-export const LegacyDataTable = <T extends TableItem>({
+export const LegacyDataTable = <TData extends RowData>({
```
0xAlunara, line 53: I've renamed `T extends RowData` to `TData extends RowData` in quite a few places so that it's in sync with TanStack's convention (and makes it less ambiguous as there's multiple generics now).

tests/cypress/support/helpers/ComponentTestWrapper.tsx
```diff
@@ -24,18 +27,22 @@ type Props = {
  * Similar to apps/main/src/app/ClientWrapper.tsx but optimized for the testing environment.
  */
 export function ComponentTestWrapper({ config, children, autoConnect }: Props) {
-  // Create a minimal router for testing environment
-  const router = createRouter({
-    routeTree: createRootRoute({ component: () => children }),
-    history: createMemoryHistory({ initialEntries: ['/'] }),
-  })
+  // Keep the stateful router stable when test children trigger a wrapper re-render.
```
0xAlunara, line 30: Truth be told this fix was suggested by gippety and seems to work; wasn't really sure what was going on initially. This PR also updates TanStack Router, so something internally must've changed there that this suddenly became an issue.

packages/evm-ui/src/shared/ui/DataTable/data-table.utils.ts
```diff
@@ -40,133 +42,79 @@ export const CLICKABLE_IN_ROW_CLASS = 'clickable-in-row'
...
-  options: Omit<TableOptions<TData>, 'data'> & { query: QueryProp<TData[]> },
-) => {
-  const table = useReactTable<TData>({ ...options, data: options.query.data ?? EMPTY_ARRAY })
-  return { ...table, isLoading: options.query.isLoading, error: options.query.error }
+/** Select a table-specific faceting strategy, falling back to TanStack's standard implementation. */
+const createDispatchedFacetedRowModel = <TFeatures extends TableFeatures, TData extends RowData>(
+  table: Table<TFeatures, TData>,
+  columnId: string,
+): (() => RowModel<TFeatures, TData>) => {
+  // This fixed feature set intentionally omits the type-only `tableMeta` slot, so the global generic augmentation wins.
```
0xLokiB: Why?

packages/evm-ui/src/shared/ui/DataTable/data-table.utils.ts
```diff
@@ -40,133 +42,79 @@ export const CLICKABLE_IN_ROW_CLASS = 'clickable-in-row'
...
 
-export const getHiddenCount = <T>(table: Table<T>): number =>
-  table.getPreFilteredRowModel().rows.length - table.getFilteredRowModel().rows.length
+/** Query-aware wrapper around the Curve app table hook. */
+export const useCurveTable = <TData extends RowData>({
+  query: { data, isLoading, error },
+  ...tableOptions
+}: Omit<TableOptions<CurveTableFeatures, TData>, 'data' | 'features'> & {
+  query: QueryProp<TData[]>
+}) => ({
```
0xLokiB, line 105: Shame we can't use mapQuery due to the hook

packages/evm-ui/src/shared/ui/DataTable/data-table.utils.ts
0xLokiB: I mean it should be in the code

packages/evm-ui/src/shared/ui/DataTable/LegacyDataRow.tsx
```diff
@@ -28,33 +27,33 @@ const onCellClick = (target: EventTarget, url: string, routerNavigate: (href: st
...
   verticalAlign = 'middle',
-}: LegacyDataRowProps<T>) => {
+}: LegacyDataRowProps<TData>) => {
   const isMobile = useIsMobile()
   const [element, setElement] = useState<HTMLTableRowElement | null>(null) // note: useRef doesn't get updated in cypress
   const push = useNavigate()
-  const url = row.original.url
-  const hasUrl = Boolean(url?.trim())
+  const href = table.options.meta?.getRowHref?.(row.original)
+  const hasHref = Boolean(href?.trim())
```
0xLokiB: I dont think we need to trim, who passes urls with spaces?

0xAlunara: Fair, figured I'd leave it untouched but now you leave me no choice!

packages/evm-ui/src/shared/ui/DataTable/data-table.utils.ts
0xAlunara: Markets have a custom faceted row model factory, I clarified the docs a bit, hopefully it's more clear. I don't know if the custom override is the best design but I don't want to think about that for now 😄

0xLokiB (approved):

0xtutti (approved):

PR #3103, opened 2026-08-28
fix: artifact download parent folder

0xLokiB:
- Create the parent folder before checking if the same run ID is already there.
- This fixes script crashes when downloading to an existing parent folder.


0xAlunara (approved):
Heh, I thought I saw a PR this very week where you made it more strict to fail when no folder was there 😄

0xLokiB:
> Heh, I thought I saw a PR this very week where you made it more strict to fail when no folder was there 😄

Exactly. But when the parent folder is there we still want it to work!

0xtutti (approved):

PR #3105, opened 2026-08-28
refactor: pool page form tabs

0xLokiB:
- Use the FormTabs in the pool page.
- Redirect users from old URLs.


apps/main/src/dex/components/PagePool/index.tsx
```diff
@@ -20,59 +25,148 @@ import { useTokensMapper } from '@/dex/hooks/useTokensMapper'
...
+      },
+      {
+        value: 'DEPOSIT_STAKE',
+        label: t`Deposit & Stake`,
+        component: transferProps => (
+          <>
+            {transferProps.poolDataCacheOrApi.gauge.isKilled ? (
+              <AlertBox alertType="warning">{t`Staking is disabled due to inactive Gauge.`}</AlertBox>
+            ) : (
+              <FormDepositStake {...transferProps} />
```
0xPearce: The deposit alert `transferProps.poolAlert?.isDisableDeposit` was also disabling `Stake` and `Deposit & Stake` tabs. You can try with this pool for example /dex/ethereum/pools/0x862CB4E988FB66E72f128d1183829f8c05B6c6A0/deposit we shouldn't be able to deposit and stake but show the alert instead

Image

apps/main/src/dex/components/PagePool/index.tsx
```diff
@@ -20,59 +25,148 @@ import { useTokensMapper } from '@/dex/hooks/useTokensMapper'
...
+      ) : (
+        <Swap poolAlert={poolAlert} {...props} />
+      ),
+  },
   {
     value: 'manage-gauge',
     label: t`Gauge`,
-    visible: (p: TransferTabsParams) => p.isAvailableManageGauge && !!p.poolData,
-    component: ManageGaugeTab,
+    visible: p => p.isGaugeManager ?? p.isRewardsDistributor,
```
0xPearce: So if  `isGaugeManager` is false we aren't rendering the Gauge tab? What about rewards distributor should be able to see the `Deposit Reward` tab? Not sure if this is a fix following a thread on Slack I read this morning 🤔 . Also it renders when no wallet is connected: 

Image

apps/main/src/routes/dex.routes.tsx
```diff
@@ -86,6 +80,11 @@ export const dexRoutes = dexLayoutRoute.addChildren([
     }),
     ...layoutProps,
   }),
+  createRoute({
+    path: '$network/pools/$poolIdOrAddress/$formType',
+    loader: ({ params: { network, poolIdOrAddress } }) => redirectTo(`/dex/${network}/pools/${poolIdOrAddress}`),
```
0xPearce, line 85: Oh nice we are dropping the `/deposit`, `/withdraw` ending to make it consistent with llamalend? I suppose we don't care about old bookmarked links

0xLokiB: There is a redirect set up

apps/main/src/dex/components/PagePool/index.tsx
0xLokiB: That's the bug tutti also found I'll check

apps/main/src/routes/dex.routes.tsx
0xPearce: Is there? `/dex/ethereum/pools/0xc061caa073f3d95F80f8e5428d32D2d76F5e1622/withdraw` redirect to `/dex/ethereum/pools/0xc061caa073f3d95F80f8e5428d32D2d76F5e1622` with the Deposit form as default. I don't think it's a problem though

0xPearce (approved):

0xtutti (approved):

PR #3109, opened 2026-08-29
feat: apr apy toggle

0xAlunara:
Pure gippety slop, do not yet review!!!


0xAlunara:
Replaced with https://github.com/curvefi/curve-frontend/pull/3149

PR #3111, opened 2026-08-29
refactor: proposals react table

0xAlunara:
Goal of this PR is to get rid of the some legacy stuff and most importantly the slice.

This is a funny case, as the proposal list is everything that a table is, except the actual table. Search, filters, sorting, pagination, refresh; except the 'table' is a 1-column list of a custom component. I tried to use as many reusable `DataTable` support components and code as I could, except for the actual `DataTable` component itself. For now it's still a simple list of the `Proposal` component (which has not been changed!)

image


apps/main/src/dao/components/PageProposals/components/ProposalsToolbar.tsx
```diff
@@ -0,0 +1,149 @@
...
+      />
+
+      <TableFiltersOverlay
+        title={t`Filter proposals`}
+        anchorRef={filterChipRef}
+        hasActiveFilters={hasActiveFilters}
+        open={filtersOpen}
+        resetFilters={resetFilters}
+        setOpen={setFiltersOpen}
+        disableSticky
```
0xAlunara, line 133: I didn't want to waste too much time and adding complexity by making this sticky like we do for the other tables, and it's not really important either. I had enough of working on this list 💀

packages/evm-ui/src/shared/ui/DataTable/TableFiltersOverlay.tsx
```diff
@@ -67,6 +69,8 @@ export const TableFiltersOverlay = ({
       onClose={() => setOpen(false)}
       anchorEl={() => anchorRef.current}
       anchorOrigin={{ vertical: 'bottom', horizontal: 'left' }}
+      marginThreshold={disableSticky ? null : undefined}
+      sx={disableSticky ? { zIndex: theme => theme.zIndex.tableFilters } : undefined}
```
0xAlunara, line 73: I tried the `{...disableSticky && { /** */ }}` method, but it doesn't really work well when we need to modify a non  `sx` prop and the `sx` prop with a `theme` function.

`sx={disableSticky && { zIndex: theme => theme.zIndex.tableFilters }}` doesn't work either

0xPearce (dismissed):

apps/main/src/dao/components/PageProposals/components/ProposalsToolbar.tsx
```diff
@@ -0,0 +1,149 @@
...
+  setSearch: (value: string) => void
+  status: string
+  setStatus: (value: string) => void
+  sortBy: string
+  sortDirection: SortDirection
+  resetFilters: () => void
+  onSortingChange: OnChangeFn<SortingState>
+}) => {
+  const isMobile = useIsMobile()
+  const [filtersOpen, , , , setFiltersOpen] = useSwitch(false)
```
0xLokiB: Might as well useState

apps/main/src/dao/components/PageProposals/components/ProposalsToolbar.tsx
```diff
@@ -0,0 +1,149 @@
...
+import { useSwitch } from '@ui/hooks/useSwitch'
+import { ArrowDownIcon } from '@ui/icons/ArrowDownIcon'
+import { RotatableIcon } from '@ui/icons/RotatableIcon'
+import { t } from '@ui/lib/i18n'
+import { ProposalColumnId } from '../columns'
+
+const { Spacing } = SizesAndSpaces
+
+const PROPOSAL_FILTERS = [
+  { value: 'all', label: 'All' },
```
0xLokiB, line 26: I'd use null as value since all means no filter

packages/ui/src/features/tables/hooks/useTableScroll.ts
```diff
@@ -38,7 +38,7 @@ export function useScrollToTopOnFilterChange<TData extends RowData>({
   })
 }
 
-/** Scrolls to the top of the table whenever the page changes. */
+/** Scrolls to the container or table top, or the page top when neither ref is provided, after pagination changes. */
```
0xLokiB: I'm not sure it's a good idea to scroll to the top of the page by default. Do we really want that?

apps/main/src/dao/components/PageProposals/components/ProposalsToolbar.tsx
0xAlunara: Ok but I'm going to do it consistently for the pools and markets table too then

apps/main/src/dao/components/PageProposals/components/ProposalsToolbar.tsx
0xAlunara: Makes sense, but it becomes a mess down the line because the components down the road are generic on `T extends string`. Because options then become `string | null` you'd have to add checks everywhere, making thing s very complex. I'd rather keep it like this.

packages/ui/src/features/tables/hooks/useTableScroll.ts
0xAlunara: Aight, reverted that change, but the `Proposals` component now needs to use `useLayoutStore` and a `useRef` for scrolling to properly work.

0xPearce (approved):

apps/main/src/dao/components/PageProposals/index.tsx
```diff
@@ -1,237 +1,151 @@
...
+    resetFilters,
+  } = useFilters({ columns: { Status: ProposalColumnId.Status }, resetPageOnChange: true })
+
+  const status = columnFiltersById[ProposalColumnId.Status] ?? 'all'
+
+  const [pagination, onPaginationChange] = usePageFromQueryString(PROPOSALS_PAGE_SIZE)
+
+  const [sorting, onSortingChange] = useSortFromQueryString(DEFAULT_SORT)
+  const globalFilterFn = useFuzzyFilterFn(tableQuery.data ?? EMPTY_PROPOSALS, search, [
+    { name: 'id', getFn: proposal => String(proposal.id) },
```
0xLokiB, line 63: Is this callback just to convert int to string? I think we could just support numbers in useFuzzyFilterFn?

apps/main/src/dao/components/PageProposals/index.tsx
```diff
@@ -1,237 +1,151 @@
...
+  ])
+
+  const table = useCurveTable({
+    query: tableQuery,
+    columns,
+    state: {
+      // Finished proposals all have undefined for Ending Soon; break that tie by creation time, newest first.
+      sorting: useMemo(() => [...sorting, ...DEFAULT_SORT], [sorting]),
+      columnFilters: useMemo(
+        () => notFalsy(status !== 'all' && { id: ProposalColumnId.Status, value: status }),
```
0xLokiB, line 76: the magic string smells

0xLokiB (approved):

apps/main/src/dao/components/PageProposals/index.tsx
0xAlunara: Fuse requires it to be a string and won't accept number, which the id is ☹️

apps/main/src/dao/components/PageProposals/index.tsx
0xAlunara: I might give stricter typing another shot in the follow up PR for the proposal component itself. Used to be stricter but it gave me a headache so loosing status to be `string` was easier for now. But yeah `all` is suddenly more magic that way.

PR #3114, opened 2026-08-31
feat: temporarily hide TAC from network list

0xAlunara:


apps/main/src/dex/lib/networks.ts
```diff
@@ -324,6 +324,7 @@ export async function getNetworks() {
     const isUpgraded = !!baseConfig // networks upgraded from lite to full
     const isOnlyPoolRewardsUpgraded = poolRewardsUpgradedChains.includes(chainId)
     const isLiteFxswapEnabled = fxSwapUpgradedChains.includes(chainId)
+    const showInSelectNetwork = chainId !== Number(Chain.Tac) // temporarily disabled as the chain's halted
```
0xAlunara, line 327: Creating a `Chain.Tac` object with `showInSelectNetwork  false` above in `defaultNetworks` breaks the site, partially because it's a lite network. I tried to get it to work but the whole networks thing needs a do-over, it's such a complicated mess...

This will do for now as it's temporary anyway

0xLokiB: The change is fine, but this comment should be in the code not in the PR.
Plus it's not super clear why we expect the chain to be back

0xLokiB (approved):

apps/main/src/dex/lib/networks.ts
0xAlunara: I've created a ticket for the simplification of the networks configs, been wanting to do this for a long time now anyway.

0xtutti (approved):

PR #3116, opened 2026-08-31
feat: disable default pools TVL filter

0xAlunara:
We've decided to disable the default $10k TVL filter for now as we wait for an alternative solution, so we can release the new pools list. But perhaps one is not needed, given we have pagination now anyway.

Also fixes some tests (I hope)

### DIFF
--- a/apps/main/src/dex/features/pool-list/PoolsTable.tsx
+++ b/apps/main/src/dex/features/pool-list/PoolsTable.tsx
@@ -80,6 +80,7 @@ export const PoolsTable = ({ network }: { network: NetworkConfig }) => {
       globalFilter,
       ...(!isLite && { pagination, columnFilters }),
     },
+    getRowId: row => row.address,
     onExpandedChange: setExpanded,
     ...(!isLite && { onPaginationChange }),
     onSortingChange,
--- a/apps/main/src/dex/features/pool-list/filters/utils.ts
+++ b/apps/main/src/dex/features/pool-list/filters/utils.ts
@@ -19,8 +19,13 @@ export enum PoolFilterId {
   Apy = 'apy',
 }
 
-// Hide small pools by default, without treating the default min as an active UI filter on its own.
-export const POOL_DEFAULT_TVL_MIN = 10_000
+/**
+ * Hide small pools by default, without treating the default min as an active UI filter on its own.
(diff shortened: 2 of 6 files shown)


tests/cypress/e2e/dex/dex-markets.cy.ts
```diff
@@ -160,18 +159,6 @@ describe('DEX Pools', () => {
       return waitForPoolListResponse()
     }
 
-    it('applies the default TVL min without URL or active filter state', () => {
```
0xAlunara, line 163: These tests no longer apply and fail as the default behavior isn't anymore to hide the default $10k filter

apps/main/src/dex/features/pool-list/PoolsTable.tsx
```diff
@@ -80,6 +80,7 @@ export const PoolsTable = ({ network }: { network: NetworkConfig }) => {
       globalFilter,
       ...(!isLite && { pagination, columnFilters }),
     },
+    getRowId: row => row.address,
```
0xAlunara, line 83: A bit unrelated, but by default row ids are the index based which I think can cause some flakyness due to unnecessary re-renders based on just the ordering.

0xtutti (approved):

apps/main/src/dex/features/pool-list/PoolsTable.tsx
0xLokiB: Sounds good, but the comment should be in the code

apps/main/src/dex/features/pool-list/filters/utils.ts
```diff
@@ -19,8 +19,13 @@ export enum PoolFilterId {
...
 
-// Hide small pools by default, without treating the default min as an active UI filter on its own.
-export const POOL_DEFAULT_TVL_MIN = 10_000
+/**
+ * Hide small pools by default, without treating the default min as an active UI filter on its own.
+ * Used to be $10k but has been set to 0 as it's not immediately obvious to users why some pools are missing.
+ * Tutti is working on an alternative approach, in the menatime the TVL filter should not block the release of the new pools list.
+ * Besides, we have pagination now anyway so do we really need this filter to begin with?
+ */
+export const POOL_DEFAULT_TVL_MIN = 0
```
0xLokiB, line 28: I'd remove the whole thing

0xLokiB (approved):

PR #3123, opened 2026-09-01
refactor: network inversion part 1

0xAlunara:
Removes a few network config properties in favor of the defaults provided by wagmi chains.

### DIFF
--- a/apps/main/src/bridge/features/bridge/components/BridgeForm.tsx
+++ b/apps/main/src/bridge/features/bridge/components/BridgeForm.tsx
@@ -1,5 +1,6 @@
 import { useMemo } from 'react'
 import { useConnection } from 'wagmi'
+import { wagmiChainsMap } from '@evm-ui/features/connect-wallet/lib/wagmi/chains'
 import { useNavigate, usePathname } from '@evm-ui/hooks/router'
 import { useTokenUsdRate } from '@evm-ui/lib/model/entities/token-usd-rate'
 import { getCurrentApp, getInternalUrl } from '@evm-ui/shared/routes'
@@ -58,7 +59,7 @@ export const BridgeForm = ({
             bridgeCost={q(bridgeCost)}
             gas={q(gas)}
             isApproved={isApproved.data}
-            nativeTokenSymbol={networks[chainId].symbol}
+            nativeTokenSymbol={wagmiChainsMap[chainId]?.nativeCurrency.name}
           />
           <BridgeInfoAlert />
         </>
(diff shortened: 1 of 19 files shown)


apps/main/src/bridge/features/bridge/components/BridgeForm.tsx
```diff
@@ -58,7 +59,7 @@ export const BridgeForm = ({
             bridgeCost={q(bridgeCost)}
             gas={q(gas)}
             isApproved={isApproved.data}
-            nativeTokenSymbol={networks[chainId].symbol}
+            nativeTokenSymbol={wagmiChainsMap[chainId]?.nativeCurrency.name}
```
0xAlunara, line 62: In a later PR (part 3) I've removed the export of `wagmiChainsMap` in favor of helper functions

apps/main/src/dex/hooks/usePoolTotalStaked.tsx
```diff
@@ -63,9 +60,8 @@ export const usePoolTotalStaked = (poolDataCacheOrApi: PoolDataCacheOrApi): Pool
   useEffect(() => {
     const shouldCallApi = staked?.timestamp ? dayjs().diff(staked.timestamp, 'seconds') > 30 : true
 
-    if (address && rpcUrl && shouldCallApi) {
+    if (address && provider && shouldCallApi) {
```
0xAlunara, line 63: Might be a regression as you need a connected wallet for now, but it's odd to have this one-off edge case where we construct our own Json RPC provider... Can revisit later perhaps.

0xtutti (approved):

apps/main/src/dex/components/PageDashboard/components/ClassicPoolsOnlyDescription.tsx
```diff
@@ -1,35 +1,37 @@
...
+  ],
 }
 
+export const ClassicPoolsOnlyDescription = ({ chainId }: { chainId: number }) =>
+  maybe(MISSING_POOLS[chainId], missingPools => (
+    <MissingPoolDescription>
+      {t`*This UI does not support the following pools:`}{' '}
+      {missingPools.map((pool, idx) => (
+        <Fragment key={pool.name}>
+          {idx === 0 ? '' : ', '}
```
0xLokiB, line 25:
```suggestion
          {idx > 0 && ', '}
```

apps/main/src/dex/hooks/usePoolTotalStaked.tsx
0xLokiB: We should definitely remove this ethers contract from the hook

0xLokiB (approved):

apps/main/src/bridge/features/bridge/components/BridgeForm.tsx
0xPearce: Shouldn't we use the symbol? `Ether` feels odd
Image

0xAlunara: oh lol now that you mention it... Will fix it in a follow up

PR #3124, opened 2026-09-01
refactor: network inversion part 2

0xAlunara:
Removes the custom mainnet and hyperliquid chain definitons, wagmi provides those. `createChainFromNetwork` also overrides the default RPCs with `getRpcUrls` so that whole WalletConnect thing isn't needed anymore for mainnet.

### DIFF
--- a/packages/evm-ui/src/features/connect-wallet/lib/wagmi/custom-chains.ts
+++ b/packages/evm-ui/src/features/connect-wallet/lib/wagmi/custom-chains.ts
@@ -1,26 +1,8 @@
-/**
- * WalletConnect ignores the configured HTTP transport URLs from Wagmi config
- * and instead creates a provider using only the first URL from chain.rpcUrls.default.http.
- *
- * This is why we need to define custom chains with specific RPC URLs that will work
- * properly when WalletConnect initializes its provider with:
- * `new WalletConnectProvider({ rpc: chain.rpcUrls.default.http[0] })`
- *
- * blockExplorers are not defined here, they get set in `createChain`.
- */
 import { chainConfig } from 'viem/op-stack'
 import { defineChain } from 'viem/utils'
 import { Chain as ChainId } from '@evm-ui/utils/network'
 import { mainnet } from '@wagmi/core/chains'
 import { RPC } from './rpc'
-
-export const ethereum = defineChain({
-  ...mainnet,
-  rpcUrls: {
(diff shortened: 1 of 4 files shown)


0xtutti (approved):

0xPearce (approved):

packages/evm-ui/src/features/connect-wallet/lib/wagmi/custom-chains.ts
```diff
@@ -1,36 +1,9 @@
-/**
- * WalletConnect ignores the configured HTTP transport URLs from Wagmi config
```
0xLokiB, line 2: won't this break walletconnect??

0xAlunara: I don't think so, inside `chains.ts` we define chains such that
`defineChain({ ...wagmiChainsMap[chainId], rpcUrls: { default: { http: getRpcUrls(chainId) } } })`

Which already overrides the RPC on a wagmi chain level. WalletConnect will use that very same RPC we set it to. So the custom `mainnet` definition was unnecessary.

In other words, the custom chain only existed to override the custom RPC for WalletConnect, which we also do inside `chains.ts` when we define the chains for wagmi.

0xLokiB: The comment said `WalletConnect ignores the configured HTTP transport URLs from Wagmi config`
which was a real issue when written

0xAlunara: I think it's outdated. Or rather, it fixed a particular issue back then before we had custom RPCs for all chains.

The only thing this custom chain does is

```typescript
export const ethereum = defineChain({
  ...mainnet,
  rpcUrls: { default: { http: RPC[ChainId.Ethereum] }, },
})
```

which is the same as
```typescript
defineChain({ ...wagmiChainsMap[chainId], rpcUrls: { default: { http: getRpcUrls(chainId) } } })
```

as
```typescript
export const defaultGetRpcUrls = <ChainId extends number>(chainId: ChainId) =>
  lodash.uniq([...(RPC[chainId] ?? []), ...getChainDefaultRpcUrls(chainId)])
```

where `chainId = ChainId.Ethereum`

The `public` RPC isn't a thing anymore afaik.

PR #3122, opened 2026-09-01
feat: price impact alert threshold

0xPearce:
- use fixed 1% warning and 25% blocking thresholds instead of the selected slippage tolerance
- added caution, warning, and error colors to price-impact depending on the values 
- reuse the same price-impact row and alert across leveraged LlamaLend, router swaps, and pool swaps for consistency
Screenshot 2026-09-03 at 11 25 34
Screenshot 2026-09-03 at 11 26 11

PR #3128, opened 2026-09-02
feat: restore reference price impact

0xPearce:
- depends on #3122
- calculate LlamaLend price impact against a $0.10 reference quote from the same router

Made this PR to test the price impact calculation against a tiny $0.10 trade, instead of using the USD rate


0xPearce:
Let’s keep the USD price impact calculation method and improve the token USD value instead

PR #3129, opened 2026-09-02
feat: table visibility migration

0xAlunara:
Adds a table visibility migration helper function that keeps columns on/off based on what they were, with an optional override for new defaults. This prevents new versions from erasing all previously configured column visibilities.

Pool version has been bumped to fix missing columns from the visibility settings if they tested beta before it got deployed to stable.

### DIFF
--- a/apps/main/src/llamalend/features/market-list/hooks/useMarketsVisibility.tsx
+++ b/apps/main/src/llamalend/features/market-list/hooks/useMarketsVisibility.tsx
@@ -3,7 +3,10 @@ import { useMemo } from 'react'
 import type { LlamaMarketsResult } from '@/llamalend/queries/market-list/llama-markets'
 import { useIsMobile } from '@evm-ui/hooks/useBreakpoints'
 import type { MigrationOptions } from '@evm-ui/hooks/useStoredState'
-import { useVisibilitySettings } from '@evm-ui/shared/ui/DataTable/hooks/useVisibilitySettings'
+import {
+  preserveVisibilityChoices,
+  useVisibilitySettings,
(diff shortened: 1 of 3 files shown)


apps/main/src/llamalend/features/market-list/hooks/useMarketsVisibility.tsx
```diff
@@ -17,28 +20,23 @@ import {
...
-  })
+const resolveMarketActive = (
+  preservedActive: boolean,
+  currentOption: VisibilityGroup<MarketColumnId>['options'][number],
+) =>
+  isEqual(currentOption.columns, [MarketColumnId.NetBorrowRate])
+    ? currentOption.active
+    : isEqual(currentOption.columns, [MarketColumnId.BorrowRate])
+      ? false
+      : preservedActive
```
0xLokiB:
```suggestion
    : !isEqual(currentOption.columns, [MarketColumnId.BorrowRate]) && preservedActive
```

0xLokiB (dismissed):

apps/main/src/llamalend/features/market-list/hooks/useMarketsVisibility.tsx
0xAlunara: 🥴

0xtutti (approved):

0xLokiB (approved):

apps/main/src/llamalend/features/market-list/hooks/useMarketsVisibility.tsx
0xLokiB: Sorry, just seeing `? false` triggers my smell sense. Maybe your version was more clear.

PR #3136, opened 2026-09-02
lint: isolate modules

0xLokiB:
- The way that yarn workspaces work, apparently you can import packages that are not declared as dependencies.
- The builtin import rules end up requiring you to declare are lot of transitive dependencies.
- So I created this custom rule to enforce module separation while allowing importing deep dependencies
- Creating a custom lint rule allows us to deduplicate the dependency declarations, simplifying the setup, and still make sure that we are strict by only importing allowed dependencies
- After fixing the rule, I was forced by the rules to move some code around.


apps/router-api/src/router.utils.ts
```diff
@@ -1,5 +1,4 @@
 import { BigNumber } from 'bignumber.js'
-import { formatUnits, parseUnits } from 'viem'
```
0xLokiB: router api should actually depend on viem for these

.eslint/isolated-packages.rule.mjs
```diff
@@ -0,0 +1,166 @@
+import fs from 'node:fs'
```
0xAlunara, line 1: I'll believe gippety here

apps/main/src/dex/components/PagePool/Page.tsx
```diff
@@ -56,9 +56,7 @@ export const PagePool = () => {
    */
   const { data: blacklist } = usePoolsBlacklist({ blockchainId: blockchainId as Chain })
   const isBlacklisted = useMemo(
-    () =>
-      isAddress(rPoolIdOrAddress, { strict: false /* address comes from URL which might be lowercase */ }) &&
-      blacklist?.some(badPool => isAddressEqual(badPool, rPoolIdOrAddress)),
+    () => isAddress(rPoolIdOrAddress) && blacklist?.some(badPool => isAddressEqual(badPool, rPoolIdOrAddress)),
```
0xAlunara: Why remove `strict: false` and the corresponding comment? I don't see how that's suddenly not an issue any more.

apps/main/src/dex/hooks/usePoolIdByAddressOrId.ts
```diff
@@ -12,7 +12,7 @@ export function usePoolIdByAddressOrId({ chainId, poolIdOrAddress }: { chainId:
 
   return useMemo(() => {
     // If not an address format, assume it's already a pool ID
-    if (!isAddress(poolIdOrAddress, { strict: false })) {
+    if (!isAddress(poolIdOrAddress)) {
```
0xAlunara: Similarly here, not sure we can afford this strictness

apps/main/src/dex/queries/pools-prices-api.query.ts
```diff
@@ -20,6 +21,6 @@ export const usePoolPricesApi = ({
   blockchainId,
   poolAddress,
 }: ChainNameParams & { poolAddress: string | undefined }) =>
-  mapQuery(usePoolsPricesApi({ blockchainId }, isAddress(poolAddress as Address, { strict: false })), pools =>
+  mapQuery(usePoolsPricesApi({ blockchainId }, isAddress(poolAddress as Address)), pools =>
```
0xAlunara: For the sake of bookkeeping leaving the strict comment here as well

packages/evm-ui/src/hooks/useCopyToClipboard.ts
```diff
@@ -1,11 +1,11 @@
 import { useCallback } from 'react'
-import { isAddress } from 'viem'
 import { copyToClipboard } from '@evm-ui/utils'
 import { showToast } from '@evm-ui/widgets/Toast/toast.util'
+import { isAddress } from '@primitives/address.utils'
 import { t } from '@ui/lib/i18n'
 
 const getTitle = (copyText: string, title: string | undefined) =>
-  title ?? t`${isAddress(copyText, { strict: false }) ? `Address` : `Value`} has been copied to clipboard`
+  title ?? t`${isAddress(copyText) ? `Address` : `Value`} has been copied to clipboard`
```
0xAlunara: And here

packages/ui/package.json
```diff
@@ -10,29 +10,32 @@
...
     "@mui/material": "9.4.0",
     "@sentry/core": "10.72.0",
     "@sentry/react": "10.72.0",
-    "@tanstack/query-async-storage-persister": "*",
-    "@tanstack/query-persist-client-core": "*",
-    "@tanstack/react-query": "*",
-    "@tanstack/react-query-persist-client": "*",
+    "@tanstack/query-async-storage-persister": "5.102.8",
+    "@tanstack/query-persist-client-core": "5.102.8",
+    "@tanstack/react-query": "5.102.8",
```
0xAlunara, line 25: I'm tripping over the fact we have to define a package's version more than once again. Isn't it now too easy again to forget updating package versions in one package.json out of many when necessary?

0xtutti (approved):

0xPearce (approved):

PR #3138, opened 2026-09-02
feat: stellar query client and sentry setup

0xLokiB:
- Move query provider, client and utils to `ui`.
- Move logging and environment handling to `ui`.
- Render the query client in the stellar app.
- Update imports everywhere.


packages/ui/src/utils/env.ts
0xLokiB: Move

tests/cypress/component/dex/refuel-form-list.cy.tsx
```diff
@@ -1,7 +1,7 @@
 import { RefuelFormList } from '@/dex/features/manage-pool/components/RefuelFormList'
 import { ComponentTestWrapper } from '@cy/support/helpers/ComponentTestWrapper'
 import { getActionValue } from '@cy/support/helpers/llamalend/action-info.helpers'
-import { constQ } from '@evm-ui/types/util'
+import { constQ } from '../../../../packages/ui/src/lib/util'
```
0xLokiB: Ni

package.json
```diff
@@ -32,6 +32,14 @@
     "@eslint-react/eslint-plugin": "5.18.0",
     "@eslint/js": "10.0.1",
     "@tanstack/eslint-plugin-query": "5.102.7",
+    "@tanstack/query-async-storage-persister": "5.102.7",
```
0xLokiB: Duplicate

packages/evm-ui/src/features/queries/query-client.ts
```diff
@@ -0,0 +1,4 @@
+import { hashFn } from 'wagmi/query'
+import { createQueryClient } from '@ui/features/queries/query-client'
+
+export const queryClient = createQueryClient(hashFn)
```
0xLokiB: what

packages/ui/src/features/queries/query-client.ts
```diff
@@ -9,7 +8,6 @@ export const queryClient = new QueryClient({
     queries: {
       retryDelay: attemptIndex => Math.min(1000 * 2 ** attemptIndex, 30000), // Double each attempt, from 1s up to 30s
       retry: 3, // retry failed requests before displaying an error
-      queryKeyHashFn: hashFn,
```
0xLokiB, line 12: Apparently this is only necessary if we are using BigInt in query keys. That doesn't seem to be the case right now. For internal wagmi stuff, their hashFn is already applied.

0xAlunara: I have to recall from memory, and I think the original issue was me trying to use the TanStack Query devtools in component tests. For some reason there were bigints in query keys back then. Wagmi internally uses bigint, but we might've used bigints in our own `queryFactory` queries at some point. Maybe not anymore, let's just keep an eye out.

0xLokiB: We can add it back if necessary

0xAlunara (approved):
Image

0xPearce (approved):
Couple of lint warnings 😅 
Image

PR #3145, opened 2026-09-03
feat: add llamalend market top participants

0xPearce:
- add paginated market borrower and vault depositor query
- add Top Borrowers and Top Suppliers tables in beta LlamaLend Activity 
In order to implement sorting we need the backend to expose the params

Screenshot 2026-09-04 at 10 33 11

Screenshot 2026-09-04 at 10 32 31


0xtutti (dismissed):

apps/main/src/lend/components/ChartAndActivityComp/index.tsx
```diff
@@ -89,23 +92,38 @@ export const ChartAndActivityComp = ({ previewPrices }: ChartAndActivityCompProp
   )
 }
 
+const MarketParticipants = () => (
+  <>
```
0xLokiB, line 96: In case you created this just to add a key, this would be an option too (just FYI)
```suggestion
  <React.Fragment key="...">
```

apps/main/src/lend/components/ChartAndActivityComp/index.tsx
```diff
@@ -89,23 +92,38 @@ export const ChartAndActivityComp = ({ previewPrices }: ChartAndActivityCompProp
...
+          chainId,
+          blockchainId,
+          ammAddress,
+          collateralToken,
+          borrowToken,
+          endpoint: 'lending',
+        }}
+      />
+      <MarketParticipants
+        // Remount when the market changes so both tables reset their local pagination to page one.
```
0xLokiB, line 124: Why would the chain or addresses change while the chart is mounted? 🤔

apps/main/src/llamalend/features/market-participants/market-participants.utils.tsx
```diff
@@ -0,0 +1,131 @@
...
+}) => (
+  <Stack direction="row" sx={{ gap: Spacing.xs, alignItems: 'center' }}>
+    {label}
+    <TokenIcon blockchainId={blockchainId} address={tokenAddress} size="mui-md" />
+  </Stack>
+)
+
+export const Percentage = ({ value }: { value: number | null | undefined }) => (
+  <Typography variant="tableCellMBold">
+    {value == null ? UNAVAILABLE_NOTATION : formatNumber(value, 'percent.value')}
```
0xLokiB, line 53: shouldn't we use the existing fallbacks in the formatNumber categories?

apps/main/src/llamalend/widgets/TokenAmount.tsx
```diff
@@ -0,0 +1,72 @@
...
+import Stack from '@mui/material/Stack'
+import Typography from '@mui/material/Typography'
+
+const { Spacing } = SizesAndSpaces
+
+type TokenAmountProps = {
+  amount: number | undefined
+  amountUsd: number | null | undefined
+  blockchainId?: string
+  tokenAddress?: string
```
0xLokiB, line 17: Why should this be optional? I don't see the point of a token amount without a token. Also, what happened to Address? And why is this specific to Llamalend?
```suggestion
  tokenAddress: Address | QueryProp<Address>
```

apps/main/src/llamalend/features/market-participants/market-participants.utils.tsx
```diff
@@ -0,0 +1,131 @@
...
+        label: t`View on explorer`,
+        href: explorerUrl,
+        size: 'extraSmall',
+        color: 'ghost',
+      },
+    )}
+  />
+)
+
+export const BorrowerExpandedPanelActions: ExpandedPanelComponent<BorrowerRow> = ({ row: { original } }) => (
```
0xLokiB, line 125: Why do we need these wrapper components?

apps/main/src/llamalend/features/market-participants/MarketParticipantsCards.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const { pagination, onPaginationChange, apiPage } = useManualPagination(PAGE_SIZE)
+  const suppliersQuery = useMarketSuppliers({
+    blockchainId,
+    contractAddress: vaultToken?.address,
+    page: apiPage,
+    perPage: PAGE_SIZE,
+  })
+  const { data: borrowTokenUsdRate } = useTokenUsdRate({ chainId, tokenAddress: tokens.borrowToken?.address })
+  const isMobile = useIsMobile()
+  const query = mapQuery(suppliersQuery, ({ depositors }) =>
```
0xLokiB, line 88: shouldn't this be combining the data with useTokenUsdRate's state? Or do we wanna leave usd rates empty while loading/error?

apps/main/src/llamalend/features/market-participants/market-participants.columns.tsx
```diff
@@ -0,0 +1,122 @@
+import { TokenAmount } from '@/llamalend/widgets/TokenAmount'
```
0xAlunara, line 1: I wonder if we should do this more often, to put all column code in one `.tsx` rather than multiple `colum.*.ts` files 🤔

apps/main/src/lend/components/ChartAndActivityComp/index.tsx
0xPearce: mb

apps/main/src/llamalend/widgets/TokenAmount.tsx
0xPearce: The `tokenAddress` is optional because it renders optionally the token icon.

apps/main/src/llamalend/features/market-participants/MarketParticipantsCards.tsx
0xPearce: probably better to leave it like this so usd rate doesn't block the table

0xtutti (approved):

0xLokiB (approved):

PR #3150, opened 2026-09-03
feat: dex speedup

0xAlunara:
All networks currently available support prices API if they're not lite networks.

First of all, there was an incorrect and out of date legacy check that would load more info than necessary. Secondly, all networks use the api so that proeprty could be removed from the `NetworkConfig`, which means `fetchPoolIds` doesn't rely on networks anymore.


apps/main/src/dex/hooks/useAutoRefresh.tsx
```diff
@@ -24,6 +21,6 @@ export const useAutoRefresh = (chainId: number | undefined) => {
   usePageVisibleInterval(async () => {
     if (!curveApi || !poolIds || !chainId) return
     const poolVolumes = await refetchPoolVolumes({ chainId })
-    await fetchPools(curveApi, poolIds, poolVolumes, isLegacyList(releaseChannel, networks?.[chainId]))
```
0xLokiB, line 27: We shouldn't check the release channel directly

0xLokiB (dismissed):

0xtutti (dismissed):

apps/main/src/dex/lib/pool-ids.ts
```diff
@@ -17,17 +18,15 @@ import type { CurveApi } from '../types/main.types'
  * depends on pool addresses, so we still apply it later while building `poolsMapper`.
  */
 export async function fetchPoolIds(curve: CurveApi, { chainId }: ChainQuery): Promise<string[]> {
```
0xPearce: unused arg!!

apps/main/src/dex/store/createGlobalSlice.ts
```diff
@@ -99,8 +96,7 @@ export const createGlobalSlice = (set: StoreApi<State>['setState'], get: StoreAp
     // update network settings from api
     state.setNetworkConfigFromApi(curveApi)
 
-    const networks = await fetchNetworks() // Pool ids have a dependency on networks
-    const isLegacy = isLegacyList(releaseChannel, networks[chainId])
+    const isLegacy = releaseChannel === ReleaseChannel.Legacy
```
0xPearce: Same here (see previous comment of 0xLoki)

0xPearce (approved):

0xLokiB (approved):

PR #3154, opened 2026-09-04
refactor: volume out of tokensmapper

0xAlunara:
Instead of fetching volumes (prematurely) during hydration, we now make it a proper hook that's only called when actually needed.

This speeds up the initial load, especially when you refresh the pool list.


packages/primitives/src/objects.utils.ts
```diff
@@ -24,6 +24,20 @@ export function mapRecord<K extends string, V, R>(obj: Record<K, V>, mapper: (ke
   return Object.fromEntries(entries) as Record<K, R>
 }
 
+/**
+ * Receives an object and a function to convert each key while preserving its value.
+ * @param obj The object to map.
+ * @param mapper The function to convert each key.
+ * @return a new object with converted keys and the same values.
+ */
+export function mapKeys<K extends string, V, R extends PropertyKey>(
```
0xLokiB, line 33: Interesting, I never needed this one yet. But I'm glad it's so much clearer than doing it locally!

0xLokiB (dismissed):

apps/main/src/dex/components/PagePool/Swap/index.tsx
```diff
@@ -118,7 +118,14 @@ export const Swap = ({
 
   const { selectList, swapTokensMapper } = useMemo(() => {
     const { selectList, swapTokensMapper } = getSwapTokens(tokensMapper, poolDataCacheOrApi)
-    return { selectList: selectList.map(toTokenOption(network?.blockchainId)), swapTokensMapper }
+    return {
+      selectList: selectList.map<TokenOption>(token => ({
+        address: getAddress(token.address),
```
0xAlunara: This used to be lowercase, checksumming it breaks token selection when trying to swap on the pool page itself. Need to fix this by simply not checksumming

apps/main/src/dex/components/PageRouterSwap/index.tsx
```diff
@@ -138,7 +139,12 @@ export const QuickSwap = ({
   }, [curve, blacklist, userAddress])
 
   const tokens = useMemo(
-    () => notFalsy(...Object.values(tokensMapper ?? {})).map(toTokenOption(network?.blockchainId)),
+    () =>
+      notFalsy(...Object.values(tokensMapper ?? {})).map<TokenOption>(token => ({
+        address: getAddress(token.address),
```
0xAlunara, line 144: Checksumming works fine here fwiw 🤔

apps/main/src/dex/features/add-gauge-reward-token/ui/TokenSelector.tsx
```diff
@@ -55,7 +54,11 @@ export const TokenSelector = ({
               crvAddress,
             ].some(rewardToken => isAddressEqual(rewardToken, token.address as Address)),
         )
-        .map(toTokenOption(network?.blockchainId)),
+        .map<TokenOption>(token => ({
+          address: getAddress(token.address),
```
0xAlunara, line 58: Checksum works fine here too

apps/main/src/dex/features/deposit-gauge-reward/ui/AmountTokenInput.tsx
```diff
@@ -54,7 +54,6 @@ export const AmountTokenInput = ({
       chain: blockchainId,
       address,
       symbol: tokensMapper[address.toLowerCase()]?.symbol ?? shortenAddress(address),
-      volume: tokensMapper[address.toLowerCase()]?.volume ?? 0,
```
0xAlunara, line 57: Lost this functionality so you don't have to lag up the interface just for volumes in the rare case you add a gauge reward token. Just use the search field 🙃

0xtutti (dismissed):

0xLokiB (approved):

0xPearce (approved):

PR #3158, opened 2026-09-04
refactor: move error boundary to ui

0xLokiB:
- Moves error boundaries and related files to the ui package.
- Depends on https://github.com/curvefi/curve-frontend/pull/3165


packages/evm-ui/src/utils/index.ts
```diff
@@ -6,7 +6,7 @@ export * from './network'
 export * from './number'
 export * from './pagination'
 export * from './decimal'
-export * from './errors'
+export * from '@ui/features/errors/errors.util'
```
0xAlunara: Not wrong but looks out of place lmao

0xLokiB: This is incorrect. I'll fix it

0xAlunara (approved):

0xPearce (approved):

PR #3159, opened 2026-09-04
refactor: remove pool cache

0xAlunara:
The idea is to reduce complexity and remove the pool cache. That cache is not useful anyway as the website refetches everything during hydration anyway. Instead of using old cached data, we should show (a) skeleton(s) when stuff's loading. In the future the pools mapper will be replaced by us constructing pools from prices API data anyway, like we do in llamalend. That prices API data will be cached by Tanstack.

To make life easier and lessen the amount of raw pool data being passed, I've also introduced a pool context, just like we have a market context already.

I tried to keep the diff as small as possible. This means most of the retarded logic has remained untouched. The goal of this PR is to ease the future refactor where we construct `PoolData` from a Prices API query.

PS: I did not implement a skeleton yet for the pool page. Also the pool page is the only one using the context as for now.


apps/main/src/dex/features/pool-context/PoolContextValue.tsx
```diff
@@ -0,0 +1,26 @@
+import type { CurveApi, PoolData } from '@/dex/types/main.types'
+import type { Address } from '@primitives/address.utils'
+
+export const createPoolContextValue = ({
```
0xLokiB, line 4: I'd just leave this in the provider file

apps/main/src/dex/widgets/page-header/PoolPageHeader.tsx
```diff
@@ -11,45 +8,38 @@ import { PoolMetricsRow } from './PoolMetricsRow'
...
       titleLoading={isLoading}
-      subtitle={subtitle ?? (isLoading ? t`Token symbols` : undefined)}
+      subtitle={tokenList?.map(({ symbol }) => symbol).join(' / ') ?? (isLoading ? t`Token symbols` : undefined)}
       subtitleLoading={isLoading}
       icon={
-        (isLoading || tokenList.length > 0) && (
+        (isLoading || (tokenList && tokenList.length > 0)) && (
           <WithSkeleton loading={isLoading} variant="rectangular" width={ICON_SIZE} height={ICON_SIZE}>
-            <TokenIcons blockchainId={blockchainId} tokens={tokenList} overflowMode="stack" />
+            <TokenIcons blockchainId={blockchainId} tokens={tokenList ?? []} overflowMode="stack" />
```
0xLokiB, line 42: ??[] smells, I guess TokenIcons could accept undefined

0xLokiB (dismissed):
I closed #3137

0xtutti (dismissed):

apps/main/src/dex/features/pool-context/PoolContextValue.tsx
0xAlunara: Sure but I'll make a new PR as I want to harmonize it with the market context provider then

0xLokiB (dismissed):

0xPearce (approved):
Are you sure we shouldn't bump the version of the store?

0xAlunara:
Should be fine I think? All the previously uncached data is just going to sit there not being used. I think we could delete the entire cache store soon in its entirety.

0xtutti (approved):

PR #3161, opened 2026-09-05
refactor: evm-independent forms

0xLokiB:
- Refactor forms so that they don't depend on evm-specific hooks.
- Includes the connect wallet button.
- Depends on #3156
- Does not move the forms to ui yet, that's done in https://github.com/curvefi/curve-frontend/pull/3162

### DIFF
--- a/packages/evm-ui/src/shared/ui/EmptyStateCard.tsx
+++ b/packages/evm-ui/src/shared/ui/EmptyStateCard.tsx
@@ -1,5 +1,5 @@
 import { ReactNode } from 'react'
-import { ConnectWalletButton } from '@evm-ui/features/connect-wallet/ui/ConnectWalletButton'
+import { ConnectEvmWalletButton } from '@evm-ui/features/connect-wallet/ui/ConnectEvmWalletButton'
 import { Box, Button, ButtonProps, Skeleton } from '@mui/material'
 import Stack from '@mui/material/Stack'
 import Typography from '@mui/material/Typography'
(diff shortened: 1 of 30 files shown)


0xtutti (dismissed):

0xAlunara (dismissed):

packages/evm-ui/src/shared/ui/EmptyStateCard.tsx
```diff
@@ -58,7 +58,7 @@ const EmptyStateButton = ({
     ...(testId && { 'data-testid': testId }),
   } as const
   return type === 'connect-wallet' ? (
-    <ConnectWalletButton {...sharedProps} label={label} />
+    <ConnectEvmWalletButton {...sharedProps} label={label} />
```
0xPearce, line 61: We are not using this empty state component for Stellar?

0xPearce (approved):

packages/evm-ui/src/shared/ui/EmptyStateCard.tsx
0xPearce: I think same answer as my previous comment 😅

0xAlunara (approved):

packages/evm-ui/src/shared/ui/EmptyStateCard.tsx
0xLokiB: Indeed, left for a later refactor (https://github.com/curvefi/curve-frontend/pull/3165)

PR #3167, opened 2026-09-06
feat: reimplement user pool position detection using prices api

0xAlunara:
Replaces the (disabled) curve-js based user pool positions detection code with a Prices API query.

At the time of writing however, the back-end seems to be incomplete; some user positions are not yet correctly returned by the API. Might complicate testing a bit.

Convex has a nice address to test with: `0x989AEb4d175e16225E39E87d0D97A3360524AD80`


packages/prices-api/src/pools/index.ts
```diff
@@ -112,6 +113,19 @@ export async function listPoolRegistries({ chainId }: { chainId: number }, optio
   return Schema.listPoolRegistriesResponse.parse(response)
 }
 
+export async function getUserPoolPositions(
+  { chainId, userAddress, newTx }: { chainId: number; userAddress: Address; newTx?: Hex },
+  options?: Options,
+) {
+  const host = getHost(options)
+  const query = maybe(newTx, newTx => addQueryString({ new_tx: newTx })) ?? ''
```
0xAlunara, line 121: I think not using `''` would leave `undefined` in the final URL? 🤔

0xLokiB: addQueryString should just ignore undefined

apps/main/src/dex/queries/user-pool-positions.query.ts
```diff
@@ -0,0 +1,20 @@
...
+export const {
+  useQuery: useUserPoolPositions,
+  invalidate: invalideUserPoolPositions,
+  reset: resetUserPoolPositions,
+} = queryFactory({
+  queryKey: ({ chainId, userAddress }: UserChainParams) =>
+    [...rootKeys.userChain({ chainId, userAddress }), 'getUserPoolPositions'] as const,
+  queryFn: ({ chainId, userAddress }: UserChainQuery) => getUserPoolPositions({ chainId, userAddress }),
+  validationSuite: createValidationSuite((params: UserChainParams) => {
+    chainValidationGroup(params)
```
0xLokiB, line 16: This is repeated I assume

0xtutti (dismissed):

packages/prices-api/src/pools/index.ts
0xAlunara: Ah this already is the case; removed the `maybe`

apps/main/src/dex/queries/user-pool-positions.query.ts
0xAlunara: Only for locking veCRV by the looks of it, not enough to warrant its own generic validation suite yet imo

0xLokiB (approved):

0xtutti (approved):

PR #3168, opened 2026-09-07
Sync banner color with Figma

0xtutti:
## Summary

- Add `caution` banner severity with matching icon, styles, and sorting priority
- Align banner borders and action colors with design tokens
- Add alert feedback color tokens across light, dark, and Chad themes
- Expand Banner stories with caution and Figma variant coverage

## Testing

- Not run (not requested)


packages/ui/src/features/themes/design/1_surfaces_text.ts
```diff
@@ -1,5 +1,7 @@
 import { Blues, Grays, Greens, Reds, Violets, Yellows, Oranges } from './0_primitives'
 
+const FeedbackAlert = { Light: '#ff0606', Dark: '#ff6644', Chad: '#ff0606' } as const
```
0xLokiB, line 3: These should be in the 0_primitives I assume? Is that not the case in Figma?
Also, the different themes usually have different color definitions. The way this is declared globally is inconsistent.

0xLokiB (approved):

0xAlunara (approved):

PR #3174, opened 2026-09-07
refactor: removed unused health query

0xPearce:
- remove the unused market liquidation-band query from position-status calculation
  - proximity only depends on the user’s first band and oracle-price band
  - retain balance and band queries because the health delta cannot distinguish soft liquidation from below-range conversion
- disable new health-value queries when in legacy channel
- use position status to distinguish active `Soft Liquidation` below-range `Liquidation Protection`, and `Hard Liquidation`
- smaller liquidation badge 
Note: The alert copy still need to be updated according to the new health / liq. buffer

**Soft liquidation:**
Screenshot 2026-09-08 at 10 39 55

**Fully converted**
Screenshot 2026-09-08 at 10 50 44


0xtutti (dismissed):

apps/main/src/llamalend/llama.utils.ts
```diff
@@ -412,18 +412,8 @@ export function getLiquidationStatus(
...
-) {
-  if (userLiquidationBand !== null && typeof oraclePriceBand !== 'number') {
-    return false
-  } else if (typeof oraclePriceBand === 'number') {
-    return userFirstBand <= oraclePriceBand + 2
-  }
-  return false
-}
+export const getIsUserCloseToSoftLiquidation = (userFirstBand: number, oraclePriceBand: number | null | undefined) =>
+  typeof oraclePriceBand === 'number' && userFirstBand <= oraclePriceBand + 2
```
0xLokiB:
```suggestion
  oraclePriceBand != null && userFirstBand <= oraclePriceBand + 2
```

apps/main/src/llamalend/features/market-position-details/BorrowPositionDetails.tsx
```diff
@@ -17,17 +17,17 @@ export const BorrowPositionDetails = () => {
...
     getPositionStatusContent(collateralToken?.symbol, borrowToken?.symbol)[liquidationStatus.data]
   const softLiquidation = mapQuery(liquidationStatus, positionStatus => positionStatus === 'softLiquidation')
-  const useNewHealth = useNewLlamalendHealth()
   return (
     <Stack sx={{ padding: Spacing.md, gap: Spacing.xs }}>
       <Stack sx={{ gap: Spacing.sm }}>
         {useNewHealth ? (
-          <HealthDetails healthQuery={q(healthQuery)} />
+          <HealthDetails healthQuery={q(healthQuery)} positionStatus={liquidationStatus.data} />
         ) : (
```
0xLokiB, line 31: We should prefer to pass queries

apps/main/src/llamalend/features/market-position-details/health/HealthAndBufferBar.tsx
0xLokiB, line 1: The 400k% when liquidated in your screenshot doesn't make sense imo

0xPearce: Yes this is an edge case, but unrelated to this PR

apps/main/src/llamalend/features/market-position-details/health/HealthDetails.tsx
```diff
@@ -18,20 +19,26 @@ const PRIMARY_METRIC_SIZE = 2.5
 
 const HEALTH_PRECISION_THRESHOLD = 1.1
 
-export const HealthDetails = ({ healthQuery }: { healthQuery: HealthQuery }) => {
+export const HealthDetails = ({
+  health,
+  positionStatus,
+}: {
+  health: HealthQuery
```
0xPearce, line 26: I’m having second thoughts about the naming of the query props. As we’re passing them around more and more, I find the `Query` suffix a bit heavy to read every time. You can just hover over the prop, see the `QueryProp` type and know that it’s a query. What do you think?

0xLokiB: Definitely, I always try to avoid any query suffix

apps/main/src/llamalend/features/market-position-details/health/HealthAndBufferBar.tsx
```diff
@@ -65,22 +66,24 @@ const SEGMENT_CONFIG: Record<
...
 }: {
-  positionStatus: UserPositionStatus
+  positionStatus: QueryProp<UserPositionStatus>
   type: HealthType
-  query: HealthQuery
+  health: HealthQuery
 }) => {
   const { size, tooltip, getValue, getColor, getPercentage } = SEGMENT_CONFIG[type]
-  const { data, isLoading } = mapQuery(query, getValue)
+  // error already shown in the sibling Metric component
```
0xLokiB, line 78: This is fragile tho. I wouldn't mind seeing the error twice.

apps/main/src/llamalend/features/market-position-details/health/HealthAndBufferBar.tsx
```diff
@@ -65,22 +66,24 @@ const SEGMENT_CONFIG: Record<
...
+  const { data, isLoading: isHealthLoading } = mapQuery(health, getValue)
+  const isHealth = type === 'health'
```
0xLokiB, line 80: can we use objects to avoid ternaries everywhere that would break when we add another `type`?

apps/main/src/llamalend/features/market-position-details/health/HealthAndBufferBar.tsx
```diff
@@ -101,17 +104,33 @@ export const HealthAndBufferBar = ({
...
+                alignItems: 'center',
+                gap: Spacing.xs,
               }}
-            />
+            >
+              {statusError && <ErrorIconButton error={statusError} size="extraExtraSmall" />}
+              <WithSkeleton
+                loading={isStatusLoading}
+                variant="rectangular"
+                width="8rem"
```
0xLokiB: hardcoded width - is this even going to work on small screens?

apps/main/src/llamalend/features/market-position-details/health/HealthDetails.tsx
0xAlunara: Pearce discovering hate for Hungarian notation from first principle lol

apps/main/src/llamalend/features/market-position-details/health/HealthAndBufferBar.tsx
0xAlunara: It's only for the skeleton dimensions right? We could maybe omit the need for `width` altogether and let mui infer its size by rendering the badge and having placeholder values inside in case `label == null` (compared to not rendering it at all)

apps/main/src/llamalend/features/market-position-details/health/HealthAndBufferBar.tsx
0xPearce: I already thought about this but it's only for the health type of bar that we show the badge, I don't think an object is needed if it's just to have `undefined` for the other types

apps/main/src/llamalend/features/market-position-details/health/HealthAndBufferBar.tsx
0xPearce: ugh not sure how to show an error for a Bar component, will ask Tutti and do this separately

apps/main/src/llamalend/features/market-position-details/health/HealthAndBufferBar.tsx
0xPearce: I can lower the width of the Skeleton but yeah been repeating this kind of logic multiple times `variant="rectangular" width="8rem"`  maybe we can think about something better

apps/main/src/llamalend/features/market-position-details/health/HealthAndBufferBar.tsx
0xPearce: but for the testId you are right

0xAlunara (approved):

0xLokiB (approved):

PR #3175, opened 2026-09-07
refactor: evm-independent headers and footers

0xLokiB:
- Make the ConnectWalletModal and Header components independent on evm-functions.
- Separate app-specific logic from the UI elements.
- Resolve app links and wagmi calls before passing them to the UI elements.
- Banners and user profile are left for #3182


0xtutti (dismissed):

packages/ui/src/features/layout/PageLayout.tsx
0xPearce, line 1: nice 🔥

0xPearce (approved):
I thought Sonic leaking to Llamalend was introduced in your PR but it appears it's already in production 
Image

packages/ui/src/features/layout/PageLayout.tsx
0xLokiB: btw sonic comes from [here](https://prices.curve.finance/v1/lending/markets/sonic)

0xPearce: oh they are all deprecated that's why I wasn't seeing them, thanks

0xtutti (approved):

PR #3151, opened 2026-09-04
feat: llamalend rate breakdowns tables

0xPearce:
- add borrow cost and yield breakdown tables for LlamaLend rates with incentives, rebasing yield, token prices, boosted CRV totals, and points campaigns
- show both breakdowns for lend markets and only the borrow breakdown for mint markets
- refactor the points-campaign table and token source cell into shared EVM UI components reused by DEX and LlamaLend.

Screenshot 2026-09-07 at 16 31 40
Screenshot 2026-09-07 at 16 32 30

PR #3180, opened 2026-09-08
fix: show price query states in rate breakdown cells

0xPearce:
- depends on #3151
- add `TokenPriceCell` to show price loading and errors without blocking the table 

Screenshot 2026-09-08 at 13 11 07

Screenshot 2026-09-08 at 13 11 14

### DIFF
--- a/packages/evm-ui/src/shared/ui/DataTable/inline-cells/TokenPriceCell.tsx
+++ b/packages/evm-ui/src/shared/ui/DataTable/inline-cells/TokenPriceCell.tsx
@@ -8,7 +8,7 @@ import { InlineTableCell } from './InlineTableCell'
 /** Displays a price query without letting its loading or error state affect the rest of the table. */
 export const TokenPriceCell = ({ query: { data, error, isLoading } }: { query: QueryProp<number | undefined> }) => (
   <InlineTableCell sx={{ alignItems: 'end' }}>
-    <WithSkeleton loading={isLoading} sx={{ width: '3rem', height: '1lh' }}>
+    <WithSkeleton loading={isLoading} variant="rectangular" width="3rem" height="1lh">
       {data == null && error ? (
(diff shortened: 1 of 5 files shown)


packages/evm-ui/src/shared/ui/DataTable/inline-cells/TokenPriceCell.tsx
```diff
@@ -0,0 +1,19 @@
...
+import Typography from '@mui/material/Typography'
+import { WithSkeleton } from '@ui/components/WithSkeleton'
+import type { QueryProp } from '@ui/features/queries/util'
+import { InlineTableCell } from './InlineTableCell'
+
+/** Displays a price query without letting its loading or error state affect the rest of the table. */
+export const TokenPriceCell = ({ query: { data, error, isLoading } }: { query: QueryProp<number | undefined> }) => (
+  <InlineTableCell sx={{ alignItems: 'end' }}>
+    <WithSkeleton loading={isLoading} variant="rectangular" width="3rem" height="1lh">
+      {data == null && error ? (
```
0xLokiB, line 12: We should probably show the error + data in case they are both present tbh

0xLokiB (approved):

packages/evm-ui/src/shared/ui/DataTable/inline-cells/TokenPriceCell.tsx
0xPearce: can that happen?

0xLokiB: In some places yes it can, we can combine queries and provide partial data

0xLokiB (approved):

0xtutti (approved):

PR #3188, opened 2026-09-08
feat: llamalend max ROE column

0xPearce:
- depends on #3196
- add `Max RoE` column to LlamaLend markets, hidden by default and tooltip breakdown
- add reusable `getRoE` calculation (for future `ActionInfo`)
- derive borrow APY from borrow APR instead of getting it from the prices API to be able in the future to change the compounding frequency


apps/main/src/llamalend/queries/market-list/llama-markets.ts
```diff
@@ -54,7 +54,7 @@ export type LlamaMarket = {
   version: MarketVersion
   minBand?: number
   maxBand?: number
-  maxLtv: number
+  maxLtv: number | undefined
   loans: number
```
0xLokiB, line 58: The other props use null instead of undefined

apps/main/src/llamalend/features/market-list/columns/column.titles.ts
```diff
@@ -27,6 +27,7 @@ export const MARKET_TITLES: Record<MarketColumnId, string> = {
   [MarketColumnId.BorrowChart]: t`${AVERAGE_CATEGORIES['llamalend.marketList.rate'].period} Borrow APR`,
   [MarketColumnId.MaxLtv]: t`Max LTV`,
   [MarketColumnId.MaxLeverage]: t`Max Leverage`,
+  [MarketColumnId.MaxRoe]: t`Max RoE`,
```
0xLokiB, line 30: Another acronym ☠️ 
Is it explained somewhere?

0xPearce: it is explained in the tooltip

0xPearce:
@0xtutti since you are working on the tooltips copy maybe you have some feedback for this one

apps/main/src/llamalend/queries/market-list/llama-markets.ts
0xPearce: used `null`, but then used `undefined` for the columns so that they can be sorted last. I wanted to do the shortcut

apps/main/src/llamalend/features/market-list/columns/column.definitions.tsx
```diff
@@ -163,7 +163,7 @@ export const MARKET_COLUMNS = columnHelper.columns([
     },
     sortUndefined: 'last',
   }),
-  columnHelper.accessor('maxLtv', {
+  columnHelper.accessor(row => row.maxLtv ?? undefined, {
```
0xLokiB, line 166: Is the ?? necessary?

apps/main/src/llamalend/queries/market-list/llama-markets.ts
```diff
@@ -311,7 +311,7 @@ const convertMintMarket = (
         rebasingYieldApr: collateralToken.rebasingYieldApr,
       },
     },
-    maxLtv: maxLtv ?? undefined,
+    maxLtv: maxLtv ?? null,
```
0xLokiB: Is the ?? necessary?

0xLokiB (dismissed):

apps/main/src/llamalend/features/market-list/cells/MaxRoeCell.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const MaxRoeTooltipContent = ({ market }: { market?: LlamaMarket }) => (
+  <TooltipWrapper>
+    <TooltipDescription
+      text={t`The Maximum Return on Equity is an estimated annualized return on your own capital at maximum leverage, after borrowing costs.`}
+    />
+    <TooltipDescription text={t`Max RoE = M × C − (M − 1) × B`} />
+    <TooltipItems secondary>
+      <TooltipItem title={t`Max multiplier (M)`}>
+        {market && formatNumber(market.leverage, { unit: 'multiplier', abbreviate: false, fallback: '-' })}
```
0xAlunara, line 22: 3x `market` check could've been one at the top 😄

0xAlunara (dismissed):

apps/main/src/llamalend/features/market-list/cells/MaxRoeCell.tsx
0xPearce: the tooltip is also used without the `market` in the column header

0xAlunara (approved):

0xLokiB (approved):

PR #3195, opened 2026-09-09
refactor: evm-independent copy utilities

0xLokiB:
- Pass a formatting function to the copy utilities instead of depending directly on viem functions
- Stack:
  - #3182
  - #3175 
  - #3187
  - #3190
  - #3191
  - #3194
  - #3195 (this PR)
  - #3198
  - #3199
  - #3203

PR #3191, opened 2026-09-08
refactor: evm-free datatable

0xLokiB:
- Create evm datatable for existing users
- Moved datatable and its non-evm dependencies to ui package
- Stack:
  - #3182
  - #3175 
  - #3187
  - #3190
  - #3191 (this PR)
  - #3194
  - #3195
  - #3198
  - #3199

PR #3190, opened 2026-09-08
feat: basic stellar layout

0xLokiB:
- Create basic stellar routes for dex pools and pool page
- Move `redirectTo`, `useBodyThemeClass`, `useLayoutStoreResponsive` to ui
- Move and split basic css
- Update stellar root and network aware layouts to render the app
- Create placeholder data and work in progress components
- **Note**: The icons are missing from the [assets repository](https://github.com/curvefi/curve-assets/blob/main/chains/) for now.
  - I've refactored `default-crypto.png` and implemented it as fallback.
  - Fix is in https://github.com/curvefi/curve-assets/pull/1092
- Dependencies:
  - #3182
    - #3175 
      - #3187
- Follow-ups:
  - #3200
  - #3191
    - #3194
      - #3195
        - #3198
          - #3199

PR #3182, opened 2026-09-08
refactor: evm-independent profile and banners

0xLokiB:
- Split the global banners into evm-specific an evm-independent ones.
- Update UserProfile components to make them evm-independent.
- Depends on #3175.

PR #3199, opened 2026-09-09
feat: stellar wallet connection

0xLokiB:
- Implements wallet connection with stellar chain
- Stack:
  - #3182
  - #3175 
  - #3187
  - #3190
  - #3191
  - #3194
  - #3195
  - #3198
  - #3199 (this PR)
  - #3203
  - #3209


apps/stellar/src/features/connect-wallet/useWallet.ts
0xLokiB, line 1: I'd like to refactor this into proper mutations later on, just like wagmi

eslint.config.mjs
```diff
@@ -206,6 +204,27 @@ const config = [
     },
   },
 
+  // Keep direct wallet SDK access inside its wrapper.
+  {
+    ignores: ['apps/stellar/src/features/connect-wallet/stellar-wallet-kit.ts'],
```
0xLokiB, line 209: #3202 merges this with the rules above

apps/stellar/package.json
```diff
@@ -15,6 +15,7 @@
     "typecheck": "tsc --tsBuildInfoFile ../../.cache/typecheck/apps-stellar.tsbuildinfo"
   },
   "dependencies": {
+    "@creit-tech/stellar-wallets-kit": "jsr:^2.6.0",
```
0xLokiB, line 18: This comes from the jsr repository so the check supply chain fails.
However we have rules in yarn now and the supply check will not fail in other PRs. So we can solve that separately

apps/stellar/src/features/connect-wallet/address.ts
```diff
@@ -0,0 +1,14 @@
+import type { Address } from '@primitives/address.utils'
+import { maybe } from '@primitives/objects.utils'
+import { shortenString } from '@primitives/string.utils'
+
+export type StellarAddress = `G${string}`
```
0xAlunara, line 5: A g-string 👉👈😳

apps/stellar/src/features/connect-wallet/stellar-wallet-kit.ts
```diff
@@ -0,0 +1,23 @@
...
+  )
+
+export const refreshSupportedWallets = () => StellarWalletsKit.refreshSupportedWallets()
+
+export const connectWallet = async (connector: WalletConnector) => {
+  StellarWalletsKit.setWallet(connector.id)
+  await StellarWalletsKit.fetchAddress()
+}
+
+export const disconnectWallet = () => StellarWalletsKit.disconnect()
```
0xAlunara, line 23: Should this also unsubscribe from the event in `onWalletAddressChanged` or does the library do that for us on disconnect?

0xAlunara (dismissed):

apps/stellar/src/features/connect-wallet/address.ts
0xPearce: Alu 🤣

0xPearce: As this will probably be used a lot, couldn't we make it shorter? `SAddress` or `GAddress`

0xPearce (dismissed):

apps/stellar/src/features/connect-wallet/address.ts
0xLokiB: GAddress could be. But I generally try to avoid acronyms

apps/stellar/src/features/connect-wallet/stellar-wallet-kit.ts
0xLokiB: The subscription stays, I don't see why we'd need to cancel

0xAlunara (approved):

0xPearce (approved):

PR #3201, opened 2026-09-09
fix: always fetch gauge data

0xAlunara:
I think JJ thought this could be omitted in the new pool list, but we also use the (legacy) gauge data on the pool page itself as that still uses curve-js. Until we use prices API there as well we should always include gauge data when fetching pools.

You can test this by going into production and clicking a page with an inactive gauge. It won't show as inactive in the advanced details as it lacks this data.

### DIFF
--- a/apps/main/src/dex/hooks/useAutoRefresh.tsx
+++ b/apps/main/src/dex/hooks/useAutoRefresh.tsx
@@ -17,6 +17,6 @@ export const useAutoRefresh = (chainId: number | undefined) => {
 
   usePageVisibleInterval(async () => {
     if (!curveApi || !poolIds || !chainId) return
-    await fetchPools(curveApi, poolIds, true)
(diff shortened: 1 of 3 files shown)


apps/main/src/dex/hooks/useAutoRefresh.tsx
```diff
@@ -19,6 +17,6 @@ export const useAutoRefresh = (chainId: number | undefined) => {
 
   usePageVisibleInterval(async () => {
     if (!curveApi || !poolIds || !chainId) return
-    await fetchPools(curveApi, poolIds, !usePoolListV2)
+    await fetchPools(curveApi, poolIds, true)
```
0xLokiB: A naked true like that cannot be read properly

0xLokiB (approved):

0xPearce (approved):

PR #3204, opened 2026-09-09
feat: user pools position table

0xAlunara:
Adds a new user pool positions table on top of the pools list. For now by default it shows just the pool name, net APR and the new deposits column.

The USD values are not returned by the API, so we'll have to use `useTokenUsdRates` for the time being. Using `useTokenUsdRates` may take a bit of time, which shouldn't block the table in a loading state. Because of this, both sorting and pagination happen client side, even though there's pagination in the endpoint call already.

A good address for testing is Convex with `0x989AEb4d175e16225E39E87d0D97A3360524AD80`.
I did not decide to hide this feature behind a beta flag; if it works it's good enough to immediately roll out and extend later.

image

### DIFF
--- a/apps/main/src/dex/features/pool-list/PoolsTable.tsx
+++ b/apps/main/src/dex/features/pool-list/PoolsTable.tsx
@@ -26,15 +26,19 @@ import { usePoolsGlobalFilterFn } from './hooks/usePoolsGlobalFilter'
 import { usePoolsPagination } from './hooks/usePoolsPagination'
 import { usePoolsSorting } from './hooks/usePoolsSorting'
 import { usePoolsTable } from './hooks/usePoolsTable'
-import { type PoolColumnVariant, usePoolsVisibility } from './hooks/usePoolsVisibility'
-import type { PoolRow } from './types'
+import { usePoolsVisibility } from './hooks/usePoolsVisibility'
+import type { PoolRow, PoolTableMeta } from './types'
 
 const LOCAL_STORAGE_KEY = 'dex-pool-list'
 const EMPTY_POOL_ROWS: readonly PoolRow[] = []
-const POOL_EXPANDED_PANEL_BODIES = {
-  full: props => <PoolExpandedPanel {...props} variant="full" />,
-  lite: props => <PoolExpandedPanel {...props} variant="lite" />,
-} satisfies Record<PoolColumnVariant, ExpandedPanelComponent<PoolRow>>
+
+const FullPoolExpandedPanel: ExpandedPanelComponent<PoolRow> = ({ row }) => (
+  <PoolExpandedPanel pool={row.original} variant="full" />
+)
+
+const LitePoolExpandedPanel: ExpandedPanelComponent<PoolRow> = ({ row }) => (
+  <PoolExpandedPanel pool={row.original} variant="lite" />
+)
 
 export const PoolsTable = ({ network }: { network: NetworkConfig }) => {
   const isLite = isLiteChain(network.chainId)
@@ -52,8 +56,8 @@ export const PoolsTable = ({ network }: { network: NetworkConfig }) => {
 
   const [expanded, setExpanded] = useState<ExpandedState>({})
   const { columnSettings, columnVisibility, toggleVisibility, variant } = usePoolsVisibility(LOCAL_STORAGE_KEY, {
-    isLite,
-    sorting,
+    variant: isLite ? 'lite' : 'full',
+    mobileColumn: sortField,
   })
 
   const { isFetching, onReload, pageCount, userHasPositions, tableQuery } = usePoolsTable({
@@ -73,7 +77,7 @@ export const PoolsTable = ({ network }: { network: NetworkConfig }) => {
   const table = useCurveTable({
     columns: POOL_COLUMNS,
     query: tableQuery,
-    meta: { getRowHref: ({ url }) => url },
+    meta: { getRowHref: ({ url }) => url, variant } as PoolTableMeta,
     state: { expanded, sorting, columnVisibility, globalFilter, ...(!isLite && { pagination, columnFilters }) },
     getRowId: row => row.address,
     onExpandedChange: setExpanded,
@@ -100,7 +104,10 @@ export const PoolsTable = ({ network }: { network: NetworkConfig }) => {
           secondaryButton: { label: t`Telegram`, href: CURVE_SOCIALS.telegram.en },
         }}
         errorState={{ title: t`Unable to retrieve pool list`, description: tableQuery.error?.message, onReload }}
-        expandedPanel={{ Body: POOL_EXPANDED_PANEL_BODIES[variant], Actions: PoolExpandedPanelActions }}
+        expandedPanel={{
+          Body: isLite ? LitePoolExpandedPanel : FullPoolExpandedPanel,
+          Actions: PoolExpandedPanelActions,
+        }}
         shouldStickFirstColumn={Boolean(useIsTablet() && userHasPositions)}
       >
         <TableFilters
--- a/apps/main/src/dex/features/pool-list/columns/column.options.ts
+++ b/apps/main/src/dex/features/pool-list/columns/column.options.ts
@@ -1,40 +1,71 @@
+import { recordEntries } from '@primitives/objects.utils'
 import type { VisibilityGroup } from '@ui/features/tables/visibility.types'
 import { t } from '@ui/lib/i18n'
 import { POOL_TITLES } from './column.titles'
 import { PoolColumnId } from './columns.enum'
 
-const createVisibility = ({ isLite }: { isLite: boolean }): VisibilityGroup<PoolColumnId>[] => [
+type ColumnVisibility = Record<PoolColumnId, { active: boolean; enabled: boolean }>
+
+const FULL_COLUMN_VISIBILITY = {
+  [PoolColumnId.PoolName]: { active: true, enabled: true },
+  [PoolColumnId.Tokens]: { active: false, enabled: true },
+  [PoolColumnId.NetRate]: { active: true, enabled: true },
+  [PoolColumnId.BaseRate]: { active: false, enabled: true },
+  [PoolColumnId.WeeklyBaseRate]: { active: false, enabled: true },
+  [PoolColumnId.CrvRate]: { active: false, enabled: true },
+  [PoolColumnId.RewardsRate]: { active: false, enabled: true },
+  [PoolColumnId.Points]: { active: false, enabled: true },
+  [PoolColumnId.Volume]: { active: true, enabled: true },
+  [PoolColumnId.Tvl]: { active: true, enabled: true },
+  [PoolColumnId.Age]: { active: false, enabled: true },
+  [PoolColumnId.Deposits]: { active: false, enabled: false },
+} satisfies ColumnVisibility
+
+const LITE_COLUMN_VISIBILITY = {
+  [PoolColumnId.PoolName]: { active: true, enabled: true },
+  [PoolColumnId.Tokens]: { active: false, enabled: true },
+  [PoolColumnId.NetRate]: { active: true, enabled: true },
+  [PoolColumnId.BaseRate]: { active: false, enabled: false },
+  [PoolColumnId.WeeklyBaseRate]: { active: false, enabled: false },
+  [PoolColumnId.CrvRate]: { active: false, enabled: true },
+  [PoolColumnId.RewardsRate]: { active: false, enabled: true },
+  [PoolColumnId.Points]: { active: false, enabled: true },
+  [PoolColumnId.Volume]: { active: false, enabled: false },
+  [PoolColumnId.Tvl]: { active: true, enabled: true },
+  [PoolColumnId.Age]: { active: false, enabled: false },
+  [PoolColumnId.Deposits]: { active: false, enabled: false },
+} satisfies ColumnVisibility
+
+const USER_POSITIONS_COLUMN_VISIBILITY = {
+  [PoolColumnId.PoolName]: { active: true, enabled: true },
+  [PoolColumnId.Tokens]: { active: false, enabled: false },
+  [PoolColumnId.NetRate]: { active: true, enabled: true },
+  [PoolColumnId.BaseRate]: { active: false, enabled: true },
+  [PoolColumnId.WeeklyBaseRate]: { active: false, enabled: true },
+  [PoolColumnId.CrvRate]: { active: false, enabled: true },
+  [PoolColumnId.RewardsRate]: { active: false, enabled: true },
+  [PoolColumnId.Points]: { active: false, enabled: true },
+  [PoolColumnId.Volume]: { active: false, enabled: false },
+  [PoolColumnId.Tvl]: { active: false, enabled: false },
+  [PoolColumnId.Age]: { active: false, enabled: false },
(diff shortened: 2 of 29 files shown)


packages/prices-api/src/pools/schema.ts
```diff
@@ -48,17 +48,6 @@ const poolTotals = z
   .transform(camelizeKeys)
   .transform(({ totalTvl, ...data }) => ({ ...data, tvl: totalTvl }))
 
-const userPoolPosition = z
```
0xAlunara, line 51: Moved down to after the definition of the shared pool schema

apps/main/src/dex/features/pool-list/columns/column.options.ts
```diff
@@ -1,42 +1,75 @@
+import { recordEntries } from '@primitives/objects.utils'
 import type { VisibilityGroup } from '@ui/features/tables/visibility.types'
 import { t } from '@ui/lib/i18n'
 import { POOL_TITLES } from './column.titles'
 import { PoolColumnId } from './columns.enum'
 
-const createVisibility = ({ isLite }: { isLite: boolean }): VisibilityGroup<PoolColumnId>[] => [
+type ColumnVisibility = Record<PoolColumnId, { active: boolean; enabled: boolean }>
+
+const FULL_COLUMN_VISIBILITY = {
```
0xAlunara: Having a single constructor function with conditionals got a bit hairy, so while it's a lot more verbose I think having three distinct mappings for the three variants is easier to grok

packages/ui/src/features/tables/CenteredEmptyState.tsx
```diff
@@ -0,0 +1,11 @@
+import type { ReactNode } from 'react'
+import Stack from '@mui/material/Stack'
+import { SizesAndSpaces } from '@ui/features/themes/design/1_sizes_spaces'
+
+const { Spacing } = SizesAndSpaces
+
+export const CenteredEmptyState = ({ children }: { children: ReactNode }) => (
```
0xAlunara, line 7: Copy pasted from llamalend so we can use it in the new pools position table as well

apps/main/src/dex/features/pool-list/PoolsTable.tsx
```diff
@@ -26,15 +26,19 @@ import { usePoolsGlobalFilterFn } from './hooks/usePoolsGlobalFilter'
...
+import type { PoolRow, PoolTableMeta } from './types'
 
 const LOCAL_STORAGE_KEY = 'dex-pool-list'
 const EMPTY_POOL_ROWS: readonly PoolRow[] = []
-const POOL_EXPANDED_PANEL_BODIES = {
-  full: props => <PoolExpandedPanel {...props} variant="full" />,
-  lite: props => <PoolExpandedPanel {...props} variant="lite" />,
-} satisfies Record<PoolColumnVariant, ExpandedPanelComponent<PoolRow>>
+
+const FullPoolExpandedPanel: ExpandedPanelComponent<PoolRow> = ({ row }) => (
```
0xAlunara, line 35: Mapping doesn't work anymore as `userPositions` is a variant now too

apps/main/src/dex/features/pool-list/hooks/usePoolsTable.ts
```diff
@@ -37,93 +26,8 @@ class UnsupportedPoolListError extends Error {
   }
 }
 
-/** Maps Prices API pool data into the source-independent pool-list model. */
-const poolToRowData = (pool: V2Pool): PoolRowData => ({
```
0xAlunara, line 41: Moved this shit to `utils.ts` to make this hook less complex and also re-use for user positions table

apps/main/src/dex/features/pool-list/hooks/usePoolsTable.ts
```diff
@@ -146,84 +50,65 @@ export const usePoolsTable = ({
...
-    searchString: searchText || undefined,
-    ...filters,
-    sortBy,
-    sortDirection,
-  }
-  const poolListQuery = usePoolList(poolListParams, !isLite && isSupported)
-  const poolListTableQuery = useMappedQuery(
-    poolListQuery,
-    useCallback(({ pools }) => pools.map(pool => toPoolRow(poolToRowData(pool))), [toPoolRow]),
+  // constQ suppresses loading state, and ?? null allows useCombinedQueries to run even when data is not yet loaded or present.
```
0xLokiB, line 79: This explain what's done but not why. It seems wrong to ignore loading state

0xLokiB: NB the datatable supports showing data and loading at the same time.

apps/main/src/dex/features/pool-list/hooks/usePoolsTable.ts
```diff
@@ -146,84 +50,65 @@ export const usePoolsTable = ({
...
+        ),
+      [],
+    ),
   )
 
-  const tableQuery = poolListSupportQuery.data
-    ? isLite
-      ? litePoolListTableQuery
-      : poolListTableQuery
+  const tableQuery = poolListSupport.data
```
0xLokiB, line 94: Isn't this fallbackQ?

apps/main/src/dex/features/pool-list/hooks/useUserPositionsTable.ts
```diff
@@ -0,0 +1,54 @@
...
+  const campaigns = useCampaigns({ blockchainId })
+  const positions = useUserPoolPositions({ chainId, userAddress })
+
+  const { data: tokenAddresses } = useMappedQuery(
+    positions,
+    useCallback(({ positions }) => positions.map(({ lpTokenAddress }) => lpTokenAddress), []),
+  )
+  const tokenRates = useTokenUsdRates({ chainId, tokenAddresses }, !!userAddress)
+
+  // constQ suppresses loading state, and ?? null allows useCombinedQueries to run even when data is not yet loaded or present.
```
0xLokiB, line 27: It seems repeated and unnecessary. But it really needed a q helper would be much cleaner

apps/main/src/dex/features/pool-list/components/PoolExpandedPanel.tsx
```diff
@@ -123,6 +120,18 @@ export const PoolExpandedPanel = ({ row, variant }: PoolExpandedPanelProps) => {
           />
         </Grid>
       )}
+      {variant === 'userPositions' && (
```
0xPearce, line 123: We should probably be using `MetricsGrid` but outside the scope of this PR

packages/ui/src/features/tables/CenteredEmptyState.tsx
0xPearce: Nice I have a task to center the empty position on all tables (not centered on Firefox)

0xPearce (dismissed):
We show `$0` in llamalend user position summary's metrics when the wallet isn't connected, here we show `N/A` but we should be consistent imo Image

apps/main/src/dex/features/pool-list/utils.ts
```diff
@@ -0,0 +1,90 @@
...
+import { isVyperVulnerablePool } from './alerts'
+import type { PoolRow, PoolRowData } from './types'
+
+/** Maps Prices API pool data into the source-independent pool-list model. */
+export const poolToRowData = (
+  pool: Pick<V2Pool, 'address' | 'name' | 'tradeableCoins' | 'extraRewardsApr'> & Partial<V2Pool>,
+): PoolRowData => ({
+  address: pool.address,
+  baseDailyApr: pool.baseDailyApr ?? undefined,
+  baseWeeklyApr: pool.baseWeeklyApr ?? undefined,
```
0xLokiB, line 16: That's a lot of ??undefined. And we use null for empty fields in the other tables, if I'm not mistaken

apps/main/src/dex/features/pool-list/utils.ts
```diff
@@ -0,0 +1,90 @@
...
+import type { LitePool, V2Pool } from '@curvefi/prices-api/pools'
+import type { CampaignRewards } from '@evm-ui/entities/campaigns'
+import { DEX_ROUTES } from '@evm-ui/shared/routes'
+import { notFalsy } from '@primitives/objects.utils'
+import { isVyperVulnerablePool } from './alerts'
+import type { PoolRow, PoolRowData } from './types'
+
+/** Maps Prices API pool data into the source-independent pool-list model. */
+export const poolToRowData = (
+  pool: Pick<V2Pool, 'address' | 'name' | 'tradeableCoins' | 'extraRewardsApr'> & Partial<V2Pool>,
```
0xLokiB, line 12: Obj destructing would clean this up I think

apps/main/src/dex/features/pool-list/utils.ts
```diff
@@ -0,0 +1,90 @@
...
+          name: reward.name,
+          price: reward.tokenPrice,
+          symbol: reward.symbol,
+        },
+      ),
+    ),
+    gauge: gauges[0],
+    gauges,
+    isMetapool: pool.isMetaPool,
+    name: pool.name ?? '',
```
0xLokiB: ??'' 😭

packages/ui/src/features/tables/CenteredEmptyState.tsx
0xLokiB: This doesn't look empty state specific at all 
And didn't I already move the empty state component anyway?

apps/main/src/dex/features/pool-list/PoolsTable.tsx
```diff
@@ -100,7 +104,10 @@ export const PoolsTable = ({ network }: { network: NetworkConfig }) => {
           secondaryButton: { label: t`Telegram`, href: CURVE_SOCIALS.telegram.en },
         }}
         errorState={{ title: t`Unable to retrieve pool list`, description: tableQuery.error?.message, onReload }}
-        expandedPanel={{ Body: POOL_EXPANDED_PANEL_BODIES[variant], Actions: PoolExpandedPanelActions }}
+        expandedPanel={{
+          Body: isLite ? LitePoolExpandedPanel : FullPoolExpandedPanel,
```
0xLokiB, line 108: I'm thinking we might wanna have a total separate implementation of the list for lite. Wdyt?

0xAlunara: No, I went specifically out of the way to have one shared table for all three variants as they share so many things.

apps/main/src/dex/features/pool-list/hooks/usePoolsTable.ts
0xAlunara: I don't see the issue. One query is the basic data required for the table, the others are just auxiliary data supplying extra but not required information.

In the end you have a final query giving the final aggregated data, but only one subquery should show a loading state. It would be silly to show the table as loading while campaign data comes in.

`combineQueries` really simplifies the merging of all this data, but one quirk is that sets the query to loading if one of its subqueries is. I find using `constQ` easier than expanding it with options, keeps it simple and untouched.

apps/main/src/dex/features/pool-list/PoolsTable.tsx
0xLokiB: The columns are different, the pagination is different, what is it they share?
We can have reusable columns

apps/main/src/dex/features/pool-list/hooks/usePoolsTable.ts
0xLokiB: It would be more correct to show the table as loading while campaign data comes in. It's not OK to show the data as ready when it's still going to change

0xtutti (dismissed):

apps/main/src/dex/features/pool-list/hooks/usePoolsTable.ts
0xLokiB: Anyway, not a blocker. We can discuss this together in a separate moment, but it would be nice to use consistent patterns.

apps/main/src/dex/features/pool-list/PoolsTable.tsx
0xAlunara: Only temporary; lite pools will have server-side everything too as soon as Oleg gets to it. I don't want to extract a table that works fine only to glue it back together later again.

apps/main/src/dex/features/pool-list/hooks/usePoolsTable.ts
0xAlunara: Fwiw campaign data is non-blocking in the existing table list as well, so making this table non-blocking would be consistent 😄

apps/main/src/dex/features/pool-list/PoolsTable.tsx
0xLokiB: Then I'll need to exclude curvejs/evm dependencies from the whole thing - the lite table must work without evm dependencies

apps/main/src/dex/features/pool-list/components/PoolExpandedPanel.tsx
0xAlunara: Yeah, need to fix conflicts so might as well do it. That PR wasn't merged yet when this was made.

packages/ui/src/features/tables/CenteredEmptyState.tsx
0xAlunara: What you moved was the `EmptyStateCard`, which is a bit different and more complex. Horrible naming though, we should probably harmonize this with Pearce's PR as it's very confusing rn

apps/main/src/dex/features/pool-list/hooks/usePoolsTable.ts
0xAlunara: Technically yes, but our pool list uses `keepPreviousData` to keep old data visible while refetching. But this is independant of the query key, so if we first show a full list of chain 1 and then switch to lite chain 2, the list of chain 1 would still be shown when using `fallbackQ` as chain 1 still has data.

But that's weird as well; I'll make a diff ticket to tackle that problem altogether.

apps/main/src/dex/features/pool-list/utils.ts
0xAlunara: This is us masking inconsistencies in the Prices API. Sometimes they return `null`, sometimes they return `undefined`. We merely force it to be `undefined`, there is shouldn't be an extra `null` option.

apps/main/src/dex/features/pool-list/utils.ts
0xAlunara: I tried to but the lines of code explode because of the many repeated property names. Not much better.

apps/main/src/dex/features/pool-list/utils.ts
0xAlunara: I don't see the issue, even though the string type can be `undefined` want to make sure it's actually not, and `null` and `undefined` are converted to a proper `string`. I guess I could make it `<No Name>`

apps/main/src/dex/features/pool-list/hooks/useUserPositionsTable.ts
0xAlunara: Idk, `useCombinedQueries` is the helper in this case; it's a bit akin to `zip` for queries where we just pass an array of properties and we combine it into one single entity.

0xtutti (approved):

0xPearce (approved):

PR #3207, opened 2026-09-10
refactor: add generic MetricsGrid component

0xAlunara:
Slightly changes the sizing of a few existing metric grids from 5 to 4 columns, but all in the name of consistency.

image


apps/main/src/llamalend/widgets/MarketMetricGrid.tsx
```diff
@@ -1,35 +0,0 @@
...
-      mobile: 'repeat(1, minmax(0, 1fr))',
-      tablet: 'repeat(2, minmax(0, 1fr))',
-      desktop: 'repeat(4, minmax(0, 1fr))',
-    },
-  },
-} as const
-
-/**
- * TODO: 1. there's a notion ticket to create a proper metrics grid component
- * 2. Ideally, we should define the category of a Metric in the parent, and make the Metric children inherit,
```
0xAlunara, line 26: Decided against point #2 for the sake of avoiding coupling

packages/ui/src/components/MetricsGrid.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { SizesAndSpaces } from '@ui/features/themes/design/1_sizes_spaces'
+import { applySxProps } from '@ui/lib/mui'
+
+const { Grid } = SizesAndSpaces
+
+const columnTemplate = (count: number) => `repeat(${count}, minmax(0, 1fr))`
+
+const METRICS_GRID_VARIANTS = {
+  default: { gridTemplateColumns: { mobile: columnTemplate(2), tablet: columnTemplate(4) } },
+  mobileRows: {
```
0xPearce, line 11: don't know about the breakpoint in the variant's name 🤔  Couldn't we use something similar to the `Metric's` variant name since we are assuming the children to be that?

0xPearce: Maybe we can make the `MetricsGrid` a React context and `Metric` read from that context so we define the category/variant only once and not for the wrapper and the children? Just thinking out loud

apps/main/src/llamalend/features/market-list/UserPositionsSummary.tsx
```diff
@@ -2,43 +2,39 @@ import { useMemo } from 'react'
...
+    <MetricsGrid
+      variant="fillMobile"
       sx={{ paddingBlock: Spacing.sm, paddingInline: Spacing.md, backgroundColor: t => t.design.Layer[1].Fill }}
     >
       {summary.map((item, index) => (
-        // eslint-disable-next-line @eslint-react/no-array-index-key -- Existing violation before enabling this rule.
-        <UserPositionStatisticItem key={index} itemSize={{ mobile: 12, tablet: 12 / summary.length }} {...item} />
+        <Metric
+          // eslint-disable-next-line @eslint-react/no-array-index-key
+          key={index}
```
0xLokiB: Can't we use the market id as key?

packages/ui/src/components/MetricsGrid.tsx
0xLokiB: The categories should be functional i.e. they shouldn't be describing how they look like but what's their function in the app. Variants are purely visual so I think it's fine. I don't think we need a context

0xLokiB (dismissed):

apps/main/src/llamalend/features/market-list/UserPositionsSummary.tsx
0xAlunara: Huh? This is about these values, they're not per market

0xLokiB: ah lol ofc.. how about the label?

0xLokiB (dismissed):

apps/main/src/llamalend/features/market-list/UserPositionsSummary.tsx
0xAlunara: Works for me

0xPearce (approved):

0xtutti (approved):

PR #3212, opened 2026-09-11
fix: improve mobile DEX performance

0xPearce:
In the DEX table with 50 rows, opening one panel caused about 500 row renders. The panel’s animation repeatedly changed the table’s height, triggering sticky-header updates because `useResizeObserver` was watching both height and width. Each update caused the table to render its rows again, even though its width had not changed.

Watching only width reduced this to 50 row renders. All 50 rows still render once because opening a panel updates the table’s shared expansion state, causing the parent table and its rows to render again. We could potentially improve this further by allowing unchanged rows to skip rendering, so opening one panel would only render the affected row.


0xAlunara (approved):
Oh shit was already wondering myself why the animations were so sluggish. Smooth as butter now 🧈

mimaklas (approved):

PR #3222, opened 2026-09-14
fix: card sizes and padding

0xPearce:
This started as a small fix but now refactors the `CardHeader` and `Card` usage to make the apps feel more consistent 

- fix the height of the LlamaLend user position table
- `CardHeader` default size is `small`, moved the size to the `Card` parent rather than the `CardHeader` child
- separate card sizes from inline variants so each variant can be used with different sizes. The `inline` card variant removes the background color and inline padding and adds a bottom border to the header
- split pool information into stacked cards like LlamaLend
Screenshot 2026-09-15 at 13 02 22
Screenshot 2026-09-15 at 13 02 14


apps/main/src/llamalend/features/market-list/UserPositionsMarketRateTable.tsx
0xLokiB: Sizing is now unused

0xLokiB (dismissed):

packages/evm-ui/src/widgets/ListPageWrapper.tsx
```diff
@@ -1,22 +1,15 @@
 import type { ReactNode } from 'react'
 import { Stack } from '@mui/material'
 import Box from '@mui/material/Box'
+import { PAGE_BLOCK_MARGIN } from '@ui/features/layout/constants'
 import { SizesAndSpaces } from '@ui/features/themes/design/1_sizes_spaces'
 import { useIsTiny } from '@ui/hooks/useBreakpoints'
 
 const { Spacing, MaxWidth, MinHeight } = SizesAndSpaces
 
 export const ListPageWrapper = ({ footer, children }: { children: ReactNode; footer?: ReactNode }) => (
```
0xLokiB: Unrelated, but wrapper doesn't add anything
```suggestion
export const ListPage = ({ footer, children }: { children: ReactNode; footer?: ReactNode }) => (
```

0xLokiB (dismissed):

packages/evm-ui/src/widgets/ListPageWrapper.tsx
0xAlunara: `ListPageWrapperBoxContainer`

packages/ui/src/features/themes/components/card-header/mui-card-header.ts
```diff
@@ -1,55 +1,95 @@
...
+import type { CardHeaderProps } from '@mui/material/CardHeader'
 import type { Components, TypographyVariantsOptions } from '@mui/material/styles'
+import { objectKeys } from '@primitives/objects.utils'
 import { handleBreakpoints } from '@ui/features/themes/basic-theme'
 import { DesignSystem } from '@ui/features/themes/design'
+import { TRANSPARENT } from '@ui/features/themes/design/0_primitives'
 import { SizesAndSpaces } from '@ui/features/themes/design/1_sizes_spaces'
 
-const { BorderWidth, Spacing, Sizing, ButtonSize, Tab } = SizesAndSpaces
+const { ButtonSize, Spacing, Sizing: ResponsiveSizing, Tab } = SizesAndSpaces
```
0xAlunara, line 9: Not sure why we are renaming `Sizing` to `ResponsiveSizing`

0xPearce: because there was another Sizing in the meantime but I removed it after, forgot to change back the name

packages/evm-ui/src/widgets/ListPageWrapper.tsx
0xLokiB: :sob:

0xLokiB (approved):

0xtutti (approved):

PR #3224, opened 2026-09-15
fix: remove curve router from syrup market

0xPearce:
- removed Curve router option from syrup market leverage providers as the router need to be whitelisted to wrap the token


tests/cypress/component/llamalend/market-constants.cy.tsx
```diff
@@ -44,7 +44,7 @@ const mountMarketAlert = ({
 const ALL_MARKET_ALERTS = recordValues(MARKETS_ALERTS)
 const ALL_DEPRECATED_LLAMAS = recordValues(DEPRECATED_LLAMAS)
 const STABLE_LEVERAGE_MARKETS = {
-  [Chain.Ethereum]: ['0x2fb54c8eae57767A9A509A395b9C4FA0702e2675', '0xC77d97cF01737EB7aCE46cAb7cd9F60eC51a40c0'],
+  [Chain.Ethereum]: ['0x3cD4d86a2c65e57ce4b4121b67E2D2224BA41bbe', '0xC77d97cF01737EB7aCE46cAb7cd9F60eC51a40c0'],
```
0xLokiB, line 47: is it really a good idea to duplicate config in tests?

0xLokiB (approved):

tests/cypress/component/llamalend/market-constants.cy.tsx
0xPearce: In one of my next task I will add to all llamalend markets a type (like crypto, stable, etc), I will do it then

0xtutti (approved):

PR #3232, opened 2026-09-15
fix: exception for pool advanced details padding block no longer needed

0xAlunara:
Appears that this exception due to inline card headers is no longer needed.

**Before**
image

**After**
image


0xLokiB (approved):

0xPearce (approved):
fixed it as well here 😄  https://github.com/curvefi/curve-frontend/pull/3222

PR #3233, opened 2026-09-15
refactor: tokens query and api endpoint

0xAlunara:
Instead of

1. Hydrating
2. Creating the pools mapper
3. Creating the tokens mapper
4. Fetching token volumes

We create an `/api/router/v1/tokens` endpoint for the router API that generates the token list using curve-js, and hydrates it with volume data when available. The front-end can then just use a TanStack query to fetch, cache and refresh that data. Some legacy code can then be simplified with `useMappedQuery` rather than raw `useMemo`.

Byebye token slice


apps/main/src/dex/components/PageCreatePool/TokensInPool/index.tsx
```diff
@@ -82,37 +82,46 @@ export const TokensInPool = ({ curve, chainId, haveSigner }: Props) => {
...
-  }, [tokensMapper, createQuickList, userAddedTokens, basePools])
+  const tokens = useTokens({ chainId })
+  const { data: selTokens = uniqBy(userAddedTokens, token => token.address) } = useMappedQuery(
+    tokens,
+    useCallback(
+      (tokens): CreateToken[] => {
+        const catalogTokens = Object.entries(tokens)
+          .map(([address, token]) => ({
+            ...token,
+            address: address.toLowerCase(),
```
0xAlunara, line 93: Some of the legacy code is still scuffed but I don't want to touch that shit (yet)

apps/main/src/dex/components/PageCreatePool/types.ts
```diff
@@ -18,9 +18,8 @@ import type { Decimal } from '@primitives/decimal.utils'
 export type CreateToken = {
   address: string
   symbol: string
-  ethAddress?: string
```
0xAlunara, line 21: These two properties weren't used at all on this page

apps/main/src/dex/utils/utilsSwap.ts
```diff
@@ -190,3 +196,58 @@ export async function routerGetToStoredRate(routes: IRoute, curve: CurveApi, toA
 
   return toStoredRate
 }
+
+export function getRouterSwapsExchangeRate(
```
0xAlunara, line 200: This has been extracted from the pools slice and is untouched aside from how the old `tokenMapper` was being used. I don't dare to touch and refactor it further rn.

apps/main/src/dex/features/pool-list/hooks/useLegacyPoolsTable.tsx
```diff
@@ -58,17 +56,12 @@ export function useLegacyPoolsTable({ blockchainId: network, chainId }: NetworkC
   const fetchPoolsRewardsApy = useStore(state => state.pools.fetchPoolsRewardsApy)
   const fetchMissingPoolsRewardsApy = useStore(state => state.pools.fetchMissingPoolsRewardsApy)
   const poolsData = useMemo(() => poolDataMapper && recordValues(poolDataMapper), [poolDataMapper])
-  const isLite = isLiteChain(chainId)
 
   const { address: userAddress } = useConnection()
   const { data: userPools } = useUserPools({ chainId, userAddress })
-  const { data: volumes, isLoading: isVolumesLoading } = usePoolVolumes({ chainId })
```
0xAlunara, line 65: Sorry, but I intend to delete this legacy pool list soon anyway as it's a drag on refactoring, so I'm not feeling sad removing volume support.

0xtutti (dismissed):

0xPearce (dismissed):
Nice!

apps/main/src/dex/components/PagePool/Swap/index.tsx
```diff
@@ -110,20 +105,17 @@ export const Swap = ({
...
-  const { selectList, swapTokensMapper } = useMemo(() => {
-    const { selectList, swapTokensMapper } = getSwapTokens(tokensMapper, poolData)
-    return {
-      selectList: selectList.map<TokenOption>(token => ({
-        address: token.address as Address, // not checksummed!
-        symbol: token.symbol,
+  const tokens = useMemo(
+    () =>
+      poolData.tokenAddresses.map<TokenOption>((address, index) => ({
+        address: address as Address, // not checksummed!
```
0xLokiB, line 111: So lower case?

apps/main/src/dex/components/PageRouterSwap/index.tsx
```diff
@@ -109,16 +107,40 @@ export const QuickSwap = ({
...
-  const slippageType = routesAndOutput && getSlippageType(routesAndOutput)
+  const quote = userAddress ? rpcRoutesAndOutput : apiRoutes
+  const slippageType = quote && getSlippageType(quote)
   const storeSlippage = useUserProfileStore(state => state.maxSlippage)
   const maxSlippage = slippageType && storeSlippage[slippageType]
+
+  const tokensMapper = useTokens({ chainId })
+  const routesAndOutput = useMemo(() => {
+    if (!fromToken || !toToken) return undefined
+    if (!userAddress) return apiRoutes
```
0xLokiB, line 124: I think we can move this one up and merge the undefined cases.
It feels like the code is too defensive btw, what can cause confusion and hide bugs. I think it could be simplified, but I haven't tried so I'm not sure.

apps/main/src/dex/hooks/useRouterApi.ts
```diff
@@ -77,74 +73,105 @@ const convertRoute = (
...
+      return
+
+    const nextError = error?.message ?? (data?.length === 0 ? 'error-swap-not-available' : '')
+    const outputAmount =
+      route && (isFrom ? fromWei(route.amountOut[0], toDecimals) : fromWei(route.amountIn[0], fromDecimals))
+    if (
+      quickSwap.formStatus.error === nextError &&
+      (outputAmount == null || outputAmount === (isFrom ? currentForm.toAmount : currentForm.fromAmount))
+    )
+      return
```
0xLokiB, line 140: All these complicated checks and returning undefined smell a bit do we really need them?

apps/main/src/dex/hooks/useRouterApi.ts
```diff
@@ -77,74 +73,105 @@ const convertRoute = (
...
 
-  return {
-    data: useMemo(
-      () => route && convertRoute(route, { fromAddress, toAddress, chainId, isPending }),
-      [route, fromAddress, toAddress, isPending, chainId],
+  return useMappedQuery(
+    { data: route, isLoading: enabled && (tokensLoading || isLoading), error: tokensError ?? error },
+    useCallback(
+      route =>
+        enabled && !tokensError && fromAddress && toAddress && fromDecimals != null && toDecimals != null && tokens
```
0xLokiB, line 154: Same here

apps/main/src/dex/queries/tokens.query.ts
```diff
@@ -0,0 +1,38 @@
...
+export const { useQuery: useTokens } = queryFactory({
+  queryKey: ({ chainId }: ChainParams) => [...rootKeys.chain({ chainId }), 'tokens'] as const,
+  queryFn: ({ chainId }: ChainQuery) => fetchJson<TokenMapper>(`/api/router/v1/tokens${addQueryString({ chainId })}`),
+  validationSuite: createValidationSuite(chainValidationGroup),
+  category: 'dex.pools',
+})
+
+/** Look up token data from potentially unsanitized input. */
+export const getToken = (tokens: TokenMapper | undefined, tokenAddress: string | undefined) => {
+  if (!tokenAddress || !isAddress(tokenAddress, { strict: false })) return undefined
```
0xLokiB, line 25: This looks sketchy why would we get tokens with invalid addresses?

apps/router-api/src/tokens/tokens.ts
```diff
@@ -0,0 +1,97 @@
...
+  const poolVolumes = curve.getIsLiteChain()
+    ? undefined
+    : await curve.getPoolVolumes().catch(error => {
+        request.log.error({ message: 'Error fetching token volumes', error, chainId: curve.chainId })
+        return undefined
+      })
+
+  const tokenVolumes = pools.reduce<Partial<Record<Address, number>>>((volumes, pool) => {
+    const volume = Number(poolVolumes?.[pool.address.toLowerCase() as Address])
+    if (Number.isNaN(volume) || volume <= 0) return volumes
```
0xLokiB: Too defensive imo

apps/router-api/src/tokens/tokens.ts
```diff
@@ -0,0 +1,97 @@
...
+  const pools = curve.getPoolList().map(id => curve.getPool(id))
+
+  const poolVolumes = curve.getIsLiteChain()
+    ? undefined
+    : await curve.getPoolVolumes().catch(error => {
+        request.log.error({ message: 'Error fetching token volumes', error, chainId: curve.chainId })
+        return undefined
+      })
+
+  const tokenVolumes = pools.reduce<Partial<Record<Address, number>>>((volumes, pool) => {
```
0xLokiB, line 20: Why reduce instead of fromEntries?

apps/router-api/src/tokens/tokens.ts
```diff
@@ -0,0 +1,97 @@
...
+  const nativeWrappedAddress = getAddress(nativeToken.wrappedAddress)
+
+  return fromEntries(
+    notFalsy(
+      // Native token
+      [
+        nativeAddress,
+        {
+          symbol: nativeToken.symbol,
+          decimals: decimals[nativeToken.address] ?? DEFAULT_DECIMALS,
```
0xLokiB: Default decimals could be dangerous

apps/router-api/src/tokens/tokens.ts
```diff
@@ -0,0 +1,97 @@
...
+              volume: tokenVolumes[address],
+            },
+          ] as const
+        }),
+        ...pool.wrappedCoinAddresses.map((stringAddress, index) => {
+          const address = getAddress(stringAddress)
+          return [
+            address,
+            { symbol: pool.wrappedCoins[index], decimals: pool.wrappedDecimals[index], volume: tokenVolumes[address] },
+          ] as const
```
0xLokiB, line 72: No volumes?

apps/router-api/src/tokens/tokens.ts
```diff
@@ -0,0 +1,97 @@
...
+        const address = getAddress(pool.lpToken)
+        return (
+          !isAddressEqual(address, zeroAddress) &&
+          ([
+            address,
+            {
+              symbol: pool.symbol,
+              decimals: decimals[pool.lpToken] ?? DEFAULT_DECIMALS,
+              lp: true,
+              volume: tokenVolumes[address],
```
0xLokiB: This line is repeated, maybe we should get the volumes after gathering addresses?

apps/main/src/dex/components/PagePool/Swap/index.tsx
0xAlunara: Yeah, that's what not checksummed means. In a previous PR I tried to remove lowercasing it but it broke the UI and I didn't want to waste time figuring out why and decided to keep it as it is.

apps/main/src/dex/components/PageRouterSwap/index.tsx
0xAlunara: This page will get an entire do over when we refactor the form itself. I try to avoid refactoring too much, as I don't want to break anything when it's unrelated to the actual token mapper refactoring.

apps/main/src/dex/queries/tokens.query.ts
0xAlunara: The app often tried to get tokens from (incomplete) user input or URL params. Need to gracefully handle those situations. If you don't need that safety you can use `useTokens()[myTokenAddress]` directly

apps/router-api/src/tokens/tokens.ts
0xAlunara: I found the imperative way easier to read and comprehend given we have an internal summing modifying values of multiple different object keys. Could even argue to have a `for of` loop rather than `reduce`.

apps/router-api/src/tokens/tokens.ts
0xAlunara: How so? If there are no pool volumes for the given pool `Number(undefined)` gives `NaN` (and we want to convert to number as the values are of type `Decimal`). If volume is `0` (or less, figured why not) then we want to omit the property altogether so just early return.

apps/router-api/src/tokens/tokens.ts
0xAlunara: I need *something* though. I suppose I could also just omit it if there's no decimals, because without it the token is useless anyway.

apps/router-api/src/tokens/tokens.ts
0xAlunara: Not sure what you mean, all pool tokens have a `volume: tokenVolumes[address]`

apps/router-api/src/tokens/tokens.ts
0xAlunara: That's how I started, but mutating an existing token list or created a new one with extra data was more complex than this. There's a bit of repetition but I favored that over the extra complexity.

apps/main/src/dex/hooks/useRouterApi.ts
0xAlunara: Probably not, but this is copy pasted from legacy code inside the stores. I'd rather clean it up later than risk breaking something now and not knowing if it's something else in this PR or the cleanup.

apps/router-api/src/tokens/tokens.ts
0xAlunara: EDIT: I guess `if (!volume)` works too lol

apps/router-api/src/tokens/tokens.ts
0xLokiB: yeah sorry I didn't read properly

0xLokiB (approved):

0xtutti (approved):

PR #3235, opened 2026-09-16
feat: temporary pool list

0xLokiB:
- Add a new temporary pool list that retrieves pools from a factory.
- This will be replaced by the backend data soon.
- Move `TableHeader`, `ListPageLayout` and `TableButton` to `ui`
- Get rid of the `--width` variable that was not available for the stellar app.


0xtutti (approved):

apps/stellar/src/queries/factories/pools.query.ts
```diff
@@ -0,0 +1,41 @@
...
+    `Missing pool ${index} in factory ${factory}`,
+  )
+
+/** Note: temporary factory pools query, it will be replaced by API when available. */
+export const { useQuery: useFactoryPools } = queryFactory({
+  queryKey: (params: NetworkParams) => [...rootKeys.network(params), 'factories', 'pools'] as const,
+  queryFn: async ({ network }: NetworkQuery): Promise<FactoryPool[]> =>
+    (
+      await Promise.all(
+        STELLAR_NETWORKS[network].factories.map(async factory =>
```
0xAlunara, line 28: `.flatMap` would've worked too ig

0xAlunara (approved):
Was expecting the pool list to also work for the deployment but not the case I guess, probably component tests only?

image

apps/stellar/src/queries/factories/pools.query.ts
0xLokiB: good point I can address in a separate PR. 
 It works for [the testnet](https://curve-frontend-stellar-git-feat-pool-list-curvefi.vercel.app/dex/stellar-testnet/pools) but the stellar mainnet has no factories yet:
https://github.com/curvefi/curve-frontend/blob/d5214c5e0a1ba53ba5e24b6f2381f54dc071f2c2/apps/stellar/src/lib/networks.ts#L14

0xLokiB: The `flatMap` doesn't work btw.
Each factory returns `Promise<Promise<FactoryPool>[]>`. The outer Promise.all resolves those factory promises, producing `Promise<FactoryPool>[][]`, rather than `FactoryPool[]`.

PR #3245, opened 2026-09-16
feat: speed up token balances prefetch by only storing non zero balances

0xAlunara:
Did some performance analysis and found that `prefetchTokenBalances` causes a serious lag spike during load for Ethereum.

Turns out that Ethereum has around 1300 tokens, and calling `setQueryData` for 1300 balances sucks as every time you call `setQueryData` our persistor dehydrates the cache, which consumes a lot of resources.

I've explored a few alternatives with gippety, but none is suitable:

1. `notifyManager.batch`: this doesn't help, as the issue is not related to the notifcations and re-rendering, but is purely a persistor issue.
2. `setQueriesData`: this unironically is internally a composition of `notifyManager.batch` and `setQueryData`, which doesn't alleviate the issue. To make matters worse, it also only updates existing keys and does not allow adding new ones.
3. Setting `{ meta: { persist: false } }` did not prevent the persistor from still calling `dehydrate`.

For now the best I could come up with is to only store balances that are non 0, that should at least cut down 1300 calls however many tokens a user holds, which is significantly less.

A final alternative would be to not store a balance per query key, but my body's not ready for such refactor yet.

Enso token balance fetching might help but only if we'd avoid it populating our TanStack cache, so I don't think that's a 'proper' solution.

User attachment


packages/evm-ui/src/hooks/useTokenBalance.ts
```diff
@@ -220,11 +221,17 @@ export const prefetchTokenBalances = async (
...
-  // { status: 'failure' } entries. Downstream consumers (useTokenBalance, fetchTokenBalance)
-  // already handle per-token failures gracefully.
+  /**
+   * Each token uses 2 contracts (balanceOf + decimals), so chunk results by 2
+   * Failures are fine — allowFailure defaults to true, so failed calls are seeded as
+   * { status: 'failure' } entries. Downstream consumers (useTokenBalance, fetchTokenBalance)
+   * already handle per-token failures gracefully.
+   *
+   * We only care about non-zero balances in order not to overload the TanStack Query
+   * persistor with too many setQueryData calls. There's sadly no good way to batch them. (see commit message)
```
0xAlunara, line 231: I pasted the PR description in the extended commit message, thought it'd be too big for docs

0xtutti (approved):

0xLokiB (approved):
I think we need the backend to provide all the token prices at once.

0xAlunara:
Perhaps yes, but this is also about balances which differ per user.

PR #3250, opened 2026-09-17
fix: cleared tokens did not get their balance reset

0xAlunara:
The previous optimization didn't allow balances to be reset. So if you had 1000 USDC, then spend it all, the token selector would still show a balance of 1000 because its 0 balance would get filtered out of the update set.


packages/evm-ui/src/hooks/useTokenBalance.ts
```diff
@@ -227,11 +226,14 @@ export const prefetchTokenBalances = async (
...
-   * persistor with too many setQueryData calls. There's sadly no good way to batch them. (see commit message)
+   * Only seed new entries for non-zero balances to limit persistence work,
+   * except if they were previously populated and should now be set to zero.
    */
   zip(tokenContracts, chunk(results, 2))
-    .filter(([, tokenResults]) => maybe(tokenResults?.[0].result, balance => balance > 0n) ?? false)
     .map(([contracts, tokenResults]) => [readContractsQueryOptions(config, { contracts }), tokenResults] as const)
+    .filter(
+      ([{ queryKey }, tokenResults]) =>
+        (tokenResults?.[0].result ?? 0n) > 0n || queryClient.getQueryData(queryKey) !== undefined,
```
0xLokiB:
```suggestion
        tokenResults?.[0].result || queryClient.getQueryData(queryKey),
```

0xLokiB (dismissed):
Lint might complain but we should disable that rule tbh

packages/evm-ui/src/hooks/useTokenBalance.ts
0xAlunara: Doh, I tried without the `?? 0n`, but ofc `> 0n` is not necessary to begin with

0xtutti (dismissed):

packages/evm-ui/src/hooks/useTokenBalance.ts
```diff
@@ -227,11 +226,11 @@ export const prefetchTokenBalances = async (
...
    *
-   * We only care about non-zero balances in order not to overload the TanStack Query
-   * persistor with too many setQueryData calls. There's sadly no good way to batch them. (see commit message)
+   * Only seed new entries for non-zero balances to limit persistence work,
+   * except if they were previously populated and should now be set to zero.
    */
   zip(tokenContracts, chunk(results, 2))
-    .filter(([, tokenResults]) => maybe(tokenResults?.[0].result, balance => balance > 0n) ?? false)
     .map(([contracts, tokenResults]) => [readContractsQueryOptions(config, { contracts }), tokenResults] as const)
+    .filter(([{ queryKey }, tokenResults]) => tokenResults?.[0].result || queryClient.getQueryData(queryKey))
```
0xPearce: Shouldn't we filter the cached data that is `0n` instead of only checking if the cached data exists? So we avoid updating unchanged `0n` values
```suggestion
    .filter(([{ queryKey }, tokenResults]) => tokenResults?.[0].result || queryClient.getQueryData(queryKey).[0].result)
```

0xLokiB (dismissed):

packages/evm-ui/src/hooks/useTokenBalance.ts
0xLokiB: this is not true anymore I think?

packages/evm-ui/src/hooks/useTokenBalance.ts
0xLokiB: I think you are right, the query data should be equal to `tokenResults`
```suggestion
    .filter(
      ([{ queryKey }, tokenResults]) =>
        tokenResults?.[0].result || queryClient.getQueryData<typeof tokenResults>(queryKey)?.[0].result,
    )
```

0xLokiB (approved):

0xtutti (approved):

PR #3221, opened 2026-09-14
refactor: market page refinements

0xPearce:
Sorry for the big diff. Most of it is moving metrics around and refactoring, but it also includes a few small fixes.

- the chart header now always shows the token pair instead of sometimes showing “Oracle price”
- implemented the chart sizes from Figma
- added market participants to Mint markets
- grouped the historical rates and market participant cards into tabs
- added oracle price, current LLAMMA price, Max RoE, and total debt metrics
- removed the market overview card and moved most of its data into Advanced Details

PR #3254, opened 2026-09-18
feat: add RoE to leverage action info

0xPearce:
- depends on #3221
- show current and future RoE in leverage action info using collateral yield and borrow APY
- moved RoE calculations in shared rate utilities


apps/main/src/llamalend/widgets/action-card/LoanActionInfoList.tsx
```diff
@@ -221,6 +224,20 @@ export const LoanActionInfoList = ({
...
             />
           )}
+          {collateralApy && (
+            <ActionInfo
+              label={t`Return on Equity (RoE)`}
+              value={formatReturnOnEquity(prevLeverageValue, prevRates, collateralApy)}
+              futureValue={formatReturnOnEquity(
+                leverageValue,
+                /** Collateral-only actions have no future rate query, so future return on equity uses the current rate. */
+                rates?.data === undefined && !rates?.isLoading && !rates?.error ? prevRates : rates,
```
0xAlunara, line 234: Would `rates?.data ? prevRates : rates` suffice? Because I don't think it's going to be properly visible that `isLoading` is `true` or if there's an error?

0xAlunara (dismissed):

apps/main/src/llamalend/widgets/action-card/LoanActionInfoList.tsx
0xLokiB: I think it's nice to have a single way of handling fallbacks, e.g.
```suggestion
                fallbackQ(rates, prevRates).data,
```

apps/main/src/llamalend/rates.utils.ts
```diff
@@ -40,6 +42,43 @@ type BorrowRateMetricsParams<TSnapshot extends WithTimestamp = WithTimestamp> =
...
 
+/** Annualized return on equity at the given leverage. Input APYs and output are percentage. */
+export const getReturnOnEquity = (
+  leverage: number | Nullish,
+  collateralApy: number | Nullish,
+  borrowApy: number | Nullish,
+): number | undefined =>
+  // Total collateral / equity = leverage, so debt / equity = leverage - 1.
+  maybes([leverage, collateralApy, borrowApy], (lev, colApy, borApy) =>
+    lev < 1 ? undefined : lev * colApy - (lev - 1) * borApy,
```
0xLokiB, line 53: why do we return undefined when leverage is <1?

0xtutti (dismissed):

0xtutti (approved):

apps/main/src/llamalend/widgets/action-card/LoanActionInfoList.tsx
0xPearce: @0xAlunara what you mean error or loading not visible? we either pass `prevRates` or `rates` which both are `QueryProp`, so `ActionInfo` receives its loading and error states. @0xLokiB  `fallbackQ` cannot be used either, if `rates` fails and `prevRates` has data, the future roe will be silently calculated with `prevRates` data

apps/main/src/llamalend/rates.utils.ts
0xPearce: to avoid making this part of the equation become a income instead of a cost: `- (lev - 1) * borApy`

apps/main/src/llamalend/widgets/action-card/LoanActionInfoList.tsx
0xLokiB: @0xPearce fallbackQ supports nonFalsy notation
```suggestion
                fallbackQ(rates, !rates.error && prevRates).data,
```

0xLokiB: anyway this looks like duplicated logic already in the ActionInfo component
https://github.com/curvefi/curve-frontend/blob/8a8abb034724d25a6ce80f8ac8d045cd3fd004ad/packages/ui/src/features/forms/action-info/ActionInfo.tsx#L140-L141

0xLokiB (approved):

apps/main/src/llamalend/widgets/action-card/LoanActionInfoList.tsx
0xPearce: This concerns a different fallback. `ActionInfo` falls back to the already calculated current RoE. Here collateral changes leverage, so we must calculate future RoE using future leverage and the unchanged borrowing rate

PR #3247, opened 2026-09-17
refactor: stellar forms consistency

0xLokiB:
- This PR originally contained the new swap form, but that's been separated since I ended up doing a few unrelated refactors.
- Stellar deposit and withdrawal preview are now dedicated action-info components and a centralized `PoolActionInfoList`, following patterns from the llamalend loan forms.
  - Move token-selector from `evm-ui` to `ui`.
  - Minor type and layout cleanup
  - Remove unused code
- Improve Stellar tests 
- The actual swap form is moved to #3255

PR #3255, opened 2026-09-18
feat: stellar swap form

0xLokiB:
- Add the swap form to the stellar network.
- Depends on #3247


apps/stellar/src/features/swap/useSwapForm.ts
```diff
@@ -0,0 +1,121 @@
...
+  const form = useForm<SwapFormValues>(formOptions)
+  const { formState, reset } = form
+  const values = form.watchValues()
+  const { fromIndex, toIndex, editedSide } = values
+  const maxAmount = maxAmounts.data?.[fromIndex]
+  const maxOutput = maybes([reserves.data?.[toIndex], decimals.data?.[toIndex]], fromWei)
+
+  const [params, isDebouncing] = useFormDebounce<SwapFormQuery, 'inputAmount' | 'outputAmount' | 'editedSide'>(
+    useMemo(
+      () => ({
```
0xAlunara, line 54: There ought to be a better way for this haha

apps/stellar/src/queries/swap/swap-quote.query.ts
```diff
@@ -0,0 +1,84 @@
...
+  return { amount: { inputAmount, outputAmount }[amountField], amountDecimals, calculatedDecimals }
+}
+
+export const {
+  useQuery: useSwapQuote,
+  fetchQuery: fetchSwapQuote,
+  getQueryOptions: getSwapQuoteQueryOptions,
+} = queryFactory({
+  queryKey: ({
+    network,
```
0xAlunara, line 33: Similarly for this. I know this is an already existing thing, just funny we spend 30 LoC just for var names 😄

0xAlunara (dismissed):

apps/stellar/src/features/swap/useSwapForm.ts
0xLokiB: we could use the same strategy as on useFormSync to at least remove the dependency array, with the trade-off of ignoring a lint rule 🤔 then it would be safe to `...` the form values

apps/stellar/src/queries/swap/swap-quote.query.ts
0xLokiB: Query keys are annoying for sure! Not sure how to improve this one though. A `pick` helper maybe?

0xtutti (dismissed):

apps/stellar/src/features/swap/useSwapForm.ts
0xLokiB: I created a task to do just that.

apps/stellar/src/queries/swap/swap-quote.query.ts
0xLokiB: I created a task to simplify the query keys

0xAlunara (dismissed):

0xAlunara (approved):

0xtutti (approved):

PR #3258, opened 2026-09-19
refactor: pool header & metrics

0xLokiB:
- Move Metric component, pool header and pool composition to UI package.


0xtutti (dismissed):

apps/main/src/llamalend/features/rate-breakdown/market-rate-breakdown.columns.tsx
```diff
@@ -28,6 +30,7 @@ const rateColumns = (rateHeader: string) =>
             source={tokenInfo}
             address={address}
             explorerUrl={explorerUrl}
+            displayAddress={shortenAddress(address)}
```
0xPearce: address can be `undefined`, as we are using this column to display the `borrow APR` label as well for example

```suggestion
            ...(address && {displayAddress: shortenAddress(address)})
```

0xPearce: <table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>

Image

</td>
    <td>Image</td>
  </tr>
</table>

packages/ui/src/features/pools/PoolCompositionCard.tsx
```diff
@@ -34,7 +44,7 @@ export const PoolComposition = ({ pricesApiPoolData }: { pricesApiPoolData?: Pri
               visibleColumns={table.getVisibleLeafColumns()}
               isLoading={isLoading}
               totalUsd={totalUsd}
-              hasBalance={rows.some(row => row.amount)}
+              hasBalance={rows.some(row => row.amount != null)}
```
0xPearce: why do we want 0 balances to be `hasBalance=true`?

packages/ui/src/components/TokenCell.tsx
```diff
@@ -17,17 +16,19 @@ const { Spacing } = SizesAndSpaces
...
   explorerUrl?: string
   /** Optional content rendered after the token information, such as a badge. */
   endAdornment?: ReactNode
 }
 
 /** Displays token information with copy-address and optional explorer interactions. */
-export const TokenCell = ({ source, address, explorerUrl, endAdornment }: TokenCellProps) => {
+export const TokenCell = ({ source, address, explorerUrl, endAdornment, displayAddress }: TokenCellProps) => {
   address = address ?? ('address' in source ? source.address : undefined)
-  const copyAddress = useCopyToClipboard({ copyText: address, format: getAddress })
```
0xPearce, line 30: llamalend still uses lowercase addresses, this will remove the checksum when copying

apps/main/src/llamalend/features/rate-breakdown/market-rate-breakdown.columns.tsx
0xLokiB: It should not be optional but I'll fix

packages/ui/src/components/TokenCell.tsx
0xLokiB: I added that to all calls

packages/ui/src/features/pools/PoolCompositionCard.tsx
0xLokiB: That doesn't look correct indeed

apps/main/src/llamalend/features/rate-breakdown/market-rate-breakdown.columns.tsx
0xLokiB: In the end there was an usage that omitted addresses altogether, so I have made it optional after all.

packages/ui/src/components/TokenCell.tsx
0xLokiB: Alternatively, I could create a wrapper EvmTokenCell component to handle the formatting.

0xtutti (dismissed):

0xAlunara (approved):

0xtutti (approved):

PR #3265, opened 2026-09-21
feat: stellar basic pool page

0xLokiB:
- Create a very basic pool page for the stellar network using the header and metrics from #3258


tests/cypress/component/stellar/pool-page.testnet.cy.tsx
0xLokiB, line 1: I used a component test because we don't yet run stellar e2e in ci

0xtutti (dismissed):

apps/stellar/src/features/pool/usePoolComposition.ts
```diff
@@ -0,0 +1,51 @@
...
+  decimals: QueryProp<(number | undefined)[]>
+}) {
+  const reserves = usePoolReserves({ network, pool })
+  const prices = useQueries({
+    queries: tokenAddresses.data?.map(token => getTokenUsdRateQueryOptions({ network, token })) ?? [],
+  })
+
+  const totalUsd = combineQueries([reserves, aggregateQueries(prices)], (reserves, rates) =>
+    isComplete(rates)
+      ? decimalSum(...zip(reserves, rates).map(([reserve, rate]) => decimalMultiply(reserve, rate)))
```
0xPearce, line 32: Was testing the loading of the pool page, and when triggering the loading for the config query, it throws with this error. I believe it comes from here. To reproduce go to any pool page in the testnet and trigger the loading of the corresponding config query

Image

0xPearce: Also shouldn't this be 
```suggestion
      ? decimalSum(...zip(reserves, rates).map(([reserve, rate]) => decimalMultiply(fromWei(reserve, decimals), rate)))
```
Can probably be done at the definition so it's not repeated twice

0xPearce (commented):
Also empty pool gives 100% total balance? 

Image

0xtutti (approved):

0xAlunara (approved):

PR #3267, opened 2026-09-21
chore: fix failing gauge rpc test

0xAlunara:
We have a token endpoint now which we need to mock for the gauge tests

RPC tests: [1](https://github.com/curvefi/curve-frontend/actions/runs/35628274163/job/106427921652)


0xtutti (approved):

tests/cypress/component/dex/gauge-management.rpc.cy.tsx
```diff
@@ -128,7 +136,11 @@ describe('Gauge Management (RPC)', () => {
   it('adds a gauge reward token', () => {
     cy.mount(<GaugeManagementTestCase vnet={getVirtualNetwork()} privateKey={privateKey} form="addReward" />)
 
-    cy.get('[data-testid="add-reward-token-selector"]', LOAD_TIMEOUT).should('contain', 'DAI').click()
+    cy.get('[data-testid="add-reward-token-selector"]', LOAD_TIMEOUT)
+      .should('contain', 'DAI')
+      .find('[role="combobox"]', LOAD_TIMEOUT)
+      .should('not.have.attr', 'aria-disabled', 'true')
```
0xLokiB, line 142: I don't think this is necessary. Cypress checks disabled state before clicking

0xLokiB (approved):

tests/cypress/component/dex/gauge-management.rpc.cy.tsx
0xAlunara: I think it's more about the timeout, for some reason it defaulted to 5 seconds. Gippety just added this and now the timeout is longer, so I just went with it.

0xLokiB: Yeah it's fine to fix the test but we have other places passing timeout directly when clicking

0xLokiB: Just smells cause AI always adds these irrelevant guards when Cypress handles that out of the box

PR #3268, opened 2026-09-21
feat: apply pool blacklist when building token list and filter lp tokens on quickswap

0xAlunara:
See title. Didn't bother creating a 'pretty' cache as it's temporary until we migrate to prices api.

EDIT: now re-uses the loaded blacklist from the curve js instance


packages/prices-api/src/chains/schema.ts
```diff
@@ -67,11 +67,16 @@ export const getPoolFiltersResponse = z
...
         chain: z.string(), // API proxies Lite API filters for integrators, so chains are not limited to package support
+        chain_id: z.number(),
         pools: z.array(z.object({ name: z.string(), address })),
       }),
     ),
   })
-  .transform(({ data }) => data.flatMap(item => item.pools.map(pool => ({ chain: item.chain, address: pool.address }))))
+  .transform(({ data }) =>
+    data.flatMap(item =>
+      item.pools.map(pool => ({ chain: item.chain, chainId: item.chain_id, address: pool.address })),
```
0xAlunara, line 77: Nested `camelizeKeys` making my head spin, not gonna bother for a sec

apps/router-api/src/tokens/tokens.ts
```diff
@@ -1,14 +1,37 @@
 import type { FastifyRequest } from 'fastify'
 import { type Address, getAddress, zeroAddress, isAddressEqual } from 'viem'
+import { getPoolFilters, type PoolFilter } from '@curvefi/prices-api/chains'
 import { fromEntries, notFalsy } from '@primitives/objects.utils'
 import { loadCurve } from '../curve-router/curvejs'
 import type { TokensQuery } from './tokens.schemas'
 
+const BLACKLIST_CACHE_DURATION = 10 * 60 * 1000 // 10 minutes
+
+let poolFiltersCache: { expiresAt: number; promise: Promise<Partial<Record<number, PoolFilter[]>>> } | undefined
```
0xLokiB: We have the memoize dependency already, might as well use it?

0xLokiB (dismissed):

0xtutti (dismissed):

apps/router-api/src/tokens/tokens.ts
0xAlunara: Yeah only just noticed it, gonna do a smol rewrite to re-use that info

mimaklas (approved):

0xLokiB (approved):

PR #3276, opened 2026-09-22
refactor: extract isWrapped from pools mapper

0xAlunara:
Extracts the 'wrapped' pools toggle from the slice to the context provider. I consider this a temporary measure as we'll refactor / redo the pool forms in its entirety.

This is about the many forms that have a checkbox that change the behavior of the entire page. This has a profound effect on what tokens are actually being used all over the place.

In a follow up PR I might finally be able to get rid of the pools mapper in its entirety.

I've primarily tested this with the Tricrypto pool on `polygon`: `0xc7c939A474CB10EB837894D1ed1a77C61B268Fa7`

| Not wrapped | Wrapped  |
|---|---|
| image  | image |


apps/main/src/dex/components/PagePool/components/AddGaugeLink.tsx
```diff
@@ -17,7 +17,8 @@ import { t } from '@ui/lib/i18n'
 export const AddGaugeLink = () => {
   const {
     chainId,
-    poolData: { pool, tokens },
+    poolData: { pool },
```
0xAlunara, line 20: Pool data is now just an object with one property: `pool`. But for the sake of diff size I haven't simplified it yet; that's for a follow up PR.

apps/main/src/dex/components/PagePool/Page.tsx
```diff
@@ -54,7 +54,7 @@ export const PagePool = () => {
     />
   ) : (
     poolId && poolData?.pool?.id === poolId && isHydrated && (
-      <PoolContextProvider network={network} poolIdOrAddress={rPoolIdOrAddress}>
+      <PoolContextProvider key={`${chainId}:${poolId}`} network={network} poolIdOrAddress={rPoolIdOrAddress}>
```
0xAlunara, line 57: Gippety said I needed this key for a proper re-render but I'm not 100% sure 🤔

apps/main/src/dex/components/LegacyTooltipBaseApy.tsx
```diff
@@ -17,9 +17,6 @@ export const LegacyTooltipBaseApy = ({
 
   if (poolData?.pool.isLending) {
     label = t`Pool APY + Lending APY`
-  } else if (poolData?.tokenAddresses?.includes('0xae7ab96520de3a18e5e111b5eaab095312d7fe84')) {
```
0xAlunara, line 20: Yeah couldn't be arsed to add `tokenAddresses` support for this legacy edge case

apps/main/src/dex/components/PoolLabel.tsx
```diff
@@ -31,21 +33,33 @@ export const PoolLabel = ({
...
+    tokens: poolTokens,
+    tokenAddresses,
+    tokenAddressesAll,
+  } = useMemo(
     () =>
-      lodash
-        .zip(poolData?.tokens, poolData?.tokenAddresses)
-        .map(([symbol, address]) => ({ symbol: symbol!, address: address! })),
-    [poolData?.tokens, poolData?.tokenAddresses],
+      maybe(pool, pool => getTokens(pool, { wrapped: isWrappedOnly(pool) })) ?? {
```
0xAlunara, line 44: This component is only used in the dashboard which will be removed so I didn't care much about code quality here

apps/main/src/dex/features/pool-context/PoolContextProvider.tsx
```diff
@@ -19,12 +20,13 @@ export const PoolContextProvider = ({
   const { curveApi: api = null } = useCurve()
   const poolId = usePoolIdByAddressOrId({ chainId, poolIdOrAddress })
   const poolData = useStore(state => state.pools.poolsMapper[chainId]?.[poolId ?? ''])
+  const [isWrapped, setIsWrapped] = useState(() => isWrappedOnly(poolData.pool))
```
0xAlunara, line 23: Again, temporary measure until slices are gone and we get new forms

0xtutti (dismissed):

0xtutti (approved):

0xtutti (dismissed):

apps/main/src/dex/components/PagePool/Deposit/components/FieldsDeposit.tsx
```diff
@@ -87,48 +86,44 @@ export const FieldsDeposit = ({
...
     <FieldsWrapper>
-      {poolData.tokens.length === amountsInput.length &&
-        poolData.tokens.map((token, idx) => {
-          const tokenAddress = poolData.tokenAddresses[idx]
+      {tokens.length === amountsInput.length &&
+        tokens.map((token, idx) => {
+          const tokenAddress = tokenAddresses[idx]
           const addressBalanceAmount = userPoolBalances.data?.[tokenAddress] ?? '0'
-          const haveSameTokenName = poolData.tokensCountBy[token] > 1
+          const haveSameTokenName = countBy(tokens)[token] > 1
```
0xLokiB, line 126: Am I reading this correctly that every token will count every other token in each iteration?

apps/main/src/dex/features/pool-context/PoolContextProvider.tsx
0xLokiB: It's not clear why though, it's weird to have this in state

apps/main/src/dex/components/PagePool/Page.tsx
0xLokiB: Did it fix any issues? Seems unnecessary.

apps/main/src/dex/store/createPoolsSlice.ts
0xLokiB, line 1: Is it really worth the risk and work to mess with these slices? I was actually planning to just build a new form next to it like was done for Llamalend.

0xPearce (commented):
I think we have lost the initial token load on the pool swap

Image

apps/main/src/dex/store/createPoolsSlice.ts
0xAlunara: I can't remove the poolsmapper otherwise. Note that I don't really touch these slices other than to remove dependencies on the pools mapper.

apps/main/src/dex/components/PagePool/Deposit/components/FieldsDeposit.tsx
0xAlunara: Yeah I should prob memoize it

apps/main/src/dex/features/pool-context/PoolContextProvider.tsx
0xAlunara: It's a page wide toggle hidden deep inside multiple forms and also used by slices. The inly alternatives are prop drilling a lot, or keeping a zustand store, which I like neither. Putting a bookean state in the context looked like the least evil option to me.

0xLokiB (approved):

0xtutti (approved):

